PR #10, opened 2016-05-29
Simple redirector for rustdocs

onur:
A simple redirector web server for rustdocs. This server will redirect URL's to (`rand` crate used in this example):

* `/rand/` to latest version of rand: `/rand/0.3.14/rand/`
* `/rand/0.3` to latest point release of rand 0.3: `/rand/0.3.14/rand/`

I am planning to use `staticfile` crate to serve files but looks like I can't mount same urls to different handlers. Basically this server should handle:

* `/rand/0.3.14/` with `crate_redirector` function
* `/rand/0.3.14/rand/index.html` with `staticfile` crate


onur:
This pull request evolved to #12, closing this.

PR #18, opened 2016-07-28
Don't 500 on a route with no match

mattyhall:
Previously a bad route (eg. /crate/regex) would be a 500 because the Iron router returned it's own error, not a `Nope`. I guessed it should be a `Nope::ResourceNotFound`

Apologies if this is unidiomatic.


onur:
This is great thanks!

My original implementation was similar, now we only need to ged rid of panic! and show error description in 500 page.

PR #33, opened 2016-08-27
Autofocus on search box

passy:
Hi! 👋

I hope I didn't miss any previous discussion about this. I really prefer it for search pages when I can type as soon as I open them and right now it requires four Tab presses.

Thanks for the amazing project! 💯

---
This change is [Reviewable](https://reviewable.io/reviews/onur/docs.rs/33)


onur:
Thanks!

Issue #38, opened 2016-08-30
Show owners, not only authors


carols10cents:
On the "Crate" tab, I *love* the link to the author in the sidebar, and I love that the author link goes to a page listing all the other things that person has authored.

What I'd really like is the same thing for "owners" in crates.io, basically [this crates.io issue](https://github.com/rust-lang/crates.io/issues/409). 

For example, I'd love to be able to see all the crates that the pistondevelopers github organization is listed as an owner on. This would let me discover crates that are officially supported to work well together, as well as find other crates by authors who I trust.

onur:
@carols10cents docs.rs saved all owner info during build but I didn't want to add owners. But sure, if you want to see owners I'll add them, but I believe "Links" section must have more priority than owners.

What do you think about [this?](https://i.imgur.com/VGDaiL1.png)

carols10cents:
Sure! As long as it's there somewhere :)

PR #39, opened 2016-08-30
Add owners and list crates by owners

onur:
This branch is currently live on docs.rs.

cc: @carols10cents what do you think?

There is only one issue. In this page: <https://docs.rs/releases/@github:pistondevelopers:pistoncollaborator> name looks a bit odd but that's the side effect of using an organization as a user.

Fixes: #38


onur:
It's much better now

carols10cents:
ZOMG this is awesome!! And that was fast!!! Thank you! ❤️

PR #41, opened 2016-09-01
[WIP] Update dependencies and iron 0.4 transition

onur:
This patch updates dependencies, but looks like there is a problem on templates.

@sunng87 I hope you can help :-)

`cargo update` performed:

```
    Updating handlebars v0.16.1 -> v0.18.2
    Updating handlebars-iron v0.15.2 -> v0.15.3
    Updating iron v0.3.0 -> v0.3.1
```

But looks like [this](https://github.com/onur/docs.rs/blob/0564859f8482c000ac935ec3945c634cd61f7ef3/templates/crate_details.hbs) template is no longer working. Problem is getting items from a vector of tuples. I have `authors: Vec<(String, String)>` and I was getting them with:

```
{{#each authors}}
<li class="pure-menu-item"><a href="/releases/{{this.[1]}}" class="pure-menu-link">{{this.[0]}}</a></li>
{{/each}}
```

This syntax is a `RenderError` now and `this.[0]` is no longer working.


sunng87:
Could you please update to handlebars-iron 0.18 which uses handlebars 0.20.5? 
This should be fixed in that version.

onur:
@sunng87 looks like I have to upgrade iron to `0.4` as well.

sunng87:
I see. Is there any road blocker for upgrading to iron 0.4? 
If you have issue with upgrading, I will make a back-port later.

onur:
Since everything is working fine with the current version, there is no need to back-port.

Upgrading iron is causing some problems. I was manipulating `iron::Url.path` in rustdoc handlers but it is no longer possible in iron 0.4 and I have to write a workaround.

I'll focus on more important issues rather than upgrading dependencies for now.

Issue #65, opened 2016-10-06
Fails to build latest serde because of out-of-date nightly


xrl:
I am hitting an issue with serde-derive building on docs.rs: https://docs.rs/crate/rtps/0.2.0/builds/31750

```
error: the `#[rustc_macro_derive]` attribute is only usable with crates of the `rustc-macro` crate type
 --> .cargo/registry/src/github.com-1ecc6299db9ec823/serde_derive-0.8.11/src/lib.rs:9:3
  |
9 | #[rustc_macro_derive(Serialize)]
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The relevant issue in the serde repository: https://github.com/serde-rs/serde/issues/567 .

It looks like the nightly in my build is from September -- perhaps updating it would help?

```
$ rustc --version
rustc 1.14.0-nightly (289f3a4ca 2016-09-29)
```

dtolnay:
Yes updating would help, the rustdoc bug is fixed in rustc 1.14.0-nightly (9c31d76e9 2016-10-03).

onur:
I updated nightly but still getting same error: https://docs.rs/crate/rtps/0.2.1/builds/31776

dtolnay:
I cloned rtps 0.2.1 locally and ran `cargo doc` with rustc 1.14.0-nightly (3210fd5c2 2016-10-05) and it worked. I can try to debug tonight.

onur:
@dtolnay can you give me `rustc` command used to compile `serde_derive-0.8.11/src/lib.rs` from `cargo doc --verbose` output?

dtolnay:
> rustc /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_derive-0.8.11/src/lib.rs --crate-name serde_derive --crate-type rustc-macro -C prefer-dynamic -g -C metadata=a27bc5a26505dfc7 -C extra-filename=-a27bc5a26505dfc7 --out-dir /media/david/coding/github/rtps/target/debug/deps --emit=dep-info,link -L dependency=/media/david/coding/github/rtps/target/debug/deps --extern serde_codegen=/media/david/coding/github/rtps/target/debug/deps/libserde_codegen-d813637bf949d9ed.rlib --cap-lints allow

The difference is `--crate-type lib` vs `--crate-type rustc-macro`.

[...]

Issue #46, opened 2016-09-11
Invalid version requirement


onur:
Some crates (`google-*`) failed to build with an error:

```
2016/09/11 15:38:19 [ERROR] cratesfyi::docbuilder::queue: Failed to build package google-licensing1-cli-0.3.6+20150901 from queue: the given version requirement is invalid
```

`google-*` crates are using a versioning scheme like: `0.3.6+20160617`. cratesfyi's build_doc cannot handle this version string.

onur:
Looks like this issue is also fixed with #69 

`google-*` crates are building fine now.

PR #69, opened 2016-10-17
Update cargo to latest revision

onur:
I believe this patch is ready, but needs to be properly tested.

Fixes: #65

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,7 +3,7 @@ name = "cratesfyi"
 version = "0.2.0"
 dependencies = [
  "badge 0.1.0",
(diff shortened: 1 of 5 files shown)


bluss:
I don't know if it's the news you want to get, but no new releases have been rendered since this was merged.

onur:
Yes @bluss, this update also fixed #46.

docs.rs is building missing `google-*` crates right now and its taking a while.

bluss:
Aha, so we are stuck in the backlog, that's fine. Would it be an idea to mark crates as being equivalent in all arches? It seems like a waste and I'd do it if it leads to faster publishing. I mean, it's a very useful thing when it matters, but often it doesn't.

onur:
I am not sure if it is possible to detect arch-specific documentation differences before building a crate.

PR #70, opened 2016-10-17
style.css: Allow bold text

### LINKED ISSUE #63: Bold (`<strong>`) text in docs is not bold
It looks like the weight for **strong** is overridden in docs.rs. This removes emphasis that may be present in the crate docs.

bluss:
This simply removes the default rule that made bold style invisible. Rustdoc pages must be able to use bold style.

Fixes #63

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -37,10 +37,6 @@ html, button, input, select, textarea,
     color: $color-standard;
 }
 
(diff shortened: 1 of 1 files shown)


onur:
Thanks for the PR.

I only need to fix few `strong` tags in templates before merging this.

Issue #29, opened 2016-08-26
Set default target for certain crates, also features


retep998:
For certain crates I'd like it if I could set the default target, so that when someone looks up a certain crate, they'd get the documentation I want them to get by default. Particularly useful for crates that only define things for certain targets, such as `winapi`.

Also useful would be a way to tell docs.rs which features to enable when building a crate. `winapi` 0.3 will have almost everything disabled by default, and then opting in with features, but that would end up with docs.rs generating useless documentation.

Yamakaky:
Same idea, but different problem: https://docs.rs/crate/clang-sys fails to find libclang (obviously). There is a feature to do a runtime link instead of a compile time link, so it should fix the problem.

onur:
I will add building crates with extra features but where is best place to define docs.rs specific options?

Maybe a `docs.rs` section in `Cargo.toml`?

steveklabnik:
IIRC, we made a `meta` section of Cargo.toml for this kind of thing

steveklabnik:
Oh, it's `metadata` http://doc.crates.io/manifest.html#the-metadata-table-optional

frewsxcv:
A use-case I have: [cocoa-rs](https://github.com/servo/cocoa-rs) is a crate managed by Servo. Servo currently does not build documentation on the OSX builders (https://doc.servo.org is Linux IIRC). It'd be nice if I could link to the documentation for cocoa-rs for the latest published version *that was built on OSX*. That said, I juts checked it looks like [cocoa-rs](https://docs.rs/crate/cocoa/0.5.0) isn't building on docs.rs, which makes my whole comment moot anyways.

bluss:
Doc features are great. What I do in ndarray is to build with the "docs" feature that depends on the optional crate features. Then I've made sure each of the items enabled by that mentions this in the docs [(“*Requires crate feature "serde"*”)](https://bluss.github.io/rust-ndarray/master/ndarray/struct.ArrayBase.html#method.serialize). Adding this to docs.rs would make it have "feature parity" and I'll switch.

onur:
This is partially fixed, looks like a commit closed this.

Docs.rs now supports building non-default features but it is not redirecting to default-target.

retep998:
Would be *really* cool to have the default-target redirect working. It's incredibly frustrating when people look for the documentation for `winapi` only to find empty documentation.

est31:
#255 fixed this

[...]

Issue #50, opened 2016-09-19
Set a specific feature during doc generation


dflemstr:
Use case: if your library depends on some native library or otherwise does something funky during build, it won't be exposed on `docs.rs`.

The build could check for the presence of a `docs` feature in the `Cargo.toml` file, for example like this:

```
cargo read-manifest | rq 'select"$.features"|map(o)=>{_.keys(o)}|spread|some(x)=>{"docs"==x}'
```

(Shameless plug: [rq](https://github.com/dflemstr/rq))

Setting this `docs` feature would allow the library to "defuse" itself, by not running a build script or not trying to link a native library.

onur:
This is a good idea but I was planning to add an option for `extra-features` and `no-default-features`. This is a most requested feature for docs.rs. See: #18 

Checking `docs` feature can also be useful.

Issue #48, opened 2016-09-18
Allow setting `RUSTDOCFLAGS` somehow


hauleth:
It can be useful for adding some custom JS libraries like KaTeX

Relevant issues:
- rust-num/num#225
- libOctavo/octavo#90

onur:
Thanks for reporting.

I'll add this feature along with #29

PR #73, opened 2016-10-20
Add docs.rs metadata and custom builds

onur:
This patch is adding `Metadata` type used in `cargo::ops::CompileOptions` to customize
docs.rs builds.

An example metadata in Cargo.toml:

``` text
[package]
name = "test"

[package.metadata.docs.rs]
all-features = true
no-default-features = true
rustc-args = [ "--example-rustc-arg" ]
dependencies = [ "example-system-dependency" ]
```

This patch is still work in progress and aiming to fix: #29, #48 and #50
#### TODO
- [x] Save default target to database.
- [ ] Install system dependencies before building a package.
- [ ] Redirect to default-target if its specified.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,6 +29,7 @@ dependencies = [
  "staticfile 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.37 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ libc = "0.2"
 badge = { version = "0", path = "src/web/badge" }
 error-chain = "0.10"
 comrak = { version = "0.1", default-features = false }
+toml = "0.4"
 
 # iron dependencies
 iron = "0.5"
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -1,5 +1,6 @@
 
 use ChrootBuilderResult;
+use Metadata;
 use utils::source_path;
 use regex::Regex;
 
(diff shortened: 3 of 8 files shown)


briansmith:
Do you have a staging server? I'd be happy to test this out for you with the *ring* crate, which needs to be built with `--features=rsa_signing` to get correct documentation.

onur:
Unfortunately there isn't any staging server and docs.rs is only working with central repository (crates.io). I'll upload a dummy package to test this.

lucab:
Coming here from https://github.com/onur/docs.rs/issues/23#issuecomment-242943910.

We are facing a similar issue while trying to automate Debian packaging, and we'd like to have some hints from crate author regarding which system packages needs to be installed in order to properly build.

The `package.metadata.docs.rs.dependencies` could be a good start, but I think it is possibly missing:
 * distro selector (a library may have a different package name in debian, ubuntu, fedora, etc...)
 * suite selector (a library may get renamed/split in debian-jessie->debian-stretch->debian-sid)
 * a version selector (a crate may depends on symbols introduces in newer versions of a library, even without soname changes)

May I suggest putting together a strawman proposal of a metadata schema somewhere (eg. https://wiki.debian.org/Teams/RustPackaging/)?

/cc @infinity0 @joshtriplett

infinity0:
opam, the ocaml package manager, is able to deal with non-ocaml system dependencies and interact generically with a system package manager. I will dig out docs for that later today and comment in more detail.

sfackler:
What is this blocked on?

onur:
Actually there is nothing blocking this issue. I wanted docs.rs to build everything automatically without requiring any user input but looks like its not enough. I'll finish this feature and add into docs.rs next week.

lucab:
@onur have you seen https://github.com/joshtriplett/metadeps? It is supposed to solve this problem in a distro-agnostic way.

/cc @joshtriplett

SergioBenitez:
This is so very sorely needed. Can this be merged soon?

onur:
@SergioBenitez I am working on it.

SergioBenitez:
Checking in. How's it going, @onur?

alexreg:
Good work with this. This would be grate for a number of projects that are currently failing to build on *docs.rs*. If there's any way we can help, let us know.

PR #82, opened 2016-11-15
Fix sass on recent releases name.

Nemikolh:
I realize that there's an issue if the crate name is too long.
Please let me know if you would like this done differently.

Before:
img

After:
img_after


onur:
Thanks! I also made release name column a bit wider.

PR #85, opened 2016-11-30
Added Open Search integration…

phaazon:
to browse directly from your URL bar! How fancy; much wow; many
searches!

I haven’t tested it yet (I only know it compiles) because I’m on my lunch break at work. I shall test a local server tonight if I find enough spare time or if anyone tries it.


phaazon:
I fucked up. I’ll re-PR haha :D

PR #109, opened 2017-03-20
Add style tag pre

hngnaig:
This fix response when view on device :).

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -93,6 +93,10 @@ body {
     font: 16px/1.4 $font-family-sans;
 }
 
+pre {
+    overflow: auto;
+}
+
 div.container {
     max-width: 1160px;
     margin: 0 auto;


hngnaig:
@onur ?

onur:
Thanks for PR @hngnaig.

Can you describe the issue a little more? I am trying to stick to rustdoc.css as much as possible in docs.rs.

I believe you want to show scrolls for code blocks in documentations. But adding overflow to pre tags is not achieving what you want because of `white-space: pre-wrap;` defined in [rustdoc.css](https://github.com/rust-lang/rust/blob/a559452b05c20041a27f518d262573c56b876b64/src/librustdoc/html/static/rustdoc.css#L131).

hngnaig:
Hi @onur, i'm view on device mode and i see the tag `<pre>` not response, this same when i view on sony xperia. I think this PR will make response tag `<pre>` on device mode. And i see another crate when use tag `table` more than three column not reponse on device :(

onur:
@hngnaig I see now thanks.

I think you can also send same PR to [rust-lang/rust](https://github.com/rust-lang/rust), it is an issue in rustdoc.

PR #120, opened 2017-05-03
Add real-time PubSubHubBub to the feed

samueltardieu:
[PubSubHubBub](https://en.wikipedia.org/wiki/PubSubHubbub) is a protocol that lets readers of an atom or RSS feed subscribe to a hub so that they get warned when the feed information has been updated.

The hub takes care of managing the subscription information, subscription refreshing, and so on. When it receives a "ping" from the web site, indicating that the feed has been updated, it checks it and contacts all subscribers with the new information (including the feed itself in some cases, a.k.a. fat ping).

To use the two major public hubs ([Google](https://pubsubhubbub.appspot.com/) and [SuperFeedr](https://pubsubhubbub.superfeedr.com/)), a site only needs to add two `<link rel="hub" href="…" />` lines to its feed, and ping the hubs through a `POST` request when the feed has been modified. Those two steps are implemented in the two commits of this PR.

As an example, people using the [SuperFeedr Telegram bot](https://telegram.me/superfeedr_bot) to get information on newly built documentation will get it as soon as it gets published instead of having the bot poll the docs.rs feed periodically.

### DIFF
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -35,8 +35,9 @@ impl DocBuilder {
 
 
     /// Builds packages from queue
-    pub fn build_packages_queue(&mut self) -> Result<()> {
+    pub fn build_packages_queue(&mut self) -> Result<usize> {
         let conn = try!(connect_db());
+        let mut build_count = 0;
 
         for row in &try!(conn.query("SELECT id, name, version
                                      FROM queue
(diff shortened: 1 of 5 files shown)


onur:
This seems useful, thanks for the effort!

jyn514:
Hi @samueltardieu, are you still using this feature? It seems ... not terribly useful to me, since you can get the list of crate publishes from crates.io already.

samueltardieu:
I don't use the feed much. Do you want to remove it?

jyn514:
I do, yes. I'll make a PR for that shortly.

Thanks for the prompt response!

Issue #121, opened 2017-05-09
Vagrant image is broken


onur:
Due to: rust-lang/rust@31a4d73 docs.rs started using latest version of nightly `rustc 1.19.0-nightly (f1140a331 2017-05-08)`. Relevant commit: cf4d28b. But looks like rusttype no longer builds in this version of rust compiler.

This issue lead to a broken vagrant provision. Currently there is no way to build docs.rs with provided vagrant image.

I'll try to fix rusttype and make it work in latest nightly.

PR #122, opened 2017-05-09
Use our own version of rusttype

onur:
Closes: #121 

This must fix nightly rusttype compilation problem but vagrant image needs to be tested.


onur:
Tested vagrant image and everything is working fine now.

Issue #103, opened 2017-02-07
The requested resource does not exist


rushmorem:
For some reason, some crates on docs.rs are coming up with this error message. This used to be a mystery to me but now even one of my own crates is returning this error for the latest version. Right now docs for [v0.0.6-alpha2](https://docs.rs/reql/0.0.6-alpha2/reql/) are running into this error while those for [v0.0.6-alpha1](https://docs.rs/reql/0.0.6-alpha1/reql/) and below are fine.

onur:
@rushmorem looks like v0.0.6-alpha2 [failed to build](https://docs.rs/crate/reql/0.0.6-alpha2/builds/41328). Some package also failed with same error message. I am not sure what is causing this. Maybe you can help, did you change any dependency in new version?

rushmorem:
I had a hunch this might be the problem. This log confirms that. The thing is `reql-io` is an implementation detail for `reql` . It depends on `reql` for some datatypes. However, to make it easier for users, `reql` also depends on `reql-io` to reexport it. Cargo doesn't seem to have a problem with this. However, thanks to this issue, I'm now working on a better design which should work on `docs.rs`.

I'm going to leave this issue open in case someone wants to try and fix this. @onur If docs.rs is working as intended here, feel free to close this issue.

onur:
I believe docs.rs is working as intended. I tried to run example code in README:

```rust
extern crate reql;

use reql::r;

fn main() {
    r.connection().connect().expect("Failed to connect to the database server");
}
```

with `reql = "0.0.6-alpha2"` dependency I am getting same error:

```
$ cargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: cyclic package dependency: package `reql v0.0.6-alpha2` depends on itself
```

And `v0.0.6-alpha3` is working fine. Docs.rs should build this package without any issue (it's in queue right now).

Issue #58, opened 2016-09-26
Idea: quick dropdown menu to switch between versions


dashed:
It wasn't immediately obvious on how to switch between versions of docs until switching it via the "Crates" tab.

Hence, I think it might be more convenient to switch between versions via a dropdown menu. This dropdown menu may be placed next to the "Platform" dropdown menu.

onur:
I think we can improve #57 and add all versions as a dropdown menu.

dashed:
Yeah I saw that issue and wanted to suggest the 'version dropdown menu' idea in that issue; but instead opted to create a new one. 😄

bluss:
I think it's important to remember to make it stand out using color or otherwise, when the docs are not the latest version.

PR #131, opened 2017-06-07
Next

onur:
This is first major upgrade to docs.rs

## Container OS upgrade

 docs.rs is now using Debian "stretch" to build crates. This will allow us to use more recent version of system dependencies.

## External crate links

Docs.rs started using a slightly modified [rust compiler](https://github.com/onur/rust). This compiler is setting `html_doc_root` for all crates and it's dependencies. This is fixing every single external crate link (along with types, traits etc.) in documentations. To achieve this docs.rs is also using a slightly modified [cargo](https://github.com/onur/cargo).

You can see this feature in documentation of [CargoError](https://docs.rs/cargo/0.18.0/cargo/util/errors/trait.CargoError.html) implementors. Docs.rs is now resolving every external link properly.

Note: `i686-pc-windows-msvc` and `x86_64-pc-windows-msvc` targets are no longer supported, docs.rs will try to build documentation for `i686-pc-windows-gnu` and `x86_64-pc-windows-gnu` targets for windows operating system.

## Non-default features

This is one of the most requested feature for docs.rs. You can now tell docs.rs to build documentation for the features you want. A simple `package.metadata.docs.rs` section in your `Cargo.toml` will do the job:

```toml
[package.metadata.docs.rs]
all-features = false
no-default-features = true
```

You can check [docsrs-test](https://docs.rs/docsrs-test/0.0.3/docsrs_test/) crate for example.

Note: `default-target` is not implemented yet.

## Shared target directory

Docs.rs no longer builds every single dependency from scratch. It is using a shared `deps` directory to store all build artifacts. This change improved builds times significantly. A simple comparison from builds logs of cargo crate:

* [cargo old](https://docs.rs/crate/cargo/0.18.0/builds/47824): 111.97 secs
* [cargo new](https://docs.rs/crate/cargo/0.18.0/builds/61381): 6.48 secs

This allowed me to build entire crates.io-index in 5 days (only the latest versions of crates).

## Crate details in navigation

You can see crate details in navigation:

I believe this is providing more essential information in documentations. It looks a bit odd in chrome, hope someone help to make it look better in major browsers.

## Transition

* [x] Rebuild every single crate (only the latest versions).
* [ ] Replacement of old `cratesfyi-prefix` directory with the new one.

### Issues

* Closes: #48
* Closes: #50
* Closes: #58
* Closes: #103 
* Partially: #29

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,6 +1,6 @@
 [root]
 name = "cratesfyi"
-version = "0.4.0-dev"
+version = "0.4.0"
 dependencies = [
  "badge 0.1.1",
  "cargo 0.20.0 (git+https://github.com/onur/cargo.git?branch=docs.rs)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,7 +3,7 @@ name = "cratesfyi"
 version = "0.4.0-dev"
 dependencies = [
  "badge 0.1.1",
- "cargo 0.20.0 (git+https://github.com/rust-lang/cargo.git?rev=9fe7f07579d6d2672dd7e95c70adaea65d8e9a2e)",
+ "cargo 0.20.0 (git+https://github.com/onur/cargo.git?branch=docs.rs)",
  "clap 2.23.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "comrak 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "crates-index-diff 3.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -158,11 +158,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "cargo"
 version = "0.20.0"
-source = "git+https://github.com/rust-lang/cargo.git?rev=9fe7f07579d6d2672dd7e95c70adaea65d8e9a2e#9fe7f07579d6d2672dd7e95c70adaea65d8e9a2e"
+source = "git+https://github.com/onur/cargo.git?branch=docs.rs#a2996e83fe5eea05b26b7f95465a4936c5171348"
 dependencies = [
  "advapi32-sys 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "crates-io 0.9.0 (git+https://github.com/rust-lang/cargo.git?rev=9fe7f07579d6d2672dd7e95c70adaea65d8e9a2e)",
+ "crates-io 0.9.0 (git+https://github.com/onur/cargo.git?branch=docs.rs)",
  "crossbeam 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "curl 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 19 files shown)


retep998:
Why are `i686-pc-windows-msvc` and `x86_64-pc-windows-msvc` no longer supported?

onur:
@retep998 I have to build rust compiler for all tier one platforms myself and it is taking tremendous amount time. Is there any crate that only supports `windows-msvc` and not `windows-gnu`? I assumed `windows-gnu` should be enough for windows related crates.

retep998:
@onur The one case I can think of where the API of a crate would change is bindings to the CRT. MinGW and VC++ both provide a different CRT, MinGW using the ancient `msvcrt.dll` with some extra stuff statically linked in, while VC++ has their newer Universal CRT. The set of APIs available differs between those two CRTs, as well as the layouts of various types. That said, neither `winapi` nor `libc` actually care about that distinction at the moment, so it's not a huge deal.

Does building the Rust compiler really take that long? I'd imagine you'd simply build the compiler once and have all the tier one platforms configured as targets, and each target only needs a standard library which should only take a few minutes.

steveklabnik:
>  Is there any crate that only supports windows-msvc and not windows-gnu?

It's ultimately that msvc is the preferred target, so having the docs be for gnu is not great.

gabdube:
`Is there any crate that only supports windows-msvc and not windows-gnu? I assumed windows-gnu should be enough for windows related crates.`

One of my creates do not build on gnu (  https://github.com/gabdube/native-windows-gui ) . Its only because of a missing symbol (and I can create a workaround) but I guess the deeper you go in the winapi, the more likely these issues can popup.

steveklabnik:
@onur oh one more thing; we should talk about what it would take to get you off of a custom rustdoc. Is there a particular reason you went with it this way instead of sending changes upstream?

onur:
@steveklabnik my rustdoc is basically ignoring html_root_url from crate root and [setting it to docs.rs](https://github.com/onur/rust/commit/4640db7dc233cee819d30bf78fce9fbafb5bf3e9) even if its not defined. Also, I might be wrong but rustc only cares about `lib.name` (`--crate-name` and `--extern NAME=PATH` argument passed by cargo). There is no way to get `package.name` and docs.rs is using a url scheme like: `docs.rs/$PACKAGE_NAME/$VERSION/$LIB_NAME`. Because of this reason I had to add a new argument to rustc, rustdoc and cargo: `--extern-versions`. I don't think upstream would accept this changes.

@retep998 unfortunately building windows targets are only possible on a windows host, and apple targets requires apple host etc. I am using 3 different host machine to build rust compiler for all tier 1 platforms. It's just not automated like nightly builds and manual steps are taking too much time.

But if I am not mistaken a windows-gnu host can build windows-msvc targets and it should be easy to add windows-msvc target. I'll add windows-msvc target again when I am free.

steveklabnik:
>  I don't think upstream would accept this changes.

We might; I'd want to talk about the details, but I'd much prefer that we extend rustdoc for your use-case than have you on a custom rust.

lilianmoraru:
This is pretty awesome work!

> You can see crate details in navigation:

Huh, good, I am not blind.
I was always going to `Source` -> `Crate`, in order to get to the `GitHub` page.
I observed the dropdown menu on the navigation a few days ago and was wondering how I didn't observe before this nice menu with all the important info nicely packed in there...

On another note: I see that the metadata supports only a default target.
Would you please consider adding support for multiple targets?
I am thinking of crates like `simd` here: https://github.com/rust-lang-nursery/simd/blob/3ce0b70860f0ccb6b93002ed66654c76cbfbb234/src/lib.rs#L218 - that currently presents the docs only for `x86`.
When I looked at the docs, I thought it doesn't support Neon, AVX, etc... I switched between platforms(I was actually searching for AVX, so I was hoping it will pop-up) but it was still presenting only `x86`(only sse2).

onur:
I re-enabled MSVC targets.

bluss:
Thanks for this. The intercrate linking is also really nice.

PR #140, opened 2017-07-19
remove line wrapping in <pre> tags

### LINKED ISSUE #139: source line wrapping breaks line numbering
docs.rs seems to modify the line wrapping behavior of the source pages somehow, and it causes the line numbers to be out of sync. See [example](https://docs.rs/postgres/0.14.2/src/postgres/rows.rs.html#395). This doesn't happen when viewing the crate docs normally (e.g. `cargo doc -p postgres --open`).

durka:
Should fix #139.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -91,6 +91,7 @@ body {
 
 pre {
     overflow: auto;
+    white-space: pre;
 }
 
 div.container {


onur:
Thanks for the report and PR!

This is an issue I was aware of for a while but I wasn't able to find a solution. Docs.rs is also showing source codes in a tight container unlike original rustdoc. I think preserving `class` attribute for body tag should fix this issue.

PR #149, opened 2017-08-24
Small improvements for the crate search input

GuillaumeGomez:
Here's the output (the "important" part is at the top right):

screenshot from 2017-08-24 13-45-52

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -140,6 +140,9 @@ div.nav-container {
             display: none;
             @media #{$media-sm} {
                 display: block;
+                border-right: 1px solid $color-border;
+                border-left: 1px solid $color-border;
+                padding-right: 3px;
             }
         }
         input.search-input-nav:focus {


templates/style.scss
```diff
@@ -140,6 +140,9 @@ div.nav-container {
             display: none;
             @media #{$media-sm} {
                 display: block;
+                border-right: 1px solid #ddd;
```
aulisius: The hardcoded color can be replaced with the already existing $color-border variable.

GuillaumeGomez: Oh indeed. Updating!

GuillaumeGomez:
Updated.

onur:
Thanks for the PR @GuillaumeGomez, but I am not sure about this change, is it really necessary? I prefer current look of it.

GuillaumeGomez:
It took me a while to:

1. See this input.
2. Understand it was an input.

By adding borders, it makes it more obvious and easier to spot.

onur:
Yeah, I agree it's hard to see, but adding borders is just making it look like another link in navbar. I don't think it's solving visibility problem.

GuillaumeGomez:
Might be worth giving it a try and ask users what they think about it no? Reverting the change is easy after all.

bluss:
@onur I think you have named the problem -- it looks like toolbar text with or without the lines.

quadrupleslap:
I *think* I'd prefer it with a spyglass icon to the left instead. Anyway, why not post the screenshot on r/rust *before* pushing instead of afterwards?

Zexbe:
I think this should be merged, because it is better then the current solution, and it does not prevent a different solution in the future. I didn't even know it was a search box until I read this pull request.
r? @jyn514

jyn514:
I don't think it's bad, I just don't think it helps very much. I like the spyglass suggestion.

jyn514:
Closing in favor of https://github.com/rust-lang/docs.rs/pull/577

Issue #145, opened 2017-08-15
Dash integration


hobofan:
I contacted @Kapeli regarding a docs.rs integration for Dash, similar to ones of rubygems.org, etc..

He said that while it is on the roadmap, he can't say when he will get to it. However, to make the process as smooth as possible, docs.rs should be ready for when the time comes. 😉 

Quote from the email:
> What I would need from docs.rs is:
> 1. A way to search for packages
> 2. A way to get the available versions for a given package
> 3. A way to download the docs in HTML format for a given package and version

Current status:
1. should be pretty straightforward, since a similar route already exists, so we just need a JSON version of it
2. AFAIK there is no specfic route for that yet, but since the same info is already shown at other places, so I guess the groundwork for that is there, and it just needs to be exposed on its own route
3. Route also doesn't exist yet (related to #88). I guess the best way to achieve it would be to generate a `.tar.gz` archive on the fly when it is first requested, and then cache that for subsequent requests. From what I can see, the [tar crate](https://crates.io/crates/tar) makes it pretty easy to compress directories into a file.

---

@onur If the proposed approaches sound okay to you, I'll prepare a PR for each of routes.

emoon:
Any update on this? I would love to have docs.rs support in Dash.

hobofan:
@Kapeli also created an issue regarding a Dash integration: https://github.com/rust-lang-nursery/docs.rs/issues/174

But other than that, no progress really. My first pull request is still open, and I'd still like to do further work towards it, if that one is accepted.

Maybe now that the repo has been moved under the rust-lang-nursery umbrella, things can move forward a bit...

zargony:
I'd like to see this happen too. I'm voting for crate docs support in Dash for years now (almost thought I'm the only one who'd like it).

Kapeli stated that he just needs docs to be downloadable in some way. But since docsets seem to be pretty common (e.g. Zeal on Linux also uses them), wouldn't it be more versatile for docs.rs to provide docsets directly? (or maybe some other filetype if there's a more popular one).
Additional processing for creating docs upon a crate release isn't probably much. However, additional disk storage size would be significant.

hobofan:
I'm not sure if doc.rs should have native support for docsets. I think the current way of having one canonical doc format, that can then be transformed into others by the user is a more proper way, and also places less maintenance burden on the docs.rs team.

[...]

Issue #83, opened 2016-11-17
Support to download docs offline


lilianmoraru:
I found today this site: http://devdocs.io/offline
It has support to download the documentation for offline navigation.
Would be nice if `docs.rs` would support that too.

jyn514:
Duplicate of https://github.com/rust-lang/docs.rs/issues/174

PR #161, opened 2017-10-13
Add json version of search route; #145

hobofan:
First step towards #145.

I based the implementation of the .json route of the existing`/crate/:name/:version/builds.json` route.

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -4,6 +4,7 @@ use iron::Handler;
 use iron::status;
 use web::page::Page;
 use std::fmt;
+use time;
 
 #[derive(Debug, Copy, Clone)]
 pub enum Nope {
@@ -48,7 +49,21 @@ impl Handler for Nope {
             Nope::NoResults => {
                 use params::{Params, Value};
                 let params = req.get::<Params>().unwrap();
-                if let Some(&Value::String(ref query)) = params.find(&["query"]) {
+
+                if req.url.path().join("/").ends_with(".json") {
+                    use iron::status;
+                    use iron::headers::{Expires, HttpDate, CacheControl, CacheDirective, ContentType,
+                                        AccessControlAllowOrigin};
+
+                    let mut resp = Response::with((status::Ok, "[]"));
+                    resp.headers.set(ContentType("application/json".parse().unwrap()));
+                    resp.headers.set(Expires(HttpDate(time::now())));
+                    resp.headers.set(CacheControl(vec![CacheDirective::NoCache,
+                                                       CacheDirective::NoStore,
+                                                       CacheDirective::MustRevalidate]));
+                    resp.headers.set(AccessControlAllowOrigin::Any);
+                    Ok(resp)
+                } else if let Some(&Value::String(ref query)) = params.find(&["query"]) {
                     // this used to be a search
                     Page::new(Vec::<super::releases::Release>::new())
                         .set_status(status::NotFound)
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -125,6 +125,9 @@ impl CratesfyiHandler {
         router.get("/releases/search",
                    releases::search_handler,
                    "releases_search");
+        router.get("/releases/search.json",
+                   releases::search_handler,
+                   "releases_search_json");
         router.get("/releases/queue",
                    releases::build_queue_handler,
                    "releases_queue");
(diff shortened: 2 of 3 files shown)


hobofan:
Hey @onur! Would be nice to get some feedback on this or any of the related issues regarding efforts for making offline docs possible (#83, #145, #174), as this seems to be one of the most wanted features for docs.rs, going by the Github reactions.

reeze:
Is there any reviewer to check this out?

hobofan:
As docs.rs has become an official Rust project by now, and there is one maintainer now, I'll try again:
@QuietMisdreavus, @onur, is there any chance to get this reviewed?

QuietMisdreavus:
Running through the backed-up PRs; this one (and the related issues) has been a little intimidating to look at because of the ardent support behind it. However, now that i've actually looked at it, it's a rather small start, so it doesn't *seem* too bad. However, there's a question of precedent-setting here, since if we start introducing a machine-readable API there's going to be concerns about stability.

For example, i'm not sure i want to expose the direct type that we use in our search results. It feels rather cumbersome to actually use, since it doesn't assemble any proper URLs that you can easily follow. For example, i built your branch (after rebasing it to the current master to make sure it still worked), and ran a query for "test":

```json
[
  {
    "description": "Helper crate for testing symbolic.\n",
    "name": "symbolic-testutils",
    "release_time": "Oct 24, 2018",
    "release_time_rfc3339": "2018-10-24T12:55:54Z",
    "rustdoc_status": true,
    "stars": 0,
    "target_name": "symbolic_testutils",
    "version": "5.5.2"
  },
  {
    "description": "Meta crate for testcontainers, a library for integration-testing against docker containers from within Rust.",
    "name": "testcontainers",
    "release_time": "Oct 23, 2018",
    "release_time_rfc3339": "2018-10-23T23:39:22Z",
    "rustdoc_status": true,
    "stars": 0,
    "target_name": "testcontainers",
    "version": "0.3.1"
  },
  {
    "description": "A test for a specific prisoners dilemna problem",
    "name": "prisoners_switches",
    "release_time": "Oct 23, 2018",
    "release_time_rfc3339": "2018-10-23T09:16:01Z",
    "rustdoc_status": false,
    "stars": 0,
    "target_name": "prisoners_switches",
    "version": "0.1.0"
  },
  {
    "description": "Core crate of testcontainers, a library for integration-testing against docker containers from within Rust.",
    "name": "tc_core",
    "release_time": "Oct 23, 2018",
    "release_time_rfc3339": "2018-10-23T23:26:16Z",
    "rustdoc_status": true,
    "stars": 0,
    "target_name": "tc_core",
    "version": "0.2.0"
  },
  {
    "description": "Simple ad-hoc server with SPA support based on Warp!. Excellent for testing React, Angular, Vue apps and the like.",
    "name": "microserver",
    "release_time": "Oct 23, 2018",
    "release_time_rfc3339": "2018-10-23T20:04:32Z",
    "rustdoc_status": false,
    "stars": 0,
    "target_name": "microserver",
    "version": "0.1.1"
  }
]
```

You would have to know ahead of time that to assemble the documentation URL, you would need to stitch together the domain name (which isn't even present here) with the path `/{name}/{version}/{target_name}`. From an API design perspective, that doesn't seem like a reasonable expectation to require of users, to me. I understand this was supposed to be a starting point from which further iteration can be done, but this doesn't even feel useful.

...but maybe we can still get this in. The code here all looks fine; it's just the signalling that bothers me. Can you change the route for the JSON search path to `"/unstable/releases/search.json"`? Calling it "unstable" would make me feel better incrementally building an API on the production service.

hobofan:
@QuietMisdreavus Thanks for the feedback!

I can very much understand your concerns regarding providing a stable API, and agree. I'd be happy to change the root to a `/unstable` one.

Regarding the response content I also agree. Maybe back then the fields were different and more useful? I'll look into it again, and think about which fields make sense to actually expose, and which others would make sense to generate, and adjust the PR.

Out of all issues in the Rust ecosystem #145 is in my top 5 most wanted, so I'm glad that there finally has been some response!

jyn514:
Triage: I think we should decide what fields it makes sense to expose before we merge anything, as per your comment.

> Regarding the response content I also agree. Maybe back then the fields were different and more useful? I'll look into it again, and think about which fields make sense to actually expose, and which others would make sense to generate, and adjust the PR.

pietroalbini:
In my opinion docs.rs should not provide a search API. crates.io already provides one, and we should not aim to duplicate their functionality.

jyn514:
I don't think this is going to get merged as-is. If you're interesting in continuing to work on a JSON API, feel free to comment and I can re-open.

PR #176, opened 2018-02-08
Get build_package_in_chroot target from metadata.default_target.

quadrupleslap:
- The only test I did was to make sure it built.
- It silently reverts to the platform native target if...
  - there was an error getting the metadata.
  - the user gave a target that was unsupported.

Again, sorry if this is an incredibly stupid PR and I'm just missing something.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,38 +1,3 @@
-[root]
-name = "cratesfyi"
-version = "0.4.2"
-dependencies = [
- "badge 0.1.1",
- "cargo 0.20.0 (git+https://github.com/onur/cargo.git?branch=docs.rs)",
- "clap 2.23.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "comrak 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


onur:
Hi @quadrupleslap this is great, thanks for the PR!

But there is a problem. By default (without any `target` in [CompileOptions](https://docs.rs/cargo/0.20.0/cargo/ops/struct.CompileOptions.html) docs.rs' cargo is  building documentation into `/home/cratesfyi/cratesfyi`. But when you provide some target, it will build into `/home/cratesfyi/cratesfyi/{TARGET}`. Your patch is working for building non-default target but docs.rs is failing to copy default documentation since it's not available in default location.

But this is right approach and I was exactly thinking the same when I first added `Metadata`. I'll fix it when I find some time, thanks again!

quadrupleslap:
Thank you! I hope you get around to it sometime.

QuietMisdreavus:
This looks like it got superseded by https://github.com/rust-lang/docs.rs/pull/255, so i'm going to close this. If you think that this PR should still be considered, feel free to comment and reopen!

PR #177, opened 2018-02-14
Upstream changes to badge crate

srijs:
Hi!

I've recently started to use the `badge` crate on [deps.rs](https://deps.rs/).

This PR is to upstream some of the improvements I made, including:

- updating `rusttype` from 0.2 to 0.4
- removing the unused lifetype parameter from `Badge` (breaking change)
- adding a `Badge::to_svg_data_uri` method

I can also break this down into multiple PRs if you'd like.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,7 +68,7 @@ dependencies = [
 
 [[package]]
 name = "badge"
-version = "0.1.1"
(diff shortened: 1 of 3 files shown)


onur:
Hi @srijs, sorry it took a while but thanks for the PR!

PR #196, opened 2018-04-14
Document docs.rs metadata on the 'about' page.

frewsxcv:
Someone online asked me what this metadata section was in one of my Cargo.toml files, and it'd be helpful if this documentation was somewhere on the docs.rs site.

Docs/example taken from:

* https://github.com/onur/docs.rs/blob/8a31bf722cb416fec86866e3dd54ba59762e31fa/src/docbuilder/metadata.rs#L15-L27

### DIFF
--- a/templates/about.hbs
+++ b/templates/about.hbs
@@ -99,6 +99,24 @@
     </tbody>
   </table>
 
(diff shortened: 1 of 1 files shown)


onur:
Thanks!

PR #199, opened 2018-04-29
Allow 'latest' alias for most recent crate version

### LINKED ISSUE #169: Support /cratename/latest as alias for newest version
Maintaining links to `docs.rs` documentation is quite annoying since, as far as I can tell, there is no way to reference the latest version by url.

eg: Let's say I want to link to documentation for a certain type like https://docs.rs/diesel/1.0.0-beta1/diesel/pg/struct.Pg.html. I have to update that link every time a new version is released, which can be done automatically sometimes, but also can be cumberosme in some situations.

It would be great to have an alias like , for the example above: ttps://docs.rs/diesel/latest/diesel/pg/struct.Pg.html, where `latest` is recognized as an alias for the newest released version.

theduke:
Allow "latest" as an additional URL alias (together with 'newest') for accesing the most recent crate version.

Fixes #169

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -232,7 +232,7 @@ fn match_version(conn: &Connection, name: &str, version: Option<&str>) -> Option
                 Err(_) => None,
             }
         })
-        .map(|v| if v == "newest" { "*".to_owned() } else { v })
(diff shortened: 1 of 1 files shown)


onur:
Thanks!

PR #218, opened 2018-07-04
move the built `cratesfyi` into the vagrant container

QuietMisdreavus:
Works around https://github.com/rust-lang/rust/issues/49710 when developing docs.rs with a Windows host.

This sets the `CARGO_TARGET_DIR` environment variable within the vagrant box to a folder inside the container rather than in the share. This means that when building the `cratesfyi` executable or the docs.rs service, the files will be placed inside the container (in `/home/cratesfyi/target`) rather than in `/vagrant`. This does mean that the binaries will be unavailable outside the container without manually copying them out. However, without this change, any build will fail with an error like the one in the linked issue. Combined with https://github.com/onur/docs.rs/pull/216, I can fully build docs.rs from a Windows host.


onur:
Thanks @QuietMisdreavus, but that vagrant image really needs some work. I am planning make a great return to this project really soon.

PR #224, opened 2018-07-22
add support for default-target

QuietMisdreavus:
This changes gives an effect to the `default-target` option. A crate that adds a `default-target` to their docs.rs metadata in Cargo.toml will cause `docs.rs/my_crate/$VERSION` (and `docs.rs/my_crate`) to redirect to `docs.rs/my_crate/$VERSION/$DEFAULT_TARGET/my_crate` (assuming a build target of `my_crate`).

~~Caveat: the redirect doesn't check that the target has docs available. I'm not sure why but my local clone doesn't properly build and save docs for all targets, so accessing `localhost:3000/winapi` redirects properly to `localhost:3000/winapi/0.3.5/x86_64-pc-windows-msvc/winapi/`, but immediately hits a 404. This would work in docs.rs's production environment, but i want to look into adding a check, regardless.~~ whoops that was easier than i thought, all good now. I mean, the bug where it doesn't seem to get the docs for all the targets in a local clone still exists, but at least we don't get bogus redirects now.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -99,17 +99,28 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
     // get target name
     // FIXME: This is a bit inefficient but allowing us to use less code in general
     let target_name: String = {
-        let query = ctry!(conn.query("SELECT target_name, default_target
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
Closing in favor of https://github.com/rust-lang-nursery/docs.rs/pull/255, which has already merged

PR #225, opened 2018-07-25
Update nightly rustc to 2018-07-24

kennytm:
The previous nightly 2018-06-03 suffers from rust-lang/rust#50561, which caused several packages failed to build (see https://github.com/onur/docs.rs/issues/23#issuecomment-397496013). This bug has been fixed on Jun 12th by rust-lang/rust#50617, so we just need to update the nightly beyond this date.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -47,7 +47,7 @@ Vagrant.configure("2") do |config|
     ############################################################
     lxc-attach -n cratesfyi-container -- apt-get update
     lxc-attach -n cratesfyi-container -- apt-get install -y --no-install-recommends curl ca-certificates binutils gcc libc6-dev libmagic1
-    lxc-attach -n cratesfyi-container -- su - cratesfyi -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-06-03'
+    lxc-attach -n cratesfyi-container -- su - cratesfyi -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-07-24'
 
     ############################################################
     # Creating rustc links for cratesfyi user                  #


onur:
Hi @kennytm, thanks for your PR. I am always getting this kind of PR's to update rustc but unfortunately updating rustc on docs.rs is a little bit more complicated process. After implementing all docs.rs specific changes to upstream rustc, it will be much easier (it's a work in process atm). Currently I need to update my [internal](https://github.com/onur/rust) rustc and build for 6 different platform and test latest rustdoc changes to update rustc on docs.rs.

So, this PR is only updating rustc that required to build docs.rs.

I'll try to update rustc this weekend when I find some free time and after implementing changes to upstream rustc we will be able to use rustup for future updates.

kennytm:
I see, thanks :)

Diggsey:
@onur can I (or anyone else) help move this along? The changes to rustc look fairly minor - what needs doing to upstream them? I'm not quite sure what the debian stuff is for, are those changes still needed if the other changes are upstreamed?

arbitrix:
One concern here is that the patch can lead to an explosion of different rust versions being downloaded and perhaps taking up diskspace.

A second concern is what happens when packages aren't updated and keep referencing very old nightlies. (Nightlies that have long since gone stable.)

arbitrix:
That said, supporting multiple versions of rust of course is highly desirable. Especially since -as a Rust newbie- I just uploaded a Rust 2018 package that requires beta which of course breaks docs.rs.

Perhaps one could start out with small curated library of rustc versions that are installed on docs.rs. This list presumably would contain a recent stable version (because none of the other channels are guaranteed to work) and one or more flavors of nightly and/or beta.

If stable is updated to the latest version and a crate breaks this presumably needs to be addressed by the crate maintainers. If this holds the list only needs to contain one stable.

If nightly is updated to the latest version and a crate breaks you could perhaps fall back to the "previously used nightly that didn't yield errors on the crate" in the docs.rs library.

So if the Rust compiler library held one stable and two or three nightlies (with buggy nightlies being replaced with less problematic ones) perhaps this could compile enough crates without errors?

It would still be useful if a crate could specify that it needed nightly or beta (or stable?). Or perhaps a minimum version required.

QuietMisdreavus:
Closing this PR in favor of https://github.com/rust-lang-nursery/docs.rs/pull/248 which updates to a newer Rust (and has been merged).

Discussions about allowing crates to choose their nightly can go into https://github.com/rust-lang-nursery/docs.rs/issues/228.

PR #229, opened 2018-08-18
[WIP] Adding a configuration option to specify rustc version

spacekookie:
This code is maybe not as elegant as it could be but should hopefully work. 

I'm not 100% sure how the crates are built (i.e. how the working directories are
built and scoped) so I hope that `rustup` is invoked in the correct directory.

If that's not the case, then this can be changed.

Additionally I am aware that this is a **work in progress** and further changes
are required on docs.rs and the compilers part to make "hot-swapping" versions
easier. See #225 for more details.

This PR addresses #228

### DIFF
--- a/src/docbuilder/metadata.rs
+++ b/src/docbuilder/metadata.rs
@@ -45,6 +45,14 @@ pub struct Metadata {
     /// is always built on this target. You can change default target by setting this.
     pub default_target: Option<String>,
 
+    /// Define the Rust version that will be used to build documentation
+    ///
+    /// You can either provide a specific nightly (such as `2018-08-18`) or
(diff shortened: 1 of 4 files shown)


jyn514:
Thank you for your contribution! Unfortunately, this PR was written for the old builder, which we stopped using last September for sandboxing reasons (see https://blog.rust-lang.org/2019/09/18/upcoming-docsrs-changes.html for details). It looks like the PR would have to be rewritten from scratch to work with Rustwide, so I'm closing this. If you would like help porting it, feel free to reply here or [on discord](https://discordapp.com/invite/f7mTXPW/) and I'd be happy to help you out.

PR #240, opened 2018-10-01
Don't override the user's fonts

jonhoo:
This patch is similar in spirit to rust-lang/blog.rust-lang.org#263, though it's also less invasive; it removes the specific font overrides in style.css so that the browser will use the user's configured fonts.

The net result of the change is that the page will look less "uniform" across different browsers, OSes, and devices, but the overall aesthetic should be preserved. It also means that user preferences are respected, and that legibility is improved for users with particular font needs, or on platforms that sometimes gets confused about how to render fonts nicely (ehrm, Linux).

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -1,10 +1,4 @@
 
-// FONTS
-$font-family-sans: "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
-$font-family-serif: "Source Serif Pro",Georgia,Times,"Times New Roman",serif;
-$font-family-mono: "Source Code Pro", Menlo, Monaco, Consolas, "DejaVu Sans Mono", Inconsolata, monospace;
-
-
 // COLORS - Guess what?
 $color-standard: #000;             // pure black
 $color-url: #4d76ae;               // blue
@@ -33,7 +27,7 @@ $media-xl: "screen and (min-width: 80em)";
 
 html, button, input, select, textarea,
 .pure-g [class *= "pure-u"] {
-    font-family: $font-family-sans;
+    font-family: sans-serif;
     color: $color-standard;
 }
 
@@ -44,12 +38,12 @@ html, button, input, select, textarea,
 }
 
 .description {
-    font-family: $font-family-serif;
+    font-family: serif;
 }
 
 // rustdoc overrides
 div.rustdoc {
-    font-family: $font-family-serif;
+    font-family: serif;
     padding: 10px 15px 20px 15px;
 
     .sidebar {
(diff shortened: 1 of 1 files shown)


onur:
Hi @jonhoo, thanks for the PR.

I love my user fonts more than fonts we used in rustdoc/docs.rs but I'd like to stick to rustdoc decision and use same fonts on docs.rs.

jonhoo:
In that case, I think the font list should at least be changed like this:
```diff
-$font-family-sans: "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
-$font-family-serif: "Source Serif Pro",Georgia,Times,"Times New Roman",serif;
-$font-family-mono: "Source Code Pro", Menlo, Monaco, Consolas, "DejaVu Sans Mono", Inconsolata, monospace;
+$font-family-sans: "Fira Sans", sans-serif;
+$font-family-serif: "Source Serif Pro",serif;
+$font-family-mono: "Source Code Pro", monospace;
```
The intermediate fonts can be quite ugly on Linux machines :)

jonhoo:
You may also need to do the same thing as rustdoc does, and pull in the primary fonts in case the user doesn't have them.

jimbo1qaz:
i'm too lazy at the moment and don't feel like grepping through the CSS to extract every single selector which forces Fira Sans, to assemble a single selector to change to sans-serif. I'm sure this can be automated as a program or JS webapp, but I don't feel like writing it now. Can this be merged (as well as upstream rustdoc)?

In the meantime I configured Firefox to disable web font overrides (breaks Google Docs fonts, but I don't care at the moment).

QuietMisdreavus:
@jimbo1qaz [The upstream rustdoc PR was eventually closed without merging](https://github.com/rust-lang/rust/pull/54772), so it's unlikely that this will remove the overrides either. May i ask why you want to remove them?

jimbo1qaz:
Fira Sans has an uncomfortably large x-height for my tastes (lowercase characters feel disproportionately huge and uppercase/ascenders small), which grates with me on PC displays. This is a design choice of Fira Sans (designed for small phone screens on Firefox OS). And the use of SCSS string-replacements, not CSS variables, makes it painfully difficult for me to override the CSS myself.

I could define a @font-face named Fira Sans and referencing my own preferred font (sadly not sans-serif), but getting the right names of all weights and italic right is a hassle. Google Fonts CSS's local() uses the wrong name half the time.

jyn514:
I don't think we plan to merge this. If something changes upstream in rustdoc, feel free to reopen.

PR #247, opened 2018-10-22
Better error messages in options.rs

jonas-schievink:
This also removes a redundant check - I don't know if there is an actual "logs" path that should be checked instead.

### DIFF
--- a/src/docbuilder/options.rs
+++ b/src/docbuilder/options.rs
@@ -3,7 +3,6 @@
 use std::{env, fmt};
 use std::path::PathBuf;
 use error::Result;
-use failure::err_msg;
 
 #[derive(Clone)]
 pub struct DocBuilderOptions {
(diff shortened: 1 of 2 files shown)


src/docbuilder/options.rs
```diff
@@ -94,16 +93,13 @@ impl DocBuilderOptions {
...
+            bail!("destionation path '{}' does not exist", self.destination.display());
         }
         if !self.chroot_path.exists() {
-            return Err(err_msg("Chroot path not exists"));
+            bail!("chroot path '{}' does not exist", self.chroot_path.display());
         }
         if !self.crates_io_index_path.exists() {
-            return Err(err_msg("crates.io-index path not exists"));
-        }
-        if !self.crates_io_index_path.exists() {
```
QuietMisdreavus: Can you replace this with a check for `self.sources_path` instead? Looks like that path is not being checked.

src/docbuilder/options.rs
```diff
@@ -94,16 +93,13 @@ impl DocBuilderOptions {
 
     pub fn check_paths(&self) -> Result<()> {
         if !self.destination.exists() {
-            return Err(err_msg("Destination path not exists"));
+            bail!("destionation path '{}' does not exist", self.destination.display());
```
QuietMisdreavus: Typo, should be `destination`.

PR #254, opened 2018-10-24
Remove pc-windows-gnu targets

onur:
When I switched custom Rust builds, I never provided a build for pc-windows-gnu targets and nobody asked for them. I think it's safe to remove this targets from docs.rs to save up some space.

### DIFF
--- a/src/docbuilder/chroot_builder.rs
+++ b/src/docbuilder/chroot_builder.rs
@@ -15,13 +15,11 @@ use error::Result;
 
 
 /// List of targets supported by docs.rs
-const TARGETS: [&'static str; 8] = [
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
cc @retep998, do you know of any crates that would be negatively affected if docs.rs stopped generating docs for windows-gnu targets?

retep998:
In general there should be no API differences between `-gnu` and `-msvc`. However there are some crates that fail to build under `-msvc` such as the GTK crates.

onur:
docs.rs can only build gtk crates for `x86_64-unknown-linux-gnu` platform since they need an external system dependency. Removing pc-windows-gnu targets doesn't make any difference for those crates.

PR #265, opened 2018-11-22
Update github links from rust-lang-nursery to rust-lang

KodrAus:
cc @QuietMisdreavus

I think GitHub should be managing redirects for these links, but I did a global search for `rust-lang-nursery`. Hopefully I haven't missed anything :)


onur:
Thanks!

Issue #257, opened 2018-10-25
Display actual rust version used on docs.rs about page


dekellum:
Assuming that for stability reasons, docs.rs is not going to be updated every night to the very latest nightly, it would be helpful if there was a known/easy place to see what version docs.rs was actually using.  This would allow testing `cargo doc` on the same version, manually or via CI, to confirm it will build on docs.rs before a release, instead of being surprised by a docs.rs build failure after release. 

As per this #233 [comment](233#issuecomment-433082216) by @QuietMisdreavus, on 2018-10-25:
> Now that #243 is merged, docs.rs is using 1.31.0-nightly (78ff609d7 2018-10-19) as its rustc/rustdoc  version.

Consider changing the following line on https://docs.rs/about:

> Docs.rs automatically builds crates' documentation released on crates.io using the nightly release of the Rust compiler. 

To something like:

> Docs.rs automatically builds crates' documentation released on crates.io, currently using rust 1.31.0-nightly (78ff609d7 2018-10-19).  If your project needs a later version, check existing [issues](.) and create a new issue if appropriate.

In addition, there could be some CI or other automation utility for something like https://docs.rs/about/version returning a text/plain or application/json version string.

PR #272, opened 2018-12-05
print current rust version on the about page

QuietMisdreavus:
Closes https://github.com/rust-lang/docs.rs/issues/257

Now that we're using a real nightly compiler to build docs, let's show the version somewhere. This PR takes the suggestion from the linked issue and prints the version of the rustc inside the container on the "about" page.

Open question: Would it be better to cache this version string somewhere (say, in the database or in some `lazy_static`) so we don't have to attach to the container and run `rustc --version` every time someone loads the About page?

### DIFF
--- a/src/web/sitemap.rs
+++ b/src/web/sitemap.rs
@@ -1,11 +1,11 @@
-
-
+use std::collections::BTreeMap;
 use iron::prelude::*;
 use iron::headers::ContentType;
+use rustc_serialize::json::Json;
 use super::page::Page;
 use super::pool::Pool;
 use time;
-
+use db::connect_db;
 
 pub fn sitemap_handler(req: &mut Request) -> IronResult<Response> {
     let conn = extension!(req, Pool);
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
Output on my personal test server (which is using an older rustc than production, but the same principle should apply):

image

dekellum:
LGTM, thanks!

onur:
Like you said in your comment, problem with this approach is its running rustc (and lxc-attach etc.) every time when someone visits about page. I think its much better to cache rustc version. We can store it in our `config` table in database (it's a simple table with key and value members designed for this kind of things), and since we are running `cratesfyi build add-essential-files` every time when we update rustc, we can use this function to update value in database.

QuietMisdreavus:
I've updated the PR to use the `config` table like you suggested. Loading it during `add-essential-files` is perfect. I've tested it on my dev server and it works great.

src/web/sitemap.rs
```diff
@@ -31,3 +31,20 @@ pub fn robots_txt_handler(_: &mut Request) -> IronResult<Response> {
     resp.headers.set(ContentType("text/plain".parse().unwrap()));
     Ok(resp)
 }
+
+pub fn about_handler(_: &mut Request) -> IronResult<Response> {
+    let mut content = BTreeMap::new();
+
+    let conn = ctry!(connect_db());
```
onur: We are using our rd2d connection pool in our iron handlers, instead of creating a new connection. Getting `req` parameter and using `let conn = extension!(req, Pool);` should do the trick (similar to sitemap).

PR #277, opened 2018-12-18
Remove contributors and sponsors

onur:
I am really grateful that Leaseweb initially sponsored this project and made it possible. Unfortunately our barter deal has become to an end and we are moving to a different home.

Contributors table is also hard to maintain since we are getting new contributors every day.

This patch is removing sponsors and contributors from README and about page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,32 +147,3 @@ It is designed to run inside a secure container.
 ```
 cargo run -- doc <CRATE_NAME>
 ```
-
-
-#### Contributors
-
-* [Onur Aslan](https://github.com/onur)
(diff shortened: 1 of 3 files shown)


QuietMisdreavus:
Why don't we include a link to the [Rustdoc team](https://www.rust-lang.org/governance/teams/dev-tools#Rustdoc%20team-info) in place of the contributors list? That way we can at least provide some listing of who is "responsible" for the site.

onur:
@QuietMisdreavus I added contact header and improved about page a little bit.

Issue #278, opened 2018-12-20
An error occurs while trying to build add-essential-files


aleksuss:
I try to start docs.rs from a vagrant image. But I got stuck when try to build `add-essential-files` stage and have an error: 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorMessage { msg: "sources path \'/home/cratesfyi/cratesfyi-prefix/sources\' does not exist" }
```

onur:
Thanks for reporting @aleksuss. This error is caused by: https://github.com/rust-lang/docs.rs/pull/247#discussion_r227836744

You can create that directory manually with `mkdir /home/cratesfyi/cratesfyi-prefix/sources` and run `vagrant up --provision` again until we fix this issue.

onur:
yea nightly rustc used in Vagrant also needs an update. Sorry @aleksuss I am prepared a patch to fix Vagrantimage. You can wait until it gets merged or you can directly use that branch.

aleksuss:
@onur yes, with latest nightly rustc ok.

PR #279, opened 2018-12-20
Remove sources path and fix Vagrant vm

onur:
docs.rs is been using `~/.cargo/registy/src` instead of `sources_path` for a while. This patch removes `sources_path`, updates nightly compiler used in Vagrant, removes some unused dependency from Cargo.lock and removes unused dependency from sitemap.rs.

Closes: #278

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -47,7 +47,7 @@ Vagrant.configure("2") do |config|
     ############################################################
     lxc-attach -n cratesfyi-container -- apt-get update
     lxc-attach -n cratesfyi-container -- apt-get install -y --no-install-recommends curl ca-certificates binutils gcc libc6-dev libmagic1
-    lxc-attach -n cratesfyi-container -- su - cratesfyi -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-12-19'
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -47,7 +47,7 @@ Vagrant.configure("2") do |config|
     ############################################################
     lxc-attach -n cratesfyi-container -- apt-get update
     lxc-attach -n cratesfyi-container -- apt-get install -y --no-install-recommends curl ca-certificates binutils gcc libc6-dev libmagic1
-    lxc-attach -n cratesfyi-container -- su - cratesfyi -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-10-20'
+    lxc-attach -n cratesfyi-container -- su - cratesfyi -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-12-19'
```
QuietMisdreavus: I wonder if we can just set this to `nightly` instead of pinning a date to it. We're updating rustdoc on the production server often enough that updating this file every time may start to become too cumbersome.

onur: Yes that makes sense, ok changed to use nightly.

PR #281, opened 2018-12-21
updates to handle new nightlies

QuietMisdreavus:
This PR adds three things:

* Starts copying `noscript.css` with "essential files"
* Sends `--static-root-path /` to rustdoc to make it load CSS/JS/font files from the root of the server and let people cache them better
* Sends `--disable-per-crate-search` to rustdoc to stop creating the crate listing drop-down on search boxes, since we only ever build one crate at a time


QuietMisdreavus:
Since this PR is still open, i'm adding `--static-root-path` and `--disable-per-crate-search` to it as well.

Issue #175, opened 2018-02-03
Show dev-dependencies separately


bluss:
Under dependencies for a crate, both dev-dependencies and dependencies are listed together.

It would be good to mark dev-dependencies separately: pulling in too many dependencies isn't always popular, and a crate's dev-dependencies do not become your own dependencies, when you use the crate, so they should be distinguished.

spk:
This can be closed I guess its on production cf https://docs.rs/pyo3/0.6.0-alpha.4/pyo3/

onur:
Yes, thank you @spk!

PR #282, opened 2018-12-26
Show dev/build dependencies info on menu and detail

spk:
I decided to go with the simple fix without JS; `nomal` kind is hidden `dev` and `build` is shown (only for new crate)

ref #175


QuietMisdreavus:
Ah, looks like the CSS that was meant to hide normal deps didn't quite work:

image

QuietMisdreavus:
Aha, looks like an issue with the template that i missed. I've opened https://github.com/rust-lang/docs.rs/pull/308 to fix it.

PR #287, opened 2019-01-18
update cargo

QuietMisdreavus:
The `recrypt` crate had [a confusing build error](https://docs.rs/crate/recrypt/0.4.0/builds/137824) that doesn't reproduce when building locally. With this Cargo update, it builds on my test server, so i'm hoping that this will let it build on prod.


QuietMisdreavus:
I combined this with https://github.com/rust-lang/docs.rs/pull/290 and tested it on my dev server; crates seem to build alright, so i'm pushing this through.

Issue #270, opened 2018-12-04
changes required in future nightlies


QuietMisdreavus:
The following PRs add new files to the rustdoc distribution:

* ~~https://github.com/rust-lang/rust/pull/55707 (merged, in nightly) added `source-script.js` as a new static file, and `source-files.js` as a new per-doc-bundle file.~~ (added in https://github.com/rust-lang/docs.rs/pull/274)
* ~~https://github.com/rust-lang/rust/pull/56332 (merged, but not in nightly yet) added `down-arrow.svg` as a new static file.~~ (added in https://github.com/rust-lang/docs.rs/pull/274)

When both of these are in nightly (may take a while for the latter one), we'll need to update docs.rs to make sure it adds these files to the site-wide cache.

QuietMisdreavus:
Once https://github.com/rust-lang/rust/pull/56577 lands, we'll want to start passing `--generate-redirect-pages` to rustdoc.

QuietMisdreavus:
https://github.com/rust-lang/rust/pull/56005 adds a new static file: `noscript.css`.

QuietMisdreavus:
I'm reopening this to keep as a running list of "changes necessary in new nightlies", especially since the last two weren't included in #274.

QuietMisdreavus:
- After https://github.com/rust-lang/rust/pull/57011 merges, we can start passing `--static-root-path /` to rustdoc to adjust how CSS/JS files are loaded on documentation pages and increase browser cache usage.
- After https://github.com/rust-lang/rust/pull/57006 merges, we can start passing `--disable-per-crate-search` to disable the crate filter on the in-doc search box, which is useless for docs.rs anyway because of how it generates docs.

QuietMisdreavus:
https://github.com/rust-lang/rust/pull/57400 changes the font files distributed/used by rustdoc. Since docs.rs loads these as "essential files" but *doesn't* apply a resource suffix, this may be problematic, but i think we can just get away with keeping the old font files in the database and loading in the new ones when we get them.

QuietMisdreavus:
https://github.com/rust-lang/rust/pull/57552 adds two new static files: `favicon.ico` and `rust-logo.png`. Since docs.rs uses its own favicon, that one can technically be ignored, but new pages will start trying to reference the logo image.

QuietMisdreavus:
[...]

PR #290, opened 2019-01-29
add new essential files from new nightly

QuietMisdreavus:
Adds in the files from the last couple comments in https://github.com/rust-lang/docs.rs/issues/270.

The font updates are weird, but since it's not *changing* any of the existing font files, just *adding/removing* them, we can just switch to this new list of essential files, leave the old font files in place for existing documentation, and make it load in the new font files when the new rustc is used.


QuietMisdreavus:
I combined this with https://github.com/rust-lang/docs.rs/pull/287 and tested it on my dev server; crates seem to build alright, so i'm pushing this through to attempt to diagnose some of the build failures.

Issue #291, opened 2019-02-08
Build queue hickup?


hellow554:
The build queue is stalled for 8-ish hours. Is there something wrong?

https://docs.rs/releases/queue

> learn_rust-0.1.2
pentest-toolbox-improved-0.1.7
learn_rust-0.1.3
enpass-cli-0.1.1
pentest-toolbox-improved-0.1.8
pentest-toolbox-improved-0.1.9
diesel-citext-0.1.0
rustc-ap-syntax_pos-368.0.0
rustc-ap-syntax-368.0.0
rustc-ap-serialize-368.0.0
rustc-ap-rustc_target-368.0.0
rustc-ap-rustc_errors-368.0.0
rustc-ap-rustc_data_structures-368.0.0
[...]

https://docs.rs/releases

> pentest-toolbox-improved-0.1.6 7 hours ago
learn_rust-0.1.1 7 hours ago
pentest-toolbox-improved-0.1.5 7 hours ago
pentest-toolbox-improved-0.1.4 7 hours ago
pentest-toolbox-improved-0.1.3 7 hours ago
gog-0.3.0 8 hours ago
x86-0.14.0 8 hours ago
[...]

hellow554:
Ok, the queue has been shortened now. Interestinly learn-rust version 0.1.2  hasn't been built although version 0.1.3 has released by now (which hasn't been built either).
Don't know if it's worth some investigation.

QuietMisdreavus:
I manually kicked `learn_rust` out of the queue - it was triggering a bug in docs.rs which was causing a crash, which was holding up the queue. I'm testing a branch for it right now.

QuietMisdreavus:
I've opened https://github.com/rust-lang/docs.rs/pull/292 to fix the crash that that crate was causing. Once it's deployed, i'll rebuild it..

QuietMisdreavus:
Alright, #292 has been deployed and `learn_rust` has been built successfully. That crate shouldn't block the queue in the future.

PR #292, opened 2019-02-08
handle docs that don't use spaces in doc comments

QuietMisdreavus:
Fixes https://github.com/rust-lang/docs.rs/issues/291

The `learn_rust` crate had a line in its crate docs which had no space between the `//!` and the first character - and the first character was a multibyte character. Combined with the code this PR is changing, docs.rs started slicing a string in an improper location, causing the build to panic and fail.

I ran the code on my test server and built `learn_rust-0.1.2`, and it worked just fine.


onur:
:+1:

PR #293, opened 2019-02-09
Show daily build failure count in release activity page

onur:


QuietMisdreavus:
Thanks so much for working on this!

I think, ideally, this information may be better off as a separate page, so we can pull it more often (say, every 3-4 hours?). I'd also love to have the list of most recent crates that failed cached so we can link to them on the page and try to diagnose the failure.

This is worth merging as-is, but if it's not too much trouble to make the extra page and the listing of recent failures, i'd love to have that.

src/utils/release_activity_updater.rs
```diff
@@ -11,30 +11,45 @@ pub fn update_release_activity() -> Result<()> {
     let conn = try!(connect_db());
     let mut dates = Vec::new();
     let mut crate_counts = Vec::new();
+    let mut failure_counts = Vec::new();
 
-    for day in 1..31 {
+    for day in 0..30 {
```
QuietMisdreavus: Did you mean to change the loop range here? It doesn't seem like the code farther down changed how it used it, so i'm wondering whether it was supposed to be like this originally.

onur: Yes, currently that loop is off by a day, this change is fixing it. I have been aware of this for a while but never had a chance to fix. This PR is a good excuse to fix it.

Issue #296, opened 2019-02-13
Source browser only shows 95 characters


iliekturtles:
Clicking the `Source` link (e.g. https://docs.rs/crate/uom/0.21.0/source/src/storage_types.rs, line 7) in the docs.rs header goes to a page where the source view only shows 95 characters. The default `rustfmt` maximum line width is 100 which means horizontal scrolling is sometimes necessary. Note that clicking a `src` link (e.g. https://docs.rs/uom/0.21.0/src/uom/storage_types.rs.html#7) within the documentation goes to a different view with a collapsed sidebar that shows over 100 characters.

After some experimentation I found that in Firefox/Chrome/Edge changing the `font-size` from the default of `16px` inherited from `body` to `15px` allows for 101 characters to be seen. There didn't seem to be a great spot in the scss file to make the change so I have not submitted a PR, but am willing to do so with a bit of guidance:

- Is this a desired change?
- Would using an `rem` or `em` font be preferred? I only used `15px` in my experimentation because `body`'s `font-size` was also specified in `px`.
- Should both source views be updated so that the font-size is consistent?
  - Would this require a `rustdoc` change?
- Where should the change be applied?
  - For all `pre` elements?
  - `.package-page-container pre, .example-wrap pre`?

iliekturtles:
Resolved by #297.

iliekturtles:
Not actually resolved by https://github.com/rust-lang/docs.rs/pull/297#issuecomment-464172067.

PR #297, opened 2019-02-14
Improve readability of build logs

onur:
[Before](https://user-images.githubusercontent.com/345828/52778644-8a2cf180-3057-11e9-9abd-93a267e447cd.png) - [After](https://user-images.githubusercontent.com/345828/52778634-800af300-3057-11e9-9c31-310dd0f49120.png)

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -276,6 +276,11 @@ div.recent-releases-container {
         font-weight: 500;
     }
 
+    pre {
+        font-size: 0.8rem;
+        white-space: pre-wrap;
+    }
+
     .release {
         display: block;
         border-bottom: 1px solid $color-border;


iliekturtles:
#296 resolved by this PR. I'll close. Thanks!

QuietMisdreavus:
Oh wait, i just realized that it may not fix that issue, since this is meant to fix build logs, not source views. Still, i'll merge this; we can check out whether it fixes that issue once it's deployed.

iliekturtles:
Good point. I forgot to check the full context of the scss change and it only applies to the `.recent-releases-container`. The fix for #296 would need the change applied to `pre` elements in `.package-page-container` (~line 546). The change could be applied there, but a fuller review of all the `pre` styles is probably a better idea.

PR #299, opened 2019-02-14
skip loading readme files if it won't fit in the DB

QuietMisdreavus:
Some crates have excessively long README files, that are too big for the database column. This PR changes the README loading function to discard a README if it wouldn't fit in the database, using a similar check that the `read_rust_doc` function uses.

### DIFF
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -232,7 +232,14 @@ fn get_readme(pkg: &Package) -> Result<Option<String>> {
     let mut reader = try!(fs::File::open(readme_path).map(|f| BufReader::new(f)));
     let mut readme = String::new();
     try!(reader.read_to_string(&mut readme));
-    Ok(Some(readme))
+
+    if readme.is_empty() {
+        Ok(None)
+    } else if readme.len() > 51200 {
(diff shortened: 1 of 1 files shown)


onur:
Interesting I thought 50KB should be enough for README files (I have no idea why I assumed first TBH). Maybe instead of ignoring this files, we can increase storage size in the [database](https://github.com/rust-lang/docs.rs/blob/cee3f821bce5e236639b2478096de0fee58b32cb/src/db/mod.rs#L104).

QuietMisdreavus:
That would be a schema change; should we wait for #305 to land first?

QuietMisdreavus:
Actually, i'm not sure increasing the size of the column is a good long-term fix - *someone* will wind up making an even longer README than what we decide on, whether accidentally or deliberately. We should have a way to properly deal with those crates. Right now they just sit in the queue at maximum attempts and don't even register as failed.

QuietMisdreavus:
I'm going to go ahead and merge this in, since it echoes the code in `read_rust_doc` farther down the file. #305 is merged now, so we can easily increase the storage size for READMEs and doc headers in the future, but for now, i'd rather put this guard in. This will allow more crates to be loaded in.

PR #300, opened 2019-02-14
log the URL that caused a 404

QuietMisdreavus:
When inspecting logs for the server process, there are a lot of log lines that are simply

```
2019/02/14 21:24:16 [DEBUG] cratesfyi::web: Requested resource not found
```

This PR adds an extra `debug!()` call to the error handler that also prints the URL, to aid in debugging whether it's a missing static file, a misspelled crate, or a drive-by Wordpress attack. This is mainly for my own curiosity, since i want to see if there's a pattern in the amount of 404s we get.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -45,12 +45,12 @@ mod builds;
 mod error;
 mod sitemap;
 
-use std::env;
+use std::{env, fmt};
 use std::error::Error;
 use std::time::Duration;
 use std::path::PathBuf;
 use iron::prelude::*;
-use iron::{Handler, status};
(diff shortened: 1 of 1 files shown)


onur:
Thanks. I also think its much better if we remove that debug message in [src/web/mod.rs](https://github.com/rust-lang/docs.rs/blob/94aa7160fad6138f743adf5dee36d2932638bbc3/src/web/mod.rs#L232) in the future. That message is causing so much noise in the our main log file. We can just use web servers log file for 404 requests.

Also IIRC there was some bug in that message but I can't remember :smile:

PR #308, opened 2019-02-18
fix navigation_rustdoc template

QuietMisdreavus:
In https://github.com/rust-lang/docs.rs/pull/282, there was an error in the `navigation_rustdoc` template that i missed in the review. This fixes that so that normal dependencies don't have the "*normal*" tag next to them.


onur:
Great!

spk:
Sorry for the bug and thanks for the fix!

PR #313, opened 2019-02-26
Make font size consistent for all `pre` elements.

### LINKED ISSUE #296: Source browser only shows 95 characters

iliekturtles:
The change also has the effect of allowing 100 characters (`rustfmt`'s
default) to be visible on all source views. Resolves #296.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -129,6 +129,7 @@ body {
 }
 
 pre {
+    font-size: 0.9rem;
     overflow: auto;
     white-space: pre;
 }
@@ -314,7 +315,6 @@ div.recent-releases-container {
     }
 
     pre {
-        font-size: 0.8rem;
         white-space: pre-wrap;
     }
 


templates/style.scss
```diff
@@ -98,6 +98,7 @@ body {
 }
 
 pre {
+    font-size: 0.8rem;
```
QuietMisdreavus: I had a chance to test this today. `0.8rem` seems a little small, personally; what do you think about `0.9rem`? It still fits 100 characters, but is a little bigger and easier to read.

iliekturtles:
Adjusted to `0.9rem` and rebased on `master`!

PR #318, opened 2019-03-19
templates/about: mention that "dependencies" key isn't being used yet

decathorpe:
Adding a comment that the `dependencies` key in the `package.metadata.docs.rs` key isn't yet used for builds should clear up some confusion.

### DIFF
--- a/templates/about.hbs
+++ b/templates/about.hbs
@@ -117,7 +117,7 @@ no-default-features = true
 default-target = "x86_64-unknown-linux-gnu"
 rustc-args = [ "--example-rustc-arg" ]
 rustdoc-args = [ "--example-rustdoc-arg" ]
-# the dependencies key is not being used for builds yet
+# the dependencies key is currently unused, but is being saved for future support
 dependencies = [ "example-system-dependency" ]</pre></code>
 
   <h4>Version</h4>
--- a/templates/about.hbs
+++ b/templates/about.hbs
@@ -116,9 +116,7 @@ all-features = true
 no-default-features = true
 default-target = "x86_64-unknown-linux-gnu"
 rustc-args = [ "--example-rustc-arg" ]
-rustdoc-args = [ "--example-rustdoc-arg" ]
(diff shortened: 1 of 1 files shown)


templates/about.hbs
```diff
@@ -117,6 +117,7 @@ no-default-features = true
 default-target = "x86_64-unknown-linux-gnu"
 rustc-args = [ "--example-rustc-arg" ]
 rustdoc-args = [ "--example-rustdoc-arg" ]
+# the dependencies key is not being used for builds yet
```
QuietMisdreavus: Minor phrasing nit, just personal preference:

```suggestion
# the dependencies key is currently unused, but is being saved for future support
```

decathorpe:
I'm not a native speaker, so suggestions for improvements are welcome :)

onur:
I wasn't planning to add/show this feature until we start using it, and we will start using it soon.

decathorpe:
> I wasn't planning to add/show this feature until we start using it, and we will start using it soon.

But it *IS* already documented here, which is leading to confusion: https://docs.rs/about :thinking:

QuietMisdreavus:
> I wasn't planning to add/show this feature until we start using it, and we will start using it soon.

Would you rather we remove the entry entirely from these docs? As @decathorpe mentioned, it is currently being shown on the "About" page.

onur:
Yes looks like its been added with this #196, I didn't realize dependencies until now sorry :nauseated_face: 

It's better if we remove that section until we start using it.

QuietMisdreavus:
Revisiting this PR. I agree with https://github.com/rust-lang/docs.rs/pull/318#issuecomment-478045021 - it's best to remove it from documentation entirely until it becomes useful. Can you update the file to remove that entry from the page?

QuietMisdreavus:
And it's live! https://docs.rs/about

decathorpe:
Thanks! :smile:

Issue #221, opened 2018-07-14
docs.rs/futures redirects to yanked version


carllerche:
Going to https://docs.rs/futures redirects to 0.2.1, however all 0.2 versions have been yanked. 0.1.23 is the latest release. See http://crates.io/crates/futures.

Issue #112, opened 2017-03-24
Hide yanked crate versions.


dpc:
Recently crates.io added quite a bit of fixes around yanked crates. Eg. when the latest version were yanked, the default one will be first version that is not yanked.

Also, generally skipping yanked versions in UI elements would be great. They can still be displayed if explicitly picked through a URL though.

crumblingstatue:
I agree that docs.rs should at least warn the user if they are looking at the docs for a yanked crate.

I often don't even look at crates.io anymore, and just copy the version straight from docs.rs.

Except I didn't realize that the version I copied has been yanked.

cramertj:
Relatedly, the `https://docs.rs/<crate_name>` link should probably go to the latest non-yanked version. For example,  https://docs.rs/futures  currently points to `futures` 0.2.1, which was yanked.

cc @onur

PR #322, opened 2019-03-27
start tracking when crates are yanked

QuietMisdreavus:
There's currently a column in our database to see whether a release has been yanked, but we don't set it when tracking updates. This PR changes that by adding:

* A change to the build queue that handles yank events by setting the release as yanked,
* A change to the "crate details" screen that adds a banner when the release is marked as yanked in our database, and
* A change to the version resolution code to look up the list of releases of the crate that are marked as yanked in our database, and avoid resolving to yanked releases unless explicitly asked for.
  * (Side note: if all of a crate's versions have been yanked, this will make it appear to not exist if you go to `docs.rs/crate`. Is this behavior we want? I can make it check whether all versions have been yanked before filtering them out, if desired.)

I also added a test to ensure that the templates are properly formatted, since i messed that up initially. `>_>`

Fixes https://github.com/rust-lang/docs.rs/issues/112, Fixes https://github.com/rust-lang/docs.rs/issues/221

### DIFF
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -18,11 +18,23 @@ impl DocBuilder {
         // I belive this will fix ordering of queue if we get more than one crate from changes
         changes.reverse();
 
-        for krate in changes.iter().filter(|k| k.kind != ChangeKind::Yanked) {
-            conn.execute("INSERT INTO queue (name, version) VALUES ($1, $2)",
-                         &[&krate.name, &krate.version])
-                .ok();
-            debug!("{}-{} added into build queue", krate.name, krate.version);
+        for krate in &changes {
+            match krate.kind {
+                ChangeKind::Yanked => {
+                    // FIXME: remove built doc files? set build as failed?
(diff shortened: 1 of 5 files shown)


QuietMisdreavus:
It's worth noting that this will only track yanks that occur after this is merged. I'll look into ways to backfill all the yank events into our database - i bet we can do it from our local clone of the crates.io index, but if that turns out to be infeasible i can ask the Crates.io team for guidance.

QuietMisdreavus:
After a quick chat on Discord, apparently it's easy enough for the crates.io maintainers to generate a CSV of currently-yanked crates, so whenever this is deployed we can ask for that to backfill yank information.

jyn514:
From my reading, this does 3 different things:

1. Update the `yanked` field in the DB when we get a yanked change from `crates_index_diff`.
2. Feed that `yanked` field throughout the rest of the code and show a warning on the doc page if it's set.
3. Make the `handlebars` templates testable.

I would love to have all of these changes, but the PR needs to be rebased. I think @Nemo157 has expressed interest in working on this.

PR #324, opened 2019-03-28
Add some keyboard interactions with the website first pages

GuillaumeGomez:
It'd be nice to navigate using the keyboard only on the first pages (on the crates pages, I prefer to not touch them by adding new shortcuts for the moment).

If someone could double-check too, it'd be very appreciated. :)

### DIFF
--- a/templates/releases.hbs
+++ b/templates/releases.hbs
@@ -1,11 +1,20 @@
 {{> header}}
 
+<script type="text/javascript" charset="utf-8">
+  function getKey(ev) {
+    if ("key" in ev && typeof ev.key != "undefined") {
+       return ev.key;
+    }
+    return String.fromCharCode(ev.charCode || ev.keyCode);
+  }
+</script>
+
 {{#if varsb.show_search_form}}
 <div class="container landing">
   <h1 class="brand"><i class="fa fa-cubes fa-fw"></i> Docs.rs</h1>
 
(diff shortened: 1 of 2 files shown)


templates/releases.hbs
```diff
@@ -31,6 +54,59 @@
     <div class="release">
       <a href="/releases"><strong>Recent Releases</strong></a> <a href="/releases/feed" title="Atom feed"><i class="fa fa-rss-square"></i></a>
     </div>
+    {{else}}
```
QuietMisdreavus: Why is the JS that registers the key handlers in the `else` branch of the template? Wouldn't that make it not appear on the front page?

GuillaumeGomez: That's the whole point! I want to be able to navigate through my search results, not on the front page.

QuietMisdreavus: But doesn't this also register the "S" shortcut?

GuillaumeGomez: No, the "S" shortcut is "registered" through `handleShortcut` whereas the up and down arrows are "registered" through `handleKey`.

QuietMisdreavus: Aha, i see it now. I was confused because this also registered "S", but that's probably because the handler overwrites `document.onkeypress`/`document.onkeydown`. This looks fine, then.

Issue #232, opened 2018-09-03
docs.rs/cratename redirect to version with failed build gives "The requested resource does not exist" and no links


huonw:
The short-links like https://docs.rs/im seem to do a redirect to the latest published version of the crate (which is 12.0.0 for that crate at the time of filing, i.e. https://docs.rs/im/12.0.0/im/). If that version failed to build, then the page is rather useless: just "The requested resource does not exist". It would be nice to give more information, such as a link to the crate page (e.g. https://docs.rs/crate/im/12.0.0) which at least has info about build failures, and/or a link to the latest successful build.

(A more general form of this issue is arbitrary URLs that are invalid but still "obviously" referring to a specific crate, e.g. https://docs.rs/crate/im/100.0.0. These could also contain a link back to the crate they refer to.)

dpc:
This is really bad user experience. Many of places all over the web have links like https://docs.rs/mycrate and then if the get "The requested resource does not exist" it's a big WAT. There are not even any pointers how to find last version that did work.

eddyb:
I stumbled onto this and discussed it with @QuietMisdreavus - I would prefer a redirect from a version that failed to build, to the `crate/` page.
For the versionless link, I think it's fine for it to go to the last-known-good version, and the "go to the latest version" link would trigger the aforementioned redirect.

Issue #226, opened 2018-07-28
How to jump to last version where docs built successfully?


Boscop:
A lot of crates' docs don't build successfully but there are earlier versions that bbuilt successfully.
E.g.:
[failed] https://docs.rs/actix-web/0.7.2/actix_web/
[works] https://docs.rs/actix-web/0.6.4/actix_web/ (but is it the latest working version of docs? you'd have to click through the versions to find out)

[failed] https://docs.rs/yew/0.4.0/yew/
[works] https://docs.rs/yew/0.2.0/yew/

I don't see a way to jump to the latest working docs. There should be a button at the top, when the docs failed to build.

Boscop:
The "Go to latest version" could still go to the docs of the most recent version, even if they failed to build.
There, it will show the button to "jump to latest version whose docs built successfully".

jyn514:
This was added in https://github.com/rust-lang/docs.rs/pull/543 :) You can see it in action on e.g. https://docs.rs/crate/actix-web/0.7.2.

PR #325, opened 2019-03-29
redirect from docs.rs/crate-name to crate details if doc build failed

QuietMisdreavus:
Fixes https://github.com/rust-lang/docs.rs/issues/232

This PR adds a check to see whether the version being loaded by the crate-docs redirect has docs available. If the `rustdoc_status` indicates that there are no docs available (whether because the crate is a binary, or the last attempted build failed), then instead of redirecting to the docs page (and immediately hitting an empty 404 page), it redirects to the crate-info page instead.

Currently this doesn't help the question "what is the last version where the docs actually worked" (https://github.com/rust-lang/docs.rs/issues/226), but this should be a decent quality-of-life improvement nonetheless.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -87,6 +87,23 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
         Ok(resp)
     }
 
+    fn redirect_to_crate(req: &Request,
+                         name: &str,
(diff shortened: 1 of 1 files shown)


dpc:
Thank you @QuietMisdreavus ! Much appreciated. I had at least two people that I tried to convert, hitting this issue.

PR #326, opened 2019-04-05
Check if status and subject in badge generator is not empty

bnjjj:
After using this crate in my project, it seems that if we don't check that status or subject is not empty we encounter a panic due to line https://github.com/rust-lang/docs.rs/blob/master/src/web/badge/badge.rs#L130 because we make a cast to u32 and so we return a negative number which is not possible when we use u32


bnjjj:
up please :(

QuietMisdreavus:
Should be live now!

bnjjj:
Thanks a lot :)

PR #328, opened 2019-04-11
fix keyboard nav on releases pages

QuietMisdreavus:
After deploying rust-lang/docs.rs#324 and testing it out, it looks like @GuillaumeGomez got the arrow keys backwards. This switches them around and makes it work properly.

### DIFF
--- a/templates/releases.hbs
+++ b/templates/releases.hbs
@@ -63,6 +63,7 @@
         return;
       }
       if (ev.which === 40) { // Down arrow
+        ev.preventDefault();
         if (active === null) {
           active = document.getElementsByClassName("recent-releases-container")[0].getElementsByTagName("li")[0];
         } else if (active.nextElementSibling) {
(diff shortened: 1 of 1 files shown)


templates/releases.hbs
```diff
@@ -62,15 +62,15 @@
       if (ev.ctrlKey || ev.altKey || ev.metaKey || document.activeElement.tagName === "INPUT") {
         return;
       }
-      if (ev.which === 38) { // Down arrow
+      if (ev.which === 40) { // Down arrow
         if (active === null) {
```
GuillaumeGomez: You need to add `ev.preventDefault()`.

templates/releases.hbs
```diff
@@ -62,15 +62,15 @@
...
           active = document.getElementsByClassName("recent-releases-container")[0].getElementsByTagName("li")[0];
         } else if (active.nextElementSibling) {
           active.classList.remove("selected");
           active = active.nextElementSibling;
         }
         active.classList.add("selected");
-      } else if (ev.which === 40) { // Up arrow
+      } else if (ev.which === 38) { // Up arrow
         if (active === null) {
```
GuillaumeGomez: You need to add `ev.preventDefault()`.

QuietMisdreavus:
Tested on dev server and it works, merging and deploying.

PR #329, opened 2019-04-12
serve all in-crate js files

QuietMisdreavus:
Starting with rust-lang/rust#59776, the search index, aliases, and source file list are getting the resource suffix added to their file names. This means we can't route them from a static path any more.

Since we always check the static file router first, this will still allow serving `main.js` for rustdoc versions prior to the addition of `--static-root-path`, and also allows rustdoc to add per-crate javascript files without requiring docs.rs to update to handle it.

(It also allows us to start hosting things *other* than rustdoc output as docs - e.g. mdbook output - in the future without changing the routing table massively.)

**WARNING**: Do not merge/deploy until rust-lang/rust#59776 is in nightly rustdoc. Deploy with `update-both-rustc-docsrs` script to ensure that docs.rs doesn't use the wrong rustdoc to build/serve docs!

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -170,9 +170,6 @@ impl CratesfyiHandler {
         router.get("/:crate/:version/",
                    rustdoc::rustdoc_redirector_handler,
                    "crate_version_");
-        router.get("/:crate/:version/*.js",
-                   rustdoc::rustdoc_redirector_handler,
-                   "crate_version_js");
         router.get("/:crate/:version/settings.html",
                    rustdoc::rustdoc_html_server_handler,
                    "crate_version_settings_html");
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
I had a chance to test this on my dev server and it unfortunately does not work. I'll probably need to change the redirection logic to make the new search index work.

(...also we missed `sidebar-items.js` :sweat:)

QuietMisdreavus:
I had a chance to look at it again, and it seems like the `/:crate/:version/*.js` route i added wasn't getting hit at all - it seemed to be hitting the `/:crate/:version/:target` one first, which made it redirect those files to the crate root docs instead. In the last commit i pushed, i took advantage of that to just inspect the file from that handler and defer to the proper handler if it was hit with a JS file. (This could be extended in the future for CSS files, if necessary, but for now i just kept it to JS files.)

PR #330, opened 2019-04-12
strip out references to rustdoc's custom favicon

QuietMisdreavus:
Starting in https://github.com/rust-lang/rust/pull/57552, rustdoc started packaging in a default column logo and favicon. We started hosting the logo in https://github.com/rust-lang/docs.rs/pull/290, but ignored the favicon, since docs.rs has its own. However, since rustdoc includes a `<link rel="shortcut icon">` tag in the `<head>`, this overrides the behavior of using `/favicon.ico`, and creates a lot of extra 404s as people try to access a favicon that isn't there. This PR strips out that favicon reference, which puts a favicon back into documentation pages again.

(Note that this will also strip out custom favicons provided by crates, like the one used by [Rocket](https://docs.rs/rocket/0.4.0/rocket/). If desired, i could add in a different check that would only filter out ones rooted locally (path either begins with `/` or `..`) which would allow custom favicons set by crates to still work.)

### DIFF
--- a/src/utils/html.rs
+++ b/src/utils/html.rs
@@ -49,6 +49,27 @@ fn extract_from_rcdom(dom: &RcDom) -> Result<(Handle, Handle)> {
 
     let head = head.ok_or_else(|| err_msg("couldn't find <head> tag in rustdoc output"))?;
     let body = body.ok_or_else(|| err_msg("couldn't find <body> tag in rustdoc output"))?;
+
+    {
+        head.children.borrow_mut().retain(|node| {
+            match node.data {
+                NodeData::Element { ref name, ref attrs, .. } => {
+                    // `<link rel="shortcut icon">` is rustdoc's custom favicon, but we want to
+                    // use our own, so ditch it so browsers request the favicon at the domain root
+                    // instead
+                    !(name.local == *"link" &&
+                      attrs.borrow()
+                           .iter()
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Alongside this, wouldn't it be simpler to generate crates with your own favicon?

QuietMisdreavus:
We don't currently have the option for the favicon URL to be set via a CLI argument. If we did that, then we could replace this PR with one that passed `--html-favicon-url /favicon.ico` when generating docs, instead. However, in that case we would still have the current docs that wouldn't have been generated with that option, which would still 404 when trying to load the favicon.

QuietMisdreavus:
Another option is to change the web server to redirect all `*.ico` routes to the static `favicon.ico` path. That would also mean not having to change upstream rustdoc, nor would it require this PR. (Finding ways to not need this PR is helpful because i'm afraid it will increase resource usage for doc page loads. `>_>`)

GuillaumeGomez:
Well, that last solution sounds good! :) Meanwhile, I'll add an option to allow to pass a favicon through CLI if you want?

QuietMisdreavus:
I managed to get the `*.ico` redirect working, though maybe not super-cleanly. It's posted as https://github.com/rust-lang/docs.rs/pull/340.

Issue #213, opened 2018-06-27
Settings not working


callym:
Recently Rustdoc has started generating a Settings page, which works fine on the std website, but on docs.rs, never saves the settings.
The browser console gives this error:
`Uncaught SyntaxError: Unexpected token <`
But doesn't give any more information to track it down any further.

pravic:
Same here. It has been broken for a while, but it's really annoying.

To reproduce:

* Go to https://docs.rs/kay/0.2.11/kay/struct.ActorSystem.html (as one of the recently updated crates)
* Click on Settings: https://docs.rs/kay/0.2.11/settings.html
* Here we go, there is an error.

While rustdoc itself works fine (or may be it has been fixed): 

https://doc.rust-lang.org/nightly/std/iter/trait.Iterator.html
https://doc.rust-lang.org/nightly/settings.html

pravic:
No idea?

onur:
Maybe @GuillaumeGomez can help us with this.

GuillaumeGomez:
I'll take a look.

sunjay:
Any update on this? I ran into this today.

whatisaphone:
Workaround:

1. Teach yourself javascript in 21 days
2. Go to https://doc.rust-lang.org/settings.html
3. Configure things as you like
4. Make a note of what's in localStorage
5. Go to https://docs.rs/futures-preview/0.3.0-alpha.13/futures/
6. Copy the localStorage entries over one by one

GuillaumeGomez:
I didn't look yet. But I assume it comes from crates that have been generated with a previous rustdoc version.

> Teach yourself javascript in 21 days

That's a pretty bad joke  (if it's a joke).

Nemo157:
It looks like (at least for some instances of this) the settings page JS file is missing, e.g. https://docs.rs/futures/0.1.26/settings.html tries to load https://docs.rs/futures/0.1.26/settings-20190318-1.35.0-nightly-0f88167f8.js which just returns a redirect to the futures homepage. Trying an absolute URL like the rest of the loaded resources (in this case https://docs.rs/settings-20190318-1.35.0-nightly-0f88167f8.js) returns a 404 instead.

QuietMisdreavus:
[...]

PR #332, opened 2019-04-15
add settings.js to essential files

QuietMisdreavus:
Fixes https://github.com/rust-lang/docs.rs/issues/213, though all the existing `settings.js` files need to be loaded in separately. Can be deployed without updating rustdoc at the same time, since this file has been there for a while `>_>`

In https://github.com/rust-lang/rust/pull/49954, rustdoc added a settings menu, which added the separate JavaScript file `settings.js`. This is a static file just like the other JS/CSS files bundled with rustdoc, but docs.rs never picked it up as an essential file.

A separate change will need to be made to rustdoc to use `--static-root-path` for this file, but this can at least let users begin using the settings page.

### DIFF
--- a/src/docbuilder/chroot_builder.rs
+++ b/src/docbuilder/chroot_builder.rs
@@ -383,6 +383,7 @@ impl DocBuilder {
                       "normalize.css",
                       "rustdoc.css",
                       "settings.css",
+                      "settings.js",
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
CI failure looks spurious.

PR #336, opened 2019-04-25
Set navigation bar of rustdoc fixed

lzutao:
Make the navigation bar on top of the page fixed.
image

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -117,6 +117,9 @@ div.nav-container {
     height: 32px;
     border-bottom: 1px solid $color-border;
     background-color: #fff;
+    left: 0;
+    right: 0;
+    position: fixed;
 
     li {
         border-left: 1px solid $color-border;
@@ -200,7 +203,7 @@ div.nav-container {
             div.right-border {
                 border-right: 1px solid $color-border;
             }
-            
+
             a.pure-menu-link {
                 word-wrap: normal;
                 white-space: normal;
@@ -223,7 +226,7 @@ div.nav-container {
 }
 
 div.nav-container-rustdoc {
-    position: absolute;
+    position: fixed;
     left: 0;
     right: 0;
     top: 0;


GuillaumeGomez:
A little screenshot and an explanation about what changed please? :)

lzutao:
@GuillaumeGomez Added.

GuillaumeGomez:
It is how it was at first. I assume there is a reason why we changed this behavior.

lzutao:
#186 changed it. But I don't see the rationale behind it.

GuillaumeGomez:
I didn't even remember doing it. XD I personnaly prefer with the top bar fixed like done in your PR, but let's wait for @QuietMisdreavus and @onur's opinion on the topic.

QuietMisdreavus:
Just to confirm, that screenshot is scrolled down a little? In other words, the top bar doesn't overlap the in-crate search box if you've scrolled all the way up on the page? I'm in favor of this, i just want to make sure it doesn't make the docs page more awkward to use.

GuillaumeGomez:
It's back to what it was. But it also means that the docs page must have a margin top now, otherwise anchors will be under the fixed top bar.

lzutao:
Yeah, I scrolled down that page a little bit then screenshotted it.

lzutao:
Any suggestions to this changes?

QuietMisdreavus:
I had a chance to test this a little bit more before deploying it, and it looks like it makes anything that's *not* a docs page shift up underneath the top bar:

image

image

image

I haven't deployed it yet, so i don't need to revert it immediately, but it's something i'd like to investigate before updating docs.rs on the server.

GuillaumeGomez:
I'll quote myself. 😃 

> It's back to what it was. But it also means that the docs page must have a margin top now, otherwise anchors will be under the fixed top bar.

PR #340, opened 2019-04-29
redirect all *.ico paths to /favicon.ico

QuietMisdreavus:
Closes https://github.com/rust-lang/docs.rs/pull/330

This is an alternate version of that PR that routes ICO files to the default favicon instead of modifying documentation pages on-the-fly to remove the reference to the nonexistent favicon.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -9,7 +9,7 @@ macro_rules! ctry {
     ($result:expr) => (match $result {
         Ok(v) => v,
         Err(e) => {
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
Coming back to this PR, i want to take this version over the one in https://github.com/rust-lang/docs.rs/pull/330 (or a potential future one to use a planned CLI flag in rustdoc). This is probably the least invasive and least complicated way to load the favicon in, even if it creates more redirects.

Issue #316, opened 2019-03-13
Search does not complete 


hellow554:
https://docs.rs/rocket/0.4.0/rocket/?search=foobar gives me: 
![grafik](https://user-images.githubusercontent.com/921462/54290652-a3b84f00-45ab-11e9-9f75-379ddefccef4.png)

https://docs.rs/regex/1.1.2/regex/?search=foobar gives me:
![grafik](https://user-images.githubusercontent.com/921462/54290691-b599f200-45ab-11e9-88f8-f302b2caf442.png)

QuietMisdreavus:
I can't reproduce this. Can you check that all the requested resources are loading on your end? Try reloading while bypassing cache (Ctrl-F5), since this fixes some issues on the official docs site.

![image](https://user-images.githubusercontent.com/5217170/54611493-0f277400-4a25-11e9-9a0c-6b22fc756e22.png)

hellow554:
I can reproduce it (maybe you should hit ctrl+f5 ;) )

Reproduceable in Firefox, chrome and edge under windows. Even with my mobile and fennec (firefox) and mobile data.

hellow554:
![Screenshot_Fennec_F-Droid_20190319-150907](https://user-images.githubusercontent.com/921462/54612338-0ea8d080-4a59-11e9-855b-7fb6c47bb4f5.png)

QuietMisdreavus:
I wonder if it's a CDN thing... I've rebuilt the rocket docs, try reloading again. If it still doesn't work, can you check to see if you can load <https://docs.rs/rocket/0.4.0/search-index.js>? I want to make sure that you're seeing all the required files.

hellow554:
Still the same issue. Have you seen, that the console gives me an TypeError? 
```
TypeError: paths is undefined main-20190318-1.35.0-nightly-0f88167f8.js:1:29247
```

Is there a way to get a non-minified version of `main.js`?

QuietMisdreavus:
That version of `main.js` is available [on the rust-lang/rust repo](https://github.com/rust-lang/rust/blob/0f88167f89fffe321590c5148f21b7d51d44388d/src/librustdoc/html/static/main.js). Chrome can also pretty-print it in the dev tools; i don't know about Firefox.

hellow554:
Thanks for the hint with the prettification! A french person has the same issue, a US person can see it just fine. It may really be a CDN issue here? Weird.
[...]

PR #342, opened 2019-04-30
always redirect version-catchall URLs to their latest version

QuietMisdreavus:
Fixes https://github.com/rust-lang/docs.rs/issues/316 (the iteration where semver-catchall URLs were getting updated after they were cached the first time)

One of docs.rs's powerful features is the ability to provide semver requirements in lieu of actual version numbers in many places. However, when combined with a CDN that likes to cache content at certain URLs, that can cause problems when not all URLs for a certain crate are cached at the same time. For example, if you load `docs.rs/tokio/0.1/tokio/index.html`, it's not certain whether you'll load the `index.html` from `0.1.0` or `0.1.19` (the current latest version in that series), since the CDN could have cached the page from an earlier version and hasn't updated the cache since then. (Or perhaps more confusing, it has updated the cache, but only in certain regions, and not everywhere has the right version yet, which happened on a different issue.)

This PR aims to solve the problem by never serving a URL where the "version" doesn't match the actual version number of the docs it's serving. Instead it aims to issue redirects to the URL with the exact version information. This way, the URL that actually serves the content is the one that's not expected to change over time.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -66,7 +66,8 @@ impl ToJson for RustdocPage {
 }
 
 
-
+/// Handler called for `/:crate` and `/:crate/:version` URLs. Automatically redirects to the docs
+/// or crate details page based on whether the given crate version was successfully built.
 pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
 
     fn redirect_to_doc(req: &Request,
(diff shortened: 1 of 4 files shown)


src/web/rustdoc.rs
```diff
@@ -260,7 +278,17 @@ pub fn badge_handler(req: &mut Request) -> IronResult<Response> {
                 }
             }
         }
-        None => {
+        MatchVersion::Semver(version) => {
+            let url = ctry!(Url::parse(&format!("{}://{}:{}/{}/badge.svg?version={}",
```
QuietMisdreavus: I'm not actually super sure if it's necessary to redirect the badge? Since the version is added by a query string, i'm not sure if it's actually cached in the first place. On the other hand, since the bare URL is meant to serve the latest version, there's a chance that the path may be cached already.

QuietMisdreavus: [Some cursory reading](https://stackoverflow.com/questions/9692665/cache-busting-via-params) tells me that caches treat the same query string as the same URL for the purposes of caching, so it's worth redirecting so that they load the proper badge. (Whether we want to alter the cache time on badges so that they're always up to date is another matter, though...)

PR #344, opened 2019-05-01
add "priority" to the build queue, and decouple builds from reading new crates

QuietMisdreavus:
This PR implements the first couple steps of https://github.com/rust-lang/docs.rs/issues/301, paving the way for manually batching builds while still allowing for new crate releases to be posted in a timely manner.

The largest change is the first commit, which breaks up the queue thread in the daemon into two: one which loads new crates from crates.io once a minute, and one which checks the build queue once a minute to start builds. This turned out to be a little more complicated than i initially imagined, because of the way the `DocBuilder` caches what has been built already. Juggling the caches made it a little more awkward than it otherwise would have been, but i think it turned out alright.

This alone allows docs.rs to avoid the confusing scenario where a long-running build job jams up the queue and makes it appear like the builder died (whereas now the queue will start accumulating entries and the builder will merely appear stuck, which is more reflective of reality). cc https://github.com/rust-lang/docs.rs/issues/335

The later commits are a logically-distinct change which is the core of https://github.com/rust-lang/docs.rs/issues/301 - the notion of a "priority" in the build queue, and a change in how builds are sorted. This allows us to queue up a batch of rebuilds without blocking new crate releases from being built.

Finally, for ease of maintenance, i added a `cratesfyi queue add` command, which allows administrators to add a crate to the build queue without having to manually enter a row into the database for it. (This means we can easily wire up a script for it, a la the current `build-from-psql-select`.) By default, crates entered through this command get priority 5 (which means they'll be sorted below the priority 0 that new crates get), though there is an available `--priority` flag you can use to override that. Since the column (and the types used in the code) are signed, it's *technically* possible to add queue entries above new crates, so some care is necessary. However, this is a much easier scheme than the current "lock the world so we can run builds outside the main log" system.

### DIFF
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -10,10 +10,11 @@ use crates_index_diff::{ChangeKind, Index};
 impl DocBuilder {
     /// Updates crates.io-index repository and adds new crates into build queue.
     /// Returns size of queue
-    pub fn get_new_crates(&mut self) -> Result<i64> {
+    pub fn get_new_crates(&mut self) -> Result<usize> {
         let conn = try!(connect_db());
         let index = try!(Index::from_path_or_cloned(&self.options.crates_io_index_path));
         let mut changes = try!(index.fetch_changes());
+        let mut add_count: usize = 0;
 
         // I belive this will fix ordering of queue if we get more than one crate from changes
         changes.reverse();
@@ -23,16 +24,19 @@ impl DocBuilder {
                          &[&krate.name, &krate.version])
                 .ok();
             debug!("{}-{} added into build queue", krate.name, krate.version);
+            add_count += 1;
         }
 
-        let queue_count = conn.query("SELECT COUNT(*) FROM queue WHERE attempt < 5", &[])
-            .unwrap()
-            .get(0)
-            .get(0);
-
-        Ok(queue_count)
+        Ok(add_count)
     }
 
+    pub fn get_queue_count(&self) -> Result<i64> {
+        let conn = try!(connect_db());
+        Ok(conn.query("SELECT COUNT(*) FROM queue WHERE attempt < 5", &[])
+            .unwrap()
+            .get(0)
+            .get(0))
+    }
 
     /// Builds packages from queue
     pub fn build_packages_queue(&mut self) -> Result<usize> {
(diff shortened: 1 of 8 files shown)


src/docbuilder/queue.rs
```diff
@@ -5,35 +5,38 @@ use super::DocBuilder;
...
             .get(0)
-            .get(0);
-
-        Ok(queue_count)
+            .get(0))
     }
 
-
     /// Builds packages from queue
     pub fn build_packages_queue(&mut self) -> Result<usize> {
```
QuietMisdreavus: I'm curious whether the old `build_packages_queue` is still necessary, TBH. It's currently not used outside of a unit test.

src/utils/daemon.rs
```diff
@@ -44,80 +45,159 @@ pub fn start_daemon() {
...
+        }
+    });
+
+    // build new crates every minute
+    thread::spawn(move || {
+        let opts = opts();
+        let mut doc_builder = DocBuilder::new(opts);
+
+        /// Represents the current state of the builder thread.
+        enum BuilderState {
```
QuietMisdreavus: When i initially wrote this enum, i had thought that the logic would be more complicated than it wound up being. This can probably be reduced to an `Option<usize>`, but i left this state enum in to make the usage code easier to read.

src/utils/daemon.rs
```diff
@@ -44,80 +45,159 @@ pub fn start_daemon() {
...
+                    Ok(n) => debug!("Succesfully pinged {} hubs", n)
                 }
-            };
 
-            // Only build crates if there is any
-            if queue_count == 0 {
-                debug!("Queue is empty, going back to sleep");
-                continue;
-            }
+                if let Err(e) = doc_builder.load_cache() {
```
QuietMisdreavus: This load/save dance is here in case the on-disk cache has been updated since the builder thread has last saved it. Since it's represented in-memory as a `BTreeSet`, reloading everything in doesn't create any duplicates in the cache when it's saved into disk later.

src/utils/daemon.rs
QuietMisdreavus: Update: I tried to reduce this to a `usize` and then to an `Option<usize>`, and proceeded to introduce bugs into the implementation. I think i'll keep this state enum as-is, since it provides a more readable implementation, IMO. (It'll take up the same space on the stack as an `Option<usize>` regardless.)

PR #345, opened 2019-05-02
Fix top navbar

GuillaumeGomez:


templates/style.scss
```diff
@@ -114,11 +116,14 @@ div.container-rustdoc {
 }
 
 div.nav-container {
+    // Nothing is supposed to be over or hovering the top navbar. Maybe add a few others '('? :)
+    z-index: 999999999;
```
QuietMisdreavus: This feels a little outlandish. A z-index of 1 should suffice, but you can also match `div.nav-container-rustdoc` by using 999.

QuietMisdreavus:
Context: this fixes the issue described in https://github.com/rust-lang/docs.rs/pull/336#issuecomment-488799226 that was introduced by that PR.

lzutao:
I'm sorry for not testing thoroughly in #336.

PR #346, opened 2019-05-03
only build crates for default target unless asked

QuietMisdreavus:
fixes https://github.com/rust-lang/docs.rs/issues/343

This PR makes it so that we don't automatically build crates for every target in our build container. See discussion in the linked issue for rationale.

I've also included an expansion of the documentation of the package metadata, since it wasn't very well explained before.

### DIFF
--- a/templates/about.hbs
+++ b/templates/about.hbs
@@ -101,22 +101,44 @@
     </tbody>
   </table>
 
-  <h4>Metadata for custom builds</h4>
+  <h4 id="metadata"><a href="#metadata">Metadata for custom builds</a></h4>
 
   <p>You can customize docs.rs builds by defining <code>[package.metadata.docs.rs]</code> table in your crates' `Cargo.toml`.</p>
 
-  <p>An example metadata:</p>
+  <p>The available configuration flags you can customize are:</p>
 
   <pre><code>[package]
 name = "test"
 
 [package.metadata.docs.rs]
+
+# Features to pass to Cargo (default: none)
 features = [ "feature1", "feature2" ]
+
+# Whether to pass `--all-features` to Cargo (default: false)
 all-features = true
+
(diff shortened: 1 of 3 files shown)


nagisa:
*How* to enable alternate targets should be made very visible. Not squirrelled somewhere in the annals of documentation, but as a link on the pop-out where the list of targets was originally. FWIW majority of my crates are cross-platform and expose platform-specific modules and this is the first time I hear about the docs.rs specific cargo.toml section.

templates/about.hbs
```diff
@@ -101,21 +101,44 @@
...
+# Features to pass to Cargo (default: none)
 features = [ "feature1", "feature2" ]
+
+# Whether to pass `--all-features` to Cargo (default: false)
 all-features = true
+
+# Whether to pass `--no-default-features` to Cargo (default: false)
 no-default-features = true
+
+# Target to test build on, used as the default landing page (default: "x86_64-unknown-linux-gnu")
```
dekellum: Thanks, this now looks very clear to me. Very helpful.

dekellum:
Regarding the docs discovery problem,  perhaps the [cargo reference, manifest format][https://doc.rust-lang.org/cargo/reference/manifest.html] could a special note and link to https://docs.rs/about#metadata-for-custom-builds (ideally with that anchor id working)?

jyn514:
What is this waiting on? I would like to see this to reduce build times.

pietroalbini:
Well, for sure this PR will have to be rewritten from scratch, as `chroot_builder` (thankfully) doesn't exist anymore.

jyn514:
Closing in favor of https://github.com/rust-lang/docs.rs/pull/532 (as mentioned above, this was obsoleted by Rustwide)

PR #358, opened 2019-05-20
Fix weird layout workflow issues on firefox

### LINKED ISSUE #356: Content is hidden by the top navbar
For instance, when scrolling through the page, when pressing 'S' to go back to the search bar, we only partially see it because it's hidden under the top navbar. We should add a top margin to the container to avoid this situation.

GuillaumeGomez:
Currently, on firefox, on a big enough page, when you scroll and press 'S', the page scrolls back to the search bar but it's still hidden under the top navbar. It's because firefox considers that we're scrolling the body and not the content, which is an issue fixed by this PR.

I need to check on chrome if it doesn't break anything and works as expected.

A little screenshot:

Screenshot from 2019-05-20 16-52-57

Fixes #356

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -19,7 +19,7 @@ $color-lifetime-incode: #B76514;   // orangish
 $color-comment-in-code: #8E908C;   // light gray
 $color-background-code: #F5F5F5;   // lighter gray
 $color-border: #ddd;               // gray
-
+$top-navbar-height: 32px;          // height of the floating top navbar
 
 
 // pure compatible media queries
@@ -58,7 +58,7 @@ div.rustdoc {
 
     .sidebar {
         @media (min-width: 701px) {
-            padding-top: 32px;
+            margin-top: $top-navbar-height;
         }
 
         .block > ul > li {
@@ -67,7 +67,7 @@ div.rustdoc {
     }
 
     #source-sidebar {
-        top: 32px;
+        top: $top-navbar-height;
     }
 
     #sidebar-toggle {
@@ -76,14 +76,14 @@ div.rustdoc {
 
     @media (max-width: 700px) {
         .sidebar.mobile {
-            top: 32px;
+            top: $top-navbar-height;
 
             .sidebar-elems.show-it {
                 top: 77px;
             }
 
             #sidebar-filler {
-                top: 32px;
+                top: $top-navbar-height;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
I confirm that it works as expected on chrome as well!

GuillaumeGomez:
cc @QuietMisdreavus  @onur

GuillaumeGomez:
Thanks to @onur, I was able to see that the right scrollbar wasn't where it was supposed to be. The second commit fixes this issue.

onur:
@GuillaumeGomez it's really a nasty hack and unfortunately didn't work. Here is the result for different screen resolutions, click them to see in full resolution:

Screenshot_2019-06-06 rand - Rust
Screenshot_2019-06-06 rand - Rust(1)
Screenshot_2019-06-06 rand - Rust(2)
Screenshot_2019-06-06 rand - Rust(3)

GuillaumeGomez:
Didn't think that the resolution would change that... I need to investigate further. Thanks again for the bug reports! :3

GuillaumeGomez:
I found an even easier way to fix the problem: removing the limit on the top element to let sub elements handle it themselves. :)

GuillaumeGomez:
ping @onur

templates/style.scss
```diff
@@ -141,14 +143,13 @@ div.container {
 }
 
 div.container-rustdoc {
-    max-width: 1200px;
```
QuietMisdreavus: What does removing this rule do?

QuietMisdreavus: After talking with Guillaume on Discord, it seems like this was the thing causing the bug Onur had seen, with the scrollbar appearing next to the content rather than on the edge of the viewpoint. However, removing this also causes the "crate search" box in the docs.rs navbar to move all the way to the side:

image

Maybe we can move this rule to `div.nav-container-rustdoc` instead, rather than deleting it entirely? The value itself may need to be changed, since the rustdoc content is a little wider than 1200px.

templates/style.scss
```diff
@@ -140,15 +142,15 @@ div.container {
     text-align: left;
 }
 
-div.container-rustdoc {
+div.nav-container-rustdoc {
```
QuietMisdreavus: When i said "move this rule to `div.nav-container-rustdoc`, i meant to use the existing block on line 265. By renaming the block like this, the div loses its `text-align: left` rule:

image

GuillaumeGomez: If I did that, the width of the navbar would be maximum 1200px, which wouldn't be nice. I changed something else instead. :)

GuillaumeGomez:
It seemed to have fix the issue (at least locally). I'll merge it for now but don't hesitate to come back to me if any other issue appears!

PR #365, opened 2019-06-04
Fix http redirects

### LINKED ISSUE #363: badge links unnecessarily redirect to an HTTP endpoint
Fetching badge links (such as e.g. https://docs.rs/py-comp/badge.svg) requires going through two redirects:

```
$ wget https://docs.rs/py-comp/badge.svg
--2019-06-03 23:14:30--  https://docs.rs/py-comp/badge.svg
Resolving docs.rs (docs.rs)... 52.222.167.82, 52.222.167.16, 52.222.167.22, ...
Connecting to docs.rs (docs.rs)|52.222.167.82|:443... connected.
HTTP request sent, awaiting response... 302 Moved Temporarily
Location: http://docs.rs/py-comp/badge.svg?version=0.1.3 [following]
--2019-06-03 23:14:36--  http://docs.rs/py-comp/badge.svg?version=0.1.3
Connecting to docs.rs (docs.rs)|52.222.167.82|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://docs.rs/py-comp/badge.svg?version=0.1.3 [following]
--2019-06-03 23:14:37--  https://docs.rs/py-comp/badge.svg?version=0.1.3
Connecting to docs.rs (docs.rs)|52.222.167.82|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 884 [image/svg+xml]
Saving to: ‘badge.svg’
```

In trying to fetch the badge, three HTTP calls are made:

1. https://docs.rs/py-comp/badge.svg --> 302 Moved Temporarily
2. http://docs.rs/py-comp/badge.svg?version=0.1.3 --> 301 Moved Permanently
3. https://docs.rs/py-comp/badge.svg?version=0.1.3 --> 200 OK

Besides the fact that this is unnecessary (because the first endpoint can just return the badge for the latest version by default), it messes with downstream uses.
Specifically, VSCode don't follow HTTP links when trying to fetch badges for display in markdown previews when "strict" security is enabled. This means that it stops following links at (2) and refuses to display the badge.

See https://github.com/microsoft/vscode/issues/73531 for reference.

pietroalbini:
All the docs.rs redirects are absolute, including the protocol and the host name. This is fine if docs.rs is exposed directly on the Internet, but the production instance sits behind CloudFront which terminates TLS connections. This means the application thinks it's serving HTTP requests instead of HTTPS ones, and it issues wrong redirects.

This PR fixes the issue by taking the Cloudfront-Forwarded-Proto header into account when building the redirect URL. This also adds `dotenv` to the project. It's not *necessary* but it's handy when you want to set environment variables and you aren't in the vagrant VM.

Fixes #363, and thanks @HeroicKatora for investigating this!

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -466,6 +466,24 @@ fn redirect(url: Url) -> Response {
     resp
 }
 
+pub fn redirect_base(req: &Request) -> String {
+    // Try to get the scheme from CloudFront first, and then from iron
+    let scheme = req.headers
+        .get_raw("cloudfront-forwarded-proto")
+        .and_then(|values| values.get(0))
+        .and_then(|value| std::str::from_utf8(value).ok())
+        .filter(|proto| *proto == "http" || *proto == "https")
+        .unwrap_or_else(|| req.url.scheme());
+
+    // Only include the port if it's needed
+    let port = req.url.port();
+    if port == 80 {
+        format!("{}://{}", scheme, req.url.host())
+    } else {
+        format!("{}://{}:{}", scheme, req.url.host(), port)
+    }
+}
+
 fn style_css_handler(_: &mut Request) -> IronResult<Response> {
     let mut response = Response::with((status::Ok, STYLE_CSS));
     let cache = vec![CacheDirective::Public,
@@ -495,10 +513,7 @@ fn ico_handler(req: &mut Request) -> IronResult<Response> {
     } else {
         // if we're looking for something like "favicon-20190317-1.35.0-nightly-c82834e2b.ico",
         // redirect to the plain one so that the above branch can trigger with the correct filename
-        let url = ctry!(Url::parse(&format!("{}://{}:{}/favicon.ico",
-                                            req.url.scheme(),
-                                            req.url.host(),
-                                            req.url.port())[..]));
+        let url = ctry!(Url::parse(&format!("{}/favicon.ico", redirect_base(req))[..]));
 
         Ok(redirect(url))
     }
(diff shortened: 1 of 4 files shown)


onur:
Thats great, thanks @pietroalbini and @HeroicKatora. We have been aware of the issue since we started using CloudFront but we weren't able to provide a solution for it. This is so useful.

onur:
like I said in the discord; I am only skeptical about dotenv, we are already doing same thing with systemd, it's reading our environment variables from (~/.cratesfyi.env) and passing it to our process. So I don't think we need dotenv.

pietroalbini:
Moved the dotenv commit in its own PR: #366

I still think it's useful but we don't need to block this PR on it.

onur:
Awesome, thanks again!

src/web/mod.rs
```diff
@@ -466,6 +466,24 @@ fn redirect(url: Url) -> Response {
...
+    let scheme = req.headers
+        .get_raw("cloudfront-forwarded-proto")
+        .and_then(|values| values.get(0))
+        .and_then(|value| std::str::from_utf8(value).ok())
+        .filter(|proto| *proto == "http" || *proto == "https")
+        .unwrap_or_else(|| req.url.scheme());
+
+    // Only include the port if it's needed
+    let port = req.url.port();
+    if port == 80 {
```
QuietMisdreavus: Shouldn't this be 80 or 443?

pietroalbini: Uh, probably yes, but it shouldn't cause any issue with our production setup, so fixing it is not that high priority.

onur: I think only 80 is fine since CloudFront is sending requests to 80.

PR #366, opened 2019-06-04
Add dotenv for local development

pietroalbini:
dotenv allows configuring environment variables loaded by the program in a text file, removing the need to set them manually every time or to add them to a bashrc (or similar). It has no side effects if the .env file is not present.

The use case for it is local development, not production.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 *.css.map
 .sass-cache
 .vagrant
+.env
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
I'm generally in favor. We already use environment files in production, but they're loaded via systemd. This allows someone to run the project outside of systemd, [without using a weird copy/paste in the bash profile to set the environment for all their sessions](https://github.com/rust-lang/docs.rs/wiki/Self-hosting-outside-the-Vagrant-VM#environment-for-the-cratesfyi-user). I've been considering putting this configuration in a file of some kind, and this is probably easier than manually parsing some TOML file or the like.

pietroalbini:
We're going to switch to docker-compose instead of vagrant, so this won't be needed anymore.

PR #371, opened 2019-06-25
add option to wait for build queue to see lock file

QuietMisdreavus:
When performing maintenance on the server, it can be useful to know that the build queue has stopped without manually inspecting the logs to wait for the "lock file exists" message. This PR introduces a `cratesfyi build lock --wait` command, which adds a second lock file to the prefix directory and blocks until it is deleted. The builder thread in the daemon will then see this file and delete it whenever it finishes an iteration and sees the main lock file.

This is admittedly very crude, but in lieu of spinning up a whole new message queue service or a unix socket or some other means of communicating between processes, this should suffice. Mainly i want a way to change our "update rustc" scripts so i don't have to `tail -f` the log file every time.

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -94,8 +94,11 @@ pub fn main() {
                     .help("Version of crate")))
             .subcommand(SubCommand::with_name("add-essential-files")
                 .about("Adds essential files for rustc"))
-            .subcommand(SubCommand::with_name("lock").about("Locks cratesfyi daemon to stop \
-                                                              building new crates"))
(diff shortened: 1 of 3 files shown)


jyn514:
Now that we're no longer updating rustc with shell scripts, I think this is a little less pressing. @QuietMisdreavus do you think this is still useful to have? I don't think we're currently locking the build queue very often.

jyn514:
Closing since this doesn't seem very pressing (and I'm not sure I like the approach of an external lock file). Feel free to reopen if you run into this again.

PR #372, opened 2019-06-25
remove duplicate imports

QuietMisdreavus:
Newer versions of Rust have introduced a new warning that tells you when you import something that was already imported in a higher scope. This started triggering on some code in docs.rs, so this PR is here to clean those warnings up.


Mark-Simulacrum:
This was essentially merged as part of https://github.com/rust-lang/docs.rs/pull/373 so closing.

PR #373, opened 2019-07-04
Utilize S3 as the backing file store

Mark-Simulacrum:
I've included the relevant access keys on the docs.rs server already, so this can be freely deployed. It should not cause any regressions, other than possibly in performance, as all reads fall back to the database. We'll want to work on a script that moves the existing rows in the database to S3, but that can be done later (and possibly not in this repository at all).

r? @QuietMisdreavus

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -8,7 +8,7 @@
 use std::path::{PathBuf, Path};
 use postgres::Connection;
 use rustc_serialize::json::{Json, ToJson};
-use std::fs::File;
+use std::fs;
 use std::io::Read;
 use error::Result;
 use failure::err_msg;
@@ -52,6 +52,32 @@ pub fn get_file_list<P: AsRef<Path>>(path: P) -> Result<Vec<PathBuf>> {
     Ok(files)
 }
 
+pub struct Blob {
+    pub path: String,
+    pub mime: String,
+    pub date_added: time::Timespec,
+    pub date_updated: time::Timespec,
+    pub content: Vec<u8>,
+}
+
+pub fn get_path(conn: &Connection, path: &str) -> Option<Blob> {
+    let rows = conn.query("SELECT path, mime, date_added, date_updated, content
+                           FROM files
+                           WHERE path = $1", &[&path]).unwrap();
+
+    if rows.len() == 0 {
+        None
+    } else {
+        let row = rows.get(0);
+        Some(Blob {
+            path: row.get(0),
+            mime: row.get(1),
+            date_added: row.get(2),
+            date_updated: row.get(3),
+            content: row.get(4),
+        })
+    }
+}
 
 /// Adds files into database and returns list of files with their mime type in Json
 pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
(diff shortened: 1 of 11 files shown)


pietroalbini:
> We'll probably want some way to handle configuration of a local s3-like bucket, but I'm not sure what options exist in that space

[Minio](https://min.io/) is a mostly feature complete S3 clone, running it with Docker locally should be painless.

Mark-Simulacrum:
Latest commit implements support for setting the S3 endpoint. This makes it possible to use minio (I've tested).

Mark-Simulacrum:
I'm going to hold off on working on the upload script until this is merged; my current plan is that we can simply run a query that goes something like `select path, content from files where content != "in-s3" limit 100;` and upload the results from this query to S3; in the same transaction we would update these paths to have "in-s3" as the content. Once all paths are converted we can drop the content column, if desired.

At this point I think this PR is ready to go.

QuietMisdreavus:
Thanks for putting this together! This looks great, and i'm glad you were able to get this working.

I'm slightly against forcing the assumption that you'll either have the S3 connection or a minio server available. Even if we bundle it into the vagrant container for the dev build, i still want a way to be independent of the S3 protocol. The simplest way to do that is to keep the database fallback, and allow the use of S3/minio as an optional optimization.

To that end, what happens right now if there are no S3 credentials in the environment? It looks like the "read" code tries to run the rusoto code, which (i assume) returns an error, so it reads out the database value in the `unwrap_or` branch. If we require S3/minio in the future, would that be replaced with a straight `unwrap`? And it looks like the "write" code already uses a straight `unwrap`, so now everyone who has a current development environment is forced to somehow provision a minio instance or set up an S3 bucket to load files into.

Unless/until we reconfigure the virtual machine used with Vagrant to include a minio instance and the proper configuration to make the development instance saved to that, i can't accept this. At the very least, i would want to see directions written into the README so that people who want to work on docs.rs can set it up themselves. (I certainly know nothing about minio and vanishingly little about setting up an S3 bucket, and i would need this information to be able to keep working on docs.rs after this change.)

Mark-Simulacrum:
I can definitely provide some setup instructions for minio and try and get a patch for Vagrant put together though I don't have Vagrant installed currently (using qemu for docs.rs development instead).

With that said, I'll also implement the database fallback to ease transition.

Mark-Simulacrum:
Okay, installing minio and configuring it in vagrant without vagrant installed locally proved a little too ambitious for me, so I scaled back. We no longer require minio (or any S3-compatible client); all reads/writes to the file DB will continue to use it if AWS access keys are not present in the environment. Note, however, that if we detect AWS_ACCESS_KEY_ID in the environment we will error on writes to the DB, though not reads to maintain the ability to gradually migrate files to S3.

src/db/file.rs
```diff
@@ -5,62 +5,108 @@
...
+    let rows = conn.query("SELECT path, mime, date_updated, content
+                           FROM files
+                           WHERE path = $1", &[&path]).unwrap();
+
+    if rows.len() == 0 {
+        None
+    } else {
+        let client = s3_client();
+        let row = rows.get(0);
+        let content = client.and_then(|c| c.get_object(GetObjectRequest {
```
QuietMisdreavus: I wonder if we could use the `in-s3` hard-coded value to gate this check, so that we don't generate erroneous calls to S3.

Mark-Simulacrum: That's a great idea, it'll definitely help with the common case for now. I've pushed up a commit that does this.

PR #380, opened 2019-07-18
Use git command instead of git2 in build script

Mark-Simulacrum:
This doesn't remove git2 from our dependencies (cargo and
crates-index-diff both use it), but it does simplify the build script.

r? @QuietMisdreavus

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,7 +426,6 @@ dependencies = [
  "env_logger 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
- "git2 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "handlebars-iron 0.22.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "html5ever 0.22.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


onur:
This is great but I am really wondering why you needed to change this portion of code. When I wrote this I assumed git might not be available in the machine this is running and that is why I used git2 crate instead of git command (same crate also used in cargo and crates-index-diff like you said).

Mark-Simulacrum:
IMO, the new code is somewhat simpler and easier for me at least to understand. I also hope that we'll eventually get to a point where we don't need to depend on Cargo directly and possibly even crates-index-diff can go away for some other way of getting a "new crates list" that doesn't need a git2 dependency; basically, I'm just removing a dependency that I don't think is necessary. The chances of someone building docs.rs without git are very slim IMO.

jyn514:
@Mark-Simulacrum are you still interesting in making this change? It looks good to me, it just needs to be rebased.

>  The chances of someone building docs.rs without git are very slim IMO.

Actually it would have to be even more of an edge case: you got the .git repo somehow without having `git` installed.

I do want to think about making it possible to build docs.rs without a git repo, using something like https://docs.rs/git-testament would abstract that a bit and give reasonable version numbers even without `.git` in the same directory. That doesn't have to go in this PR, just something to think about.

Mark-Simulacrum:
I don't have time right now to rebase this, but I think it would be a good change to make.

Feel free to close this, or I'll try to get to it sometime this week (or maybe on the weekend).

jyn514:
Going to close this since I don't feel that strongly about the change and you haven't had time to pick it back up. Feel free to reopen if you find the time :)

PR #381, opened 2019-07-18
Fix uploads of large crates to S3 breaking the client

Mark-Simulacrum:
Currently, we reuse the client for all files that are uploaded per call
to the relevant code. Unfortunately, S3 will prevent connections from
being long-lived (it's unknown what the exact timeout is, but we hit it
relatively rarely). This code attempts to retry S3 uploads up to 3 times
while replacing the client each time.

r? @QuietMisdreavus

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -121,7 +121,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
     try!(cookie.load::<&str>(&[]));
 
     let trans = try!(conn.transaction());
-    let client = s3_client();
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
I ran `update queue set attempt = 100 where name = 'alt-stm32f30x-hal';` on the DB to temporarily stop building that crate, as we had been attempting to do so for the last 7-10 hours I believe and failing on the upload each time. We'll want to undo that with `update queue set attempt = 0 where name = 'alt-stm32f30x-hal';` after deploying this.

PR #383, opened 2019-07-18
Utilize multi-threaded tokio runtime for move-to-s3

Mark-Simulacrum:
The current thread runtime was previously used but there was no reason
to do so. This utilizes the default runtime.


Mark-Simulacrum:
I seem to recall getting a compilation error last time I tried this, but I must have changed something since then.

onur:
Thanks!

PR #385, opened 2019-07-22
Migration fixes

Mark-Simulacrum:
https://github.com/rust-lang/docs.rs/pull/381 will also need to be merged, and this is based on https://github.com/rust-lang/docs.rs/pull/384.

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -63,35 +63,53 @@ pub struct Blob {
 }
 
 pub fn get_path(conn: &Connection, path: &str) -> Option<Blob> {
-    let rows = conn.query("SELECT path, mime, date_updated, content
-                           FROM files
-                           WHERE path = $1", &[&path]).unwrap();
+    if let Some(client) = s3_client() {
+        let res = client.get_object(GetObjectRequest {
+            bucket: "rust-docs-rs".into(),
+            key: path.into(),
+            ..Default::default()
+        }).sync();
 
-    if rows.len() == 0 {
-        None
-    } else {
-        let row = rows.get(0);
-        let mut content = row.get(3);
-        if content == b"in-s3" {
-            let client = s3_client();
-            content = client.and_then(|c| c.get_object(GetObjectRequest {
-                bucket: "rust-docs-rs".into(),
-                key: path.into(),
-                ..Default::default()
-            }).sync().ok()).and_then(|r| r.body).map(|b| {
-                let mut b = b.into_blocking_read();
-                let mut content = Vec::new();
-                b.read_to_end(&mut content).unwrap();
-                content
-            }).unwrap();
+        let res = match res {
+            Ok(r) => r,
+            Err(err) => {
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
What was the motivation for removing the `static_files` handler? It's still useful for those who aren't using S3, who are just setting up (since the base docs.rs pages depend on an older version of the rustdoc stylesheets, which at the moment need to be downloaded from an existing host of docs.rs).

src/db/file.rs
```diff
@@ -63,35 +63,53 @@ pub struct Blob {
 }
 
 pub fn get_path(conn: &Connection, path: &str) -> Option<Blob> {
-    let rows = conn.query("SELECT path, mime, date_updated, content
-                           FROM files
-                           WHERE path = $1", &[&path]).unwrap();
+    if let Some(client) = s3_client() {
```
QuietMisdreavus: Do you think it would be useful to run the database query anyway, and instead of switching on `content = 'in-s3'`, switch on whether the row exists in the table? I know it makes no difference for the production service, but it feels more resilient for anyone who hosts their own version of it and transitions to S3 after being online for a while.

Mark-Simulacrum: I'd prefer not to, as it does add some amount of overhead and to an extent makes the code more complex. `move-to-s3` should be the preferred route for migrating local installations to S3.

Mark-Simulacrum:
I primarily removed the static files handler since independent of using S3 or the DB, it seems odd to store some files on disk and some on S3. There should be no reason to differentiate, IMO; in particular I always get annoyed when standing up a new instance as the new files need to basically be copied from production which isn't great. I can re-add it back and configure it on the staging server if you'd like, but I'd rather move us towards solely using S3.

src/db/file.rs
QuietMisdreavus: Right, i'm just thinking about the period between saving S3 credentials into the environment and the moment `cratesfyi database move-to-s3` finishes. Any calls to the website in that time will fail because we won't be looking at the database for the files.

QuietMisdreavus:
I would be more willing to ditch the `static_files` handler if i hadn't had trouble using `cratesfyi database add-directory` to manually add files to the database. The last time i tried (when i was trying to load `settings.js` files for previous versions of rustdoc) it exited successfully but didn't actually add the files to the database. I don't know if i just wasn't using it correctly, but in the absence of knowing that would work i would rather keep `static_files` around.

Mark-Simulacrum:
I've dropped the commit removing static_files. I realized that prod can just keep those files in S3, as before, since we check the database/S3 prior to looking at static files.

src/db/file.rs
Mark-Simulacrum: That's true. move-to-s3 is no longer an operation that can run in the background. Since it's likely to never be used again outside local installations though that seems fine -- and if in the future there is a desire to make that happen the code change isn't really that bad.

PR #390, opened 2019-07-30
Clear target dir

Mark-Simulacrum:
cc @pietroalbini 

r? @QuietMisdreavus

### DIFF
--- a/src/docbuilder/chroot_builder.rs
+++ b/src/docbuilder/chroot_builder.rs
@@ -8,12 +8,14 @@ use db::{connect_db, add_package_into_database, add_build_into_database, add_pat
 use cargo::core::Package;
 use cargo::util::CargoResultExt;
 use std::process::Command;
-use std::path::PathBuf;
-use std::fs::remove_dir_all;
+use std::path::{Path, PathBuf};
+use std::fs::{self, remove_dir_all};
 use postgres::Connection;
 use rustc_serialize::json::{Json, ToJson};
 use error::Result;
+use systemstat::{Platform, Filesystem, System};
 
+const MAX_DISK_USAGE: f32 = 80.0;
 
 /// List of targets supported by docs.rs
 const TARGETS: [&'static str; 6] = [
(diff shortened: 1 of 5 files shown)


src/docbuilder/chroot_builder.rs
```diff
@@ -237,6 +237,19 @@ impl DocBuilder {
                 .join("doc");
             let _ = remove_dir_all(crate_doc_path);
         }
+
+        if let Ok(fs) = System::new().mount_at(".") {
+            let disk_free = 100.0 - (fs.free.as_usize() as f32 / fs.total.as_usize() as f32);
+            if disk_free < 20.0 {
```
pietroalbini: Maybe extract this into a constant?

src/docbuilder/chroot_builder.rs
```diff
@@ -237,6 +237,19 @@ impl DocBuilder {
                 .join("doc");
             let _ = remove_dir_all(crate_doc_path);
         }
+
+        if let Ok(fs) = System::new().mount_at(".") {
```
pietroalbini: I remember having trouble with this on Crater when `.` wasn't a mount point and I had [to do this](https://github.com/rust-lang/crater/blob/c0587168029236cce79847ad0819b76049004963/src/runner/worker.rs#L203-L220). Can you double check everything works?

Mark-Simulacrum:
Review nits fixed and verified that this works locally (by setting allowed disk usage to 20% or so)

onur:
thanks @Mark-Simulacrum but I think 20% might not be enough. Some crates (stm32) generate more than 10 GB documentation (no idea about their build artifacts) and since our disk space is really limited now, maybe we can increase it to at least 30 or 40 percent.

Mark-Simulacrum:
We're currently sitting at 50 GB free on the server (out of 100 GB total). The way the logic is implemented, all 50 GB can be used by a single crate compilation (AFAICT), but before/after each crate build we will clear out the target directory if the usage exceeds 80GB. So, basically, this isn't making things any worse for an individual crate compilation.

onur:
Then its great :+1:

onur:
I am merging this before we ran out of space again :joy:

PR #391, opened 2019-08-01
allow authenticating to AWS with the EC2 instance role

pietroalbini:
This changes the credentials provider used to fetch the AWS credentials from [EnvironmentProvider](https://docs.rs/rusoto_credential/0.40.0/rusoto_credential/struct.EnvironmentProvider.html) (which just looked at environment variables) to [DefaultCredentialsProvider](https://docs.rs/rusoto_credential/0.40.0/rusoto_credential/struct.DefaultCredentialsProvider.html), which looks at:

1. Environment variables
2. `~/.aws/credentials`
3. EC2 instance roles

The old behavior is preserved when the environment variable is present, but this will also allow using EC2 instance roles which are going to be implemented on the production server.

A new FORCE_S3 environment variable was also added.

:warning: :warning: **I don't have a local environment running** so I haven't tested the changes, but the code compiles. Could anyone with a local environment test if everything is ok? :warning: :warning: 

r? @Mark-Simulacrum

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -14,7 +14,7 @@ use error::Result;
 use failure::err_msg;
 use rusoto_s3::{S3, PutObjectRequest, GetObjectRequest, S3Client};
 use rusoto_core::region::Region;
-use rusoto_credential::EnvironmentProvider;
+use rusoto_credential::DefaultCredentialsProvider;
 
 
 fn get_file_list_from_dir<P: AsRef<Path>>(path: P,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Code looks good. I'll probably not have time to test the changes locally for some time but I would loosely recommend that we just merge and deploy since the worst case is a few minutes of downtime as we rebuild after a revert on the server (we can test it without restarting docs.rs in full, too, if we wanted to).

QuietMisdreavus:
I mean, it should have the same behavior if you don't change the environment variables, right? It seems harmless to deploy in that case, especially if whoever pushes the button watches for issues afterward.

PR #392, opened 2019-08-02
update vagrant to use 2048MB ram on Windows

miller-time:
`cargo build` ran out of memory when it was set to 1024


jyn514:
This is obsoleted by #432, I don't think docker has an explicit memory limit.

miller-time:
even better, thanks! I'll have to try it out with `docker-compose` 😃

PR #393, opened 2019-08-02
Parallel s3 uploads

miller-time:


miller-time:
oops a commit snuck in from my other branch https://github.com/rust-lang/docs.rs/pull/392

src/db/file.rs
```diff
@@ -138,102 +138,96 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
-                bucket_path,
-                content,
-                mime,
-            )
-        };
-
-        // If AWS credentials are configured, don't insert/update the database
-        if client.is_none() {
+                    })
+                    .into_stream()
```
Mark-Simulacrum: This won't do what we want unfortunately. I think it should be fine to regress to just retrying once though so we can just skip the streaming here I think and just retry the PUT.

miller-time: Dang I was trying to get fancy and bundle all that behavior into a Future :)

miller-time:
>  during the main loop
can you point me to that?

Mark-Simulacrum:
Ah, the directory reading loop -- in that same function. Right after pushing onto the vector of futures we can check if we've exceeded some constant amount of futures and dispatch based on that.

src/db/file.rs
```diff
@@ -138,102 +138,97 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+                    key: bucket_path.clone(),
+                    body: Some(content.clone().into()),
+                    content_type: Some(mime.clone()),
+                    ..Default::default()
+                }).map_err(|_| {
+                    panic!("failed to upload twice, exiting");
+                })
+            }));
+
+            if futures.len() > 1000 {
```
miller-time: @Mark-Simulacrum something like this?

Mark-Simulacrum: Yep, this is almost perfect -- let's extract the constant into `const MAX_CONCURRENT_UPLOADS: usize = 1000;` at top of the module (after all the imports, though).

src/db/file.rs
```diff
@@ -138,102 +141,96 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+                    content_type: Some(mime),
+                    ..Default::default()
+                }).map_err(|_| {
+                    panic!("failed to upload twice, exiting");
+                })
+            }));
+
+            if futures.len() > MAX_CONCURRENT_UPLOADS {
+                if rt.block_on(::futures::future::join_all(futures)).is_err() {
+                    panic!("upload to s3 failed");
```
QuietMisdreavus: Should we include the error details in this panic message? I feel like that would make diagnosing issues much easier.

src/db/file.rs
```diff
@@ -138,102 +141,96 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
                                 WHERE path = $1",
-                                &[&path, &mime, &content]));
+                                &[&bucket_path, &mime, &content]));
             }
         }
+
+        file_list_with_mimes.push((mime.clone(), file_path.clone()));
+    }
+
+    if rt.block_on(::futures::future::join_all(futures)).is_err() {
```
QuietMisdreavus: Part of me is wary about calling `join_all` with an empty Vec (in the case of not having AWS credentials configured). I figure it's harmless, but should we consider wrapping this block with an `if !futures.is_empty()` block?

src/db/file.rs
```diff
@@ -138,102 +141,96 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+        panic!("upload to s3 failed");
```
QuietMisdreavus: If we're adding error information to the panic message, we should add it here as well.

src/db/file.rs
miller-time: what details would you like to see added?

Mark-Simulacrum: debug print of the error is probably enough -- instead of is_err we can do if let Err(e) = ... { panic!("upload to s3 failed: {:?}", e); } or so

src/db/file.rs
miller-time: yeah it looks like [`block_on`](https://docs.rs/tokio/0.1.22/src/tokio/runtime/threadpool/mod.rs.html#260-270) will unnecessarily create a channel, which might be good to prevent?

src/db/file.rs
miller-time: didn't mean to start a review...

```
help: the trait `std::fmt::Debug` is not implemented for `rusoto_core::RusotoFuture<rusoto_s3::PutObjectOutput, rusoto_s3::PutObjectError>`
```

I think if I can unwrap that `RusotoFuture` to get the `PutObjectError` I will be able to accomplish this. I'm not sure how to do that exactly.

Mark-Simulacrum: That seems a bit odd -- block_on shouldn't be returning a future I'd imagine? Maybe your parentheses are off or something?

miller-time: It might be a result:

https://github.com/rusoto/rusoto/blob/df96e8ce92f34248ec667f8514817158bb06ff01/rusoto/core/src/future.rs#L275

Mark-Simulacrum: Well in any case -- it sounds like this might be harder to do that @QuietMisdreavus expected so probably not worth it then. Most of the time AWS errors are really opaque anyway...

src/db/file.rs
Mark-Simulacrum: This I think would be good to implement.

src/db/file.rs
QuietMisdreavus: Hmm, now that i look at the future being inserted into the queue, it's already panicking if it's failing twice:

```rust
            futures.push(client.put_object(PutObjectRequest {
                bucket: "rust-docs-rs".into(),
                key: bucket_path.clone(),
                body: Some(content.clone().into()),
                content_type: Some(mime.clone()),
                ..Default::default()
            }).map_err(move |e| {
                log::error!("failed to upload to {}: {:?}", bucket_path, e); // logging the error after the first failure
                client.put_object(PutObjectRequest {
                    bucket: "rust-docs-rs".into(),
                    key: bucket_path,
                    body: Some(content.into()),
                    content_type: Some(mime),
                    ..Default::default()
                }).map_err(|_| {
                    panic!("failed to upload twice, exiting"); // already panicking after the second error
                })
            }));
```

I wonder if the nesting should be changed on that. As far as i remember combinator chaining working out, you can tag on the second `map_err` without introducing that rightward-drift:

```rust
            futures.push(client.put_object(PutObjectRequest {
                bucket: "rust-docs-rs".into(),
                key: bucket_path.clone(),
                body: Some(content.clone().into()),
                content_type: Some(mime.clone()),
                ..Default::default()
            }).map_err(move |e| {
                log::error!("failed to upload to {}: {:?}", bucket_path, e);
                client.put_object(PutObjectRequest {
                    bucket: "rust-docs-rs".into(),
                    key: bucket_path,
                    body: Some(content.into()),
                    content_type: Some(mime),
                    ..Default::default()
                })
            }).map_err(|e| {
                log::error!("failed to upload to {} a second time: {:?}", bucket_path, e);
                panic!("failed to upload twice, exiting");
            }));
```

Would something like this work?

miller-time: I'll give that a shot

miller-time: Nope, same issue...

```
help: the trait `std::fmt::Debug` is not implemented for `rusoto_core::RusotoFuture<rusoto_s3::PutObjectOutput, rusoto_s3::PutObjectError>`
```

QuietMisdreavus:
This looks great! One thing i noticed is that this doesn't include the "refresh the client" logic. Do we think that this will get around that issue? Should we create a new client when we execute a batch of uploads?

Mark-Simulacrum:
Oh, yeah, we probably want a new client for each batch -- that should be low enough overhead wise. I imagine that's not too difficult to add?

miller-time:
The way it was working before, I believe the client was only refreshed if an error occurred. It seems like now we call `s3_client()` for each file. So I think, if anything, we might be refreshing the client too much?

QuietMisdreavus:
> 
> 
> The way it was working before, I believe the client was only refreshed if an error occurred. It seems like now we call `s3_client()` for each file. So I think, if anything, we might be refreshing the client too much?

Oh, you're right! In this case, i think this will be fine. I don't think that creating a new client each time is that heavyweight, but i haven't looked that strongly into it. @Mark-Simulacrum, will this work, or do you want to bring back the old logic of "save the client outside the loop and overwrite it after a batch is sent"?

Mark-Simulacrum:
Oh, yeah -- I think a new client for each future is probably _not_ what we want (it seems likely to cause problems of some kind, or be generally slower) -- I'm not sure what a good way to structure this would be though. Maybe we can do a RwLock<Client> that we share between all the futures, and then replace it on error? If that proves pretty complicated, then we can likely get away with just creating a new client for every batch and re-running the batch if it was unsuccessful (so, kinda similar to before, but less fine-grained, I guess).

miller-time:
I feel like I'm getting pretty close...

```diff
diff --git src/db/file.rs src/db/file.rs
index faf66d2..2465141 100644
--- src/db/file.rs
+++ src/db/file.rs
@@ -10,6 +10,7 @@ use postgres::Connection;
 use rustc_serialize::json::{Json, ToJson};
 use std::fs;
 use std::io::Read;
+use std::sync::RwLock;
 use error::Result;
 use failure::err_msg;
 use rusoto_s3::{S3, PutObjectRequest, GetObjectRequest, S3Client};
@@ -148,6 +149,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
     try!(cookie.load::<&str>(&[]));
 
     let trans = try!(conn.transaction());
+    let client = s3_client().map(|c| RwLock::new(c));
     let mut file_list_with_mimes: Vec<(String, PathBuf)> = Vec::new();
 
     let mut rt = ::tokio::runtime::Runtime::new().unwrap();
@@ -188,11 +190,13 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
             }
         };
 
-        if let Some(client) = s3_client() {
+        if let Some(client) = client {
             let bucket_path = bucket_path.clone();
             let content = content.clone();
             let mime = mime.clone();
 
+            let mut client = client.write().unwrap();
+
             futures.push(client.put_object(PutObjectRequest {
                 bucket: "rust-docs-rs".into(),
                 key: bucket_path.clone(),
@@ -201,6 +205,11 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
                 ..Default::default()
             }).map_err(move |e| {
                 log::error!("failed to upload to {}: {:?}", bucket_path, e);
+                // Get a new client, in case the old one's connection is stale.
+                // AWS will kill our connection if it's alive for too long; this avoids
+                // that preventing us from building the crate entirely.
+                *client = s3_client().unwrap();
+
                 client.put_object(PutObjectRequest {
                     bucket: "rust-docs-rs".into(),
                     key: bucket_path,

```

...but

```
error[E0277]: `std::sync::RwLockWriteGuard<'_, rusoto_s3::S3Client>` cannot be sent between threads safely
   --> src/db/file.rs:225:23
    |
225 |                 if rt.block_on(::futures::future::join_all(futures)).is_err() {
    |                       ^^^^^^^^ `std::sync::RwLockWriteGuard<'_, rusoto_s3::S3Client>` cannot be sent between threads safely
    |
```

Also, acquiring a write lock for each file upload seems like a bad idea... **but** because of the closure with the "retry" in it, we're moving `client` during each iteration of the loop. Here's an attempt to only acquire the write lock for the "client refresh":

```diff
diff --git src/db/file.rs src/db/file.rs
index faf66d2..5f8c252 100644
--- src/db/file.rs
+++ src/db/file.rs
@@ -10,6 +10,7 @@ use postgres::Connection;
 use rustc_serialize::json::{Json, ToJson};
 use std::fs;
 use std::io::Read;
+use std::sync::RwLock;
 use error::Result;
 use failure::err_msg;
 use rusoto_s3::{S3, PutObjectRequest, GetObjectRequest, S3Client};
@@ -148,6 +149,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
     try!(cookie.load::<&str>(&[]));
 
     let trans = try!(conn.transaction());
+    let mut client = s3_client().map(|c| RwLock::new(c));
     let mut file_list_with_mimes: Vec<(String, PathBuf)> = Vec::new();
 
     let mut rt = ::tokio::runtime::Runtime::new().unwrap();
@@ -188,12 +190,12 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
             }
         };
 
-        if let Some(client) = s3_client() {
+        if let Some(client) = client {
             let bucket_path = bucket_path.clone();
             let content = content.clone();
             let mime = mime.clone();
 
-            futures.push(client.put_object(PutObjectRequest {
+            futures.push(client.read().unwrap().put_object(PutObjectRequest {
                 bucket: "rust-docs-rs".into(),
                 key: bucket_path.clone(),
                 body: Some(content.clone().into()),
@@ -201,7 +203,15 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
                 ..Default::default()
             }).map_err(move |e| {
                 log::error!("failed to upload to {}: {:?}", bucket_path, e);
-                client.put_object(PutObjectRequest {
+                // Get a new client, in case the old one's connection is stale.
+                // AWS will kill our connection if it's alive for too long; this avoids
+                // that preventing us from building the crate entirely.
+                {
+                    let mut c = client.write().unwrap();
+                    *c = s3_client().unwrap();
+                }
+
+                client.read().unwrap().put_object(PutObjectRequest {
                     bucket: "rust-docs-rs".into(),
                     key: bucket_path,
                     body: Some(content.into()),
```

As expected, the move prevents this from working:

```
error[E0382]: use of moved value
   --> src/db/file.rs:193:21
    |
193 |         if let Some(client) = client {
    |                     ^^^^^^ value moved here, in previous iteration of loop
    |
```

```
error[E0505]: cannot move out of `client` because it is borrowed
   --> src/db/file.rs:204:24
    |
198 |             futures.push(client.read().unwrap().put_object(PutObjectRequest {
    |                          ----------------------
    |                          |
    |                          borrow of `client` occurs here
    |                          a temporary with access to the borrow is created here ...
...
204 |             }).map_err(move |e| {
    |                        ^^^^^^^^ move out of `client` occurs here
...
210 |                     let mut c = client.write().unwrap();
    |                                 ------ move occurs due to use in closure
...
223 |             }));
    |                - ... and the borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `std::sync::RwLockReadGuard`
```

I'll keep poking at this, but any help or insight greatly appreciated! :grinning:

src/db/file.rs
```diff
@@ -201,6 +210,18 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
                 ..Default::default()
             }).map_err(move |e| {
                 log::error!("failed to upload to {}: {:?}", bucket_path, e);
+
+                {
+                    // Get a new client, in case the old one's connection is stale.
+                    // AWS will kill our connection if it's alive for too long; this avoids
+                    // that preventing us from building the crate entirely.
+                    let mut c = retry_lock.write().unwrap();
```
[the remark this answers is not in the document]
QuietMisdreavus: If that happens, how do we know which files to re-upload? Aren't the futures consumed in the call to `join_all`?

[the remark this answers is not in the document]
Mark-Simulacrum: We can reupload all the files; we might duplicate work but it shouldn't really matter (i.e., the files are the same).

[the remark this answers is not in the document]
miller-time: There are two places where we might want to re-create the client (two different `rt.block_on(::futures::future::join_all(futures))` calls):

* `if futures.len() > MAX_CONCURRENT_UPLOADS` - this is actually inside the `for file_path in ...` loop. after blocking on the initial futures, it clears the futures vec and continues with the next file (are you calling this the "completion of the inner loop"?)
* immediately following the `for file_path in ...` loop

were you talking about re-creating the client in both cases?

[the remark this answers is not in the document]
miller-time: when `MAX_CONCURRENT_UPLOADS` is reached, there's a partially-completed iterator that represents the files that we tried to upload (of which one or more failed). it's not immediately obvious to me how we would retry those uploads...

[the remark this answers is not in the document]
miller-time: Also, we currently have a retry for each file. It's unclear how we'd perform a reupload of all the files. It seems like we basically need to start over iterating through the file paths. Am I missing something?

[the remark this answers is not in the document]
Mark-Simulacrum: This looks like it doesn't match the code precisely -- but this is what I was envisioning in an "abstract" sense:

 * we have a list of all the file paths we need to upload (`to_upload`)
 * we remove MAX_CONCURRENT_UPLOADS elements off of this list (`currently_uploading`)
 * create a fresh client here (every time)
 * loop through these, creating a future for each upload
 * block on the completion of all these futures
   * if any fail, re-create the futures from the list `currently_uploading` (this may need an extra copy or so but that's fine) -- from the "create a fresh client" step basically
 * repeat the last several steps until we're done (i.e., `to_upload` is empty)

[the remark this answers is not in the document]
miller-time: That makes sense, I gave it a shot. It was a bit of a rewrite so hopefully it's correct :smile:

Mark-Simulacrum:
Looking at the diff is pretty horrible (indentation changed and things moved around) but based on a high-level read through of https://github.com/rust-lang/docs.rs/blob/d9beb9e6a5e85348811cee5f5884dc977b02b4c5/src/db/file.rs this looks good.

QuietMisdreavus:
This looks like it will retry forever if files fail to upload for reasons other than a stale connection, i.e. now there's no `panic!()` branch. Do we want to have a failure counter and panic on 2-3 failed upload batches?

Mark-Simulacrum:
Seems quite reasonable. We might want to also log the error somewhere (ideally in a way metrics can pick up). But that can be left for future work I think.

src/db/file.rs
```diff
@@ -182,64 +193,40 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+                                    &[&bucket_path, &mime, &content]));
+                } else {
+                    try!(trans.query("UPDATE files SET mime = $2, content = $3, date_updated = NOW() \
+                                    WHERE path = $1",
+                                    &[&bucket_path, &mime, &content]));
+                }
+            }
 
-            let content = content.expect("content never None if client is None");
+            file_list_with_mimes.push((mime.clone(), file_path.clone()));
```
miller-time: I think if this happens during a retry, it will duplicate items in the `file_list_with_mimes` vec.

This might need to be done only when the `put_object` is successful?

Something like

```
client.put_object(PutObjectRequest {
  // ...
}).and_then(|| {
  file_list_with_mimes.push((mime.clone(), file_path.clone()));
});
```

except I think now `file_list_with_mimes` needs to be in an `Arc`

Mark-Simulacrum: Might be simpler to deduplicate after uploading by file path.

miller-time: Or prevent duplicates during insert?

src/db/file.rs
```diff
@@ -145,11 +148,22 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
 
-    for file_path in try!(get_file_list(&path)) {
-        let (path, content, mime) = {
+    use std::collections::HashMap;
+    let mut file_paths_and_mimes: HashMap<PathBuf, String> = HashMap::new();
+
+    let mut rt = ::tokio::runtime::Runtime::new().unwrap();
+
+    let mut to_upload = try!(get_file_list(&path));
+    let mut currently_uploading: Vec<_> = to_upload.drain(..MAX_CONCURRENT_UPLOADS).collect();
```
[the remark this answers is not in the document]
miller-time: nice catch, I should have read the `drain` docs more closely :disappointed:

Mark-Simulacrum:
Yes, I'm happy with this overall (a few nits but not something we need to fix here, I might try to fix them up when I work on compression this weekend, presuming this lands).

Let's merge and deploy (cc @pietroalbini). I'd like to deploy this quickly after merging and monitor it for a few uploads so not doing so myself.

(Ideally, we'd collect metrics on upload time, and we'd know if this is actually an improvement).

miller-time:
squash first?

Mark-Simulacrum:
Oh, yes, it'd be good to squash this into one commit.

pietroalbini:
> Let's merge and deploy (cc @pietroalbini). I'd like to deploy this quickly after merging and monitor it for a few uploads so not doing so myself.
>
> (Ideally, we'd collect metrics on upload time, and we'd know if this is actually an improvement).

Should we add metrics for a day or two and then merge this PR then? Adding a new one for uploaded files shouldn't be hard.

miller-time:
Is that something I can help out with?

miller-time:
@pietroalbini does this look right to you?

```rs
crate::web::metrics::UPLOADED_FILES_TOTAL.inc_by(batch_size as i64);
```

pietroalbini:
@miller-time yeah, it's fine. It'd be great to increment by 1 each time a single future completes, but that's not critical and if you want it can be implemented in a later PR. I think merging the speedups first is more important than slightly more accurate metrics.

Huge thanks for the work you put into this! :tada:

Issue #396, opened 2019-08-14
"Go to latest version" should ignore yanked versions


Kinrany:
Clicking "Go to latest version" at https://docs.rs/futures/0.1.28/futures/ opens the docs for [version `0.2.3-docs-yank.4`][1] and then redirects back to `0.1.28`.

Expected behavior:  
~~should open the latest non-yanked version, `0.2.1`.~~
should not display "Go to latest version" since `0.1.28` is already the latest version;
for `0.1.27` should open `0.1.28` directly instead of opening the yanked version first.

[1]: https://docs.rs/futures/0.2.3-docs-yank.4/futures/

Related: https://github.com/rust-lang/docs.rs/issues/112, https://github.com/rust-lang/docs.rs/issues/221

Kinrany:
Yanked versions should be filtered from `&crate_details.versions` here:

https://github.com/rust-lang/docs.rs/blob/5727d15b57cb1a48c21b7c2ba64953191a24d253/src/web/rustdoc.rs#L241

QuietMisdreavus:
Would this get fixed by https://github.com/rust-lang/docs.rs/pull/322?

Kinrany:
> Would this get fixed by #322?

It's related: would allow replacing the snippet above with

```rust
let versions_without_yanked = &crate_details.versions
    .filter(|v| !&crate_details.yanked.contains(v));
let latest_version = latest_version(versions_without_yanked, &version);
```

Kinrany:
Oh wait, `yanked` is a `bool`, not a list of yanked versions.

Kinrany:
Still related. https://github.com/rust-lang/docs.rs/issues/396#issuecomment-522289026 would apply if you added [`yanked_versions`][1] to `CrateDetails` in addition to `yanked`.

[1]: https://github.com/rust-lang/docs.rs/pull/322/files#diff-0aa36516a2a42b271d0c346c642fb5eaR310

Nemo157:
Should be fixed by #739, the futures crate is no longer a good testcase unfortunately, but there's internal tests covering it.

PR #397, opened 2019-08-14
Add a test: latest_version() should ignore yanked

Kinrany:
This is a test for https://github.com/rust-lang/docs.rs/issues/396

"Go to latest version" should be displayed for yanked versions and point to the latest non-yanked version.

See also: https://github.com/rust-lang/docs.rs/blob/3bebf1c79bcda76cf8536bb536135f633b6b2ff2/templates/navigation_rustdoc.hbs#L89


Kinrany:
Hm, I guess changing `latest_version()` is not the correct fix: yanked versions may have regular names, indistinguishable from actual versions.

Issue #193, opened 2018-03-30
https://docs.rs/std should be useable


vi:
It should either contain docs for libstd of recent-ish nightly and/or stable, or just redirect to the official location of Rust standard library docs like https://doc.rust-lang.org/std/.

`libstd` always feels like a bump in doc accessiblilty because of I need to explicitly remember about it and not use the docs.rs URL alias.

de-vri-es:
This would be very useful. I know it's my own problem, but I always forget the URL for the standard library documentation. `docs.rs` on the other hand, I always remember.

(note: I have heard about bookmarks, but I'm stubborn.)

pietroalbini:
Closing as a duplicate of https://github.com/rust-lang/docs.rs/issues/56.

Issue #56, opened 2016-09-25
Consider hosting std docs


retep998:
Because the Rust team is apparently incapable of hosting docs for more than one platform (https://github.com/rust-lang/rust/issues/24658), it would be cool if docs.rs could host Rust docs for all the platforms.

sbstp:
Yeah, I would also like this. Getting docs for Windows on Linux is basically impossible. You guys already handle different platforms well, so it would be nice to integrate the docs even further.

cbarrick:
I would like to bump this issue as well. Docs.rs is a great improvement over the plane rustdoc interface. For reference, Godoc hosts the standard library for Go: https://godoc.org/-/go

bennofs:
This would also allow one to use `docs.rs` as an universal search for rust docs, right now I need to switch between `docs.rs` and `ruststd` docs

yjh0502:
How about hosting standard library documentation on https://docs.rs/std ? I'm using docs.rs extensively, and usually by typing address on browser omnibox directly. It's quite annoying to type `docs.rs/crate_name` for crates and `doc.rust-lang.org/std` for standard library.

vi:
Maybe add a quicky temporary redirect from `docs.rs/std` to `https://doc.rust-lang.org/std/` until proper implementation lands?

Arnavion:
Official libstd docs have been publishing OS-specific docs for some time now. Don't think this is needed any more.

suhr:
Having std docs on docs.rs would be *extremely* convenient.

m-ou-se:
https://docs.rs/std is now online! (As a redirect to https://doc.rust-lang.org/stable/std/)

pietroalbini:
Thanks for implementing this @m-ou-se! Closing the issue.

Progdrasil:
The redirect is online, is it possible to have the standard library to be hosted by docs.rs eventually?

jyn514:
@Progdrasil We do not plan to host the standard library.

PR #398, opened 2019-08-23
Redirect docs.rs/std to the official standard library docs.

m-ou-se:
See also #56 and #193.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -101,6 +101,7 @@ impl CratesfyiHandler {
         router.get("/robots.txt", sitemap::robots_txt_handler, "robots_txt");
         router.get("/sitemap.xml", sitemap::sitemap_handler, "sitemap_xml");
         router.get("/opensearch.xml", opensearch_xml_handler, "opensearch_xml");
+        router.get("/std", rustdoc::std_redirector_handler, "std");
         router.get("/releases", releases::recent_releases_handler, "releases");
         router.get("/releases/feed",
                    releases::releases_feed_handler,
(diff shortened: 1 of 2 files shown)


m-ou-se:
Is there anything that needs to be done or discussed before this can be merged?

GuillaumeGomez:
The problem with this PR is that it doesn't take into account the version of the std used when documenting the crate. Could you update the url to match it like `https://doc.rust-lang.org/{version}/std/` please? So for example: `https://doc.rust-lang.org/1.29.0/std/`.

m-ou-se:
> The problem with this PR is that it doesn't take into account the version of the std used when documenting the crate.

When documenting which crate? I'm not sure I understand your comment. I just want `docs.rs/std` to redirect to the latest std documentation at `doc.rust-lang.org/std`, because the first url is easier to remember and type. All generated documentation would still directly link to `doc.rust-lang.org/{version}/...`.

GuillaumeGomez:
From what I understand of [your code](https://github.com/rust-lang/docs.rs/pull/398/files#diff-db933256db133927163e75c554ffdfb6R71), it doesn't take the version into account. For example, if the item being linked doesn't exist anymore in the current std lib, then you'd end up with a dead link.

m-ou-se:
Which items? All it does is redirect `docs.rs/std` to `doc.rust-lang.org/std`, nothing else. It doesn't link to specific items.

GuillaumeGomez:
Oh my bad, I completely misunderstood what your PR was doing... Just one thing then, can you replace the current url with `doc.rust-lang.org/stable/std` please?

m-ou-se:
Sure, done!

pietroalbini:
The redirect is in place! https://docs.rs/std

Issue #400, opened 2019-09-03
regex 1.3.0 fails to build


BurntSushi:
Here's the build log:

https://docs.rs/crate/regex/1.3.0/builds/182874

I can't seem to make sense of the error and I can't reproduce it, even after installing the same version of Rust nightly as used on docs.rs:

```
$ rustc --version
rustc 1.39.0-nightly (bea0372a1 2019-08-20)

$ cargo build
   Compiling memchr v2.2.1
   Compiling lazy_static v1.4.0
   Compiling regex-syntax v0.6.12 (/home/andrew/code/rust/regex/regex-syntax)
   Compiling thread_local v0.3.6
   Compiling aho-corasick v0.7.6
   Compiling regex v1.3.0 (/home/andrew/code/rust/regex)
    Finished dev [unoptimized + debuginfo] target(s) in 6.17s

$ cargo doc
 Documenting lazy_static v1.4.0
    Checking lazy_static v1.4.0
 Documenting regex-syntax v0.6.12 (/home/andrew/code/rust/regex/regex-syntax)
    Checking regex-syntax v0.6.12 (/home/andrew/code/rust/regex/regex-syntax)
 Documenting memchr v2.2.1
    Checking memchr v2.2.1
    Checking thread_local v0.3.6
    Checking aho-corasick v0.7.6
 Documenting thread_local v0.3.6
 Documenting aho-corasick v0.7.6
 Documenting regex v1.3.0 (/home/andrew/code/rust/regex)
    Finished dev [unoptimized + debuginfo] target(s) in 4.46s
```

Any ideas on how to debug this?

pietroalbini:
cc @onur @QuietMisdreavus

Nemo157:
You have the `Cargo.lock` in the distributed package. Extracting it and attempting to build with an older compiler I get:

```console
> cargo download regex | tar xz
INFO: cargo-download v0.1.2
INFO: Latest version of crate regex=* is 1.3.0
INFO: Crate `regex==1.3.0` downloaded successfully

> cd regex-1.3.0

> cargo +stable --version
cargo 1.37.0 (9edd08916 2019-08-02)

> cargo +stable build --locked
error: failed to parse lock file at: /tmp/tmp.suvZWYqiw1/regex-1.3.0/Cargo.lock

Caused by:
  invalid serialized PackageId for key `package.dependencies`
```

[...]

PR #402, opened 2019-09-04
Bump cargo to the latest master

pietroalbini:
This also updates the docs.rs codebase to account for [this cargo change](https://github.com/rust-lang/cargo/pull/7186). Should fix regex 1.3.0 [failing to build](https://github.com/rust-lang/docs.rs/issues/400) once a build is rescheduled with this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,7 +138,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bitflags"
-version = "1.0.4"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


QuietMisdreavus:
Is this necessary, now that https://github.com/rust-lang/docs.rs/pull/407 is merged?

pietroalbini:
Hmm, cargo is still used in other parts of docs.rs, so this might still be useful. I'll rebase it when I have some spare time.

QuietMisdreavus:
Also please be aware that pushing cargo all the way to master might be counterproductive, if it's not the same commit that the rust compiler is using. Double-check that the commit you eventually rebase with is the same that the submodule on rust-lang/rust is at.

jyn514:
This is obsoleted by #461.

PR #404, opened 2019-09-09
Fix mount detection panic when the prefix is relative

pietroalbini:
I tried to setup docs.rs with a relative prefix and the mount detection panicked because it didn't receive an absolute path. This fixes the issue by canonicalizing the prefix before checking where it's mounted.


pietroalbini:
Thinking about it a bit more let's close this. This code is going to be deleted by rustwide anyway.

PR #405, opened 2019-09-09
Add queue-related prometheus metrics

pietroalbini:
This adds the `docsrs_queued_crates_count` and `docsrs_failed_crates_count` prometheus metrics to docs.rs, exposed in the `/about/metrics` page. Those metrics will be helpful to setup alerting based on the state of the queue.

I chose `/about/metrics` for the URL as it doesn't clash with any current of future crate (`/about` is reserved anyway), but other suggestions are welcome.

r? @Mark-Simulacrum


Mark-Simulacrum:
Looks good to me, feel free to merge when ready.

PR #406, opened 2019-09-09
metrics: actually return a 200 status code

pietroalbini:
For whatever reason the page is returning a 404 status code at the
moment. I haven't found exactly why, but this forces the status code on
that page to be 200 when metrics are returned.

r? @Mark-Simulacrum

### DIFF
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -1,6 +1,7 @@
 use super::pool::Pool;
 use iron::headers::ContentType;
 use iron::prelude::*;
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Huh. Would've definitely expected 200 to be a default.

Mark-Simulacrum:
er, thought that was a merge button. Anyway, r=me with CI passing

pietroalbini:
Confirmed it works with this landed.

PR #407, opened 2019-09-13
Switch to rustwide for sandboxing

pietroalbini:
This PR replaces the whole sandboxing and build system with [rustwide](https://github.com/rust-lang/rustwide) (Crater's build system extracted into a library). The practical effects of this are:

* The build environment will be the same as Crater.
* Docker is used instead of LXC to sandbox the builds.
* Nightlies are updated every day automatically.
* There is less reliance on Cargo as a library.

Because this will change the sandbox some crates will be broken. I'll write a post for blog.rust-lang.org explaining the changes in the next few days.

This is ready for review, but there are things that needs to be done before this can be deployed:

* [x] Switch this PR to use a crates.io release of rustwide: at the moment this is pointing to the git repository directly, and I want to cut a 0.3.0 release just before this is ready to land.
* [x] ~~[Align the rustwide build environment with the current chroot](https://github.com/rust-lang/crates-build-env/issues/18)~~
* [x] ~~Write a blog post explaining the sandboxing changes~~ https://github.com/rust-lang/blog.rust-lang.org/pull/403

r? @QuietMisdreavus 
cc @Mark-Simulacrum @onur


pietroalbini:
It *should* be possible to review this commit-by-commit.

src/db/add_package.rs
```diff
@@ -159,13 +158,12 @@ pub fn add_package_into_database(conn: &Connection,
         if let Some(versions_array) = versions.as_array_mut() {
             let mut found = false;
             for version in versions_array.clone() {
-                if &semver::Version::parse(version.as_string().unwrap()).unwrap() ==
-                   pkg.manifest().version() {
+                if version.as_string().unwrap() == metadata_pkg.version {
```
Mark-Simulacrum: hm so semver equality is not quite the same as string equality IIRC, so maybe this does need to parse?

src/docbuilder/limits.rs
```diff
@@ -0,0 +1,97 @@
...
+        res.insert("Available RAM".into(), size_scale(self.memory));
+        res.insert(
+            "Maximum rustdoc execution time".into(),
+            time_scale(self.timeout.as_secs() as usize),
+        );
+        res.insert("Maximum size of a build log".into(), size_scale(self.max_log_size));
+        if self.networking {
+            res.insert("Network access".into(), "allowed".into());
+        } else {
+            res.insert("Network access".into(), "blocked".into());
```
Mark-Simulacrum: All of the keys here are `&str` and that seems like it'll be true indefinitely, so maybe `BTreeMap<&'static str, String>` might make some sense?

src/docbuilder/rustwide_builder.rs
```diff
@@ -61,6 +52,9 @@ static ESSENTIAL_FILES_UNVERSIONED: &[&str] = &[
     "SourceSerifPro-It.ttf.woff",
 ];
 
+static DUMMY_CRATE_NAME: &str = "acme-client";
```
Mark-Simulacrum: Uh, could we choose a less... likely to exist dummy crate? Maybe use a rustc-dep-of-std dep which is ~guaranteed empty?

Looking at the code maybe we just want a `Limit::default()`

Mark-Simulacrum:
I would like to also not have a "number of lines" limit since that seems needless, we have a limit on how many bytes you're emitting which is all we should care about presumably?

src/docbuilder/limits.rs
pietroalbini: Heh, that's what I originally tried.

`rustc-serialize`'s [ToJson](https://docs.rs/rustc-serialize/0.3.24/rustc_serialize/json/trait.ToJson.html) is implemented for `BTreeMap<String, T>` but not `BTreeMap<&str, T>`, so this isn't really actionable without switching to serde (and changing this to serde would require a switch at least for all the frontend).

src/utils/build_doc_rustwide.rs
```diff
@@ -11,6 +11,10 @@ use std::path::Path;
 use utils::{parse_rustc_version, resolve_deps};
 use Metadata;
 
+// TODO: 1GB might not be enough
+const SANDBOX_MEMORY_LIMIT: usize = 1024 * 1025 * 1024; // 1GB
+const SANDBOX_NETWORKING: bool = false;
```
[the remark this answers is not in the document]
pietroalbini: This isn't here in the following commits, are there similar concerns with the new code?

src/docbuilder/rustwide_builder.rs
pietroalbini: > Looking at the code maybe we just want a Limit::default()

The dummy crate is not used just for limits, it's actually built to then capture the essential files after the toolchain is updated (those constants are used a few lines below to create the `Crate` struct.

> Uh, could we choose a less... likely to exist dummy crate? Maybe use a rustc-dep-of-std dep which is ~guaranteed empty?

I just ported the crate name from the old code :sweat_smile: 

The crate actually exists but in version `0.0.0` is empty, so we're basically building an empty crate to then extract the essential files.

pietroalbini:
> I would like to also not have a "number of lines" limit since that seems needless, we have a limit on how many bytes you're emitting which is all we should care about presumably?

Already removed it.

src/docbuilder/limits.rs
```diff
@@ -0,0 +1,97 @@
+use error::Result;
+use postgres::Connection;
+use std::collections::BTreeMap;
+use std::time::Duration;
+
+const DEFAULT_MEMORY_LIMIT: usize = 3 * 1024 * 1024 * 1024; // 3 GB
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(15 * 60); // 15 minutes
+const DEFAULT_NETWORKING: bool = false;
```
Mark-Simulacrum: Is it useful to have these separate from the default impl?

pietroalbini: Mostly because it's easier to see and change them when looking at the file. I don't care too much about them being there.

pietroalbini:
Addressed all of Mark's review comments.

pietroalbini:
Updated the rustwide's build environment with the list of crates installed in the current container and [wrote a draft of the blog post announcing the changes](https://github.com/rust-lang/blog.rust-lang.org/pull/403).

pietroalbini:
Also created https://github.com/rust-lang/docs.rs/pull/409 to properly alert docs.rs users of the changes in this PR.

src/bin/cratesfyi.rs
```diff
@@ -308,5 +309,6 @@ fn logger_init() {
                 record.args())
     });
     builder.parse(&env::var("RUST_LOG").unwrap_or("cratesfyi=info".to_owned()));
-    builder.init();
+
+    rustwide::logging::init_with(builder.build());
```
QuietMisdreavus: I'm assuming this initializes the global logger as well? I want to make sure we can still depend on the docs-rs log itself, since this replaced the `init` line.

pietroalbini: Yeah, rustwide's logging wraps another logger (docs.rs's env_logger in this case): it processes the records and then it forwards them to the wrapped logger.

src/docbuilder/rustwide_builder.rs
```diff
@@ -0,0 +1,459 @@
...
+    "i686-apple-darwin",
+    "i686-pc-windows-msvc",
+    "i686-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-linux-gnu",
+];
+static DEFAULT_TARGET: &str = "x86_64-unknown-linux-gnu";
+
+static ESSENTIAL_FILES_VERSIONED: &[&str] = &[
```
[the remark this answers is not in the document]
Mark-Simulacrum: I continue to believe this is not that necessary for this to land; this is a good issue to file for something we need to do, but we've not added such a file in a relatively long time to rustdoc and we can add a comment to rustdoc saying "go ping docs.rs folks if you add a file here".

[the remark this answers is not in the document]
Mark-Simulacrum: But if this _is_ a blocker, then let's lock down nightly for now -- getting rustdoc to print these out seems sort of 'spec' work to me which is always hard :)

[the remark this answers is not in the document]
QuietMisdreavus: I feel like we can get away with landing this without making rustdoc report them, but i would like to have some kind of process so we can quickly update docs.rs if/when new files are added.  Making it automatic would be ideal, but in lieu of that, i believe your suggestion of adding a comment to the spot in rustdoc where the static files are emitted should be fine as a stopgap solution.

src/db/add_package.rs
```diff
@@ -189,47 +191,41 @@ pub fn add_build_into_database(conn: &Connection,
...
-        let version = format!("{}", dependency.version_req());
-        let kind = match dependency.kind() {
-            Kind::Normal => "normal",
-            Kind::Development => "dev",
-            Kind::Build => "build",
-        };
+    for dependency in &pkg.dependencies {
+        let name = dependency.name.clone();
+        let version = dependency.req.clone();
+        let kind = dependency.kind.as_ref().map(|s| s.clone()).unwrap_or_else(|| "normal".into());
```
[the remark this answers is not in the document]
pietroalbini: Uh, yeah, probably an artifact of a refactoring. Fixed in https://github.com/rust-lang/docs.rs/pull/407/commits/50273431885e834a2cf0fc432e6f258c1a7af1c4.

src/docbuilder/rustwide_builder.rs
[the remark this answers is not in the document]
pietroalbini: Yeah, I don't think it's worth blocking on this. Another solution in the meantime is to configure highfive to ping people when that file is changed.

pietroalbini:
Ok, merging this :tada:

PR #409, opened 2019-09-16
Add global alert to notify users about upcoming breaking changes

pietroalbini:
This PR adds an alert in the navigation bar of each page to alert everyone of the changes in #407. It links to the blog post prepared in https://github.com/rust-lang/blog.rust-lang.org/pull/403, so it should be merged only after the post is published (and ideally shortly after).

The alert can be easily disabled or customized in the source code, so it's safe to keep the code even when it's not used anymore.

| Crate page | Docs page | Index |
| --- | --- | --- |
|  |  |  |

r? @Mark-Simulacrum @QuietMisdreavus

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,15 @@ pub mod utils;
 mod docbuilder;
 mod web;
 
+use web::page::GlobalAlert;
+
+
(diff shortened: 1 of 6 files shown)


pietroalbini:
Did some really basic "user research", and I changed the alert color to red.

PR #415, opened 2019-09-28
Fix search background color on dark theme

pickfire:
Using firefox with dark theme.

Before

2019-09-28-102501_1366x768_scrot

After

2019-09-28-102513_1366x768_scrot

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -186,6 +186,7 @@ div.nav-container {
             font-size: 0.8em;
             text-align: right;
             box-shadow: none;
+            background-color: #fff;
             height: 31px;
             display: none;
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
I can't reproduce on my linux so I'll retry on another computer when I have time.

pickfire:
@GuillaumeGomez I think you need to enable `widget.content.allow-gtk-dark-theme` in firefox.

I am using firefox 69 with breeze-dark as my system theme.

GuillaumeGomez:
Oh I see! I'll give it another try then. Thanks for the tip!

GuillaumeGomez:
Thanks!

Issue #417, opened 2019-09-30
Left slider not working


liborty:
Firefox 69.0.1, Linux x86-64.
Even when the left hand slider is put hard down, some of the docs menu items (functions, crates, etc) remain hidden and inaccessible.
![Screenshot from 2019-10-01 09-30-17](https://user-images.githubusercontent.com/457124/65924210-0fbab180-e42f-11e9-8ebe-1f8a36c543c9.png)

liborty:
Thank you. I would not say it was such a small matter, as it made a lot of documentation inaccessible.

GuillaumeGomez:
It's fixed in rustdoc.

@pietroalbini @QuietMisdreavus Any idea if we could/should update the current noscript stylesheets to backport this change?

pietroalbini:
That's not trivial as we'd need to change the noscript stylesheets for each affected nightly.

GuillaumeGomez:
Indeed. :)

jyn514:
Closing in favor of #464

PR #418, opened 2019-10-01
small style improvement

GuillaumeGomez:
Part of https://github.com/rust-lang/docs.rs/issues/417

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -59,7 +59,7 @@ div.rustdoc {
 
     .sidebar {
         @media (min-width: 701px) {
-            margin-top: $top-navbar-height;
+            top: $top-navbar-height;
         }
 
         .block > ul > li {


jyn514:
Could you describe why you're making this change? I tried it in developer console and didn't notice a difference.

jyn514:
Ok even after looking at that issue I don't see the difference between top and margin-top as displayed. I'm on https://docs.rs/hyper/0.13.0-alpha.4/hyper/struct.Request.html and I see that the crate is hidden under the scrollbar, but that happens even if I use `top`.

This looks related to #465, since if I take out `margin-top` altogether the scrollbar works fine but the Rust icon at the top is hidden by the top banner.

GuillaumeGomez:
I'll close it for the moment and come back to it later.

PR #421, opened 2019-10-02
Forward package rustdoc_args to rustdoc

LucioFranco:
This fixes a regression that happened in the recent docs.rs changes where the package rustdoc_args would not get forwarded to the rustdoc flags.

fixes #421


pietroalbini:
Thanks :heart:

PR #432, opened 2019-10-11
Add dockerfile

jyn514:
I went over this with @pietroalbini in Discord - Since the Vagrantfile is out of date, it's hard to build docs.rs locally. This pull allows anyone with `docker-compose` installed to get started as fast as they can build the crates.

**Overview**

Most of the build happens in the Dockerfile and is cached. The only changes that happen after the docker container starts are building the documentation for each crate (which is cached, since it's stored in the mounted Rustwide volume) and performing the database migrations, which are cached in practice since no one will be deleting the database container for the fun of it.

The webserver does run as root in the docker container, which rustwide requires in order to be able to communicate with docker. Although we could add the `cratesfyi` user to the `docker` group, there's no way to make sure that the mounted docker socket in the container has the same GID as the host.

** Changes **

- Adds `--foreground` flag for `cratesfyi daemon` (needed for docker)
- Adds CSS files to git. Not sure if there's another way these are managed, since there are timestamps on each CSS file. Additionally, I'm not sure I got them all, I basically just clicked around until I stopped getting 404 errors.
- The server now listens on `0.0.0.0` when it can't determine a hostname, not `127.0.0.1`


jyn514:
Removed the duplicate files, thanks for the review!

pietroalbini:
Thanks for making this!

PR #435, opened 2019-10-16
Also redirect the remaining rust-lang.org crates.

### LINKED ISSUE #433: Host core and alloc
In the same vein as #56, docs.rs should also host docs for the `core` and `alloc` crates.

Originally mentioned [on the users forum](https://users.rust-lang.org/t/publish-docs-for-std-on-docs-rs/32769/4?u=michael-f-bryan):

> https://docs.rs/std 2 now redirects! Looks like it has been done for std, but not core or alloc.

de-vri-es:
This PR adds redirects for `alloc`, `core`, `proc_macro` and `test`. Together with `std` that covers all crates from rust-lang.org :)

Should resolve #433.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -103,11 +103,14 @@ impl CratesfyiHandler {
         router.get("/robots.txt", sitemap::robots_txt_handler, "robots_txt");
         router.get("/sitemap.xml", sitemap::sitemap_handler, "sitemap_xml");
         router.get("/opensearch.xml", opensearch_xml_handler, "opensearch_xml");
+
+        // Redirect standard library crates to rust-lang.org
         router.get("/alloc", rustdoc::RustLangRedirector::new("alloc"), "alloc");
         router.get("/core", rustdoc::RustLangRedirector::new("core"), "core");
         router.get("/proc_macro", rustdoc::RustLangRedirector::new("proc_macro"), "proc_macro");
         router.get("/std", rustdoc::RustLangRedirector::new("std"), "std");
         router.get("/test", rustdoc::RustLangRedirector::new("test"), "test");
+
         router.get("/releases", releases::recent_releases_handler, "releases");
         router.get("/releases/feed",
                    releases::releases_feed_handler,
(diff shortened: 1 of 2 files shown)


src/web/mod.rs
```diff
@@ -103,7 +103,11 @@ impl CratesfyiHandler {
         router.get("/robots.txt", sitemap::robots_txt_handler, "robots_txt");
         router.get("/sitemap.xml", sitemap::sitemap_handler, "sitemap_xml");
         router.get("/opensearch.xml", opensearch_xml_handler, "opensearch_xml");
-        router.get("/std", rustdoc::std_redirector_handler, "std");
+        router.get("/alloc", rustdoc::RustLangRedirector::new("alloc"), "alloc");
```
GuillaumeGomez: Maybe add an empty (and a comment) before and after this block so it's easier to spot it when reading?

src/web/rustdoc.rs
```diff
@@ -65,11 +65,25 @@ impl ToJson for RustdocPage {
...
+impl RustLangRedirector {
+    pub fn new(target: &'static str) -> Self {
+        Self { target }
+    }
+}
+
+impl iron::Handler for RustLangRedirector {
+    fn handle(&self, _req: &mut Request) -> IronResult<Response> {
+        let url = url::Url::parse("https://doc.rust-lang.org/stable/")
+            .unwrap()
```
GuillaumeGomez: Nitpicking but could you replace the `unwrap` calls with `expect` calls please?

GuillaumeGomez:
Just two comments otherwise seems good to me!

src/web/mod.rs
de-vri-es: Sounds good.

src/web/rustdoc.rs
de-vri-es: Sure :)

de-vri-es:
Thanks for the quick review. Processed the feedback :)

GuillaumeGomez:
Thanks!

GuillaumeGomez:
@pietroalbini @QuietMisdreavus Can one of you update the server please? :)

pietroalbini:
Deployed the changes!

PR #445, opened 2019-10-25
add build success/fail counts to metrics

QuietMisdreavus:
This PR adds four new records to the `/about/metrics` endpoint:

- The total number of crates built in the last six hours
- The number of crates built successfully in the last six hours
- The number of crates which failed to build in the last six hours, due to a compile error
- The number of crates which failed to build in the last six hours, due to being flagged "not a library"

By making the metrics a sliding window of "the last six hours", we can better track the rate of failures, rather than any individual success/failure count. For example, calculating `failures / total builds`, we can get a rolling calculation of our build failure rate.

(It would be really cool to have some kind of inspection of the build logs to flag errors due to build scripts, `pkg-config` failing, etc, to get an idea of crates that need some kind of system dependency versus those that may be due to our own build configuration causing an issue, but queries on the build output take too long to be repeatedly polled. Perhaps something like that can be tracked as a counter that is incremented after a build finishes, rather than a query that's repeatedly executed.)

cc @pietroalbini

### DIFF
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -62,6 +62,7 @@ impl DocBuilder {
         match builder.build_package(self, &name, &version) {
             Ok(_) => {
                 let _ = conn.execute("DELETE FROM queue WHERE id = $1", &[&id]);
+                ::web::metrics::TOTAL_BUILDS.inc();
             }
             Err(e) => {
                 // Increase attempt count
@@ -70,6 +71,7 @@ impl DocBuilder {
                 let attempt: i32 = rows.get(0).get(0);
                 if attempt >= 5 {
                     ::web::metrics::FAILED_BUILDS.inc();
+                    ::web::metrics::TOTAL_BUILDS.inc();
                 }
                 error!("Failed to build package {}-{} from queue: {}",
                        name,
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -43,7 +43,7 @@ impl DocBuilder {
     ) -> Result<bool> {
         let conn = try!(connect_db());
 
-        let query = try!(conn.query("SELECT id, name, version
+        let query = try!(conn.query("SELECT id, name, version, attempt
                                      FROM queue
                                      WHERE attempt < 5
                                      ORDER BY priority ASC, attempt ASC, id ASC
@@ -58,6 +58,7 @@ impl DocBuilder {
         let id: i32 = query.get(0).get(0);
         let name: String = query.get(0).get(1);
         let version: String = query.get(0).get(2);
+        let attempt: i32 = query.get(0).get(3);
 
         match builder.build_package(self, &name, &version) {
             Ok(_) => {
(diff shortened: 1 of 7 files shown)


pietroalbini:
> By making the metrics a sliding window of "the last six hours", we can better track the rate of failures, rather than any individual success/failure count. For example, calculating failures / total builds, we can get a rolling calculation of our build failure rate.

I'd still prefer to use a counter that shows data for all time (either backed by the database or incremented on the fly, letting prometheus dealing with resets). That gives us way more flexibility while graphing and alerting. For example, the same behavior in this PR can be replicated with a counter and the query `increase(docsrs_metric_name[6h])`.

src/docbuilder/rustwide_builder.rs
```diff
@@ -310,6 +310,20 @@ impl RustwideBuilder {
                 }
 
                 let has_examples = build.host_source_dir().join("examples").is_dir();
+                let metadata = res.cargo_metadata.root();
+                let is_library = match metadata.targets[0].kind.as_slice() {
```
pietroalbini: We do the same check in `add_package`, and I worry over time the two might diverge. Could you extract both of them into an utility function?

https://github.com/rust-lang/docs.rs/blob/ffe043c00e5016cdc010bb87b8a6e1378e0fd876/src/db/add_package.rs#L40-L43

src/docbuilder/queue.rs
```diff
@@ -58,15 +58,20 @@ impl DocBuilder {
...
         match builder.build_package(self, &name, &version) {
             Ok(_) => {
                 let _ = conn.execute("DELETE FROM queue WHERE id = $1", &[&id]);
             }
             Err(e) => {
                 // Increase attempt count
-                let _ = conn.execute("UPDATE queue SET attempt = attempt + 1 WHERE id = $1",
-                                     &[&id]);
+                let attempt = attempt + 1;
+                let _ = conn.execute("UPDATE queue SET attempt = $1 WHERE id = $2",
```
pietroalbini: To keep this atomic, you can consolidate the select and the update with PostgreSQL's [`RETURNING`](https://www.postgresql.org/docs/9.5/dml-returning.html):

```
UPDATE queue SET attempt + 1 WHERE id = $1 RETURNING attempt;
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -310,6 +310,20 @@ impl RustwideBuilder {
...
+                    &[ref kind] if kind == "lib" || kind == "proc-macro" => true,
+                    _ => false,
+                };
+                if res.successful {
+                    ::web::metrics::SUCCESSFUL_BUILDS.inc();
+                } else {
+                    if is_library {
```
pietroalbini: Nit: this if can be flattened.

src/utils/queue.rs
```diff
@@ -4,6 +4,7 @@ use postgres::Connection;
 use error::Result;
 
 pub fn add_crate_to_queue(conn: &Connection, name: &str, version: &str, priority: i32) -> Result<()> {
+    ::web::metrics::TOTAL_BUILDS.inc();
```
pietroalbini: I'd move this incr to `queue`, after the package has been built/failed 5 times, otherwise if we have a long queue our metrics are going to be messed up.

src/docbuilder/rustwide_builder.rs
QuietMisdreavus: I admit to stealing that block from `add_package`, so i'll go ahead and make it a method on Packages.

src/utils/queue.rs
QuietMisdreavus: In other words, the counter would become "packages we've built", rather than "packages we've queued"? That seems fine to me; i could also see it diverging over time if we manually queued builds or reset their attempts counter, so this works for me.

pietroalbini:
Thanks!

PR #450, opened 2019-10-28
Remove unused vagrantfile

jyn514:
Since it's not working and also not mentioned in the README, there's no reason to keep it around.


pietroalbini:
Thanks!

PR #452, opened 2019-10-28
[docker] Don't rebuild when the templates change

jyn514:
This speeds up development, especially since docker always builds in release mode.


jyn514:
I tested this locally and changes to the templates are picked up by the website as long as you re-run `docker-compose build`.

pietroalbini:
Great, thanks!

Issue #200, opened 2018-05-03
"Go to latest version" should try to go to the corresponding sub-page of the latest version


Boscop:
"Go to latest version" should try to go to the corresponding sub-page that one is currently on, not to the crate doc's index page of the latest version.
Going to the index should only be done as a fallback if the current sub-page doesn't exist in the new version's docs.

jyn514:
How would I get started on fixing this? I've run into this quite a lot.

Boscop:
@jyn514 Each doc page is associated with a symbol name (ident). In a new version, the symbols are sometimes in different modules but still named the same. To find the right doc page, search through all doc pages of the new version to find the one associated with the symbol of the page you're on now (the doc of the old version).

daboross:
How about just trying out the same module/symbol name as an attempt, and then landing the users on the search page for the ident if it fails?

I'd be hesitant to unconditionally search for other things with the same name, mostly because struct names are never guaranteed to be unique and having the same name in different packages can sometimes be a positive pattern, as with `Error` structs.

jyn514:
When I try to build `docs.rs` with `vagrant up`, I get the following error:

```
    default: Running
    default:  
    default: `/home/cratesfyi/docs.rs/target/debug/cratesfyi build add-essential-files`
    default: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1165:5
    default: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

The panic is coming from the following line in `src/bin/cratesfyi.rs`: `let mut builder = RustwideBuilder::init().unwrap();`.

Any idea what's going wrong?

jyn514:
Took this to discord, currently writing a docker backend since Vagrant is very out of date - https://github.com/jyn514/docs.rs/ https://github.com/jyn514/docs.rs/tree/docker-in-docker

pickfire:
@jyn514 I have that issue as well and no idea what went wrong? It took me one hour to get to that point and it failed, vagrant is really slow even a startup takes 1-5 minutes.

jyn514:
[...]

PR #454, opened 2019-10-28
Keep the subpage when going to latest version

jyn514:
Addresses https://github.com/rust-lang/docs.rs/issues/200

If the page does not exist on the latest version, searches for a page with that title.

~~TODO: If
1. Going to a new version that
2. Deleted the existing module/item and
3. The user is looking at a platform other than the default,
then the link will be wrong (it will not include the name of the crate).~~

~~The solution is to check if `req_path[3]` is the crate name or the platform. Once this is done, https://github.com/rust-lang/docs.rs/issues/200#issuecomment-544033326 will be trivial to fix.~~

This is fixed, see below.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -15,6 +15,7 @@ use super::page::Page;
 use rustc_serialize::json::{Json, ToJson};
 use std::collections::BTreeMap;
 use iron::headers::{Expires, HttpDate, CacheControl, CacheDirective};
+use postgres::Connection;
 use time;
 use iron::Handler;
 use utils;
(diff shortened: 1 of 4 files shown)


jyn514:
Also I haven't found a test case for an item that was deleted/moved between versions, if anyone knows of one please let me know!

jyn514:
Fixed the bug with a different platform. I've tested the following manually:

- [x] same item
- [x] same item with platform
- [x] renamed item
- [x] renamed item with platform
- [x] deleted item
- [x] deleted item with platform

If anyone has suggestions for how to set up unit tests, I'd be happy to hear them! I'm not sure how to ensure that certain crates exist in the database :/

jyn514:
This does not address the following issues:

- clicking the name of the crate in the top-left does not keep the platform
- choosing a different platform goes to the root, not to the current item

jyn514:
This is ready for review.

EverlastingBugstopper:
This is fantastic, thanks so much @jyn514

src/web/rustdoc.rs
```diff
@@ -259,20 +261,46 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+        .set("latest_version", &version)
         .to_resp("rustdoc")
 }
 
-
+fn path_for_version(req_path: &[&str], target_name: &str, conn: &Connection) -> String {
+    if File::from_path(&conn, &req_path.join("/")).is_some() {
+        req_path[3..].join("/") // NOTE: this adds 'index.html' if it wasn't there before
+    } else { // this page doesn't exist in the latest version
+        let search_item = if *req_path.last().unwrap() == "index.html" { // this is a module
```
QuietMisdreavus: Style nit: Can these comments be given their own lines? I prefer to see explanatory comments on their own line above or inside the relevant line or block. For example:

```rust
    if File::from_path(&conn, &req_path.join("/")).is_some() {
        // NOTE: this adds 'index.html' if it wasn't there before
        req_path[3..].join("/")
    } else {
        // this page doesn't exist in the latest version
        let search_item = if *req_path.last().unwrap() == "index.html" {
            // this is a module
            req_path[req_path.len() - 2]
        } else {
            // this is an item
            req_path.last().unwrap().split('.').nth(1)
                .expect("paths should be of the form <class>.<name>.html")
        };
        // check if req_path[3] is the platform choice or the name of the crate
        let concat_path;
        let crate_root = if req_path[3] != target_name {
            concat_path = format!("{}/{}", req_path[3], req_path[4]);
            &concat_path
        } else {
            req_path[3]
        };
        format!("{}/?search={}", crate_root, search_item)
    }
```

src/web/rustdoc.rs
```diff
@@ -259,20 +261,46 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
-        .set_bool("is_latest_version", latest_version.is_none())
-        .set("latest_version", &latest_version.unwrap_or(String::new()))
+        .set_bool("is_latest_version", path.is_empty())
+        .set("path_in_latest", &path)
+        .set("latest_version", &version)
         .to_resp("rustdoc")
 }
 
-
+fn path_for_version(req_path: &[&str], target_name: &str, conn: &Connection) -> String {
```
QuietMisdreavus: Can this function get a doc comment? Something like this would work:

```rust
/// Checks whether the given path (including version and target information) exists. Crate target
/// name is used to confirm whether a platform triple is part of the path.
///
/// Returns a path that can be appended to `/crate/version/` to create a complete URL.
```

jyn514:
Added comments as requested :)

Thanks for taking a look at this!

daboross:
Thanks for doing this!

PR #455, opened 2019-10-29
Improve Dockerfile

pietroalbini:
This PR changes the Dockerfile, improving multiple aspects of it. The changes will allow the image to be deployed on our infrastructure.

* The base image is now Ubuntu 18.04, to be the same as the other infra ones.
* The image has been shrunk from 3.2GB to 500MB, thanks to moving the index outside of the image and adopting multi-stage builds.
* Environment variables have been shuffled around, moving docker-compose specific ones either in `docker-compose.yml` or a new gitignored `.env`.
* The entrypoint has been changed to retry migrating the database multiple times, allowing PostgreSQL to initialize during the first startup.
* Unneeded steps have been removed.


pietroalbini:
Also added Docker builds on CI, and the code to upload the image to the infra team's registry on master.

docker-compose.yml
```diff
@@ -11,6 +11,9 @@ services:
             - ".rustwide:/home/cratesfyi/rustwide"
         environment:
             CRATESFYI_RUSTWIDE_WORKSPACE: /home/cratesfyi/rustwide
+            CRATESFYI_DATABASE_URL: postgresql://cratesfyi:password@db
+        env_file:
+            - .env
```
jyn514: This prevents people from running `docker-compose up` to get started (they have to run `cp .env.sample .env` first). It looks like [docker-compose automatically reads from .env](https://docs.docker.com/compose/env-file/) even if it's not in the compose file, would that work instead? We could still document it in the README so it doesn't look like magic :)

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
         volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
             - ".rustwide:/home/cratesfyi/rustwide"
+            - "cratesio-index:/opt/docsrs/prefix/crates.io-index"
```
jyn514: nit: this should probably be part of the previous commit

Dockerfile
```diff
@@ -1,58 +1,74 @@
...
-RUN adduser --home $HOME --disabled-login --disabled-password --gecos "" cratesfyi
+FROM ubuntu:bionic AS build
 
-### STEP 3: Setup build environment as new user ###
-ENV CRATESFYI_PREFIX=/home/cratesfyi/prefix
-RUN mkdir $CRATESFYI_PREFIX && chown cratesfyi:cratesfyi "$CRATESFYI_PREFIX"
+# Install packaged dependencies
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+  build-essential git curl cmake gcc g++ pkg-config libmagic-dev \
+  libssl-dev zlib1g-dev sudo ca-certificates docker.io
```
jyn514: `sudo` and `docker.io` are no longer necessary. Most of the other commands are also not needed I think.

Dockerfile
```diff
@@ -1,10 +1,16 @@
-FROM rust:slim
+FROM ubuntu:bionic
```
jyn514: Is this just for consistency with the other infra images? I'd rather reuse the `rust` image so this one has fewer commands ...

Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN cargo build --release
 # source code didn't change thanks to mtime weirdness.
 RUN rm -rf src build.rs
 
+COPY .git .git
```
jyn514: nit: could we set `CARGO_PKG_VERSION` instead? That would save on space.

docker-entrypoint.sh
```diff
@@ -7,7 +7,26 @@ export DOCS_RS_DOCKER=true
 export RUST_LOG=cratesfyi,rustwide=info
 export PATH="$PATH:/build/target/release"
 
-cratesfyi database migrate
+# Try migrating the database multiple times if it fails
+# This avoids the docker container crashing the first time it's started with
+# docker-compose, as PostgreSQL needs some time to initialize.
+set +e
```
jyn514: Thank you for working on this! I meant to fix it but never got around to it.

.github/workflows/ci.yml
```diff
@@ -21,3 +21,25 @@ jobs:
...
         run: cargo test -- --test-threads=1
+
+  docker:
+    name: Docker
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 2
```
jyn514: Out of curiosity, why 2 instead of 1?

docker-compose.yml
pietroalbini: That's a good point, but I'm not sure if what you suggested would help.

As far as I understand yes, docker-compose reads the `.env`, but you still need to forward the variables explicitly to the container you want, and docker-compose will error out if they're missing. That would make error messages less clear.

If we want to retain just one line to get up and running, we could create a wrapper script like this:

```bash
#!/bin/sh
if ! [ -f ".env" ]; then
    cp .env.sample .env
fi
docker-compose "$@"
```

And tell people to do:

```
./compose up
```

Let's also cc @QuietMisdreavus to see what she thinks.

jyn514: Hmm ... could we set the variables to a default and have them be overriden by `.env`? We could put that in the entrypoint, something like `CRATESFYI_GITHUB_USERNAME={$CRATESFYI_GITHUB_USERNAME:-}` (there's probably a way to do that in posix syntax).

QuietMisdreavus: I think it's fine from a user-experience standpoint to require the `cp` step. Generally speaking, "repo setup" steps like this are not that bad IMO, since they're going to be performed once per clone, so you can just set it and forget it. If you miss it, i'm assuming docker-compose will yell about the missing file so you can check the README or remember the copy step there.

jyn514: Ok, that's reasonable. In that case I think this is ready to merge :)

jyn514: and yes, if it's not there you get `ERROR: Couldn't find env file: /path/to/docs.rs/.env`

pietroalbini:
Thanks for the review! Merging this then.

PR #461, opened 2019-11-02
Remove duplicated package versions

jyn514:
See #459 . This removes about 15 packages, the rest have non-trivial upgrade paths.


pietroalbini:
Yay for finally removing the Cargo dependency!

PR #465, opened 2019-11-02
Fix top banner hiding text

GuillaumeGomez:
I don't know what happened with the CSS formatting so I fixed it... But the important commit is the second one: it prevents text to go under the top banner. For example: https://docs.rs/halt/0.3.0/halt/struct.Invalid.html#implementations, the title won't "disappear" under the banner.

cc @pietroalbini

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -14,11 +14,13 @@ html, button, input, select, textarea,
 div.rustdoc {
   font-family: "Source Serif Pro", Georgia, Times, "Times New Roman", serif;
   padding: 10px 15px 20px 15px;
-  position: absolute;
+  position: fixed;
+  top: 32px;
+  bottom: 0;
   left: 0;
   right: 0;
-  top: 32px;
-  max-height: calc(100vh - 32px);
+  display: block;
+  overflow-y: auto;
 }
 @media (max-width: 700px) {
   div.rustdoc {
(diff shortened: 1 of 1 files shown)


jyn514:
I didn't even realize this was a problem with docs.rs, I thought firefox was being wacky. Thanks for fixing this!

css/style.css
```diff
@@ -13,12 +13,13 @@ html, button, input, select, textarea,
 
 div.rustdoc {
   font-family: "Source Serif Pro", Georgia, Times, "Times New Roman", serif;
-  padding: 10px 15px 20px 15px;
```
jyn514: It looks like taking out this 'padding' style makes the search bar overlap with the style button, at least in my browser.

image

GuillaumeGomez: Oh funny. Can you open an issue about this (and maybe even a PR?). I don't want to mix more things in this PR.

jyn514: I'm not sure I understand, do you mean in the main rust repo? If you just add the padding back in (like it was before) it works fine.

GuillaumeGomez: Oh crap! Completely missed the fact that I removed it. Good catch! Sorry for not understanding at first.

GuillaumeGomez:
@jyn514 I have put back what I mistakenly removed.

jyn514:
Great, thanks so much!

PR #467, opened 2019-11-04
allow building locally as well as with docker

jyn514:
If docker uses the `.rustwide` directory, it will only be readable and writable by root.
This uses a different scratch directory so that the `.rustwide` directory can be used for local builds.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,10 +8,10 @@ services:
             - "3000:3000"
         volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
-            - ".rustwide-docker:/home/cratesfyi/rustwide"
+            - ".rustwide-docker:/opt/docsrs/rustwide"
             - "cratesio-index:/opt/docsrs/prefix/crates.io-index"
         environment:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
             - "3000:3000"
         volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
-            - ".rustwide:/home/cratesfyi/rustwide"
+            - ".rustwide-docker:/home/cratesfyi/rustwide"
```
pietroalbini: Unrelated: can you change the path on the container to be `/opt/docsrs/rustwide`? You'll also need to change the environment variable below.

jyn514:
This means that artifacts aren't cached between the container and the host, which was the original idea behind using the default directory. Unfortunately I don't see a way around this, I ran into the same problem in #432 and I think we decided there's not a good way to map permissions.

jyn514:
Good catch!

jyn514:
This is ready for review.

pietroalbini:
Thanks!

PR #468, opened 2019-11-05
Remove Cargo dependency

jyn514:
Retry of https://github.com/rust-lang/docs.rs/pull/461 with fewer changes.

Hopefully this one won't slow the prod server down to a stall.


pietroalbini:
Let's try this.

PR #469, opened 2019-11-05
increment upload metric for each successful file

miller-time:
I should have incremented the metrics this way before merging pull request #393 - this should hopefully help smooth out the chart for this metric.


miller-time:
@pietroalbini r?

pietroalbini:
Thanks!

Issue #474, opened 2019-11-15
`docs.rs/std` redirect works but not `docs.rs/std/`


nagisa:
Going to `docs.rs/std/` will give a 404.

jyn514:
This looks like it's specific to the standard library, routes like `/cranelift-codegen/` redirect properly.

PR #476, opened 2019-11-15
allow trailing slash in standard-library redirects

jyn514:
fixes https://github.com/rust-lang/docs.rs/issues/474


pietroalbini:
I'd prefer not to duplicate the list of crates we redirect. A solution could be to create a static like this:

```rust
static DOC_RUST_LANG_ORG_REDIRECTS: &[&str] = &["alloc", "std", ...];
```

...and dynamically generate the routes?

jyn514:
@pietroalbini  I removed the duplicate code, want to take another look?

pietroalbini:
Looks great, thanks!

PR #478, opened 2019-11-18
Check scss

GuillaumeGomez:
I ended the scss file cleanup. To prevent having to do it once again, I added a script to check that the file continues to follow a precise format.

I also added this check into the github actions (curious to see if python 3 is already installed there or not...).

I used this script to lint the scss file:

<details>

```python
import os
import sys


def check_scss_file(file_path):
    if not file_path.endswith(".scss"):
        return 0
    errors = 0
    with open(file_path, "r") as f:
        block_level = 0
        empty_line_before = 0
        content = f.read().split("\n")
        instructions = 0
        blocks = [0]
        i = 0
        while i < len(content):
            reset_empty_line_count = True
            reset_instructions_count = True

            line = content[i].strip()
            if line.startswith("//"):
                i += 1
                continue
            if "\t" in line:
                print("[{}:{}]: Tab character found".format(file_path, i + 1))
                errors += 1
            line = line.split("//")
            line = line[0].strip() if len(line) > 0 else ""
            if len(line) == 0:
                empty_line_before += 1
                if empty_line_before > 1:
                    print("[{}:{}]: Unexpected empty line".format(file_path, i + 1))
                    errors += 1
                reset_empty_line_count = False
            elif line.endswith("{"):
                block_level += 1
                blocks.append(0)
                if empty_line_before != 1 and instructions > 0:
                    print("[{}:{}]: Expected an empty line before".format(file_path, i + 1))
                    errors += 1
            elif line.endswith("}"):
                block_level -= 1
                blocks.pop()
                blocks[-1] += 1
                if block_level < 0:
                    print("[{}:{}]: Parsing error: block level is below 0".format(file_path, i + 1))
                    errors += 1
                if empty_line_before > 0:
                    print("[{}:{}]: Unexpected empty line before block end".format(file_path, i))
                    errors += 1
            elif line.endswith(';'):
                if block_level > 0:
                    instructions += 1
                    reset_instructions_count = False
                if block_level > 0 and empty_line_before > 0 and blocks[-1] == 0:
                    print("[{}:{}]: Unexpected empty line before this one".format(file_path, i + 1))
                    errors += 1

            if reset_empty_line_count is True:
                empty_line_before = 0
            if reset_instructions_count is True:
                instructions = 0
            i += 1
    return errors


def through_folders(path, callbacks):
    errors = 0
    for f in os.listdir(path):
        full_path = os.path.join(path, f)
        if os.path.isdir(full_path):
            errors += through_folders(full_path, callback)
        elif os.path.isfile(full_path):
            for callback in callbacks:
                errors += callback(full_path)
    return errors


def run_main():
    errors = through_folders("templates", [check_scss_file])
    if errors != 0:
        print("{} errors detected, please fix them!".format(errors))
        sys.exit(1)
    print("No error detected. Success!")


if __name__ == "__main__":
    run_main()
```
</details>

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -1,9 +1,7 @@
-
 // FONTS
 $font-family-sans: "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
 $font-family-serif: "Source Serif Pro",Georgia,Times,"Times New Roman",serif;
 $font-family-mono: "Source Code Pro", Menlo, Monaco, Consolas, "DejaVu Sans Mono", Inconsolata, monospace;
-
 
 // COLORS - Guess what?
 $color-standard: #000;             // pure black
@@ -22,7 +20,6 @@ $color-border: #ddd;               // gray
 $color-red: #d93d3d;               // red
 $top-navbar-height: 32px;          // height of the floating top navbar
 
-
 // pure compatible media queries
 $media-sm: "screen and (min-width: 35.5em)";
 $media-md: "screen and (min-width: 48em)";
@@ -30,7 +27,6 @@ $media-lg: "screen and (min-width: 64em)";
 $media-xl: "screen and (min-width: 80em)";
 // usage:
 // @media #{$media-sm} { ... }
-
 
 html, button, input, select, textarea,
 .pure-g [class *= "pure-u"] {
@@ -104,7 +100,6 @@ body {
     padding-top: $top-navbar-height;
 }
 
-
 // this is a super nasty override for help dialog in rustdocs
 // see #52 for details
 body.blur > :not(#help) {
@@ -120,11 +115,9 @@ body.blur > div.rustdoc > :not(#help) {
     opacity: .7;
 }
 
-
 // rustdocs have 200px sidebar and
 // max-width 960px main pane
 // BUT I really want to make the website centered
-
 body {
     text-align: center;
     font: 16px/1.4 $font-family-sans;
@@ -189,10 +182,12 @@ div.nav-container {
             background-color: #fff;
             height: 31px;
             display: none;
+
             @media #{$media-sm} {
                 display: block;
             }
         }
+
         input.search-input-nav:focus {
             outline: unset;
         }
@@ -202,6 +197,7 @@ div.nav-container {
         border: 1px solid $color-border;
         border-radius: 0 0 2px 2px;
         margin-left: -1px;
+
         li {
             border-left: none;
         }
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
pinging someone randomly: @pietroalbini

pietroalbini:
The formatted file looks great, but I'm not sure if it's worth adding a custom linter just for scss files.

GuillaumeGomez:
Considering how bad the file was before I fix the formatting in https://github.com/rust-lang/docs.rs/pull/465/commits/96f00d9ab68814f9407e20407008b68357fe2934, I'd prefer to keep things under surveillance. The check itself is really quick to run and pretty simple making it easy to maintain.

pietroalbini:
Can we use an existing formatter like [prettier](https://prettier.io)? I'd prefer not rolling out our own.

GuillaumeGomez:
I see two reasons for this:
 1. It adds an unnecessary dependency (which itself requires to have a JS environment)
 2. It's way too big for something this small

Again, not sure to really understand your issue... :-/

pietroalbini:
There are a few problems I see with any custom linter (including `./x.py tidy` tbh):

* It only covers a small part of the syntax: as far as I can see `foo:1px` is not linted against, without asking to write it as `foo: 1px`.
* It adds maintenance overhead, as it's more code in the repository.
* It just tells you to fix the problems yourself, while an existing linter hopefully has autoformatting that can be hooked up in the editor.

Of course fixing those problems in a custom linter requires a lot more work than what's worth, while prettier or similar just works. The JS environment is already setup in CI, and prettier is actually nice since it's just a single package without dependencies (so no npm hell).

GuillaumeGomez:
> It only covers a small part of the syntax: as far as I can see `foo:1px` is not linted against, without asking to write it as `foo: 1px`.

I guess it can be debated but for me it's just formatting issue, not readability so completely secondary.

> It adds maintenance overhead, as it's more code in the repository.

Considering I'm the one who wrote it, I don't mind keeping it up-to-date.

> It just tells you to fix the problems yourself, while an existing linter hopefully has autoformatting that can be hooked up in the editor.

Well, that's the whole point of a linter. What you're talking about is a formatter and is generally a lot bigger.

> Of course fixing those problems in a custom linter requires a lot more work than what's worth, while prettier or similar just works. The JS environment is already setup in CI, and prettier is actually nice since it's just a single package without dependencies (so no npm hell).

It still requires to have `npm` and `node` installed locally (which isn't my case) which are, by themselves, a hell of dependencies. ;)

pietroalbini:
I just think if we want to do this we should do it the proper way, without custom scripts.

GuillaumeGomez:
Ok, just kept the scss cleanup. :'(

PR #481, opened 2019-11-20
add updated Source Code Pro to essential files

QuietMisdreavus:
This is a PR to update our builder to accommodate https://github.com/rust-lang/rust/pull/65665. **Do not deploy this** until a nightly is released with that PR (should be 2019-11-21, but confirm ahead of time). Pin the nightly to the current one for now, and unpin it while deploying this change.


pietroalbini:
The change is being reverted.

PR #485, opened 2019-11-24
Replace deprecated try! with `?`

jyn514:
Removes lots of warnings when building. Special thanks to https://github.com/google/rerast which made this much easier :)


GuillaumeGomez:
Thanks!

PR #486, opened 2019-11-25
web: show the priority in the queue

pietroalbini:
One of the tools docs.rs operators have to ensure a fair build queue to everyone is to manually deprioritize slow-to-build crates. Before this the priority wasn't exposed anywhere in the UI though, making the queue look weird when a bunch of crates were deprioritized.

This changes the queue page to show the priority in the build queue if it's not '0'.

r? @jyn514

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -555,14 +555,21 @@ pub fn activity_handler(req: &mut Request) -> IronResult<Response> {
 
 pub fn build_queue_handler(req: &mut Request) -> IronResult<Response> {
     let conn = extension!(req, Pool);
-    let mut crates: Vec<(String, String)> = Vec::new();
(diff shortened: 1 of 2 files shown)


jyn514:
The screenshot and code looks great, but I'm not sure how to change the priority in the site so I can test locally.

pietroalbini:
> but I'm not sure how to change the priority in the site so I can test locally.

Open a psql shell (the password is `password`):

```
docker-compose run db psql -h db -U cratesfyi
```

Then run (remember the priority is inverted for *reasons*):

```sql
UPDATE queue SET priority = 5 WHERE name = 'lazy_static';
```

jyn514:
Going to make a separate pull adding that to the README ;) Thanks!

jyn514:
LGTM

PR #491, opened 2019-11-27
Document proc-macros despite bugs in rustdoc

### LINKED ISSUE #422: Missing docs for procedural macro crate
I noted this behavior yesterday (2019/10/3). After published last version of a crate of mine (`rstest`) I took a look to crate.io to check everything.

Doc link exists https://docs.rs/crate/rstest/0.4.0 but just page is marked as _not a library_ and _Doc_ tab is missed.

I took a look to other procedural macro crates published in the last days and all have the same issue.

I missed something?

Thanks

jyn514:
Use documentation in /target/doc if /target/$target/doc doesn't exist.
In that case, also don't pass `--target` to rustdoc.

Closes #422.

I also moved `build_target` into a separate function since `build_package` was getting kind of big.

r? @QuietMisdreavus

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -26,7 +26,6 @@ static TARGETS: &[&str] = &[
     "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-gnu",
 ];
-static DEFAULT_TARGET: &str = "x86_64-unknown-linux-gnu";
 
 static ESSENTIAL_FILES_VERSIONED: &[&str] = &[
     "brush.svg",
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
Is there an issue on Cargo or rustdoc tracking the problem upstream?

I remember we talked about there sometimes being a problem with it not respecting `--static-root-path`. Does not passing `--target` on the "test build" solve that?

I also see that you don't build any extra targets for proc-macros. This gets around the immediate issue, but i can't shake the feeling that there's going to be a proc-macro somewhere that does something different per-platform. This might work for now, but we should really be reporting this to Cargo so that this can be fixed properly.

jyn514:
> Is there an issue on Cargo or rustdoc tracking the problem upstream?

Yes, https://github.com/rust-lang/rust/issues/66796.

> Does not passing --target on the "test build" solve that?

Yes.

>there's going to be a proc-macro somewhere that does something different per-platform

Probably, and it might make sense come back to this once the upstream issue is fixed, but I think this hits the vast majority of use cases without being terribly hard to change in the future.

jyn514:
Just for help reviewing - you can build an example proc macro with `docker-compose run web build crate rstest 0.4.0`, the compiled docs should look something like this:

image

PR #492, opened 2019-11-27
Extend size of `:target`ed elements

TooManyBees:
When the browser follows a link with a `#fragment`, it scrolls up or
down to put the element with the matching `id` attribute at the top
of the window. In rustdoc output, that's usually section headers.

docs.rs's floating top navbar hides that element. This PR extends
the height of the element the browser is scrolling to, in a way that
doesn't affect its visual appearance, so that it lands below the
navbar and is visible to the user.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -91,6 +91,17 @@ div.rustdoc {
     right: 0;
     top: $top-navbar-height;
     max-height: calc(100vh - #{$top-navbar-height});
+
+    // this prevents the current fragment target (the element that the browser
+    // scrolls to when following a link with a `#fragment`) from hiding under
+    // the navbar bar
+    :target::before {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
It actually has already been fixed in #465. But for some reasons, the `style.css` file we get when browsing docs.rs is an older one...

cc @pietroalbini

jyn514:
@TooManyBees This should have been fixed in https://github.com/rust-lang/docs.rs/pull/496, can you take a look at the deployed site (e.g. https://docs.rs/parze/0.5.1/parze/also/trait.Also.html#tymethod.also) and make sure it looks good? If so I'll close this.

TooManyBees:
i guess it's already fixed then

PR #496, opened 2019-11-27
Improve the CSS situation in the repository

pietroalbini:
This PR fixes some of the CSS mess contained in this repository:

* Removed `css/style.css`, as it was just an out of date copy of the generated `templates/style.scss`. The file was actually never served to users, as the SCSS generated on the fly takes precedence.
* Removed the `1.40.0-nightly` vendored CSS files, as they're not used by anything.
* Removed the `1.10.0-nightly` vendored CSS files, and changed the application to use the latest nightly's essential files instead of the 1.10 ones. This required a change in the Docker entrypoint to load the essential files the first time the container is started, and a few tweaks to our `style.scss` to workaround the changes in Rustdoc's CSS since 1.10.
* Moved `css/` to `static/` in the repository, as it now only contains `favicon.ico`.

r? @jyn514


pietroalbini:
cc @GuillaumeGomez as well

jyn514:
Can you also apply https://github.com/rust-lang/docs.rs/pull/465/commits/60ded3b7c061a1d4f7305bf46c34c7d74c058f9a to `style.scss`?

src/web/page.rs
```diff
@@ -6,6 +6,27 @@ use iron::{IronResult, Set, status};
...
+        .unwrap_or_else(|_| "???".into());
+}
+
+fn load_rustc_resource_suffix() -> Result<String, failure::Error> {
+    let conn = crate::db::connect_db()?;
+
+    let res = conn.query("SELECT value FROM config WHERE name = 'rustc_version';", &[])?;
+    if res.is_empty() {
+        failure::bail!("missing rustc version");
+    }
```
jyn514: It might be better to have this call `add_essential_files` instead of making it a manual step. You could move the logic you added above here to make sure it's not being called more than once.

pietroalbini: I'd prefer not to do that. In the future I'd like to see the frontend being a separate binary running on a separate server, instead of having both the frontend and the build daemon on the same machine. `add-essential-files` requires the build daemon, so if I make this change I would make splitting the two harder.

jyn514: Makes sense, I'd also like to see those split up at some point so the webserver isn't running as root.

pietroalbini:
> Can you also apply 60ded3b to style.scss?

Ported the configuration.

jyn514:
Hmm, I see your changes in the CSS but the toolbar still overlaps with the item ... but that's separate from your changes here, we can fix it later.

image

jyn514:
ah wait I'm on the wrong git branch lol one sec

jyn514:
Ok yeah this looks great!

PR #498, opened 2019-11-27
Give focus to rustdoc main content if available

GuillaumeGomez:
Users will be able to scroll again on the content!


jyn514:
I can't replicate this fix locally, I still can't scroll up or down without clicking on the main window. My firefox install is pretty buggy in general though, can someone else test this out?

GuillaumeGomez:
I picked the upper level element (both works in my case but maybe this one will work in your case as well?).

Koxiaet:
I also encountered this problem but added a different fix - [510](https://github.com/rust-lang/docs.rs/pull/510). Perhaps this one will work better.

jyn514:
Closing in favor of #579

Issue #504, opened 2019-12-02
Link to crates-build-env instead of asking people to make an issue


jyn514:
From Discord:

> re build failures due to dependencies: I think I'm going to write up a wiki page showing how to make a PR to https://github.com/rust-lang/crates-build-env. Hopefully we can then link to that instead of telling people to make an issue and it will be less work for everyone

Going to start with #483 just so I know how to make a PR, then I'll add it to the wiki

jyn514:
Hmm ... for this to work there needs to be a way to test changes to the `rustops/crates-build-env` container locally. Maybe add a flag to `cratesfyi build crate` that specifies a custom docker container?

jyn514:
Maybe we don't even have to go that far ... it's running a docker command, we could have people run that docker command locally. Here's the one for `arpack-ng-sys`:

```
"docker" "create" "-v" "/home/cratesfyi/workspace/builds/arpack-ng-sys-0.1.0/target:/opt/rustwide/target:rw,Z" "-v" "/home/cratesfyi/workspace/builds/arpack-ng-sys-0.1.0/source:/opt/rustwide/workdir:ro,Z" "-v" "/home/cratesfyi/workspace/cargo-home:/opt/rustwide/cargo-home:ro,Z" "-v" "/home/cratesfyi/workspace/rustup-home:/opt/rustwide/rustup-home:ro,Z" "-e" "SOURCE_DIR=/opt/rustwide/workdir" "-e" "MAP_USER_ID=1001" "-e" "CARGO_TARGET_DIR=/opt/rustwide/target" "-e" "RUSTFLAGS=" "-e" "RUSTDOCFLAGS=-Z unstable-options --resource-suffix -20191116-1.41.0-nightly-5c5b8afd8 --static-root-path / --disable-per-crate-search --extern-html-root-url bindgen=https://docs.rs/bindgen/0.52.0 --extern-html-root-url cmake=https://docs.rs/cmake/0.1.42" "-e" "CARGO_HOME=/opt/rustwide/cargo-home" "-e" "RUSTUP_HOME=/opt/rustwide/rustup-home" "-w" "/opt/rustwide/workdir" "-m" "3221225472" "--network" "none" "rustops/crates-build-env" "/opt/rustwide/cargo-home/bin/cargo" "+nightly-2019-11-17" "doc" "--lib" "--no-deps" "--target" "x86_64-unknown-linux-gnu"
```

The only bits that need to change are the mounted directories:
[...]

Issue #483, opened 2019-11-22
No Fortran compiler available


SolidTux:
**Crate name:** arpack-ng-sys
**Build failure link:** https://docs.rs/crate/arpack-ng-sys/0.1.0/builds/200378

**Additional details:** arpack-ng needs a Fortran compiler for building.

Would it be possible to make gfortran available?

jyn514:
Sorry about that, apparently github automatically closes mentioned issues even if the changes haven't been merged yet.

jyn514:
This is waiting for the new docker image to be uploaded; after that's finished I can queue a rebuild.

jyn514:
This is built!

PR #507, opened 2019-12-07
Link to docs for changing the build environment in issue templates

jyn514:
Addresses https://github.com/rust-lang/docs.rs/issues/504

The motivation here is to reduce the maintenance burder on the docs.rs team by having people submit PRs to https://github.com/rust-lang/crates-build-env directly. This has 2 main benefits:

1. Crate authors already know the packages they need, so they don't have to tell us and have us make a PR ourselves; this removes an intermediate step.
2. If crate authors are wrong about the packages they need (for instance, https://github.com/rust-lang/docs.rs/issues/483 needed liblapack-dev, not just gfortran), they get immediate feedback instead of having to wait for a deploy and republish their package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,8 +159,8 @@ docker-compose run -p 3000:3000 web daemon --foreground
 
 ### Changing the build environment
 
-To make a change to https://github.com/rust-lang/crates-build-env
-and test that it works on docs.rs, see https://github.com/rust-lang/docs.rs/wiki/Making-changes-to-the-build-environment.
+To make a change to [the build environment](https://github.com/rust-lang/crates-build-env)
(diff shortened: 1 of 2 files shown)


jyn514:
Note that this only changes the issue templates, not the links on a failed page.

README.md
```diff
@@ -157,6 +157,11 @@ docker exec -it <the container name goes here> psql -U cratesfyi
 docker-compose run -p 3000:3000 web daemon --foreground
 ```
 
+### Changing the build environment
+
+To make a change to https://github.com/rust-lang/crates-build-env
```
pietroalbini: Maybe we want to have nice link texts here? Something like "To make a change to [the build environment] and test that it works on docs.rs, see [the wiki page]"

jyn514: Added this in the next commit.

jyn514:
This is ready for review.

pietroalbini:
Thanks!

Issue #503, opened 2019-12-01
rlib and cdylib crates are marked as 'not a library'


jyn514:
Looking at https://doc.rust-lang.org/reference/linkage.html, everything except `bin` is a type of library. I ran into @MaulingMonkey on discord who was confused why his crate wasn't building when he had `rlib` and `cdylib` set: https://docs.rs/crate/bugsalot/0.1.5/builds/202497.

MaulingMonkey:
Seems to have been a regression sometime between rust 1.35 and rust 1.38+ (which also seems to try and include Cargo.lock under the impression that the crate isn't a library) since the manifests didn't change at all besides the version:

https://docs.rs/crate/bugsalot/0.1.4/source/Cargo.toml.orig (published with rust 1.35)
https://docs.rs/crate/bugsalot/0.1.5/source/Cargo.toml.orig (published with rust 1.39.  I get similar behavior from `cargo package` - including the `Cargo.lock` file - in 1.38, but not in 1.35.)

`rlib` + `cdylib` is recommended by [wasm-pack docs](https://rustwasm.github.io/docs/wasm-pack/tutorials/npm-browser-packages/template-deep-dive/cargo-toml.html#1-crate-type) - I'm not sure how necessary it is for leaf crates like mine though.  I do have a workaround:  switching from `rlib` to `lib`.

(As a side note, if I try to use both `rlib` and `lib` simultaniously, instead of replacing the former with the latter, it produces a panic in rust 1.39:)
```
warning: output filename collision.
The lib target `bugsalot` in package `bugsalot v0.1.6 (C:\local\bugsalot)` has the same output filename as the lib target `bugsalot` in package `bugsalot v0.1.6 (C:\local\bugsalot)`.
Colliding filename is: c:\local\bugsalot\target\x86_64-pc-windows-msvc\debug\deps\libbugsalot.rlib
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
thread 'main' panicked at 'assertion failed: mtimes.insert(output.clone(), mtime).is_none()', src\tools\cargo\src/cargo\core\compiler\fingerprint.rs:795:13
...bogus looking stack backtrace...
```

MaulingMonkey:
Okay, apparently I lied about `"lib"` being a workaround.  I tried this for https://docs.rs/crate/bugsalot/0.1.6/ but still no docs.

[...]

PR #508, opened 2019-12-07
Mark all crates as libraries that aren't bins

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/503

See that issue for context on why `kind` was changed to `crate_types`.

r? @pietroalbini


pietroalbini:
Thanks!

Issue #502, opened 2019-11-30
Redirect to latest version does not work if build failed


jyn514:
This is actually a combination of a number of issues:

1. If you go to https://docs.rs/isatty/0.1.0/isatty/, the link to the latest version takes you to https://docs.rs/isatty/0.2.0/isatty/?search=isatty
2. isatty 0.2.0 failed its build
3. /isatty/0.2.0/isatty/ gives a 404

https://github.com/rust-lang/docs.rs/pull/325 fixed this for /isatty/0.2.0/isatty, so the issue is that /:crate/:version/:target/ behaves differently than /:crate/:version/:target.

We can't just change the redirect to link to `/:crate/:version/:target`, since `?search=` doesn't work on that page. Maybe the easiest thing would be to link to `/:target` and have the redirect to `/:target/` keep the search query parameter?

jyn514:
Perhaps a better thing to do is not show 'go to latest version' for failed builds? That would hit the majority of use cases. For this particular case though I think we really should show 0.2.0 so people see the 'this crate is deprecated' message.

PR #509, opened 2019-12-07
Fix redirects if latest version failed to build

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/502

The redirector behaved differently for `/:crate/:version/:target` than
for `/:crate/:version/:target/`. This changes the latest version link to
use the former, since it properly accounts for failed builds.

Additionally, this changes the `/:target` redirect to keep the query
parameter so that `?search=` URLs will be kept when going to `/:target/`.

r? @QuietMisdreavus 

Some good crates to test this with:
- isatty 0.1.0, 0.2.0 (0.2.0 failed to build)
- pyo3 0.2.7, 0.8.2 (moved pyo3::exc::ArithmeticError to a different module)

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -43,10 +43,16 @@ pub(crate) fn assert_redirect(path: &str, expected_target: &str, web: &TestFront
     let response = web.get(path).send()?;
     let status = response.status();
 
+    let mut tmp;
     let redirect_target = if expected_target.starts_with("https://") {
         response.url().as_str()
     } else {
-        response.url().path()
+        tmp = String::from(response.url().path());
+        if let Some(query) = response.url().query() {
+            tmp.push('?');
+            tmp.push_str(query);
+        }
+        &tmp
     };
     // Either we followed a redirect to the wrong place, or there was no redirect
     if redirect_target != expected_target {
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
So, to double check what this PR does:

- Changes the "jump to latest version" redirect to not have a trailing slash before adding the search query
- Changes the `/:crate`/`/:crate/:version` redirect to include a query string in the redirect.

Would it be possible to check for the presence of the query string and only include the `?` in the URL if it's already there? I get a little annoyed when URLs have an unnecessary `?` or `#` sometimes.

jyn514:
Yes, that's right. I can change it to remove the '?' if there's no query, give me a second.

jyn514:
Ok should be fixed.

jyn514:
Not sure why github actions is failing. The other build/test CI is working.

jyn514:
This needs unit tests, I'll add them hopefully this weekend.

jyn514:
This depends on https://github.com/rust-lang/docs.rs/pull/570/files for the `latest_version_redirect` that was added there.

jyn514:
This is ready for review.

kvrhdn:
Sure @jyn514, I should be able to check out these changes this weekend 🙂

PR #510, opened 2019-12-07
Fix scrolling on some browsers

Koxiaet:
By making the body of rustdoc position: relative instead of fixed, scrolling on many browsers (like qutebrowser and iOS Safari) works like it should. This change has no impact on how the page looks (although you might want to test it some more).

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -85,14 +85,7 @@ div.rustdoc {
         }
     }
 
-    // this is actual fix for docs.rs navigation and rustdoc sidebar
-    position: fixed;
-    top: $top-navbar-height;
-    bottom: 0;
-    left: 0;
-    right: 0;
-    display: block;
-    overflow-y: auto;
+    position: relative;
 }
 
 body {


Koxiaet:
I just realized that this attempts to fix the same problem as [498](https://github.com/rust-lang/docs.rs/pull/498), but this fixes the actual problem instead of adding a workaround.

jyn514:
> I just realized that this attempts to fix the same problem as 498, but this fixes the actual problem instead of adding a workaround.

That's ok, I like solutions without JavaScript better :)

GuillaumeGomez:
Fixing it this way would bring back the issue that when going to an anchor, the anchor will be under the top navbar and therefore invisible.

Koxiaet:
We could add the following JavaScript that gets the best of both worlds (although this is very hacky):
```
const links = Array.from(document.getElementsByTagName("a")).filter(link => link.href.includes("#"));
for link of links {
    link.addEventListener("click", _ =>
        requestAnimationFrame(_ => scrollBy(0, -50))
    );
}
if (location.hash) {
    scrollBy(0, -50);
}
```
This means that people with JS disabled would still get the above mentioned issue, and it's also bad practice in general to use JS for hacks like this, but I think there aren't many better alternatives.

GuillaumeGomez:
That's worse from my point of view... However I know that we have another thing going on which will have as side-effect to fix this issue.

jyn514:
Closing in favor of #579

Koxiaet:
@GuillaumeGomez I still think that my solution (with the JS) is the best solution to the problem - #579 doesn't actually work for me. I (and probably many other Rust users) use a keyboard-based browser, and on most webpages I can use my regular keyboard shortcuts to navigate the main content, but not docs.rs as it is `position: fixed`. Maybe I should file an issue with the browser instead, but also from a CSS-philosophy and web accessibility point of view, having the _main content_ be `position: fixed` is not good practice. Moreover on some browsers auto-focusing the content gives an ugly yellow border around it.

Sorry about bringing this issue back, but I just really want to use my keyboard shortcuts again!

PR #514, opened 2019-12-09
Refactor the router

pietroalbini:
:warning::warning: **This PR is already deployed to production!** :warning::warning:

This PR refactors the router to use a custom wrapper. In addition to registering the route, the wrapper:

- For internal pages (like `/about`, `/releases/queue`...), it setups a redirect from the URL with a trailing slash to the URL without it.

- It registers all the prefixes of internal pages (`/about`, `/releases`, `/crate`) and blacklists those prefixes from the rustdoc routes. This prevents, for example, the `/about/0.1.0/` page from being hijacked by the author of the about crate.

r? @QuietMisdreavus

### DIFF
--- a/src/web/routes.rs
+++ b/src/web/routes.rs
@@ -71,7 +71,7 @@ pub(super) fn build_routes() -> Routes {
         super::builds::build_list_handler,
     );
     routes.internal_page(
-        "/crate/:name/:version/source",
+        "/crate/:name/:version/source/",
         super::source::source_browser_handler,
     );
     routes.internal_page(
--- a/src/web/routes.rs
+++ b/src/web/routes.rs
@@ -71,6 +71,10 @@ pub(super) fn build_routes() -> Routes {
         super::builds::build_list_handler,
     );
     routes.internal_page(
+        "/crate/:name/:version/source",
+        SimpleRedirect::new(|url| url.set_path(&format!("{}/", url.path()))),
+    );
+    routes.internal_page(
         "/crate/:name/:version/source/",
         super::source::source_browser_handler,
     );
@@ -180,8 +184,12 @@ impl Routes {
         // Automatically add another route ending with / that redirects to the slash-less route.
         if !pattern.ends_with('/') {
             let pattern = format!("{}/", pattern);
-            self.get
-                .push((pattern.to_string(), Box::new(remove_slashes)));
+            self.get.push((
+                pattern.to_string(),
+                Box::new(SimpleRedirect::new(|url| {
+                    url.set_path(&url.path().trim_end_matches('/').to_string())
+                })),
+            ));
         }
 
         // Register the prefix if it's not the home page and the first path component is not a
@@ -204,15 +212,26 @@ impl Routes {
     }
 }
 
-/// Simple Iron route that redirects to the same path without the trailing slash, for example
-/// /foo/bar/baz/ to /foo/bar/baz.
-fn remove_slashes(req: &mut iron::Request) -> iron::IronResult<iron::Response> {
-    let mut url: iron::url::Url = req.url.clone().into();
-    url.set_path(&url.path().trim_end_matches('/').to_string());
-    Ok(iron::Response::with((
-        iron::status::Found,
-        iron::modifiers::Redirect(iron::Url::parse(&url.to_string()).unwrap()),
-    )))
+#[derive(Copy, Clone)]
+struct SimpleRedirect {
+    url_mangler: fn(&mut iron::url::Url),
+}
+
+impl SimpleRedirect {
+    fn new(url_mangler: fn(&mut iron::url::Url)) -> Self {
+        Self { url_mangler }
+    }
+}
+
+impl Handler for SimpleRedirect {
+    fn handle(&self, req: &mut iron::Request) -> iron::IronResult<iron::Response> {
+        let mut url: iron::url::Url = req.url.clone().into();
+        (self.url_mangler)(&mut url);
+        Ok(iron::Response::with((
+            iron::status::Found,
+            iron::modifiers::Redirect(iron::Url::parse(&url.to_string()).unwrap()),
+        )))
+    }
 }
 
 /// Iron Middleware that prevents requests to blacklisted prefixes.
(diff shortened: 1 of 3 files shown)


pietroalbini:
Thanks to @xampprocky for disclosing this issue according to our [security policy](https://www.rust-lang.org/policies/security).

jyn514:
I know you asked Q to review but if I could chime in:

This fixes the problem in the short term, but doesn't let us add new routes without discarding the documentation from entire crates. If we want to add say a /builds route, that requires blacklisting the `builds` crate.

Would it be possible to show the documentation for these crates under /crate/:crate/docs or something similar? That would allow us to add routes in the future without breaking existing docs.

pietroalbini:
> This fixes the problem in the short term, but doesn't let us add new routes without discarding the documentation from entire crates. If we want to add say a /builds route, that requires blacklisting the builds crate.
>
> Would it be possible to show the documentation for blacklisted crates under /crate/:crate/docs or something similar? That would allow us to add routes in the future without breaking existing docs.

I think adding our own routes in the same namespace as the crates is something we don't want to do ever again. Even if we implement the alternate rustdoc route for those crates, we're still breaking all the links pointing to it before we added the route.

The approach I'd go with is to use `_` as the prefix for all *new* internal routes (for example `https://docs.rs/_/api`). It's not possible at all to register a crate with that name, and it doesn't look that bad.

src/web/routes.rs
```diff
@@ -0,0 +1,257 @@
...
+                ));
+            }
+        }
+        self.handler.handle(req)
+    }
+}
+
+/// Automatically generate a Route ID from a pattern. Every non-alphanumeric character is replaced
+/// with `_`.
+fn calculate_id(pattern: &str) -> String {
```
jyn514: This is useful, the first mistake I always see new contributors making is duplicate route IDs so it's nice to avoid it automatically :)

src/web/routes.rs
```diff
@@ -0,0 +1,257 @@
...
+    rustdoc_get: Vec<(String, Box<dyn Handler>)>,
+    page_prefixes: HashSet<String>,
+}
+
+impl Routes {
+    fn new() -> Self {
+        Self {
+            get: Vec::new(),
+            rustdoc_get: Vec::new(),
+            page_prefixes: HashSet::new(),
```
jyn514: Could you add some comments here? Something like "`rustdoc_get`s can be blacklisted while `get`s are always served as-is"? I had to read through a bit of code to figure out what these were for.

src/web/routes.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    pub(super) fn page_prefixes(&self) -> HashSet<String> {
+        self.page_prefixes.clone()
+    }
+
+    pub(super) fn iron_router(mut self) -> Router {
```
jyn514: Oh I see, this whole `Routes` struct is a builder for `Router`. That would be nice to have in comments up at the top.

src/web/routes.rs
```diff
@@ -0,0 +1,257 @@
...
+            let pattern = format!("{}/", pattern);
+            self.get
+                .push((pattern.to_string(), Box::new(remove_slashes)));
+        }
+
+        // Register the prefix if it's not the home page and the first path component is not a
+        // pattern or a wildcard.
+        if pattern != "/" {
+            if let Some(first_component) = pattern.trim_matches('/').split('/').next() {
+                if !first_component.contains('*') && !first_component.starts_with(':') {
```
jyn514: I'm not quite sure what this is doing ... looking at the places `internal_page` is called it seems none of the prefixes meet this condition.

src/web/routes.rs
```diff
@@ -71,7 +71,7 @@ pub(super) fn build_routes() -> Routes {
         super::builds::build_list_handler,
     );
     routes.internal_page(
-        "/crate/:name/:version/source",
+        "/crate/:name/:version/source/",
```
jyn514: Why was this change made? This seems to prevent navigating to the slash-less version, e.g. https://docs.rs/crate/rcc/0.2.0/source now gives a 404.

pickfire: Maybe a backward-compatible redirect would be nice? IIRC I have experienced docs.rs breaking backward compatibility.

src/web/routes.rs
pietroalbini: This code is actually not needed right now, but it's a safeguard if in the future someone tries to add an internal page that starts with a wildcard.

src/web/routes.rs
pietroalbini: Because the contents of the source page rely on a slash being at the end of the URL. I can add a redirect from slash-less to the URL with the slash.

src/web/routes.rs
pietroalbini: It's actually explained in the comment above, is that enough?

jyn514: Yes, I think that's fine.

pietroalbini:

src/web/routes.rs
```diff
@@ -204,15 +220,26 @@ impl Routes {
...
-fn remove_slashes(req: &mut iron::Request) -> iron::IronResult<iron::Response> {
-    let mut url: iron::url::Url = req.url.clone().into();
-    url.set_path(&url.path().trim_end_matches('/').to_string());
-    Ok(iron::Response::with((
-        iron::status::Found,
-        iron::modifiers::Redirect(iron::Url::parse(&url.to_string()).unwrap()),
-    )))
+#[derive(Copy, Clone)]
+struct SimpleRedirect {
+    url_mangler: fn(&mut iron::url::Url),
```
jyn514: Not super important, but is there a reason you made this a struct instead of a trait with a default implementation for `Handler`? That would allow static instead of dynamic dispatch.

jyn514: Ignore this - if it's a trait than it's really annoying to call `SimpleRedirect::new`

pietroalbini: Hmm? I'm not sure I follow what you're proposing.

jyn514:
Looks great, going to test locally real quick and then merge :)

PR #518, opened 2019-12-12
install nightly before checking what version it is

jyn514:
fixes crashes on the first run

r? @pietroalbini

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -36,6 +36,6 @@ if ! [ -d "${CRATESFYI_PREFIX}/crates.io-index/.git" ]; then
     git --git-dir="$CRATESFYI_PREFIX/crates.io-index/.git" branch crates-index-diff_last-seen
 fi
 
-cratesfyi build add-essential-files --only-first-time
+cratesfyi build update-toolchain --only-first-time
 
 cratesfyi "$@"
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -182,7 +182,7 @@ pub fn main() {
             }
 
             let mut builder = RustwideBuilder::init().unwrap();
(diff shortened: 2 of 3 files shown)


pietroalbini:
`update_toolchain` above is supposed to install the release though. Can you make it return an `Option<String>` and handle the `None` around the codebase?

jyn514:
Ah that's the issue, main calls `add_essential_files` instead of `update_toolchain`: https://github.com/rust-lang/docs.rs/blob/master/src/bin/cratesfyi.rs#L185

jyn514:
Ok I fixed that instead, @pietroalbini could you review again?

pietroalbini:
So, I think having the `add-essential-files` command unconditionally adding the essential files is something we still want: it happens that something is missing or wrong with them, and in those cases I don't want to fiddle with the database by removing the cached rustc version to make the command work.

Still, that doesn't mean the `add-essential-files` is the right command to call in the entrypoint: what I would do is creating a new `update-toolchain` command which calls the `update_toolchain` method, moving the `--only-first-time` logic to it, and changing the entrypoint to use it.

jyn514:
I made the changes you suggested, but it got me thinking - do we need the toolchain to be installed just for the web server? It looks like the server only needs access to the database, not to `rustdoc` or anything similar. `cratesfyi build crate` already runs `update_toolchain` when called: https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L225

pietroalbini:
This looks great! Thanks!

> I made the changes you suggested, but it got me thinking - do we need the toolchain to be installed just for the web server?

Yeah, because we need to have the essential files to render the pages, as every page includes rustdoc's CSS.

pietroalbini:
This was reverted due to #521.

Issue #517, opened 2019-12-12
panic when releases.files is NULL


jyn514:
Caught this from looking through the logs on the build server: https://docs.rs/crate/rips/0.2.0/source/ gives a 500 when you visit it. The build for `rips` failed, maybe that's why the files are NULL?

Backtrace:

```
web_1  | thread '<unnamed>' panicked at 'error retrieving column 5: Error(Conversion(WasNull))', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/postgres-0.15.2/src/rows.rs:201:31
web_1  | stack backtrace:
web_1  |   13: postgres::rows::Row::get
web_1  |   14: cratesfyi::web::source::source_browser_handler
web_1  |   15: <F as iron::middleware::Handler>::handle
web_1  |   16: <alloc::boxed::Box<dyn iron::middleware::Handler> as iron::middleware::Handler>::handle
web_1  |   17: router::router::Router::handle_method
web_1  |   18: <router::router::Router as iron::middleware::Handler>::handle
web_1  |   19: iron::middleware::Chain::continue_from_handler
web_1  |   20: <iron::middleware::Chain as iron::middleware::Handler>::handle
web_1  |   21: <alloc::boxed::Box<dyn iron::middleware::Handler> as iron::middleware::Handler>::handle
web_1  |   22: <cratesfyi::web::CratesfyiHandler as iron::middleware::Handler>::handle
web_1  |   23: <iron::iron::RawHandler<H> as hyper::server::Handler>::handle
web_1  |   24: hyper::server::Worker<H>::handle_connection
web_1  |   25: hyper::server::listener::spawn_with::{{closure}}
```

Note for maintainers: to find other crates with this issue, run this psql command on the server:

`SELECT releases.target_name, releases.version FROM releases WHERE releases.files IS NULL;`

jyn514:
It looks like this happens whenever the build fails.

jyn514:
Relevant code: https://github.com/rust-lang/docs.rs/blob/master/src/web/source.rs#L107

jyn514:
Hmm ... I worked around this by returning `None` in the linked code, which avoids the panic, but the rendered page is still mostly blank :(

![mostly-blank_1920x1080](https://user-images.githubusercontent.com/23638587/70746155-6710b400-1cf3-11ea-9f0f-35c02843e8c7.png)

jyn514:
I turned it into a 404 instead, at least we're giving _an_ error message now.

[...]

PR #519, opened 2019-12-12
Fix panic viewing source if crate failed to build

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/517

Gives a 404 if you try to view the source instead of panicking


pietroalbini:
This looks great! There is still the problem that the "Source" tab is present even when we know it will lead to a 404, but I opened https://github.com/rust-lang/docs.rs/issues/520 to follow up with that.

PR #523, opened 2019-12-14
Don't ignore DB cache if local cache is empty

jyn514:
This greatly speeds up builds using `--skip` + docker-compose, for which the local cache will _always_ be empty.

This will be helpful for speeding up CI builds, which will (for the most part) test the functionality of the website, not the builder.


pietroalbini:
Thanks!

PR #526, opened 2019-12-16
Fix panics when viewing source of failed build (redux)

jyn514:
This reverts all the changes from #519 except for actually fixing the panic. I tested this with directory listings as well as individual files:

- If no files exist, gives a 200 with an empty page (not ideal but better than the panic + blank screen from before). This can be tested with e.g. pam-sys 1.0.0-alpha1.
- If a directory listing exists, show that. Can be tested with most crates, e.g. rand 0.7.2.
- If a file exists, show that (no logic changed)

r? @pietroalbini

### DIFF
--- a/src/web/source.rs
+++ b/src/web/source.rs
@@ -104,7 +104,7 @@ impl FileList {
             return None;
         }
 
-        let files: Json = rows.get(0).get(5);
+        let files: Json = rows.get(0).get_opt(5).unwrap().ok()?;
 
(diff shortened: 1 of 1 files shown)


pietroalbini:
Can we still show a not found page when `None` is returned?

jyn514:
I _think_ this only shows a 404 for binaries and not libraries, but I'd appreciate you testing locally too after last time :laughing:

pietroalbini:
This looks great!

Issue #528, opened 2019-12-18
Enable removal/access restriction for specific versions of a crate, rather than the entire crate


carols10cents:
Currently, [the way to remove crates](https://github.com/rust-lang/crates-io-cargo-teams/blob/master/docs/crate-removal-procedure.md#remove-relevant-versions-andor-entire-crates-from-docsrs) is to edit the nginx file and block the whole crate:

```
location /CRATENAME {
			return 410;
		}
		location /crate/CRATENAME {
			return 410;
		}
```

There are reasons for needing to take down some versions and not others, so docs.rs needs a way to do that.

Could it be as simple as doing this in the nginx config instead...?

```
location /CRATENAME/VERSION {
    return 410;
}
```

@pietroalbini I know you already know about this, but I wanted an issue for tracking and I couldn't find an existing one.

pietroalbini:
I'm working on a command like this:

```
cratesfyi database remove-crate CRATENAME
```

...which properly removes the crate from the database and from S3, so the nginx stuff isn't required anymore.

jyn514:
@pietroalbini delete-crate was a good first step, but I think it only allowed removing the entire crate, not a specific version.  I think this issue should be kept open until we add that functionality.

jyn514:
#893 did not actually work, re-opening until I fix it.
```
thread 'main' panicked at 'failed to delete the crate: Unknown(BufferedHttpResponse {status: 400, body: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>MalformedXML</Code><Message>The XML you provided was not well-formed or did not validate against our published schema</Message><RequestId>3A0407C0B2410DCE</RequestId><HostId>WA3p1NxP9ZbBH/PyDUlcaLVbYwH3mv1FPC1aCoeV8GJ62YLLcY1AS8C0UfdCjj8H1RhGYCSI21s=</HostId></Error>", headers: {"x-amz-request-id": "3A0407C0B2410DCE", "x-amz-id-2": "WA3p1NxP9ZbBH/PyDUlcaLVbYwH3mv1FPC1aCoeV8GJ62YLLcY1AS8C0UfdCjj8H1RhGYCSI21s=", "content-type": "application/xml", "transfer-encoding": "chunked", "date": "Thu, 16 Jul 2020 14:37:33 GMT", "connection": "close", "server": "AmazonS3"} })

   0: failure::backtrace::internal::InternalBacktrace::new
[...]

PR #529, opened 2019-12-18
delete-crate command

pietroalbini:
This PR adds a new command to the CLI:

```
cratesfyi database delete-crate CRATE_NAME
```

The command will remove all the published versions of a crate, both from the database and S3.

Fixes https://github.com/rust-lang/docs.rs/issues/528


jyn514:
> The command will remove all the published versions of a crate, both from the database and S3.

I think someone mentioned removing only specific versions at some point, is that still something that's planned? That doesn't need to block this PR, but it would be nice to keep in the background.

pietroalbini:
> I think someone mentioned removing only specific versions at some point, is that still something that's planned? That doesn't need to block this PR, but it would be nice to keep in the background.

I don't think we need this at the moment, we can think about that when it's needed.

pietroalbini:
This is ready for review.

src/db/delete_crate.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn delete_from_database(conn: &Connection, name: &str, crate_id: i32) -> Result<(), Error> {
+    let transaction = conn.transaction()?;
+
+    transaction.execute(
+        "DELETE FROM sandbox_overrides WHERE crate_name = $1",
+        &[&name],
+    )?;
+    transaction.execute(
```
jyn514: I was looking for a way to use [`batch_execute`](https://docs.rs/postgres/0.15.2/postgres/transaction/struct.Transaction.html#method.batch_execute) with parameterized queries but it doesn't look like it's possible. Maybe we could query the crate id once and use `batch_execute` with a `format!` string? That doesn't seem too outrageous ... if you don't think it's worth the effort it's not a big deal though.

src/db/delete_crate.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use failure::Error;
+    use postgres::Connection;
+
+    #[test]
+    fn test_delete_from_database() {
```
jyn514: This is amazing, thank you so much for working on this!

jyn514: This is amazing, thank you so much for working on this!

.github/workflows/ci.yml
```diff
@@ -16,6 +16,14 @@ jobs:
...
         run: rustup update stable && rustup default stable
 
+      - name: Install PostgreSQL
+        run: |
+          sudo apt-get update && DEBIAN_FRONTEND=noninteractive sudo apt-get install -y postgresql
+          sudo systemctl start postgresql
+          sudo -u postgres createuser $(whoami) -w
+          sudo -u postgres createdb $(whoami) -O $(whoami)
+          echo "::set-env name=CRATESFYI_DATABASE_URL::postgresql://$(whoami)@%2Fvar%2Frun%2Fpostgresql/$(whoami)"
+
```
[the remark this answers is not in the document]
pietroalbini: It sets the `CRATESFYI_DATABASE_URL` environment variable in all future steps executed inside the job to the value after the second `::`. GitHub Actions (and Azure Pipelines, with a different syntax) calls this a [logging command](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#logging-commands)

src/db/add_package.rs
```diff
@@ -36,9 +38,7 @@ pub(crate) fn add_package_into_database(conn: &Connection,
     let dependencies = convert_dependencies(metadata_pkg);
     let rustdoc = get_rustdoc(metadata_pkg, source_dir).unwrap_or(None);
     let readme = get_readme(metadata_pkg, source_dir).unwrap_or(None);
-    let (release_time, yanked, downloads) = get_release_time_yanked_downloads(metadata_pkg)?;
```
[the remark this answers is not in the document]
pietroalbini: `get_release_time_yanked_downloads` calls crates.io's API: that prevents us from mocking those values or using crate names missing on crates.io (as the API returns an error). Also it's not a good practice to use the network in tests unless strictly necessary.

src/db/add_package.rs
```diff
@@ -435,25 +463,38 @@ fn add_owners_into_database(conn: &Connection, pkg: &MetadataPackage, crate_id:
...
+/// Adds owners into database
+fn add_owners_into_database(conn: &Connection, owners: &[CrateOwner], crate_id: &i32) -> Result<()> {
+    for owner in owners {
+        let owner_id: i32 = {
+            let rows = conn.query("SELECT id FROM owners WHERE login = $1", &[&owner.login])?;
+            if rows.len() > 0 {
+                rows.get(0).get(0)
+            } else {
+                conn.query("INSERT INTO owners (login, avatar, name, email)
+                                 VALUES ($1, $2, $3, $4)
```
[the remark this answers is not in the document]
pietroalbini: This is not my code, I just moved this code into another function. Could you open an issue to fix this function?

src/db/add_package.rs
```diff
@@ -435,25 +463,38 @@ fn add_owners_into_database(conn: &Connection, pkg: &MetadataPackage, crate_id:
...
+            });
         }
+    }
+
+    Ok(result)
+}
+
+/// Adds owners into database
+fn add_owners_into_database(conn: &Connection, owners: &[CrateOwner], crate_id: &i32) -> Result<()> {
+    for owner in owners {
```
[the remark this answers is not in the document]
pietroalbini: This is not my code, I just moved this code into another function. Could you open an issue to fix this function?

src/db/delete_crate.rs
pietroalbini: I'd prefer not to use `format!()`s just in case. Also optimizing 6 queries into one for a CLI program called every once in a few months doesn't seem too big of a priority.

pietroalbini:
Replied to or addressed all the comments.

src/db/add_package.rs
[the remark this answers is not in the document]
jyn514: Gotcha, this allows you to pass in data without making a network call.

src/db/add_package.rs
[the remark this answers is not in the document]
jyn514: It looks like `slug` is a `UNIQUE` constraint anyway so it shouldn't be an issue: https://github.com/rust-lang/docs.rs/blob/master/src/db/migrate.rs#L109

src/db/add_package.rs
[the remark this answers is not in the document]
jyn514: Done, #530.

src/db/migrate.rs
```diff
@@ -1,11 +1,29 @@
...
+use std::rc::Rc;
+
+
+enum ApplyMode {
+    Permanent,
+    Temporary,
+}
+
+struct MigrationContext {
+    apply_mode: ApplyMode,
```
jyn514: I saw lots of calls to `.clone()`, I pushed a commit that makes this `Copy` instead. A boolean is about as cheap to copy as it gets.

src/db/migrate.rs
```diff
@@ -1,11 +1,29 @@
...
+}
+
+impl MigrationContext {
+    fn format_query(&self, query: &str) -> String {
+        query.replace("{create_table}", match self.apply_mode {
+            ApplyMode::Permanent => "CREATE TABLE",
+            ApplyMode::Temporary => "CREATE TEMPORARY TABLE",
+        })
```
jyn514: This seems kind of magic to me (i.e. I won't remember how it works in a week). Not sure I see a better way to reuse all these functions though ...

src/db/migrate.rs
pietroalbini: Yeah, at the moment your approach is way better. I chose `Rc<MigrationContext>` to future-proof if we add non-`Copy` context items, but we can think about that when we need those.

src/db/migrate.rs
pietroalbini: Hmm, I'm out of ideas on this as well.

cc @QuietMisdreavus, do you have any better idea on how to dynamically switch from `CREATE TABLE` to `CREATE TEMPORARY TABLE` dynamically based on the `ApplyMode` in migrations?

jyn514: The only thing I can think of is to use an in-memory database connection, but I don't think there's a way to set that up from the client side... could we pass in a `Transaction` instead of a `Connection` and set [`read_only`](https://docs.rs/postgres/0.15.2/postgres/transaction/struct.Config.html#method.read_only) to true?

pietroalbini: @jyn514 what do you think about c7a036a? With it we won't worry about temporary tables when writing a new migration, but we can still use them behind the scenes.

jyn514: I like it! I wouldn't have thought of that, fewer changes are always nice :)

PR #532, opened 2019-12-21
Only build one target by default

### LINKED ISSUE #440: Cfg missing during build script build
**Crate name:** proc-macro2
**Old working build:** https://docs.rs/crate/proc-macro2/1.0.4/builds/186665
**Build failure link:** https://docs.rs/crate/proc-macro2/1.0.6/builds/192719

**Additional details:**
With the switch to rustwide, it seems that `--cfg` entries from our `[package.metadata.docs.rs]` are no longer seen by our build script (but still present when building the crate), which causes the build to fail.

Notice how in the working 1.0.4 build the rustc invocation that builds build.rs has `--cfg procmacro2_semver_exempt`. From tracing through the types in the failing 1.0.6 build, I believe that failure could only happen if `--cfg procmacro2_semver_exempt` is set for the lib.rs rustc invocation but not for the build.rs rustc invocation.

Relevant part of Cargo.toml: https://github.com/alexcrichton/proc-macro2/blob/1.0.6/Cargo.toml#L18-L20

Diff between working build current failing build: https://github.com/alexcrichton/proc-macro2/compare/1.0.4...1.0.6
(no relevant changes on our end).

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/343.

:warning: :warning: **DO NOT MERGE** :warning: :warning:  until we've (at the minimum):

- made an 'Inside Rust' blog post saying that we won't be building all targets anymore
- linked to that post on every page that will be affected (or every page globally if that's easier)
- left the warning up for at least a few weeks

Most of the following comments are related to https://github.com/rust-lang/docs.rs/pull/550 and can be safely ignored.

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -762,7 +762,9 @@ mod tests {
             let version = DUMMY_CRATE_VERSION;
             let default_target = "x86_64-unknown-linux-gnu";
 
-            assert_eq!(env.config().include_default_targets, true);
+            env.override_config(|config| {
+                config.include_default_targets = true;
+            });
 
             let mut builder = RustwideBuilder::init(env).unwrap();
             builder
--- a/src/config.rs
+++ b/src/config.rs
@@ -106,7 +106,7 @@ impl Config {
             local_docker_image: maybe_env("DOCS_RS_LOCAL_DOCKER_IMAGE")?,
             toolchain: env("CRATESFYI_TOOLCHAIN", "nightly".to_string())?,
             build_cpu_limit: maybe_env("DOCS_RS_BUILD_CPU_LIMIT")?,
-            include_default_targets: env("DOCSRS_INCLUDE_DEFAULT_TARGETS", true)?,
+            include_default_targets: env("DOCSRS_INCLUDE_DEFAULT_TARGETS", false)?,
             disable_memory_limit: env("DOCSRS_DISABLE_MEMORY_LIMIT", false)?,
         })
     }


jyn514:
I forgot to check if this works with a default other than x64-linux, not at my computer right now but I don't think it will.

jyn514:
Ok yeah this is broken for a different default target because cargo puts files in `target/:target/doc` when you cross compile. Maybe it would be simplest to just move that whole directory to `target/doc`?

jyn514:
Pushed a fix.

pietroalbini:
So, I think having a separate PR to avoiding building the default target twice is better: we should announce breaking changes on the main blog, and doing so is a big ask from a marketing angle. In late February or March I plan to make another breaking change, so we could bundle them together.

The separate PR removing the double build for the default target would help mitigating the build cost without visibly affecting users, while we could defer deploying this until next year.

src/docbuilder/rustwide_builder.rs
```diff
@@ -304,22 +293,25 @@ impl RustwideBuilder {
...
-                            self.build_target(
-                                target,
-                                &build,
-                                &limits,
-                                &local_storage.path(),
-                                &mut successful_targets,
-                            )?;
+                    successful_targets.push(res.target.clone());
+                    // Then build the documentation for all the targets
+                    for target in &metadata.extra_targets {
```
pietroalbini: To merge this without breaking the crates, we could gate the new behavior on the `DOCS_RS_ONLY_BUILD_DEFAULT_TARGET` environment variable: if it's not set we iterate over `TARGETS` instead. This would avoid this PR rotting over time if we wait until February/March.

jyn514: The actual breaking change should be very small, just changing `for target in TARGETS {` to `for target in &metadata.extra_targets {`. This will be easier to see when I break up the PR. I think an environment variable would be helpful for local development + testing though so I'll make the change.

jyn514: Added in https://github.com/rust-lang/docs.rs/pull/532/commits/2b8a99a768dd8d65767152eefb5c0676436469fc

jyn514:
Drafted a blog post, feedback is welcome: https://github.com/rust-lang/blog.rust-lang.org/pull/603,

jyn514:
https://github.com/rust-lang/team/pull/1051

PR #533, opened 2019-12-23
deps: upgrade to rustwide 0.4.0

pietroalbini:
r? @jyn514

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,7 +226,7 @@ dependencies = [
  "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-integer 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.84 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -1570,6 +1578,14 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "quote"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "proc-macro2 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
```
jyn514: I don't like that we're building yet another version of syn/quote, but I don't see a good way to fix this without upgrading all our  dependencies.

Cargo.lock
```diff
@@ -2574,7 +2600,15 @@ name = "toml"
 version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "serde 1.0.84 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "toml"
+version = "0.5.5"
```
jyn514: While you're upgrading, can you bump `toml` to 0.5 so it's not duplicated?

pietroalbini: Done.

pietroalbini:
Resolved all the review comments.

jyn514:
Looks great, thanks!

PR #534, opened 2019-12-24
Don't build default target twice

### LINKED ISSUE #440: Cfg missing during build script build

jyn514:
This is the non-breaking changes from #532. I also kept the additional documentation for existing metadata.

Copy/pasted from that PR for convenience:

## Architecture Changes

This removes the duplicated 'default' build which is run twice.

- The default build no longer passes `--target` to Cargo for the common case where `default-target` is not set. This will fix #440. As a side effect, it removes most of the hacking around I had to do to get proc macros to work.
- The target for the default build is now stored in the database. As before, the default target is `x86_64-unknown-linux-gnu` and can be overriden with `default-target = ...` in Cargo.toml
- The target for the default build is no longer run twice
- The web server redirects requests to `/:crate/:version/:default-target` to `/:crate/:version/`, keeping the subpage if it exists.

This has been tested both by building with the new code and by building with the old code and running `cratesfyi database migrate`. Note that if `cratesfyi database migrate` is _not_ run, the web server will panic whenever viewing a crate built with the old code.

r? @pietroalbini

### DIFF
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -26,6 +26,7 @@ pub(crate) fn add_package_into_database(conn: &Connection,
                                  metadata_pkg: &MetadataPackage,
                                  source_dir: &Path,
                                  res: &BuildResult,
+                                 default_target: &str,
                                  files: Option<Json>,
                                  doc_targets: Vec<String>,
                                  cratesio_data: &CratesIoData,
@@ -82,7 +83,7 @@ pub(crate) fn add_package_into_database(conn: &Connection,
                                          &is_library,
                                          &res.rustc_version,
                                          &metadata_pkg.documentation,
-                                         &res.target])?;
+                                         &default_target])?;
             // return id
             rows.get(0).get(0)
 
@@ -136,7 +137,7 @@ pub(crate) fn add_package_into_database(conn: &Connection,
                               &is_library,
                               &res.rustc_version,
                               &metadata_pkg.documentation,
-                              &res.target])?;
+                              &default_target])?;
             rows.get(0).get(0)
         }
     };
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -28,7 +28,6 @@ pub(crate) fn add_package_into_database(conn: &Connection,
                                  res: &BuildResult,
                                  files: Option<Json>,
                                  doc_targets: Vec<String>,
-                                 default_target: &Option<String>,
                                  cratesio_data: &CratesIoData,
                                  has_docs: bool,
                                  has_examples: bool)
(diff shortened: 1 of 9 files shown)


pietroalbini:
Build code also seem to work fine!

jyn514:
> if you click the default target you'll be sent first to /:crate/:version/:default-target, and then redirected to /:crate/:version. I'd be nice to remove that redirect.

I'm not sure how to fix this. Currently we set those links like this in `templates/navigation_rustdoc.hbs`:

```
                {{#each doc_targets}}
                <li class="pure-menu-item"><a href="/{{../../content.crate_details.name}}/{{../../content.crate_details.version}}/{{this}}/{{../../content.crate_details.target_name}}/" class="pure-menu-link">{{this}}</a></li>
                {{/each}}
```

We'd have to compare each `doc_target` to the `default_target`, but I don't know how to do that in handlebars. The [API reference says](https://handlebarsjs.com/guide/builtin-helpers.html#if) that `#if` has to be passed a single value, it can't handle `==` or anything like that.

pietroalbini:
> I'm not sure how to fix this.

I'd generate a `Map<TargetName, TargetUrl>` from Rust and pass it to the template, sidestepping the handlebars limitations.

jyn514:
I don't see a way to send a data structure to a page, we'd have to make it part of CrateDetails.

Is it ok if we add this at a later point? It's not hurting usability, you still get sent to the right page.

pietroalbini:
> I don't see a way to send a data structure to a Page, we'd have to make it part of CrateDetails.

Yep.

> Is it ok if we add this at a later point? It's not hurting usability, you still get sent to the right page.

Ok, can you open an issue about that?

jyn514:
Opened: https://github.com/rust-lang/docs.rs/issues/549

PR #538, opened 2019-12-27
Improve error messages for wrong CLI args

jyn514:
- delete-crate now gives a usage message instead of panicking when no
crate name is passed
- the following commands give a usage message when no subcommand is passed:
  - build
  - database
  - queue

PR #540, opened 2019-12-28
Add ArgRequiredElseHelp to clap commands without default behaviour

koenaad:
In `cratesfyi.rs`, the large if-else structure starting at line 137 is not exhaustive. Sometimes, if no subcommand is specified, the execution of `cratesfyi` will end without printing anything.
This PR adds the `ArgRequiredElseHelp` setting to a number of clap commands.

I have added this setting to all (sub)commands that have subcommands themselves.

For instance before this commit, running `$ cratesfyi database` would print nothing and simply exit.

After these changes it will print the help text and exit:

```
$ target/debug/cratesfyi database
cratesfyi-database 
Database operations

USAGE:
    cratesfyi database [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    add-directory              Adds a directory into database
    delete-crate               Removes a whole crate from the database
    help                       Prints this message or the help of the given subcommand(s)
    migrate                    Run database migrations
    move-to-s3                 
    update-github-fields       Updates github stats for crates.
    update-release-activity    Updates monthly release activity chart
    update-search-index        Updates search index
```

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -11,7 +11,7 @@ extern crate rustwide;
 use std::env;
 use std::path::{Path, PathBuf};
 
-use clap::{Arg, App, SubCommand};
+use clap::{Arg, App, AppSettings, SubCommand};
 use cratesfyi::{DocBuilder, RustwideBuilder, DocBuilderOptions, db};
 use cratesfyi::utils::add_crate_to_queue;
 use cratesfyi::start_web_server;
@@ -24,8 +24,10 @@ pub fn main() {
     let matches = App::new("cratesfyi")
         .version(cratesfyi::BUILD_VERSION)
         .about(env!("CARGO_PKG_DESCRIPTION"))
+        .setting(AppSettings::ArgRequiredElseHelp)
         .subcommand(SubCommand::with_name("build")
             .about("Builds documentation in a chroot environment")
+            .setting(AppSettings::ArgRequiredElseHelp)
             .arg(Arg::with_name("PREFIX")
                 .short("P")
                 .long("prefix")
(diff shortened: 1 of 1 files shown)


src/bin/cratesfyi.rs
```diff
@@ -231,12 +235,15 @@ pub fn main() {
             let conn = db::connect_db().expect("failed to connect to the database");
             db::delete_crate(&conn, &name).expect("failed to delete the crate");
         }
+
```
jyn514: Is there a reason you added this whitespace?

src/bin/cratesfyi.rs
```diff
@@ -115,6 +118,7 @@ pub fn main() {
                     .help("Name of the crate to delete"))))
```
jyn514:
```suggestion
                    .required(true)
                    .help("Name of the crate to delete"))))
```

Not directly part of your PR, but would be nice to give an error here (I did in https://github.com/rust-lang/docs.rs/pull/538)

koenaad: Good call, done.

src/bin/cratesfyi.rs
koenaad: This improved the readability significantly for me. The if-else structures feels very dense and I found it difficult to parse the different top-level subcommands. By adding a whitespace the commands `delete-crate`, `start-web-server`, `daemon` and `queue` stand out from each other more.
Since it doesn't touch any previously written code I thought it would okay to add. But in the end, it's mostly personal preference so I don't mind reverting it as well.

jyn514: I don't have a strong opinion so if it makes it easier for you we can leave it in.

PR #541, opened 2019-12-30
Drop support for building 32-bit Apple docs

pietroalbini:
Rust's [RFC 2837](https://github.com/rust-lang/rfcs/pull/2837) demoted the `i686-apple-darwin` target to Tier 3, so that target will not be downloadable anymore through rustup starting from Rust 1.42.0. The change will also land on nightly soon, and that will prevent docs.rs from updating the rustwide toolchain.

This PR removes `i686-apple-darwin` from the list of targets we build, and adds code to remove it automatically from existing rustwide workspaces.

This is ready for review, but we should hold off deploying it until we announce the demotion on the blog.

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -19,7 +19,6 @@ static USER_AGENT: &str = "docs.rs builder (https://github.com/rust-lang/docs.rs
 static DEFAULT_RUSTWIDE_WORKSPACE: &str = ".rustwide";
 
 static TARGETS: &[&str] = &[
-    "i686-apple-darwin",
     "i686-pc-windows-msvc",
     "i686-unknown-linux-gnu",
     "x86_64-apple-darwin",
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -9,8 +9,10 @@ use postgres::Connection;
 use rustc_serialize::json::ToJson;
 use rustwide::cmd::{Command, SandboxBuilder};
 use rustwide::logging::{self, LogStorage};
+use rustwide::toolchain::ToolchainError;
 use rustwide::{Build, Crate, Toolchain, Workspace, WorkspaceBuilder};
 use std::borrow::Cow;
+use std::collections::HashSet;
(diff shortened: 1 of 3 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -95,12 +96,43 @@ impl RustwideBuilder {
...
+        for target in installed_targets {
+            if !targets_to_install.remove(&target) {
+                self.toolchain.remove_target(&self.workspace, &target)?;
+            }
+        }
+
         self.toolchain.install(&self.workspace)?;
-        for target in TARGETS {
+
+        for target in &targets_to_install {
```
jyn514: It looks like this will never update toolchains that are already installed, since they will be removed on line 124. I don't think this is the intended behavior.

src/docbuilder/rustwide_builder.rs
```diff
@@ -95,12 +96,43 @@ impl RustwideBuilder {
         // Ignore errors if detection fails.
         let old_version = self.detect_rustc_version().ok();
 
+        let mut targets_to_install = TARGETS
+            .iter()
+            .map(|t| t.to_string())
+            .collect::<HashSet<_>>();
```
jyn514: Since we don't need to modify this (see above), I don't think we need a second data structure, we can just use `.contains()` on TARGETS on line 124.

src/docbuilder/rustwide_builder.rs
pietroalbini: Those targets will be updated: once a target is installed rustup will keep it up to date even without running `rustup target add` every time. Before we were installing all the targets all the time because we didn't have the list of already installed targets, but now that we have it we can use that to cut the number of `rustup` calls we make.

jyn514: Ah, I misunderstood what call to rustup this was making; I thought it was calling `rustup toolchain install`, not `rustup target add`.

pietroalbini:
We can merge this. Are you fine with the current state of the code @jyn514?

jyn514:
I have a nagging feeling I missed something ... but the code itself looks fine, I'm ok with merging.

Issue #516, opened 2019-12-10
Link to last successful build when build fails


jyn514:
I thought about this while investigating why https://docs.rs/crate/hyper/0.13.0 failed to build - the reason this usually matters to library authors is because new users/users upgrading see an error instead of the docs. If there _a_ version of the docs, even an out of date version, I imagine that would make the failures much less critical for library authors.

jyn514:
I want to make it a link, not a redirect, so people know that it's an old version.

jyn514:
Actually maybe this would be better served by implementing https://github.com/rust-lang/docs.rs/issues/341, https://github.com/rust-lang/docs.rs/issues/396 at the same time and just putting it in the 'Redirect to Latest Version' button at the top? 'Latest Version' is no longer an accurate description then ... 'latest successful stable version' is closer but way too verbose to put in a link.

jyn514:
Actually maybe this would be better served by implementing https://github.com/rust-lang/docs.rs/issues/341, https://github.com/rust-lang/docs.rs/issues/396 at the same time and just putting it in the 'Redirect to Latest Version' button at the top? 'Latest Version' is no longer an accurate description then ... 'latest successful stable version' is closer but way too verbose to put in a link.

jyn514:
Hmm ... but in this case I think people would want to know the latest version is released even if there's no docs. I think separate links is fine.

Relevant code: https://github.com/rust-lang/docs.rs/blob/master/templates/crate_details.hbs#L65

jyn514:
The way to get started with this would be to
- find the last successful build from the database (something like `SELECT version FROM releases INNER JOIN crates ON releases.crate_id = crates.id WHERE build_status = true AND yanked = false AND crates.name = $1;`)
- add the last successful build to [the variables for the page](https://github.com/rust-lang/docs.rs/blob/master/src/web/crate_details.rs#L239)
- add a message to the [error template](https://github.com/rust-lang/docs.rs/blob/master/templates/crate_details.hbs#L65) which links to the last build

koenaad:
Hi, I would like to work on this :wave: Can you assign me? 🙂

koenaad:
I did some analysis, just to make sure I'm changing the right things 😉

So the main issue: if the build of a release of a crate fails, a user should be able to find the most recent **successful** build fairly simple.
Currently we often link to the latest release which isn't necessary a successful build.

There a two places this causes a problem:

### Crate details page

[...]

PR #546, opened 2020-01-04
Crate details page: show whether a build failed in the Versions list

koenaad:
_Implements second change related to #516.
I was planning to rebase these commits on top of the commits in #543 if that PR gets approved & merged. But this can be merged indepedently._

This pull request changes the Crate Details page to display whether a build failed in the _Versions_ list.

Current design:

Screenshot 2020-01-04 at 18 59 21

Page after these commits:

Screenshot 2020-01-04 at 18 59 32

I performed the changes in two commits, this should make it a bit easier to reason about them. _Hopefully..._

521b8ae
- I added an extra test case for CrateDetails::versions.
- Most code is the same as 0a56ff7 from PR #543, this was necessary to create the test. This code should be removed from the commits when doing a rebase.

da358f8
- In `CrateDetails` I replaced the field `versions` with `releases`. The type changed from a raw String to `Release`, which holds a `version` string and a `build_status` boolean.
- I create a fn `versions()` which has the same behaviour as the removed `versions` field.
- I set `releases` in `CrateDetails::new` but I took care to change as little code as possible. Hence the introduction of `map_to_release`. This entire function can be refactored to be more efficient / perform less queries on the database, but that's for a later PR.
- Lastly, I edited `crate_details.hbs` to use the additional information and show a ⚠️  sign when appropriate.

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -38,7 +38,7 @@ pub struct CrateDetails {
     keywords: Option<Json>,
     have_examples: bool, // need to check this manually
     pub target_name: String,
-    pub versions: Vec<String>,
+    releases: Vec<Release>,
     github: bool, // is crate hosted in github
     github_stars: Option<i32>,
     github_forks: Option<i32>,
@@ -77,7 +77,7 @@ impl ToJson for CrateDetails {
         m.insert("keywords".to_string(), self.keywords.to_json());
         m.insert("have_examples".to_string(), self.have_examples.to_json());
         m.insert("target_name".to_string(), self.target_name.to_json());
-        m.insert("versions".to_string(), self.versions.to_json());
+        m.insert("releases".to_string(), self.releases.to_json());
         m.insert("github".to_string(), self.github.to_json());
         m.insert("github_stars".to_string(), self.github_stars.to_json());
         m.insert("github_forks".to_string(), self.github_forks.to_json());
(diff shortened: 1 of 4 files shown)


templates/crate_details.hbs
```diff
@@ -42,8 +42,14 @@
...
             <div class="pure-menu pure-menu-scrollable sub-menu">
               <ul class="pure-menu-list">
-                {{#each versions}}
-                <li class="pure-menu-item"><a href="/crate/{{../name}}/{{this}}" class="pure-menu-link">{{this}}</a></li>
+                {{#each releases}}
+                <li class="pure-menu-item">
+                  {{#if this.build_status}}
+                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link">{{this.version}}</a>
+                  {{else}}
+                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link warn"><i class="fa fa-fw fa-warning"></i> {{this.version}} (build failed)</a>
```
jyn514: I don't quite like how this looks. Currently it shows both the warning sign :warning: and `(build failure)` on the right. I would rather have only the warning sign and have the 'build failure' text show up when you hover over it.

src/web/crate_details.rs
```diff
@@ -149,7 +166,10 @@ impl CrateDetails {
 
             versions.sort();
             versions.reverse();
-            versions.iter().map(|semver| format!("{}", semver)).collect()
+            versions.iter()
+                .map(|semver| format!("{}", semver))
+                .map(|version| map_to_release(&conn, crate_id, version))
```
jyn514: I think there only needs to be one `.map()` here.

```suggestion
                .map(|version| map_to_release(&conn, crate_id, version.to_string()))
```

koenaad: Yes, indeed. I was bit confused by the use of `format!("{}", semver)` instead of simply `semver.to_string()`...

koenaad:
I have simplified the versions display by using a tooltip:

Screenshot 2020-01-07 at 22 58 53

src/web/crate_details.rs
```diff
@@ -237,6 +254,31 @@ impl CrateDetails {
...
+
+fn map_to_release(conn: &Connection, crate_id: i32, version: String) -> Release {
+    let rows = conn.query(
+        "SELECT build_status
+         FROM releases
+         WHERE releases.crate_id = $1 and releases.version = $2;",
+        &[&crate_id, &version],
+    ).unwrap();
+
+    let build_status = if rows.len() >= 1 {
```
```suggestion
    let build_status = if !rows.is_empty() {
```

jyn514:
I'm really looking forward to having this, thanks again for your hard work!

src/web/crate_details.rs
koenaad: Good eye, fixed it!

jyn514: Blame clippy, not me :)

koenaad: Hmm, weird. Clippy isn't saying anything on my machine... 😕

PR #550, opened 2020-01-08
Remove duplicate default target (redux)

jyn514:
Same as #534 but now it doesn't crash when viewing source code.

r? @pietroalbini

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -280,6 +280,19 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
             // downgrade query
             "ALTER TABLE sandbox_overrides ALTER COLUMN max_memory_bytes TYPE INTEGER;"
         ),
+        migration!(
+            context,
+            // version
+            8,
+            // description
+            "Make default_target non-nullable",
+            // upgrade query
+            "UPDATE releases SET default_target = 'x86_64-unknown-linux-gnu' WHERE default_target IS NULL;
+             ALTER TABLE releases ALTER COLUMN default_target SET NOT NULL",
+            // downgrade query
+            "ALTER TABLE releases ALTER COLUMN default_target DROP NOT NULL;
+             ALTER TABLE releases ALTER COLUMN default_target DROP DEFAULT",
+        ),
     ];
 
     for migration in migrations {
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -487,7 +487,8 @@ mod test {
             let base = "/dummy/0.3.0/dummy/";
             assert_success(base, web)?;
             assert_redirect("/dummy/0.3.0/x86_64-unknown-linux-gnu/dummy/", base, web)?;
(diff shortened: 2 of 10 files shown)


pietroalbini:
There are merge conflicts.

pietroalbini:
...and it'd be nice to add tests, at least for the web part.

jyn514:
> There are merge conflicts.
Yeah ... I tried to rebase, but it didn't work very well, I think because of the commit to master reverting the earlier changes. I'll fix the conflicts (again) once I've written some tests, probably tomorrow.

jyn514:
So ... I can add tests for the web changes once #553 is merged (it has the `.file()` addition to FakeReleases). However, we don't currently have a way to test the build step works properly and generates all the static files it should. I might spend some time adding a FakeBuild tomorrow, but my schedule is about to get busy again so I may not finish.

pietroalbini:
> However, we don't currently have a way to test the build step works properly and generates all the static files it should. I might spend some time adding a FakeBuild tomorrow, but my schedule is about to get busy again so I may not finish.

I'd not handle builds along with other tests in the codebase, but I'd do something separate (like rustwide's buildtest or crater's minicrater). My worry with adding built tests along with the other tests is that it encourages using a "fake build" around the codebase tests: builds are inheritely slow, as the rustwide initialization can't be parallelized, so we want to have as few of them as possible.

jyn514:
This is ready for review.

src/db/migrate.rs
```diff
@@ -269,6 +269,19 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
             // downgrade query
             "DROP TABLE blacklisted_crates;"
         ),
+        migration!(
+            context,
+            // version
+            7,
```
pietroalbini:
```suggestion
            8,
```

Migration 7 is already taken by https://github.com/rust-lang/docs.rs/pull/567.

src/web/rustdoc.rs
```diff
@@ -224,6 +225,19 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     req_path.insert(1, &name);
     req_path.insert(2, &version);
 
+    // if visiting the full path to the default target, remove the target from the path
+    // expects a req_path that looks like `/rustdoc/:crate/:version[/:target]/.*`
+    let crate_details = cexpect!(CrateDetails::new(&conn, &name, &version));
+    if req_path[3] == crate_details.metadata.default_target {
```
pietroalbini: Can we add a `&& req_path.len() > 4`? Just to be safe, I don't want this to panic.

jyn514: It works out, you can slice one past the bounds of the array without a problem: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ff716f47089c50c73b45839444f2a39b

src/db/migrate.rs
jyn514: This required a rebase.

src/web/rustdoc.rs
jyn514: I added a test just to be on the safe side.

jyn514:
Addressed review comments.

Issue #552, opened 2020-01-08
Panic in `path_for_latest`


jyn514:
This is coming from https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L312.

Somehow, there's a case where `req_path` has a `module` that's not the same as `target_name`, but also not a target. Maybe a renamed crate?

This is all the information I have from the logs:

<details><summary>backtrace</summary>

```
Jan 07 15:43:38 docsrs cratesfyi[417]: thread '<unnamed>' panicked at 'index out of bounds: the len is 4 but the index is 4', src/web/rustdoc.rs:312:53
Jan 07 15:43:38 docsrs cratesfyi[417]: stack backtrace:
... truncated ...
Jan 07 15:43:38 docsrs cratesfyi[417]:   14: core::panicking::panic_bounds_check
Jan 07 15:43:38 docsrs cratesfyi[417]:              at src/libcore/panicking.rs:96
Jan 07 15:43:38 docsrs cratesfyi[417]:   15: cratesfyi::web::rustdoc::rustdoc_html_server_handler
Jan 07 15:43:38 docsrs cratesfyi[417]:   16: <F as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   17: <alloc::boxed::Box<dyn iron::middleware::Handler> as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   18: <cratesfyi::web::routes::BlockBlacklistedPrefixes as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   19: <alloc::boxed::Box<dyn iron::middleware::Handler> as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   20: router::router::Router::handle_method
Jan 07 15:43:38 docsrs cratesfyi[417]:   21: <router::router::Router as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   22: iron::middleware::Chain::continue_from_handler
Jan 07 15:43:38 docsrs cratesfyi[417]:   23: <iron::middleware::Chain as iron::middleware::Handler>::handle
Jan 07 15:43:38 docsrs cratesfyi[417]:   24: <alloc::boxed::Box<dyn iron::middleware::Handler> as iron::middleware::Handler>::handle
[...]

PR #553, opened 2020-01-08
Fix occasional panic when visiting settings.html

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/552


jyn514:
My only concern is about other pages besides `settings.html` being one directory too high, but my thought is at least now we'll see them in the traceback as opposed to having to guess.

jyn514:
Added tests :)

src/test/fakes.rs
```diff
@@ -76,15 +77,37 @@ impl<'db> FakeRelease<'db> {
...
+
     pub(crate) fn create(self) -> Result<i32, Error> {
         let tempdir = tempdir::TempDir::new("docs.rs-fake")?;
 
+        let files: Vec<_> = self.files.into_iter().map(|(mimetype, path, data)| {
+            let file = tempdir.path().join(&path);
+            std::fs::write(file, data)?;
+
+            let metadata = format!("[\"{}\", \"{}\"]", mimetype, path);
+            Ok(Json::from_str(&metadata).expect("should be valid json"))
```
pietroalbini: Is there a reason why you're converting to string and serializing it back? This *should* work:

```rust
Ok(Json::Array(vec![Json::String(mimetype), Json::String(path)]));
```

This way you don't even need to restrict `"` in paths.

src/test/fakes.rs
```diff
@@ -76,15 +77,37 @@ impl<'db> FakeRelease<'db> {
...
+        }).collect::<Result<_, Error>>()?;
+
         let release_id = crate::db::add_package_into_database(
             &self.db.conn(),
             &self.package,
             tempdir.path(),
             &self.build_result,
-            self.files,
+            if files.is_empty() { None } else { Some(Json::Array(files)) },
```
pietroalbini: Hmm, I'm pretty sure the `files` here refers to source files, not to the rustdoc documentation.

src/test/mod.rs
```diff
@@ -8,9 +8,15 @@ use reqwest::{Client, Method, RequestBuilder};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 pub(crate) fn wrapper(f: impl FnOnce(&TestEnvironment) -> Result<(), Error>) {
+    use std::panic;
```
pietroalbini: Nit: can you move the import outside of the function? Or just using `std::panic::...` instead or just `panic::`.

src/test/fakes.rs
jyn514: There is not, this is better.

src/test/fakes.rs
jyn514: Oh whoops! I need to reshuffle some things then.

jyn514: Also, we should probably document that. It's referred to as 'files' all through the code until it goes into the DB and it's hard to find again.

pietroalbini: Definitely.

jyn514: I did some refactoring, this should be fixed now.

src/test/fakes.rs
```diff
@@ -76,15 +80,40 @@ impl<'db> FakeRelease<'db> {
...
+        self
+    }
+
     pub(crate) fn create(self) -> Result<i32, Error> {
         let tempdir = tempdir::TempDir::new("docs.rs-fake")?;
 
+        let upload_files = |prefix, files: Vec<(String, Vec<u8>)>, package: &MetadataPackage, db: &TestDatabase| {
+            for (path, data) in files {
+                let file = tempdir.path().join(&path);
+                std::fs::write(file, data)?;
```
jyn514: I just realized there is a bug here, the files from the first call will also be uploaded on the second call.

jyn514:
This is ready for another review.

It's mostly commit-by-commit. The actual fix is https://github.com/rust-lang/docs.rs/pull/553#pullrequestreview-340358622, the rest is just test infrastructure.

src/test/fakes.rs
```diff
@@ -77,33 +80,40 @@ impl<'db> FakeRelease<'db> {
...
         self
     }
 
     pub(crate) fn create(self) -> Result<i32, Error> {
         let tempdir = tempdir::TempDir::new("docs.rs-fake")?;
 
-        let files: Vec<_> = self.files.into_iter().map(|(mimetype, path, data)| {
-            let file = tempdir.path().join(&path);
-            std::fs::write(file, data)?;
+        let upload_files = |prefix, files: Vec<(String, Vec<u8>)>, package: &MetadataPackage, db: &TestDatabase| {
```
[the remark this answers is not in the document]
jyn514: No, using self.foo leads to a borrow error since we're moving `self.rustdoc_files`.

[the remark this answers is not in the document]
jyn514: Actually that might work if we pass a slice, one second.

jyn514:
This is ready for review.

pietroalbini:
The code looks good, but I don't have time to test it locally :(

jyn514:
If you like, you could give me a laundry list of things to test locally and I can do it for you?

PR #556, opened 2020-01-08
Test utilities for the web server

pietroalbini:
This PR adds a way to test the web server in `cargo test`, using the dummy database. With it, it's possible to write a test like this:

```rust
#[test]
fn test_index_is_successful() {
    crate::test::wrapper(|env| {
        let web = env.frontend();
        assert!(web.get("/").send()?.status().is_successful());
        Ok(())
    });
}
```

Additionally, to avoid having tons of wrappers, the existing `crate::test::with_database` has to be written this way now:

```rust
#[test]
fn test_db() {
    crate::test::wrapper(|env| {
        let db env.db();
        assert!(!db.conn().query("SELECT 1;", &[]).is_empty());
        Ok(())
    });
}
```

Of course it's possible to mix the two. The PR can be reviewed commit-by-commit.

r? @jyn514 
cc @koenaad

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ prometheus = { version = "0.7.0", default-features = false }
 lazy_static = "1.0.0"
 rustwide = "0.5.0"
 tempdir = "0.3"
-once_cell = "1.2.0"
 
 # iron dependencies
 iron = "0.5"
@@ -52,6 +51,9 @@ staticfile = { version = "0.4", features = [ "cache" ] }
 [dependencies.postgres]
 version = "0.15"
 features = [ "with-time", "with-rustc-serialize" ]
+
+[dev-dependencies]
+once_cell = "1.2.0"
 
 [build-dependencies]
 time = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ prometheus = { version = "0.7.0", default-features = false }
 lazy_static = "1.0.0"
 rustwide = "0.5.0"
 tempdir = "0.3"
+once_cell = "1.2.0"
 
 # iron dependencies
 iron = "0.5"
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -2,13 +2,55 @@ mod fakes;
 
 use crate::web::Server;
 use failure::Error;
+use once_cell::unsync::OnceCell;
 use postgres::Connection;
-use std::sync::{Arc, Mutex, MutexGuard};
 use reqwest::{Client, Method, RequestBuilder};
+use std::sync::{Arc, Mutex, MutexGuard};
 
-pub(crate) fn with_database(f: impl FnOnce(&TestDatabase) -> Result<(), Error>) {
-    let env = TestDatabase::new().expect("failed to initialize the environment");
-    report_error(|| f(&env));
+pub(crate) fn wrapper(f: impl FnOnce(&TestEnvironment) -> Result<(), Error>) {
+    let env = TestEnvironment::new();
+    let result = f(&env);
+    env.cleanup();
+
+    if let Err(err) = result {
+        eprintln!("the test failed: {}", err);
+        for cause in err.iter_causes() {
+            eprintln!("  caused by: {}", cause);
+        }
+
+        eprintln!("{}", err.backtrace());
+
+        panic!("the test failed");
+    }
+}
+
+pub(crate) struct TestEnvironment {
+    db: OnceCell<TestDatabase>,
+    frontend: OnceCell<TestFrontend>,
+}
+
+impl TestEnvironment {
+    fn new() -> Self {
+        Self {
+            db: OnceCell::new(),
+            frontend: OnceCell::new(),
+        }
+    }
+
+    fn cleanup(self) {
+        if let Some(frontend) = self.frontend.into_inner() {
+            frontend.server.leak();
+        }
+    }
+
+    pub(crate) fn db(&self) -> &TestDatabase {
+        self.db
+            .get_or_init(|| TestDatabase::new().expect("failed to initialize the db"))
+    }
+
+    pub(crate) fn frontend(&self) -> &TestFrontend {
+        self.frontend.get_or_init(|| TestFrontend::new(self.db()))
+    }
 }
 
 pub(crate) struct TestDatabase {
@@ -37,13 +79,6 @@ impl TestDatabase {
     }
 }
 
-pub(crate) fn with_frontend(db: &TestDatabase, f: impl FnOnce(&TestFrontend) -> Result<(), Error>) {
-    let frontend = TestFrontend::new(db);
-    let result = f(&frontend);
-    frontend.server.leak();
-    report_error(|| result)
-}
-
 pub(crate) struct TestFrontend {
     server: Server,
     client: Client,
@@ -58,23 +93,11 @@ impl TestFrontend {
     }
 
     fn build_request(&self, method: Method, url: &str) -> RequestBuilder {
-        self.client.request(method, &format!("http://{}{}", self.server.addr(), url))
+        self.client
+            .request(method, &format!("http://{}{}", self.server.addr(), url))
     }
 
     pub(crate) fn get(&self, url: &str) -> RequestBuilder {
         self.build_request(Method::GET, url)
-    }
-}
-
-fn report_error(f: impl FnOnce() -> Result<(), Error>) {
-    if let Err(err) = f() {
-        eprintln!("the test failed: {}", err);
-        for cause in err.iter_causes() {
-            eprintln!("  caused by: {}", cause);
-        }
-
-        eprintln!("{}", err.backtrace());
-
-        panic!("the test failed");
     }
 }
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1,22 +1,14 @@
 mod fakes;
 
+use crate::web::Server;
 use failure::Error;
 use postgres::Connection;
 use std::sync::{Arc, Mutex, MutexGuard};
+use reqwest::{Client, Method, RequestBuilder};
 
 pub(crate) fn with_database(f: impl FnOnce(&TestDatabase) -> Result<(), Error>) {
     let env = TestDatabase::new().expect("failed to initialize the environment");
-
-    if let Err(err) = f(&env) {
-        eprintln!("the test failed: {}", err);
-        for cause in err.iter_causes() {
-            eprintln!("  caused by: {}", cause);
-        }
-
-        eprintln!("{}", err.backtrace());
-
-        panic!("the test failed");
-    }
+    report_error(|| f(&env));
 }
 
 pub(crate) struct TestDatabase {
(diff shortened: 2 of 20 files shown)


src/test/mod.rs
```diff
@@ -44,3 +36,45 @@ impl TestDatabase {
...
+    fn build_request(&self, method: Method, url: &str) -> RequestBuilder {
+        self.client.request(method, &format!("http://{}{}", self.server.addr(), url))
+    }
+
+    pub(crate) fn get(&self, url: &str) -> RequestBuilder {
+        self.build_request(Method::GET, url)
+    }
+}
+
+fn report_error(f: impl FnOnce() -> Result<(), Error>) {
```
jyn514: It might be simpler to have this take a `Result<(), Error>` instead of passing it a closure directly. It doesn't look like there's any setup or teardown so it shouldn't affect the semantics.

src/web/mod.rs
```diff
@@ -335,24 +343,53 @@ fn latest_version(versions_json: &Vec<String>, req_version: &str) -> Option<Stri
...
+        self.inner.socket
+    }
+
+    /// Iron is bugged, and it never closes the server even when the listener is dropped. To
+    /// avoid never-ending tests this method forgets about the server, leaking it and allowing the
+    /// program to end.
+    ///
+    /// The OS will then close all the dangling servers once the process exits.
+    ///
+    /// https://docs.rs/iron/0.6.1/iron/struct.Listening.html#method.close
```
jyn514: nit: we're still on 0.5 :(

This is [fixed in hyper 0.11](https://github.com/hyperium/hyper/issues/338#issuecomment-347044047), but even the latest version of iron [depends on 0.10](https://docs.rs/crate/iron/0.6.1/source/Cargo.toml.orig).

```suggestion
    /// https://docs.rs/iron/0.5.1/iron/struct.Listening.html#method.close
```

src/web/mod.rs
```diff
@@ -335,24 +343,53 @@ fn latest_version(versions_json: &Vec<String>, req_version: &str) -> Option<Stri
...
+    #[cfg(test)]
+    pub(crate) fn leak(self) {
```
jyn514: Would it be easier to make a custom implementation of `drop` instead? That we we couldn't forget to call `leak()`.

src/web/mod.rs
```diff
@@ -516,6 +536,17 @@ mod test {
     extern crate env_logger;
     use super::*;
 
+    #[test]
+    fn test_index_returns_success() {
+        crate::test::with_database(|db| {
+            crate::test::with_frontend(db, |web| {
```
jyn514: It might be nice to have `with_frontend()` take an `Option<TestDatabase>` so we don't have patterns like this calling `with_database` then `with_frontend` all over the place.

src/web/pool.rs
```diff
@@ -4,14 +4,24 @@ use postgres::Connection;
 use r2d2;
 use r2d2_postgres;
 
+#[cfg(test)]
+use std::sync::{Arc, Mutex, MutexGuard};
+
 pub(crate) enum Pool {
     R2D2(r2d2::Pool<r2d2_postgres::PostgresConnectionManager>),
+    #[cfg(test)]
+    Simple(Arc<Mutex<Connection>>),
```
jyn514: Instead of making this an enum where we only use the `Simple` variant in test mode, would it be easier to make the whole enum a cfg() parameter?

```rust
#[cfg(not(test))]
type Pool = r2d2::Pool<r2d2_postgres::PostgresConnectionManager>;
#[cfg(test)]
type Pool = Arc<Mutex<Connection>>;
```

That would also prevent accidentally modifying the prod database during tests (not that we plan to run tests in prod,  but it would also avoid screwing up my dev environment :P)

Cargo.toml
```diff
@@ -40,6 +40,7 @@ prometheus = { version = "0.7.0", default-features = false }
 lazy_static = "1.0.0"
 rustwide = "0.5.0"
 tempdir = "0.3"
+once_cell = "1.2.0"
```
jyn514: Can you make this a `dev-dependency`?

src/test/mod.rs
```diff
@@ -2,13 +2,55 @@ mod fakes;
...
+
+    fn cleanup(self) {
+        if let Some(frontend) = self.frontend.into_inner() {
+            frontend.server.leak();
+        }
+    }
+
+    pub(crate) fn db(&self) -> &TestDatabase {
+        self.db
+            .get_or_init(|| TestDatabase::new().expect("failed to initialize the db"))
```
jyn514: Do we want the tests to see each others changes? I'm worried that will lead to non-deterministic tests, especially since `cargo test` runs tests in parallel in an arbitrary order.

src/test/mod.rs
```diff
@@ -2,13 +2,55 @@ mod fakes;
...
+        for cause in err.iter_causes() {
+            eprintln!("  caused by: {}", cause);
+        }
+
+        eprintln!("{}", err.backtrace());
 
-pub(crate) fn with_database(f: impl FnOnce(&TestDatabase) -> Result<(), Error>) {
-    let env = TestDatabase::new().expect("failed to initialize the environment");
-    report_error(|| f(&env));
+        panic!("the test failed");
```
jyn514: Why not `panic!("the test failed: {}", err)`?

src/web/crate_details.rs
```diff
@@ -348,7 +348,9 @@ mod tests {
 
     #[test]
     fn test_last_successful_build_when_last_release_failed() {
-        crate::test::with_database(|db| {
+        crate::test::wrapper(|env| {
```
jyn514: It might be easier to add a `crate::test::with_db` function that does exactly this: `crate::test::wrapper(|env| f(env.db()))`. That would make the tests easier to write.

src/web/mod.rs
```diff
@@ -538,11 +538,9 @@ mod test {
 
     #[test]
     fn test_index_returns_success() {
-        crate::test::with_database(|db| {
-            crate::test::with_frontend(db, |web| {
-                assert!(web.get("/").send()?.status().is_success());
-                Ok(())
-            });
+        crate::test::wrapper(|env| {
+            let web = env.frontend();
```
jyn514: Ditto for `with_frontend()`, see comments about `with_db()` above.

src/test/mod.rs
jyn514: Whoops, this function got removed in a later commit.

src/web/mod.rs
pietroalbini: I would prefer not to do that:

* If we add this to `Drop` unconditionally the process will quit as soon as the web server starts, breaking `cratesfyi start-web-server`.
* If we add this just in `#[cfg(test)]` we would *probably* get the behavior we'd want, but it might cause problems down the road if people forget this bit.

The method is also supposed to be called just one time (from `src/test/mod.rs`), so there is not much risk of forgetting to call it. It's also pretty obvious things are wrong when you were supposed to call the method but you didn't.

src/web/pool.rs
```diff
@@ -1,29 +1,54 @@
...
+    pub(super) fn get<'a>(&'a self) -> DerefConnection<'a> {
+        match self {
+            Self::R2D2(conn) => DerefConnection::Connection(&conn),
+        }
+    }
+}
+
+pub(crate) enum DerefConnection<'a> {
+    Connection(&'a Connection),
+}
```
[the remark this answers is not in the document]
pietroalbini: Unfortunately yes, due to lifetime errors.

In Iron's request context we don't add `Pool`, but the related `PoolConnection`. This means you can't really use lifetimes in `PoolConnection`.

Now, for R2D2 `DerefConnection` can easily be avoided (as we did before), but for `Simple` we need to use it, as we need to store the intermediate `MutexGuard<'a, Connection>`. This is because the following code doesn't (rightfully) compile:

```rust
fn foo(&self) -> &Type {
    &self.mutex.lock().unwrap()
}
```

If this code was accepted by the compiler, the lock would be released as soon as `foo()` returns, creating a reference to the mutex's data without an active lock.

src/web/mod.rs
pietroalbini: I refactored this away in a later commit.

src/web/pool.rs
pietroalbini: It would be really painful to do this code wise: during `#[cfg(test)]` all the code is compiled, not just tests, and the rest of the application still expects a connection pool (for example when starting the web server from the binary).

src/test/mod.rs
pietroalbini: Tests will not see each other changes: each test has its own instace of `TestEnvironment`, and the `OnceCell`s are fields inside it. That's the reason I added the `once_cell` crate: compared to lazy_static you can make non-global lazy fields.

src/web/mod.rs
```diff
@@ -75,11 +76,12 @@ struct CratesfyiHandler {
     router_handler: Box<dyn Handler>,
     database_file_handler: Box<dyn Handler>,
     static_handler: Box<dyn Handler>,
+    pool_factory: Box<dyn Fn() -> Pool + Send + Sync>,
 }
 
 
 impl CratesfyiHandler {
-    fn chain<H: Handler>(base: H) -> Chain {
+    fn chain<H: Handler>(pool_factory: &dyn Fn() -> Pool, base: H) -> Chain {
```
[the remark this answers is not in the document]
pietroalbini: Woops.

src/test/mod.rs
pietroalbini: Because the error message is already printed above, along with the causes and the backtrace when available: the panic is only needed to mark the test as failed.

src/web/mod.rs
pietroalbini: Apart from this smoke test, I expect most if not all the frontend tests are going to use the database as well (for example to create dummy releases).

src/web/crate_details.rs
pietroalbini: Hmm, that was the approach I went with before that commit, and I wasn't much satisfied with it: unless we want to special-case db tests (why?) we're going to eventually have a bunch of combinations (`db`, `frontend`, an eventual `build`...) or nested wrappers, both of which I don't really like.

I don't think this is too bad: if you don't want to add the `let db = env.db()` you can always use `env.db().conn()...`.

pietroalbini:
Addressed or replied to all review comments.

src/test/mod.rs
jyn514: Right, my question is why not replace the 5 or so lines of boilerplate with `panic!` ?

src/web/crate_details.rs
jyn514: I agree a combinatorial explosion would be bad. I was imagining a linear number of methods: `db()`, `frontend()`, `build()`,... and if you want to do anything fancier you call `test::wrapper` yourself. 

It came to mind because I saw you were changing this a lot below, but it's not a giant deal.

src/test/mod.rs
pietroalbini: Because panic doesn't show the causes of the error, nor the error backtrace (it shows instead the `panic!()` backtrace).

src/test/mod.rs
jyn514: I missed that, that makes a lot of sense!

src/web/mod.rs
jyn514: We could have `with_frontend` pass a `(db, frontend)` tuple to the closure ... but the complexity might not be worth it. Up to you.

pietroalbini: At this point I'd prefer to keep it generic with just the `wrapper` call.

src/test/mod.rs
jyn514: Ah right, I forgot `Error` has its own backtrace.

src/web/pool.rs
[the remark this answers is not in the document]
jyn514: I see, the problem is the _lock_ has to live at least as long as the data, not just the parent struct.

jyn514:
Looks good to me!

Issue #554, opened 2020-01-08
"Versions" dropdown no longer populated.


MaulingMonkey:
This seems to be missing in every crate I've checked so far - both those recently updated and those that haven't updated in awhile.  No console errors, DOM doesn't appear to be populated, occurs in both Chrome (including with adblock disabled) and FireFox (which I don't have adblock installed for at all) on multiple computers.

### Example: https://docs.rs/lazy_static/1.4.0/lazy_static/
![image](https://user-images.githubusercontent.com/75894/71953724-f21c8700-3197-11ea-8f81-da9cff5ef514.png)

jyn514:
Probably because of https://github.com/rust-lang/docs.rs/pull/546. cc @koenaad

jyn514:
Looks like the template is referring directly to `versions`, but we changed that field to `releases`.

https://github.com/rust-lang/docs.rs/blob/master/templates/navigation_rustdoc.hbs#L76

PR #557, opened 2020-01-08
Fix regression when showing versions in dropdown

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/554

r? @koenaad


pietroalbini:
Tested locally, it works.

koenaad:
Oops 🙈 glad this could be fixed so quickly. Thanks for the quick fix @jyn514 
Weird the .hbs template didn't complain 😕

jyn514:
Templates don't have any way to throw an error. {#each some_unknown_variable} just renders as blank.

Issue #164, opened 2017-11-07
Document that system dependencies are ignored


sfackler:
I thought I had configured docs.rs correctly to install an extra system dependency, but it doesn't seem to have worked:

https://github.com/sfackler/rstack/blob/master/unwind-sys/Cargo.toml#L14
https://docs.rs/crate/unwind-sys/0.1.0/builds/75612

Any idea what's going on there?

sfackler:
As far as I can tell, that field isn't actually used anywhere...

onur:
Yes, installing system dependencies is not a thing ATM and I am not sure it will be in future. I was planning to switch to a docker based build system and use a plain image and `dependencies` to install whatever needed similar to travis. But it will just cause more headache, people have to maintain correct dependencies all the time and it may be tricky in Debian. And there is no testing suite for docs.rs, even with current simple metadata section, people are forced to upload a new version if they mess up metadata. Because of this I didn't want to implement system dependencies and trying to keep docs.rs as simple as possible.

pkgw:
It would be helpful to update the documentation to indicate that `dependencies` doesn't do anything at the moment.

jyn514:
This was never documented and is no longer in the code at all.

PR #558, opened 2020-01-09
Remove unused code

jyn514:


jyn514:
r? @koenaad 

There's lots of messing with `pub` in the PR, but what I actually care about is getting rid of the unused functions in `utils/copy.rs` and `utils/rustc_version.rs`. Getting rid of `pub` just made the compiler give me warnings so I could find them.

koenaad:
Oh hey, sorry @jyn514. I didn't see this mention between all my Github notifications. I should probably tweak my github settings a bit...
I might be able to look at this tomorrow, otherwise Tuesday should work.

src/docbuilder/metadata.rs
```diff
@@ -53,7 +53,7 @@ pub struct Metadata {
     /// System dependencies.
     ///
     /// Docs.rs is running on a Debian jessie.
-    pub dependencies: Option<Vec<String>>,
+    dependencies: Option<Vec<String>>,
```
koenaad: `Metadata::dependencies` is never used by docs.rs and it's not mentioned in the docs: https://docs.rs/about#metadata
Wouldn't it be cleaner to completely remove this field?
(Can be done in a separate PR ofc)

Any idea what the original intention was of this data? This seems to conflict with the use of `crates-build-env`, since this container already controls all of the available system dependencies...

src/utils/copy.rs
```diff
@@ -94,7 +67,7 @@ mod test {
         let destination = tempdir::TempDir::new("cratesfyi").unwrap();
 
         // lets try to copy a src directory to tempdir
-        let res = copy_dir(Path::new("src"), destination.path());
+        let res = copy_doc_dir(Path::new("src"), destination.path());
```
koenaad: Since this test deals with `copy_doc_dir` now, the test name should also reflect this --> `test_copy_doc_dir`

Also this test is `#[ignore]` which is a bit unfortunate, it would be nice to check if we can get this test working again (or delete it otherwise). --> also for another PR?

src/utils/copy.rs
```diff
@@ -8,39 +8,15 @@ use error::Result;
...
-
-fn copy_files_and_handle_html(source: PathBuf,
-                              destination: PathBuf,
-                              handle_html: bool,
-                              rustc_version: &str)
-                              -> Result<()> {
+    let destination = destination.as_ref().to_path_buf();
 
     // FIXME: handle_html is useless since we started using --resource-suffix
     //        argument with rustdoc
```
koenaad: This FIXME is resolved now 🙂 
```suggestion
```

src/lib.rs
```diff
@@ -47,10 +47,9 @@ extern crate once_cell;
 pub use self::docbuilder::RustwideBuilder;
 pub use self::docbuilder::DocBuilder;
 pub use self::docbuilder::options::DocBuilderOptions;
-pub use self::docbuilder::metadata::Metadata;
 pub use self::web::Server;
 
-pub mod error;
+pub(crate) mod error;
```
koenaad: Since this module is defined in `lib.rs`, we don't need `pub(crate)`. All other modules are descendants of the top-level module and can access its private modules.

```suggestion
mod error;
```

src/docbuilder/metadata.rs
jyn514: Yeah, so originally docs.rs used a custom sandbox that had a very custom environment, and the idea was we would run `apt-get install <dependencies>` if you had it set. Since we're using docker now it doesn't make sense anymore, I agree we should get rid of it. I actually had someone trying to use it the other day and I had to tell them it wasn't implemented: https://github.com/rust-lang/docs.rs/issues/164

jyn514:
Addressed review comments

koenaad:
Cool! Looks good to me. Again, great work making the docs.rs codebase a bit easier to work with 😄

Issue #395, opened 2019-08-12
Incorrect "no builds" badge shown for complex version string


repi:
We have a crate that has a version string `0.1.0+4.1`, when linking to the docs.rs badge for it it shows up as "no build" when using this string: `https://docs.rs/physx-sys/badge.svg`. 

![image](https://user-images.githubusercontent.com/1262692/62884611-47628280-bd37-11e9-951c-b0700546e5ca.png)

Although the actual link works and goes to the correct built documentation for that specific crate [`https://docs.rs/physx-sys/0.1.0+4.1/physx_sys/`](https://docs.rs/physx-sys/0.1.0+4.1/physx_sys/).

Looks like there may be some specific badge picture generation error for complex version strings, maybe it tries to find version just `0.1.0`, which doesn't exist, in the picture selection instead of the full `0.1.0+4.1`. While the page link itself uses the correct one.

Cargo itself when one references a crate with such a more complex version string does only use the first semver portion (MAJOR.MINOR.PATCH) and gives a warning if one specifies the rest (which it ignores).

arirawr:
Just poking around a bit on this one - could it be related to the `+` symbol being incorrectly encoded around [here](https://github.com/rust-lang/docs.rs/blob/7ef7265012e541d1974a8ecfe9ffab7aa45d694e/src/web/mod.rs#L293-L301)?

jyn514:
Reproducible crate: https://docs.rs/zstd/0.5.1+zstd.1.4.4/zstd/

BTW @repi  the [build failure](https://docs.rs/crate/physx-sys/0.2.3+4.1.1/builds/199433) for your 2.0 versions is because cmake is trying to write to the source directory, we added more sandboxing in #407. If you write to the target directory instead (cargo sets OUT_DIR), it should work.

jyn514:
A secondary problem: if _any_ version is not a valid semver range, _all_ versions are marked as being invalid: https://github.com/rust-lang/docs.rs/blob/master/src/web/mod.rs#L267

jyn514:
@arirawr it turns out this _isn't_ a problem in `match_version`, surprisingly. Somehow, when you go to /zstd/badge.svg?version=0.5.1+zstd.1.4.4 the version that gets passed to `match_version` is `0.5.1 zstd.1.4.4` (with a space).

jyn514:
I think it might be a bug in `params`? I'm giving up for now.

jyn514:
[...]

Issue #312, opened 2019-02-25
docs.rs updates latest docs link before docs are built


LPGhatguy:
Hi! I just published a new version of the `rbx_reflection` crate. The previous version was `0.1.0` and the new version was `0.2.373`.

Almost immediately after publish, `https://docs.rs/rbx_reflection` started redirecting to `https://docs.rs/rbx_reflection/0.2.373/rbx_reflection/`, which was not generated yet and thus returned a 404. The docs stayed broken this way for several minutes! 😦 

Does it make sense for docs.rs to continue pointing to the old version docs until the new version is done generating? This is a crate that is going to be published automatically every couple weeks and I'd like to avoid breaking my documentation for a few minutes so often.

onur:
Docs.rs is currently adding a release information before adding documentation into database. This is definitely a bug I was aware of and will fix it as soon as we start working on new model.

jyn514:
@LPGhatguy I can't reproduce this bug, I published a crate recently but `/:crate` redirected to the latest built version instead of the latest published version. Can you try again and see if it's still an issue? If not, I'd like to close this bug.

LPGhatguy:
Sure, I'll check when I publish the next version of this crate. Could it be that the crate is very large, and that's what causes the race? One of the functions [was quite big](https://github.com/rust-lang/rust/issues/60753) when I filed this issue.

jyn514:
I double checked that the other crate was building when I refreshed the page, and 0.4 was definitely building at the same time 0.3 was displayed. I'll wait until I hear about your crate though :)

jyn514:
Closing this since I haven't been able to reproduce the bug. Feel free to reopen if this happens again.

Issue #223, opened 2018-07-19
Opening a crates versionless url links to prerelease versions


Nemo157:
I just pushed a test crate to https://docs.rs/nemo157, there was a `0.3.0` version published previously and I just added a `0.3.1-pre` version. I expected the versionless url to still redirect to the latest non-prerelease version, but instead it links to the prerelease version; and explicitly going to https://docs.rs/nemo157/0.3.0/nemo157/ shows a bright orange "Go to latest version" link.

I think for the purposes of promoting a stable ecosystem it would be best to treat prerelease versions as ineligible to be the latest version unless a crate has no normal versions available. Having them still listed under the versions for the crate and allowing them to be linked to directly would still be useful.

udoprog:
If like to boost the signal in this a bit. It's causing a bit of a yanking party in tokio-rs/tokio#2036

jyn514:
@Nemo157 the test crate makes this much faster to debug, thanks :)

jyn514:
I'm going to try and also fix #221 at the same time.

jyn514:
This is mostly fixed, I'm just trying to solve https://github.com/rust-lang/docs.rs/issues/395 at the same time. If I don't get it in the next half hour or so I'll just make a PR without.

PR #559, opened 2020-01-10
Refactor web::latest_version, move logic into CrateDetails

koenaad:
I noticed `web/mod.rs` contained a really complicated `latest_version` function: https://github.com/rust-lang/docs.rs/blob/master/src/web/mod.rs#L316
The bulk of the work was spent in converting the version strings into `semver::Version`, to sort the array. Something that is already done in `CrateDetails`...

The original `latest_version` would return:
- `None` if `req_version` was an invalid function or if `req_version` is the latest version
- `Some(version)` if this crate has a more recent version

I have split what this function does in two functions:
- `CrateDetails::is_latest_version` --> check if this is the most recent release
- `CrateDetails::latest_version` --> grab the most recent release

I like how it is much easier to reason about now 😊

These changes should be easy to follow commit-by-commit. The real changes happen in the second commit: 309ea7a

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -255,13 +255,6 @@ impl CrateDetails {
         Some(crate_details)
     }
 
-    /// Returns all versions of this crate.
-    pub fn versions(&self) -> Vec<String> {
-        self.releases.iter()
-            .map(|release| release.version.clone())
-            .collect()
-    }
-
     /// Returns whether this release is the latest release of this crate.
     pub fn is_latest_version(&self) -> bool {
         self.version == self.latest_version()
@@ -403,7 +396,7 @@ mod tests {
     }
 
     #[test]
-    fn test_versions() {
+    fn test_releases_should_be_sorted() {
         crate::test::wrapper(|env| {
             let db = env.db();
 
@@ -414,8 +407,6 @@ mod tests {
             create_release(&db, "foo", "0.0.2", true)?;
 
             let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
-
-            assert_eq!(details.versions(), vec!["1.0.0", "0.0.3", "0.0.2", "0.0.1"]);
             assert_eq!(details.releases, vec![
                 Release { version: "1.0.0".to_string(), build_status: true },
                 Release { version: "0.0.3".to_string(), build_status: false },
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -262,8 +262,8 @@ impl CrateDetails {
 
     /// Returns the version of the latest release of this crate.
     pub fn latest_version(&self) -> &str {
-        // this is safe to unwrap: releases will always contain at least the version of this release
-        &self.releases.get(0).unwrap().version
+        // releases will always contain at least one element
+        &self.releases[0].version
     }
 }
 
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -396,17 +396,23 @@ mod tests {
             let db = env.db();
 
             // Add new releases of 'foo' out-of-order since CrateDetails should sort them descending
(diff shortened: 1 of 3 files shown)


src/web/rustdoc.rs
```diff
@@ -262,7 +262,9 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
 
     content.full = file_content;
     let crate_details = cexpect!(CrateDetails::new(&conn, &name, &version));
-    let (path, latest_version) = if let Some(latest_version) = latest_version(&crate_details.versions(), &version) {
+
+    let (path, latest_version) = if !crate_details.is_latest_version() {
+        let latest_version = crate_details.latest_version().to_string();
```
jyn514: This now calculates the latest version twice. Not a giant deal since it's now just a field, but it would be nice to only calculate once.

src/web/crate_details.rs
```diff
@@ -255,14 +255,17 @@ impl CrateDetails {
...
+    /// Returns whether this release is the latest release of this crate.
+    pub fn is_latest_version(&self) -> bool {
+        self.version == self.latest_version()
     }
-}
 
+    /// Returns the version of the latest release of this crate.
+    pub fn latest_version(&self) -> &str {
+        // this is safe to unwrap: releases will always contain at least the version of this release
+        &self.releases.get(0).unwrap().version
```
```suggestion
        &self.releases[0].version
```

koenaad:
@jyn514 I have simplified the code a bit, avoiding the need to call `latest_version()` twice. Can you take a look again? :)

src/web/mod.rs
```diff
@@ -547,16 +513,4 @@ mod test {
...
-    #[test]
-    fn test_latest_version() {
-        let versions = vec!["1.0.0".to_string(),
-                            "1.1.0".to_string(),
-                            "0.9.0".to_string(),
-                            "0.9.1".to_string()];
-        assert_eq!(latest_version(&versions, "1.1.0"), None);
-        assert_eq!(latest_version(&versions, "1.0.0"), Some("1.1.0".to_owned()));
-        assert_eq!(latest_version(&versions, "0.9.0"), Some("1.1.0".to_owned()));
-        assert_eq!(latest_version(&versions, "invalidversion"), None);
```
jyn514: Please add these test cases to the new `test_latest_version` function. It would also be great if you could add a test case that these are sorted by semver instead of alphabetically: `1.12.0` comes lexicographically before `1.2.0`, but should come after when sorted by version.

jyn514:
Overall I like this :) It will make https://github.com/rust-lang/docs.rs/issues/396, https://github.com/rust-lang/docs.rs/issues/312, https://github.com/rust-lang/docs.rs/issues/223, and https://github.com/rust-lang/docs.rs/issues/395 easier to fix now that everything's in one place.

#395 would be a good test case for this actually, let me run that locally.

jyn514:
395 was a separate issue in `match_version`, it won't be affected.

src/web/mod.rs
koenaad: Good idea! I have added these edge-cases to `test_releases_should_be_sorted` since this test focuses solely on `releases` being sorted correctly.
I would prefer not to put the same test cases in `test_latest_version` to avoid having duplicate tests.

Is that acceptable for you?

koenaad:
> Overall I like this :) It will make #396, #312, #223, and #395 easier to fix now that everything's in one place.
> 
> #395 would be a good test case for this actually, let me run that locally.

About those issues mentioning yanked crates, I have another small-ish refactor coming up which would add a `yanked` bool to `Release`. This should make it fairly easy to improve `latest_version()` and ignore yanked crates.

src/web/mod.rs
jyn514: Looks good to me :)

PR #560, opened 2020-01-10
Refactor CrateDetails: add yanked field to Release

koenaad:
This is another small refactor from working on #516.

Changes should be easier to check looking at them commit-by-commit.

Basically, the previous implementation had to query the database to determine the last successful build.
But most of the information needed is already present in `releases`. By also fetching the `yanked` data into `Release` we avoid doing an extra query.
Additionally, the `yanked` field will be useful later to improve `latest_version()`, see the issues mentioned by https://github.com/rust-lang/docs.rs/pull/559#issuecomment-573029453.

I will have to merge / rebase these changes on top of #559. But you can already take a look :)

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -338,11 +338,13 @@ mod tests {
             db.fake_release().name("foo").version("0.0.2").create()?;
             db.fake_release().name("foo").version("0.0.3").build_result_successful(false).create()?;
             db.fake_release().name("foo").version("0.0.4").cratesio_data_yanked(true).create()?;
+            db.fake_release().name("foo").version("0.0.5").build_result_successful(false).cratesio_data_yanked(true).create()?;
 
             assert_last_successful_build_equals(&db, "foo", "0.0.1", None)?;
             assert_last_successful_build_equals(&db, "foo", "0.0.2", None)?;
             assert_last_successful_build_equals(&db, "foo", "0.0.3", Some("0.0.2"))?;
-            // don't test for foo-0.0.4, yanked crates are not displayed
+            assert_last_successful_build_equals(&db, "foo", "0.0.4", None)?;
+            assert_last_successful_build_equals(&db, "foo", "0.0.5", Some("0.0.2"))?;
             Ok(())
         });
     }
@@ -358,7 +360,7 @@ mod tests {
 
             assert_last_successful_build_equals(&db, "foo", "0.0.1", None)?;
             assert_last_successful_build_equals(&db, "foo", "0.0.2", None)?;
-            // don't test for foo-0.0.3, yanked crates are not displayed
+            assert_last_successful_build_equals(&db, "foo", "0.0.3", None)?;
             Ok(())
         });
     }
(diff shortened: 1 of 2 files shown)


jyn514:
Merged #559, time for that rebase :) I'm looking in the meantime.

src/web/crate_details.rs
```diff
@@ -235,21 +236,11 @@ impl CrateDetails {
...
-                &[&name],
-            ).unwrap();
-
-            if rows.len() >= 1 {
-                crate_details.last_successful_build = Some(rows.get(0).get(0));
-            }
+            crate_details.last_successful_build = crate_details.releases.iter()
+                .filter(|release| release.build_status && !release.yanked)
+                .map(|release| release.version.to_owned())
+                .nth(0);
```
```suggestion
                .next();
```

koenaad: Aha, Rust always has a slighter better way of doing things.

koenaad:
> I started looking at this, but in the middle I discovered that we in fact show yanked crates as if they were normal crates. With that in mind, I think we should wait to refactor yanked crates until that's fixed: #561

This PR doesn't change the behavior of `last_successful_build`. I have added some tests with 
247d1e3  to pin down the expected results before changing the implementation.
So I don't think #561 should block this PR.

The current implementation will not show a `last_successful_build` warning if the build was successful (thanks to this if case: https://github.com/rust-lang/docs.rs/pull/560/files#diff-07b7024fe121ba3b5acf483a50bf3000R239).
I can update the tests to reflect this behavior? Now there is a comment saying we don't test for yanked release, since I didn't expect this to be relevant.

jyn514:
Hmm, I suppose as long as there's no change in behavior it doesn't hurt to go ahead. Please do update the tests.

>  Now there is a comment saying we don't test for yanked release, since I didn't expect this to be relevant.

I think I got yanked versions mixed up with blacklisted versions before: yanked is by the user, blacklisted is by the infra team (docs.rs or crates.io). Sorry about that.

src/web/crate_details.rs
jyn514: This is actually the way iterators are implemented under the hood :) If you look at https://doc.rust-lang.org/std/iter/trait.Iterator.html, `next()` is the only required function and all the rest are provided methods.

koenaad: Ah, that's right! I even knew that, but I guess I didn't connect the dots at the time.

koenaad:
I have added the asserts for yanked releases now.

The test expose that the behavior is a bit funky right now... in [this test](https://github.com/rust-lang/docs.rs/pull/560/commits/d87822bd7e0c0d74ee3d5aa72a90631c56656d79#diff-07b7024fe121ba3b5acf483a50bf3000R346) `0.0.4` is the most recent successful build, but since it is yanked other releases will point to 0.0.2 as `last_successful_build`... 🤔

> I think I got yanked versions mixed up with blacklisted versions before: yanked is by the user, blacklisted is by the infra team (docs.rs or crates.io). Sorry about that.

No worries, it is pretty confusing 😅

jyn514:
> The test expose that the behavior is a bit funky right now... in this test 0.0.4 is the most recent successful build, but since it is yanked other releases will point to 0.0.2 as last_successful_build...

That seems like the right behavior to me. Is there something you think we should do instead?

koenaad:
No, I think it's okay for now. It feels inconsistent since we don't mention anything on the yanked release, but we'll address that with #561 :)

PR #562, opened 2020-01-11
navigation_rustdoc: show whether a build failed in the Versions list

koenaad:
Changes in this PR are very similar to #546. In fact, the code introduced here is exactly the same as in [`crate_details.hbs`](https://github.com/rust-lang/docs.rs/blob/master/templates/crate_details.hbs#L44)

This PR impacts the navigation bar in the rustdoc view: in the dropdown menu, versions that failed to build will also have a ⚠️ sign now.

Screenshot taken with changes from this PR:

Screenshot 2020-01-11 at 04 30 58

### DIFF
--- a/templates/navigation_rustdoc.hbs
+++ b/templates/navigation_rustdoc.hbs
@@ -74,7 +74,13 @@
                         <div class="pure-menu pure-menu-scrollable sub-menu">
                           <ul class="pure-menu-list">
                             {{#each releases}}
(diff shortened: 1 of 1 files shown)


koenaad:
We could avoid the code duplication by extracting this part into a separate .hbs file. But I'm not sure if that is worth doing for about 10 lines of code.
I also don't see any other pages doing that.

jyn514:
I think for this little code it's fine to have some duplication.

PR #565, opened 2020-01-14
Fix various bugs in match_version

jyn514:
- Ignore yanked versions
- Ignore pre-release versions
- Add lots of tests

Note: this now queries `releases.version` instead of `crates.versions`
so that I have accessed to `releases.yanked`. This _shouldn't_ affect
behavior, but if a previous bug set `crates.versions` to something
different, it will no longer be seen.

Closes https://github.com/rust-lang/docs.rs/issues/223, closes https://github.com/rust-lang/docs.rs/issues/221

r? @pietroalbini

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -500,11 +500,19 @@ impl ToJson for MetaData {
 
 #[cfg(test)]
 mod test {
-    extern crate env_logger;
+    use crate::test::*;
+    use web::{match_version, MatchVersion};
+
+    fn release(version: &str, db: &TestDatabase) -> i32 {
+        db.fake_release().name("foo").version(version).create().unwrap()
+    }
+    fn version(v: Option<&str>, db: &TestDatabase) -> MatchVersion {
+        match_version(&db.conn(), "foo", v)
+    }
 
     #[test]
     fn test_index_returns_success() {
-        crate::test::wrapper(|env| {
+        wrapper(|env| {
             let web = env.frontend();
             assert!(web.get("/").send()?.status().is_success());
             Ok(())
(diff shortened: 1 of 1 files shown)


pietroalbini:
Could someone else review this?

GuillaumeGomez:
Looks good to me but as not being familiar with this code, I think a third reviewer point of view would be appreciated. :)

jyn514:
@koenaad would you be interested in reviewing this PR?

koenaad:
@jyn514 yeah sure, I'll take a look at it later today. I don't feel _super_ confident in my rust knowledge, but I should be able to review this 🙂

jyn514:
Awesome, thank you! It shouldn't be _super_ complicated, feel free to ask here or on Discord if you have any questions.

koenaad:
These changes look good to me :)

I don't have any comments on the code itself, just some general notes (not blocking for this PR):

- While the test case looks complete and it seems to cover all edge cases. It's a bit long and thus difficult to read (in my opinion). It might be nice to split up the test into smaller tests that each cover a specific functionality, something like `test_prereleases_match_with_any`, `test_yanked_crates_are_not_matched`,... Ideally just one test should fail if some of the code changes, quickly indicating what functionality broke. It also documents how it  is supposed to work.
But in the end it's mostly personal preference, so pick whatever feels most intuitive for you while developing 😄

- If I look at where `match_version` is used, it's always followed by a query on the database fetching the `rustdoc_status`.
    - here we fetch `rustdoc_status`: https://github.com/rust-lang/docs.rs/blob/master/src/web/releases.rs#L486
    - here we fetch `rustdoc_status` and `target_name`: https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L160
  
  Maybe we can change `match_version` to expose some additional information that makes the next queries easier or, even better, unnecessary?
  A simple tweak would be to return the id of the release, making the next query easier.

jyn514:
> It might be nice to split up the test into smaller tests that each cover a specific functionality

Good call, done.

> If I look at where match_version is used, it's always followed by a query on the database fetching the rustdoc_status.

Hmm, I see the comment saying 'this should be part of match_version' but I'm not sure why - giving back the latest docs with _docs_ is very different from the latest docs, that would be a big change in behavior. One of the things I was explicitly asked for https://github.com/rust-lang/docs.rs/issues/516 was to make sure it did _not_ redirect to the latest successful version so people knew a later version was published.

Returning the id sounds like a good refactor, but I think that should go in a different PR, I'll open an issue.

koenaad:
> Hmm, I see the comment saying 'this should be part of match_version' but I'm not sure why - giving back the latest version with docs is very different from the latest version, that would be a big change in behavior. 

Yeah, I agree it does not make much sense anymore given the current context... changing the implementation would break the usage of `match_version` in other places.

> Returning the id sounds like a good refactor, but I think that should go in a different PR, I'll open an issue.

Agree 👍

PR #567, opened 2020-01-20
limits: allow overriding memory limits to more than 4GB

pietroalbini:
Before this, the overridden memory limits were stored as the allowed number of bytes in a 32-bit integer, which couldn't represent more than 4GB of RAM.

This switches the database table to use a BIGINT (64-bit integer) as the column type, fixing the issue and allowing us to add big overrides.

r? @jyn514

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -269,6 +269,17 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
             // downgrade query
             "DROP TABLE blacklisted_crates;"
         ),
+        migration!(
+            context,
+            // version
+            7,
+            // description
+            "Change sandbox_overrides's max_memory_bytes to BIGINT",
+            // upgrade query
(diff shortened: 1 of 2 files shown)


src/db/migrate.rs
```diff
@@ -269,6 +269,17 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
             // downgrade query
             "DROP TABLE blacklisted_crates;"
         ),
+        migration!(
+            context,
+            // version
+            7,
+            // description
+            "Change sandbox_overrides's max_memory_bytes to BIGINT",
```
jyn514: Nit: Don't say what we're doing, say why.
```suggestion
            "Allow memory limits of more than 4 GB",
```
Other than this looks good :)

pietroalbini: Well, I kept it consistent with the other ones :sweat_smile: 

You're right though, updated.

PR #570, opened 2020-01-24
Add unit tests from the smoke test PR

jyn514:
Obsoletes https://github.com/rust-lang/docs.rs/pull/524.

This is ready for review.

Things to do:

- [ ] Make PRs for unrelated fixes
  - [x] [don't ignore DB cache if file cache is empty](https://github.com/rust-lang/docs.rs/pull/524/files#diff-a15a6c91293e16464916ab0f956d169fR47) (https://github.com/rust-lang/docs.rs/pull/523)
  - [ ] [don't ignore passing tests](https://github.com/rust-lang/docs.rs/pull/524/commits/57e270f29031061445a6d3194aeafca369130cbd)
  - [ ] Do we want to have a [FAST_INIT env var](https://github.com/rust-lang/docs.rs/pull/524/commits/922f54a80bdfba9c08a8aae338a02c9abad70712#diff-8fac5c4dd8b0132a470c525431c6c4bc)?
- [x] Test that docs for binary crates redirect to `/crate/:crate/:version`
- [x] Make sure that viewing source code works (with or without trailing slashes)
  - [x] for rustdoc
  - [x] for our source viewer
- [x] 'Go to latest version' should
  - [x] keep the page if it exists
  - [x] keep the platform if it exists
  - [x] search for the page if the page doesn't exist
- [x] /std should redirect to https://doc.rust-lang.org/stable/std/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ features = [ "with-time", "with-rustc-serialize" ]
 
 [dev-dependencies]
 once_cell = "1.2.0"
+kuchiki = "0.7"
 
 [build-dependencies]
 time = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ features = [ "with-time", "with-rustc-serialize" ]
 
 [dev-dependencies]
 once_cell = "1.2.0"
-kuchiki = "0.7"
+kuchiki = "0.8"
 
 [build-dependencies]
 time = "0.1"
--- a/src/test/fakes.rs
+++ b/src/test/fakes.rs
@@ -111,6 +111,13 @@ impl<'a> FakeRelease<'a> {
         self
     }
 
+    pub(crate) fn platform<S: Into<String>>(mut self, platform: S) -> Self {
+        let name = self.package.targets[0].name.clone();
+        let target = Target::dummy_lib(name, Some(platform.into()));
+        self.package.targets.push(target);
+        self
+    }
+
     pub(crate) fn create(self) -> Result<i32, Error> {
         use std::fs;
         use std::path::Path;
@@ -119,8 +126,11 @@ impl<'a> FakeRelease<'a> {
         let package = self.package;
         let db = self.db;
 
-        let upload_files = |prefix: &str, files: &[(&str, &[u8])]| {
-            let path_prefix = tempdir.path().join(prefix);
+        let upload_files = |prefix: &str, files: &[(&str, &[u8])], target: Option<&str>| {
+            let mut path_prefix = tempdir.path().join(prefix);
+            if let Some(target) = target {
+                path_prefix.push(target);
+            }
             fs::create_dir(&path_prefix)?;
 
             for (path, data) in files {
@@ -129,10 +139,12 @@ impl<'a> FakeRelease<'a> {
                     fs::create_dir_all(path_prefix.join(parent))?;
                 }
                 let file = path_prefix.join(&path);
+                log::debug!("writing file {}", file.display());
                 fs::write(file, data)?;
             }
 
-            let prefix = format!("{}/{}/{}", prefix, package.name, package.version);
+            let prefix = format!("{}/{}/{}/{}", prefix, package.name, package.version, target.unwrap_or(""));
+            log::debug!("adding directory {} from {}", prefix, path_prefix.display());
             crate::db::add_path_into_database(&db.conn(), &prefix, path_prefix)
         };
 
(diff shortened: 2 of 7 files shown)


jyn514:

src/test/mod.rs
```diff
@@ -43,7 +43,11 @@ pub(crate) fn assert_redirect(path: &str, expected_target: &str, web: &TestFront
     let status = response.status();
 
     // Reqwest follows redirects automatically
```
koenaad: This might be a bit nit-picky... but I find this comment confusing now. This comment originally had a slightly different place/context. Now it seems to tell us something about the conversion of `response` to `redirect_target`, but the redirect has already happened by the time we get here.

I would propose:
- move this comment up a bit, maybe at `let response = web.get(path).send()?;`?
- perhaps add a (short) comment at `let redirect_target = if ...` to clarify why we sometimes take the full URL and sometimes the path only.

src/web/rustdoc.rs
```diff
@@ -429,6 +429,18 @@ impl Handler for SharedResourceHandler {
 #[cfg(test)]
 mod test {
     use crate::test::*;
+    fn latest_version_redirect(path: &str, web: &TestFrontend) -> Result<String, failure::Error> {
+        use html5ever::tendril::TendrilSink;
+        let data = web.get(path).send()?.text()?;
+        let dom = kuchiki::parse_html().one(data);
+        for elems in dom.select("form ul li a.warn") {
```
koenaad: I'm a bit worried about this selector. If one day a second list item is added that is also `a.warn`, then this test might start to fail.
It would be nice if we could select this element using its text ("Go to latest version"), but... I imagine that would be pretty complicated to write ☹️ 

The effort to clean this up doesn't seem to be worth it for me, so I'm okay with keeping this as is.

Cargo.toml
```diff
@@ -54,6 +54,7 @@ features = [ "with-time", "with-rustc-serialize" ]
 
 [dev-dependencies]
 once_cell = "1.2.0"
+kuchiki = "0.7"
```
koenaad: Was it a deliberate decision to take 0.7? The [latest release is 0.8.0](https://crates.io/crates/kuchiki/0.8.0) and it seems to work on my machine.

src/test/fakes.rs
```diff
@@ -111,6 +111,13 @@ impl<'a> FakeRelease<'a> {
         self
     }
 
+    pub(crate) fn platform<S: Into<String>>(mut self, platform: S) -> Self {
```
koenaad:
```suggestion
    pub(crate) fn add_platform<S: Into<String>>(mut self, platform: S) -> Self {
```

This code adds a new target for the given platform, so it seems more appropriate to name it `add_platform`.

src/web/rustdoc.rs
```diff
@@ -534,7 +534,29 @@ mod test {
...
+              .create().unwrap();
+            db.fake_release().name("dummy").version("0.2.0")
+              .platform("x86_64-pc-windows-msvc")
+              .create().unwrap();
+
+            let web = env.frontend();
+            /* TODO: this should work even without a trailing slash
+            let redirect = latest_version_redirect("/dummy/0.1.0/x86_64-pc-windows-msvc/dummy", web)?;
+            assert_eq!(redirect, "/dummy/0.2.0/x86_64-pc-windows-msvc/dummy");
+            */
```
koenaad: Could you either:
- document why this doesn't work (yet) and perhaps make an issue to track this TODO
- remove this TODO

src/web/rustdoc.rs
jyn514: Since it's a test, we can fix it when it breaks without worrying about downtime, I don't think it's fragile enough to worry about until then.

Cargo.toml
jyn514: Whoops, good catch!

src/test/mod.rs
jyn514: I pushed a change that comments this, does that make it more clear?

src/web/rustdoc.rs
jyn514: Thanks for calling me out on this :) added an issue and left a link in the comment.

jyn514: Whoops, replied to the wrong TODO, I meant to comment for `src/web/mod.rs`

src/web/mod.rs
```diff
@@ -523,6 +523,53 @@ mod test {
...
+            let db = env.db();
+            db.fake_release().name("bat").version("0.2.0").binary(true)
+              .create().unwrap();
+            let web = env.frontend();
+            assert_redirect("/bat/0.2.0", "/crate/bat/0.2.0", web)?;
+            assert_redirect("/bat/0.2.0/i686-unknown-linux-gnu", "/crate/bat/0.2.0", web)?;
+            /* TODO: this should work
+            assert_redirect("/bat/0.2.0/i686-unknown-linux-gnu/bat", "/crate/bat/0.2.0", web)?;
+            assert_redirect("/bat/0.2.0/i686-unknown-linux-gnu/bat/", "/crate/bat/0.2.0/", web)?;
+            */
```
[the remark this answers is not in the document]
jyn514: https://github.com/rust-lang/docs.rs/issues/603

src/web/rustdoc.rs
jyn514: Turns out this was fixed in https://github.com/rust-lang/docs.rs/pull/584 :)

src/test/mod.rs
koenaad: Yeah, this is great :+1:

PR #575, opened 2020-01-31
Fix order by in search

### LINKED ISSUE #489: docs.rs search is ordered randomly for non-exact matches
Search for Tokio and tokio cames out completely different result
tokio first result: `tokio-0.2.0-alpha.6`
Tokio first result: `new-tokio-smtp-0.8.1`, and DO NOT have `tokio-0.2.0-alpha.6` at all
futures first result:  `futures-0.3.1`
Futures first result: `futures_future-0.1.1`, and `futures-0.3.1` is fifth result
I don't think this is the behavior a search engine should be.

a
a-r
b
b-r
c
c-r

P.S.: I found this when I was in bed, using my phone to search tokio doc, and my keyboard auto capitalized first character of tokio

Zexbe:
Fix #489

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -274,7 +274,7 @@ fn get_search_results(conn: &Connection,
                                  INNER JOIN releases ON crates.latest_version_id = releases.id
                                  WHERE crates.name LIKE concat('%', $1, '%')
                                     OR crates.content @@ to_tsquery($1)
-                                 ORDER BY crates.name = $1 DESC,
+                                 ORDER BY crates.name DESC,
                                     crates.name LIKE concat('%', $1, '%') DESC,
                                     rank DESC
                                  LIMIT $2 OFFSET $3",


src/web/releases.rs
```diff
@@ -274,7 +274,7 @@ fn get_search_results(conn: &Connection,
                                  INNER JOIN releases ON crates.latest_version_id = releases.id
                                  WHERE crates.name LIKE concat('%', $1, '%')
                                     OR crates.content @@ to_tsquery($1)
-                                 ORDER BY crates.name = $1 DESC,
+                                 ORDER BY crates.name DESC,
                                     crates.name LIKE concat('%', $1, '%') DESC,
```
jyn514: Please also remove the useless `crates.name LIKE concat('%', $1, '%')`, which is a constant 1 because of the `WHERE` clause.

jyn514:
Actually, I tried this out and I don't think it's the right fix at all. That line was making sure the exact match came first, after this PR the exact match can be much further down.

Zexbe:
Okay going to try a different approach

PR #577, opened 2020-02-01
Make search box more noticable

Zexbe:
Adds a spyglass, and backdrop to the search box.

Alternate implementation to: https://github.com/rust-lang/docs.rs/pull/149


jyn514:
Can you post a screenshot of what it looks like now?

jyn514:
Actually I have it: image

pietroalbini:
Hmm, wait, that new search field has a completly different style than the rest of the navbar.

It'd be way better to add borders consistent with the rest of the navbar, and an icon from the same set.

PR #579, opened 2020-02-01
Workaround for keyboard scrolling

Zexbe:
Works around keyboard not scrolling on crate docs pages (without clicking)

Alternate solution to:
https://github.com/rust-lang/docs.rs/pull/498 
https://github.com/rust-lang/docs.rs/pull/510

### DIFF
--- a/templates/rustdoc.hbs
+++ b/templates/rustdoc.hbs
@@ -10,8 +10,24 @@
 </head>
 <body>
 {{> navigation_rustdoc}}
-<div class="{{content.rustdoc_body_class}}" tabindex="1">
+<div id="rustdoc_body_wrapper" class="{{content.rustdoc_body_class}}" tabindex="-1">
     {{{content.rustdoc_body}}}
 </div>
 </body>
+<script>
+    var doc_body = document.getElementById("rustdoc_body_wrapper");
+    if(window.location.hash) {
+        var notFirefox = typeof InstallTrigger === 'undefined';
+        if(notFirefox) {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
This is much simpler indeed. However, I'm not sure how much we can rely on it. If we don't have any other "taxindex", I think this is the best solution currently proposed. However, I wonder if we shouldn't have a JS solution "in case" this one doesn't work at some point. What do you think?

Zexbe:
Someone would have to change something for it not to work. That would be a unit tests job to prevent, which I don’t know how to make for this. This may fail in some unknown case, but it currently fails in all cases. Unless it is merged though, nobody is going to use it, so that unknown case will never be reported. Unless we get an example that it fails at, building a script that tries to detect, and correct it probably won’t work.

jyn514:
Can we use `autofocus` on a div? Or maybe use a `#rustdoc` hash in the URL so it's in focus? https://stackoverflow.com/a/3656569

jyn514:
I think we shouldn't add JS until we need to, I agree with @Zexbe that this should be caught by a unit test, not by two pieces of code that do the same thing.

jyn514:
Have you tried this with JavaScript disabled? I think `tabindex=-1` will prevent it from being selected with the keyboard.

Zexbe:
@jyn514 I am able to navigate with tab on the docs page with javascript disabled. -1 only controls the div, and a div cannot be tabbed to by default.

PR #581, opened 2020-02-01
Add section to about page that points to how to build a crate's documentation locally

### LINKED ISSUE #574: Better documentation for how to build crates locally
Is it possible to locally test a documentation build for a crate? If so, is there any documentation about doing this? If there is a docker container or something I'd be happy to help improve to docs with this. I am asking as I had an issue where my -sys crate was having trouble building on docs.rs and I had to rerelease several times to fix all of the issues.

jackgerrits:
Fixes #574

@jyn514 let me know if you'd like the wording to be changed.

### DIFF
--- a/templates/about.hbs
+++ b/templates/about.hbs
@@ -169,7 +169,7 @@ rustc-args = [ "--example-rustc-arg" ]
 rustdoc-args = [ "--example-rustdoc-arg" ]</pre></code>
 
   <h4>Test crate documentation build locally</h4>
-  <p>The <a href="https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand" target="_blank">docs.rs README</a> details how to build an unpublished crate's documentation locally using the same Docker image used by the build agent.</p>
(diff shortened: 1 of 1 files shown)


templates/about.hbs
```diff
@@ -168,6 +168,9 @@ rustc-args = [ "--example-rustc-arg" ]
 # Additional `RUSTDOCFLAGS` to set (default: none)
 rustdoc-args = [ "--example-rustdoc-arg" ]</pre></code>
 
+  <h4>Test crate documentation build locally</h4>
+  <p>The <a href="https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand" target="_blank">docs.rs README</a> details how to build an unpublished crate's documentation locally using the same Docker image used by the build agent.</p>
```
jyn514: nit:
```suggestion
  <p>The <a href="https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand" target="_blank">docs.rs README</a> describes how to build an unpublished crate's documentation locally using the same Docker image used by the build agent.</p>
```

jyn514:
Note that this doesn't not just use the same docker image, it's also uses the same sandboxing and `rustc` version.

templates/about.hbs
jackgerrits: Your wording is definitely an improvement.

jackgerrits:
That's good to know, being able to do this locally is super helpful!

PR #582, opened 2020-02-02
Improve link fallback

### LINKED ISSUE #288: Try to serve index.html if link points to directory
This change may fix a lot of broken links such as the ones I've found in https://github.com/hyperium/hyper/issues/1749 and https://github.com/hyperium/hyper/issues/1750

Zexbe:
Improve handling of urls that are missing a /
Ex: http://127.0.0.1:3000/hyper/0.13.2/hyper/client/conn

Fix #288

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -111,7 +111,7 @@ impl CratesfyiHandler {
 
         let shared_resources = Self::chain(&pool_factory, rustdoc::SharedResourceHandler);
         let router_chain = Self::chain(&pool_factory, routes.iron_router());
-        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX").unwrap()).join("public_html");
+        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX").expect("CRATESFYI_PREFIX environment variable does not exists")).join("public_html");
         let static_handler = Static::new(prefix)
             .cache(Duration::from_secs(STATIC_FILE_CACHE_DURATION));
 
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -20,6 +20,7 @@ use time;
 use iron::Handler;
 use utils;
 
+const DIR_DEFAULT_FILE: &str = "index.html";
 
 #[derive(Debug)]
 struct RustdocPage {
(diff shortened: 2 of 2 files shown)


src/web/mod.rs
```diff
@@ -111,7 +111,7 @@ impl CratesfyiHandler {
 
         let shared_resources = Self::chain(&pool_factory, rustdoc::SharedResourceHandler);
         let router_chain = Self::chain(&pool_factory, routes.iron_router());
-        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX").unwrap()).join("public_html");
+        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX").expect("CRATESFYI_PREFIX environment variable does not exists")).join("public_html");
```
jyn514: What do you think, is this overkill?

```suggestion
        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX")
            .expect("the CRATESFYI_PREFIX environment variable is not set, \
                     see https://github.com/rust-lang/docs.rs/wiki/Developing-without-docker-compose \
                     for instructions on developing locally")
        ).join("public_html");
```

src/web/rustdoc.rs
```diff
@@ -448,8 +452,10 @@ mod test {
             let web = env.frontend();
             assert_success("/", web)?;
             assert_success("/crate/buggy/0.1.0/", web)?;
+            assert_success("/crate/buggy/0.1.0/", web)?;
```
jyn514: I think this was a typo.

```suggestion
            assert_success("/crate/buggy/0.1.0", web)?;
```

src/web/rustdoc.rs
```diff
@@ -238,15 +238,18 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
 
-    let path = {
-        let mut path = req_path.join("/");
-        if path.ends_with('/') {
+    if let Some(&end) = req_path.last(){
+        if end == "" {
             req_path.pop(); // get rid of empty string
-            path.push_str("index.html");
             req_path.push("index.html");
+        } else if !end.contains('.') {
```
jyn514: This is not the solution I would like to see. If there is a file (not a directory) called `/:crate/:version/name`, this will prevent it from being seen, instead always redirecting to the directory. I would rather fallback to the directory in https://github.com/rust-lang/docs.rs/blob/777b37cad5ce0812a184f6dcf6603e85ced3cd26/src/web/rustdoc.rs#L254 only if the file does not already exist.

Also, this is a bit of a fragile part of the code and I'd rather not change it unless absolutely necessary.

src/web/mod.rs
Zexbe: The issue I feel with putting the link on the expect is that it is weird for this one to have it, and not the other ones, but if all of them have it, it gets pretty ugly, so maybe it needs to just check all the environment variables at the start, and print an error message, but I don't think it should be put into this pull request.

src/web/rustdoc.rs
Zexbe: It was a typo

src/web/rustdoc.rs
```diff
@@ -238,26 +239,28 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
-            path.push_str("index.html");
-            req_path.push("index.html");
-        }
-        path
+    let mut path = req_path.join("/");
+
+    //If it is a directory, we default to looking 
+    if path.ends_with("/") {
+        path.push_str(DIR_DEFAULT_FILE);
+        req_path.push(DIR_DEFAULT_FILE);
```
jyn514: I don't think this is the same behavior, it will turn `/crate/version/` into `/crate/version//index.html`.

This will mess up the redirect to latest version, which unfortunately only has unit tests in #570.

Since that's the case, I'm not comfortable merging until either #570 is merged or I get a chance to test locally which will probably be a few days.

If you instead change _only_ the `from_path` block (line 251), I would be more comfortable with the changes.

src/web/rustdoc.rs
```diff
@@ -315,7 +318,7 @@ fn path_for_version(req_path: &[&str], target_name: &str, conn: &Connection) ->
         return req_path[3..].join("/");
     }
     // this page doesn't exist in the latest version
-    let search_item = if *req_path.last().unwrap() == "index.html" {
+    let search_item = if *req_path.last().unwrap() == DIR_DEFAULT_FILE {
```
jyn514: Is there a reason you made this change?

src/web/rustdoc.rs
```diff
@@ -438,18 +441,24 @@ mod test {
             db.fake_release()
               .name("buggy").version("0.1.0")
               .build_result_successful(true)
-              .rustdoc_file("settings.html", b"some data")
+              .rustdoc_file("settings.html", b"some setting data")
               .rustdoc_file("all.html", b"some data 2")
+              .rustdoc_file("file_without_ext", b"some data 3")
+              .rustdoc_file("some_module/directory/index.html", b"some data 3")
+              .rustdoc_file("some_module/file_without_ext", b"some data 3")
```
jyn514: Please also test `/directory.html/index.html`, which I don't think will work currently.

src/web/mod.rs
jyn514:
```suggestion
        let prefix = PathBuf::from(env::var("CRATESFYI_PREFIX").expect("the CRATESFYI_PREFIX environment variable is not set")).join("public_html");
```

Issue #580, opened 2020-02-01
Allow compiling without libmagic


jyn514:
This has been an issue for various contributors. I would like to be able to run the test suite (even if not the builds) on architectures other than x86_64 debian.

See https://github.com/bjorn3/rustc_codegen_cranelift/issues/884.

cc @Zexbe

jyn514:
libmagic is used in [`add_path_into_database`](https://github.com/rust-lang/docs.rs/blob/master/src/db/file.rs#L154) to detect the file type. I wonder if we could get away with having a static database of file extensions?

Zexbe:
Do you always have the file extension? Would this work? https://crates.io/crates/mime_guess

jyn514:
Not all files have file extensions, so this isn't 100% reliable, I was mostly imagining it for local testing where you really only care about .html, .css and .js

jyn514:
But yeah that could work locally

jyn514:
Fixed in https://github.com/rust-lang/docs.rs/pull/600

PR #586, opened 2020-02-03
Build on windows support [WIP, No Merge]

Zexbe:
1. Conditionally support background, so fork does not block windows from compiling
2. Use mime_guess instead of magic crate.

Fix: https://github.com/rust-lang/docs.rs/issues/580

I am going to split this up into smaller pull requests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,6 +321,7 @@ dependencies = [
  "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "mime_guess 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "once_cell 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "params 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "postgres 0.15.2 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -346,7 +347,6 @@ dependencies = [
  "time 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "tree_magic 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 1.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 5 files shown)


jyn514:
The `fork()` stuff looks fine to me, we don't run on windows in prod and it will be very clear that it doesn't work if someone runs `cratesfyi daemon`.

I'm a little more concerned about the change to `tree_mime` since it hasn't been updated in a year. I also took a look about how we use mime types and it looks very fragile to me: [we treat anything that doesn't start with `text/` as binary](https://github.com/rust-lang/docs.rs/blob/10dc6f5eb524f6e1aacd78dbf49b66d2e6ad4a27/src/web/source.rs#L132) (even `application/json` and similar). We should add tests to be sure the behavior is the same, at least for Rust, HTML, CSS, JS, markdown, and TOML.

src/db/file.rs
```diff
@@ -352,4 +332,17 @@ mod test {
...
+
+    #[test]
+    fn test_mime_types() {
+        assert_eq!(detect_mime(Path::new(".gitignore")), "text/plain");
+        assert_eq!(detect_mime(Path::new("hello.toml")), "text/x-toml");
+        assert_eq!(detect_mime(Path::new("hello.css")), "text/css");
+        assert_eq!(detect_mime(Path::new("hello.js")), "application/javascript");
+        assert_eq!(detect_mime(Path::new("hello.html")), "text/html");
+        assert_eq!(detect_mime(Path::new("hello.hello.md")), "text/x-markdown");
+        assert_eq!(detect_mime(Path::new("hello.markdown")), "text/markdown");
```
jyn514: Out of curiosity, do you know why `.markdown` is a different mimetype from `.md`?

src/db/file.rs
```diff
@@ -352,4 +332,17 @@ mod test {
         let files = get_file_list(env::current_dir().unwrap().join("Cargo.toml")).unwrap();
         assert_eq!(files[0], std::path::Path::new("Cargo.toml"));
     }
+
+    #[test]
+    fn test_mime_types() {
```
jyn514: Have you tested this with the old `magic` library as well?

Cargo.lock
```diff
@@ -1053,13 +1030,11 @@ dependencies = [
 
 [[package]]
 name = "mime_guess"
-version = "2.0.0-alpha.6"
+version = "2.0.1"
```
jyn514: Well look at that, `multipart` was already using `mime_guess` so we don't even have any more dependencies :)

src/db/file.rs
Zexbe: I thought it might be an official, vs unofficial kind of thing. I did more research tho, and they should both just be text/markdown. https://tools.ietf.org/html/rfc7763

src/db/file.rs
Zexbe: No

jyn514:
> ---- web::rustdoc::test::settings_html stdout ----
the test failed: database error: ERROR: null value in column "mime" violates not-null constraint

Try adding a migration to `src/db/migrate.rs`, there's lots of examples in that file.

PR #587, opened 2020-02-03
Make search tab style match navigation

Zexbe:
1. Change the icon to use font awesome
2. Use the same border the other menu items are using

small

medium

large

### DIFF
--- a/templates/navigation.hbs
+++ b/templates/navigation.hbs
@@ -3,7 +3,7 @@
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
           <form action="/releases/search" method="GET" class="landing-search-form-nav">
             {{#unless varsb.show_search_form}}
-            <input class="search-input-nav" name="query" type="text" aria-label="Find crate by search query" placeholder="Find crate &#x1F50D;"{{#if varss.search_query}} value="{{varss.search_query}}"{{/if}}>
+            <input class="search-input-nav" name="query" type="text" aria-label="Find crate by search query" placeholder="&#xf002; Find crate"{{#if varss.search_query}} value="{{varss.search_query}}"{{/if}}>
             {{/unless}}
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
I guess you tried to use "fa-search" from font-awesome. Unfortunately, it doesn't work unless you add "font-family: FontAwesome" on the element. But then, the text would change as well. So not really ideal. :)

The border change looks good.

Zexbe:
Oh no, I solved that issue. I lost the change somehow, one second.

GuillaumeGomez:
Please add a screenshot too so people can see the changes.

Zexbe:
Okay, I added pictures for three different media sizes.

GuillaumeGomez:
Thanks! I'll merge once CI pass.

pietroalbini:
Could you move the icon out of the placeholder so it's shown even when something is written in the input?

Zexbe:
I'm not sure how I would make that work.
The icon would have to be all the way to the left, or right to keep it from moving as you typed.

pietroalbini:
It'd be fine for me if the icon is all the way to the left.

pietroalbini:
...and if we do that we could also left-align the input to make it less weird.

Zexbe:
It would also be a button?

Zexbe:
It is done, it looks even better now. I updated the pictures

jyn514:
Any objections to merging?

GuillaumeGomez:
Not from me. Let's wait for @pietroalbini's approval. ;)

pietroalbini:
On the label add:

```css
label {
  cursor: pointer;
  padding-left: 0.5rem;
  font-size: 0.8em;
}
```
That way is more closely aligned with the rest of the items:

Zexbe:
Okay, updated the search label

pietroalbini:
Thanks!

PR #594, opened 2020-02-06
Remove unused code redux

jyn514:
This fixes the panic from before, I added a test and also built a couple crates locally.

I also refactored `copy_doc_dir` to remove the gratuitous use of PathBuf.

r? @koenaad

### DIFF
--- a/src/utils/copy.rs
+++ b/src/utils/copy.rs
@@ -13,8 +13,7 @@ use regex::Regex;
 /// Target directory must have doc directory.
 ///
 /// This function is designed to avoid file duplications.
-pub fn copy_doc_dir<P: AsRef<Path>>(target: P, destination: P) -> Result<()> {
-    let source = PathBuf::from(target.as_ref()).join("doc");
+pub fn copy_doc_dir<P: AsRef<Path>>(source: P, destination: P) -> Result<()> {
     let destination = destination.as_ref().to_path_buf();
 
     // Make sure destination directory exists
@@ -27,7 +26,7 @@ pub fn copy_doc_dir<P: AsRef<Path>>(target: P, destination: P) -> Result<()> {
         r"(\.lock|\.txt|\.woff|\.svg|\.css|main-.*\.css|main-.*\.js|normalize-.*\.js|rustdoc-.*\.css|storage-.*\.js|theme-.*\.js)$")
         .unwrap();
 
-    for file in source.read_dir()? {
+    for file in source.as_ref().read_dir()? {
 
         let file = file?;
         let mut destination_full_path = PathBuf::from(&destination);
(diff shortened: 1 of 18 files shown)


koenaad:
The last commits looks good! Great job on tracking down the `doc/` recursion.

When I test it locally the styling is all messed up, I'm not sure if this is related to this PR or that my local setup broke :(

Screenshot 2020-02-08 at 00 03 43

In the console I get a bunch of errors when loading resources that contain `-20200206-1.43.0-nightly-442ae7f04`:
```
Loading failed for the <script> with source “http://localhost:3000/storage-20200206-1.43.0-nightly-442ae7f04.js”.
This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.
Loading failed for the <script> with source “http://localhost:3000/theme-20200206-1.43.0-nightly-442ae7f04.js”.
Loading failed for the <script> with source “http://localhost:3000/main-20200206-1.43.0-nightly-442ae7f04.js”.
ReferenceError: addSearchOptions is not defined
```

koenaad:
Nevermind the above ☝️, this also happens if I build/run from the master branch...

So... this PR seems good but I can't reliably test it now... And you might want to ask for a second opinion after the trouble the last PR caused on production 🙈

jyn514:
> So... this PR seems good but I can't reliably test it now... And you might want to ask for a second opinion after the trouble the last PR I reviewed caused on production

It's ok, [this](https://github.com/rust-lang/docs.rs/pull/466) [has](https://github.com/rust-lang/docs.rs/pull/534) [happened](https://github.com/rust-lang/docs.rs/pull/519) [before](https://github.com/rust-lang/docs.rs/pull/518) even with me and pietro reviewing each others code ... We need test cases for builds (also, it would be nice to get #570 merged).

> When I test it locally the styling is all messed up, I'm not sure if this is related to this PR or that my local setup broke :(

As you've noticed, this is not because of this PR. I'm confident this won't change any of the CSS (how could it?). The way I tested locally was by building crates (`docker-compose run web build crate regex 1.3.0`).

koenaad:
> As you've noticed, this is not because of this PR. I'm confident this won't change any of the CSS (how could it?). The way I tested locally was by building crates (docker-compose run web build crate regex 1.3.0).

Yeah, you are right. Just wanted to make sure this wasn't caused by an incorrect file copy.

src/utils/copy.rs
```diff
@@ -1,75 +1,39 @@
 
-// FIXME: There is so many PathBuf's in this module
-//        Conver them to Path
-
-use std::path::{Path, PathBuf};
```
koenaad: Nice work cleaning up all the `PathBuf`'s 🙂

PR #596, opened 2020-02-07
Format code using 'cargo fmt'

Atul9:


jyn514:
So, while I'd like to use `cargo fmt` eventually, it's going to be a big task (as you can see from the diff). It will also cause major merge conflicts in nearly every PR we have. I'd also like to configure the settings a little in .rustfmt.toml first. For all of these reasons, this makes `cargo fmt` not a very good first PR to docs.rs.

If you'd like to contribute, you can take a look at the issues page or run `cargo clippy`, I'd be very happy to have fewer clippy warnings.

PR #597, opened 2020-02-07
Add minio container to mock s3 locally

### LINKED ISSUE #463: Use `minio` instead of postgres locally
Discussed on discord:

> @QuietMisdreavus @jynelson btw... if we want to "bless" docker-compose we could add minio to it (to replicate s3) and completly ditch storing files in the db

> QuietMisdreavusYesterday at 4:32 PM
true, if there's a stock minio container then we could do that

 > pietroalbiniYesterday at 4:33 PM
yeah -- https://docs.min.io/docs/minio-docker-quickstart-guide

 > QuietMisdreavusYesterday at 4:34 PM
my current dev server would have to start running minio but that sounds fairly easy
based on your descriptions

 > jynelsonYesterday at 4:42 PM
minio is like s3 but locally? I like databases, but that would allow us to get rid of the database code which sounds nice ..

I _think_ this involves spinning up a minio container instead of a postgres container and using rusoto_s3 to interact with it. @pietroalbini would know more.

koenaad:
Extends docker-compose.yml to also launch an container with minio, backed by a persistent volume.
Closes #463

It seems to work, but I can't test it reliably due this other issue I'm having right now: https://github.com/rust-lang/docs.rs/pull/594#issuecomment-583657798

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,17 @@ information on how to setup this environment.
 
 ### Docker-Compose
 
+The services started by Docker-Compose are defined in [docker-compose.yml].
+Three services are defined:
+
+| name | access                | credentials                | description                            |
+|------|-----------------------|----------------------------|----------------------------------------|
+| web  | http://localhost:3000 | N/A                        | A container running the docs.rs binary |
+| db   | /                     | N/A                        | Postgres database used by web          |
+| s3   | http://localhost:9000 | `cratesfyi` - `secret_key` | Minio (simulates AWS S3) used by web   |
+
+[docker-compose.yml]: ./docker-compose.yml
+
 #### Rebuilding Containers
 
 To rebuild the site, run `docker-compose build`.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
         build: .
         depends_on:
             - db
+            - s3
         ports:
             - "3000:3000"
         volumes:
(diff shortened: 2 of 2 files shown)


koenaad:
TODO:

- automatically create the bucket `rust-docs-rs`

Open questions:

- Should we expose port 9000? This serves a webpage on which you can browse through the buckets/files.
  We don't expose the port of the database, so this would not be consistent.

- Should we remove the code that can fallback to the database if S3 is not available?

jyn514:
> Should we expose port 9000? This serves a webpage on which you can browse through the buckets/files.

You currently have it exposed. I kind of like this, I would find it very helpful for debugging.

> Should we remove the code that can fallback to the database if S3 is not available?

That would make it harder to work on docs.rs without using docker ... Do you know if minio is easy to get set up on Windows?

koenaad:
> That would make it harder to work on docs.rs without using docker ... Do you know if minio is easy to get set up on Windows?

I would imagine you could also use Docker on Windows?
There are instructions to install minio on your local machine, but I haven't tried them myself: https://docs.min.io/docs/minio-quickstart-guide.html

jyn514:
Yes, but you can't run `cargo test` with docker.

koenaad:
Since port 9000 is open, you could run `cargo test` and let the tests connect to the minio instance from the docker-compose.
I have done something similar for tests that requires the database (you have to expose the port and set the correct environment variables).

But then we would also have to look at cleaning up the bucket after the tests have run...
 --> seems like a good next PR: integrate S3/minio in `test::TestEnvironment` so we can test with `S3Client` as well :)

I think for now it is best to keep this code. If we find out no-one uses this fallback mechanism anymore we can clean it up.

jyn514:
Wow, browsing s3 locally is _super_ helpful. Thanks for working on this!

jyn514:
Can you add instructions to the README for how to access this locally? Just 'navigate to localhost:9000` in your browser' is fine.

koenaad:
@jyn514 done! Rendered: https://github.com/koenaad/docs.rs/blob/mock_s3/README.md

PR #598, opened 2020-02-08
Normalize windows filepath

Zexbe:
On windows, the files end up getting stored like this `something/buggy\\index.html`, which makes the lookup logic fail in other places. This change normalizes it so it saves the same way on both unix, and windows.


jyn514:
Is there any way you could only compile in `path-slash` on windows? I'd rather not add more dependencies in production.

Zexbe:
I changed it to only use the dependency on windows

jyn514:
Looks good to me. I haven't tested on Windows, I assume it works there?

PR #599, opened 2020-02-08
Only use fork on non-windows OS

Zexbe:
Windows does not have fork, so this lets it get closer to compiling on windows.

### DIFF
--- a/src/utils/daemon.rs
+++ b/src/utils/daemon.rs
@@ -5,12 +5,8 @@
 
 use std::{env, thread};
 use std::panic::{catch_unwind, AssertUnwindSafe};
-use std::process::exit;
-use std::fs::File;
-use std::io::Write;
 use std::time::Duration;
 use std::path::PathBuf;
-use libc::fork;
 use time;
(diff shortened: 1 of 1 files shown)


src/utils/daemon.rs
```diff
@@ -36,15 +38,22 @@ pub fn start_daemon(background: bool) {
...
-        // fork the process
-        let pid = unsafe { fork() };
-        if pid > 0 {
-            let mut file = File::create(dbopts.prefix.join("cratesfyi.pid"))
-                .expect("Failed to create pid file");
-            writeln!(&mut file, "{}", pid).expect("Failed to write pid");
-
-            info!("cratesfyi {} daemon started on: {}", ::BUILD_VERSION, pid);
-            exit(0);
+        #[cfg(target_os = "windows")]
```
jyn514:
```suggestion
        #[cfg(target_os = "windows")]
```

jyn514:
I tested locally and `daemon` still works fine on Linux.

PR #600, opened 2020-02-09
Use different library for mime on windows

Zexbe:
Windows can't build with magic, so this gives it an alternative crate to use.
While adding unit tests, and checking what magic returned, I found a few mimes it was returning incorrectly, and have also corrected those.

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -281,13 +281,13 @@ fn detect_mime(_content: &Vec<u8>, file_path: &Path) -> Result<String> {
 
 fn correct_mime(mime: &str, file_path: &Path) -> Result<String> {
     Ok(match mime {
-        "text/plain" | "text/troff" | "text/x-markdown" | "text/x-rust" => {
+        "text/plain" | "text/troff" | "text/x-markdown" | "text/x-rust" | "text/x-toml" => {
             match file_path.extension().and_then(OsStr::to_str) {
                 Some("md") => "text/markdown",
                 Some("rs") => "text/rust",
                 Some("markdown") => "text/markdown",
                 Some("css") => "text/css",
-                Some("toml") => "text/x-toml",
+                Some("toml") => "text/toml",
                 Some("js") => "application/javascript",
                 Some("json") => "application/json",
                 _ => mime
@@ -379,7 +379,7 @@ mod test {
     #[test]
     fn test_mime_types() {
         check_mime("/ignored", ".gitignore", "text/plain");
-        check_mime("[package]", "hello.toml","text/x-toml");
+        check_mime("[package]", "hello.toml","text/toml");
         check_mime(".ok { color:red; }", "hello.css","text/css");
         check_mime("var x = 1", "hello.js","application/javascript");
         check_mime("<html>", "hello.html","text/html");
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -16,6 +16,10 @@ use failure::err_msg;
 use rusoto_s3::{S3, PutObjectRequest, GetObjectRequest, S3Client};
 use rusoto_core::region::Region;
 use rusoto_credential::DefaultCredentialsProvider;
+#[cfg(not(windows))]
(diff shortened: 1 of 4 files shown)


src/db/file.rs
```diff
@@ -195,25 +191,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
-                    } else if e == "js" {
-                        "application/javascript".to_owned()
-                    } else {
-                        mime.to_owned()
-                    }
-                } else {
-                    mime.to_owned()
-                }
-            };
+            let mime = detect_mime(&content, &file_path)?;
```
jyn514: This loads a new cookie for every item in the loop, which I imagine could have a performance inpact. It might be better to have `detect_mime` take the cookie as a parameter on linux.

src/db/file.rs
```diff
@@ -276,6 +254,46 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+#[cfg(not(windows))]
+fn detect_mime(content: &Vec<u8>, file_path: &Path) -> Result<String> {
+    use std::ffi::OsStr;
+    use magic::{Cookie, flags};
+    let cookie = Cookie::open(flags::MIME_TYPE)?;
+    cookie.load::<&str>(&[])?;
+    let mime = cookie.buffer(&content)?;
+
+    // magic is not specific enough sometimes
+    match mime.as_str() {
```
jyn514: This makes me a little nervous ... I want to look around and see what the mime type is used for before we change it.

src/db/file.rs
```diff
@@ -356,4 +385,28 @@ mod test {
...
+    fn test_mime_types() {
+        check_mime("/ignored", ".gitignore", "text/plain");
+        check_mime("[package]", "hello.toml","text/x-toml");
+        check_mime(".ok { color:red; }", "hello.css","text/css");
+        check_mime("var x = 1", "hello.js","application/javascript");
+        check_mime("<html>", "hello.html","text/html");
+        check_mime("## HELLO", "hello.hello.md","text/markdown");
+        check_mime("## WORLD", "hello.markdown","text/markdown");
+        check_mime("{}", "hello.json","application/json");
+        check_mime("hello world", "hello.txt","text/plain");
```
jyn514: Please add a test case for rust source code. In particular, [we depend](https://github.com/rust-lang/docs.rs/blob/master/src/web/source.rs#L132) on rust source starting with `text/`.

src/db/file.rs
```diff
@@ -276,6 +262,51 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+                _ => Ok(mime.to_owned())
+            }
+        }
+        "text/troff" => {
+            match file_path.extension().and_then(OsStr::to_str) {
+                Some("css") => Ok("text/css".to_owned()),
+                _ => Ok(mime.to_owned())
+            }
+        }
+        _ => Ok(mime.to_owned())
```
jyn514: I see `to_owned()` a lot here, is there any way you could factor that out to only come at the end of the function? Might make it easier to return a borrow in the future.

src/db/file.rs
```diff
@@ -276,6 +262,51 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
+
+#[cfg(windows)]
+fn detect_mime(_content: &Vec<u8>, file_path: &Path) -> Result<String> {
+    let file_ext = file_path.extension().map(|ext| ext.to_str()).unwrap_or_default();
+    Ok(match file_ext {
+        Some("md") => "text/markdown".to_owned(),
+        Some("rs") => "text/rust".to_owned(),
+        None => "text/plain".to_owned(),
+        _ => mime_guess::from_path(file_path).first_raw().map(|m| m.to_owned()).unwrap_or("text/plain".to_owned())
+    })
```
jyn514: Ditto `to_owned()`, see above

PR #606, opened 2020-02-13
refactor match version

### LINKED ISSUE #569: Return crate id from match_version
If I look at where `match_version` is used, it's always followed by a query on the database fetching the `rustdoc_status`.
    - here we fetch `rustdoc_status`: https://github.com/rust-lang/docs.rs/blob/master/src/web/releases.rs#L486
    - here we fetch `rustdoc_status` and `target_name`: https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L160
  
  Maybe we can change `match_version` to expose some additional information that makes the next queries easier or, even better, unnecessary?
  A simple tweak would be to return the id of the release, making the next query easier.

_Originally posted by @koenaad in https://github.com/rust-lang/docs.rs/pull/565#issuecomment-577442795_

Zeegomo:
Refactor `MatchVersion` to contain also the id of the release in order to simplify successive queries.

fix #569

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -190,10 +190,10 @@ impl Handler for CratesfyiHandler {
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum MatchVersion {
     /// `match_version` was given an exact version, which matched a saved crate version.
-    Exact(String),
+    Exact((String, i32)),
     /// `match_version` was given a semver version requirement, which matched the given saved crate
     /// version.
-    Semver(String),
+    Semver((String, i32)),
     /// `match_version` was given a version requirement which did not match any saved crate
     /// versions.
     None,
@@ -202,7 +202,7 @@ enum MatchVersion {
 impl MatchVersion {
     /// Convert this `MatchVersion` into an `Option`, discarding whether the matched version came
     /// from an exact version number or a semver requirement.
-    fn into_option(self) -> Option<String> {
+    fn into_option(self) -> Option<(String, i32)> {
         match self {
             MatchVersion::Exact(v) | MatchVersion::Semver(v) => Some(v),
             MatchVersion::None => None,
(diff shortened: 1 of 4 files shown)


src/web/mod.rs
```diff
@@ -516,11 +516,25 @@ mod test {
...
     fn release(version: &str, db: &TestDatabase) -> i32 {
         db.fake_release().name("foo").version(version).create().unwrap()
     }
-    fn version(v: Option<&str>, db: &TestDatabase) -> MatchVersion {
-        match_version(&db.conn(), "foo", v)
+    fn version(v: Option<&str>, db: &TestDatabase) -> Option<String> {
+        strip_id(match_version(&db.conn(), "foo", v))
+    }
+
+    fn strip_id(v: MatchVersion) -> Option<String> {
```
jyn514: This looks useful, can you make it a method on MatchVersion instead of in test code?

src/web/mod.rs
```diff
@@ -190,10 +190,10 @@ impl Handler for CratesfyiHandler {
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum MatchVersion {
     /// `match_version` was given an exact version, which matched a saved crate version.
-    Exact(String),
+    Exact((String, i32)),
     /// `match_version` was given a semver version requirement, which matched the given saved crate
```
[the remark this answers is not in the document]
gbuglione: Well, the i32 here is the release id, not the crate id. I think all the successive queries use the release id, not the crate one.
Anyway, adding a comment seems like a good idea :)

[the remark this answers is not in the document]
Zeegomo: Well, the i32 here is the release id, not the crate id as I think that's what is used by all the successive queries (like [this](https://github.com/rust-lang/docs.rs/blob/5618f17b1937a8b76cb49a002afa8e2d939feb01/src/web/releases.rs#L491))
Anyway, adding a comment seems like a good idea :)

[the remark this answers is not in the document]
jyn514: Yes, I meant the release id, thank you.

src/web/mod.rs
```diff
@@ -581,8 +582,8 @@ mod test {
             let release_id = release("0.3.0", db);
             let query = "UPDATE releases SET yanked = true WHERE id = $1 AND version = '0.3.0'";
             db.conn().query(query, &[&release_id]).unwrap();
-            assert_eq!(version(None, db), strip_id(MatchVersion::None));
-            assert_eq!(version(Some("0.3"), db), strip_id(MatchVersion::None));
+            assert_eq!(version(None, db), MatchVersion::None.strip_id());
+            assert_eq!(version(Some("0.3"), db), MatchVersion::None.strip_id());
```
jyn514: Do you need to call `strip_id()` here if the version is already `None`?

Zeegomo: hmm, yes, because they're different types. `version(Some("0.3"))` is `Option<String>` while the other is `MatchVersion`. Here `strip_id` is just used to convert into the correct type.

jyn514: Ok, it might be easier to just say `None` instead of `MatchVersion::None` then.

Issue #595, opened 2020-02-07
iPhone scrolling is painful


cormacrelf:
Recently, docs.rs gained a fixed header bar. To accommodate this, all other content on the page was moved into an `overflow: auto` div of some kind. This is a problem. For context, there are two kinds of scrolling used on websites in Safari on an iPhone:

1.	Inertial scrolling, keeps moving until simulated friction stops it after you lift your finger. Covers a lot of ground very quickly but allows precise scrolling at low speed. Also has some priority over child type 2 elements for determining which content to scroll.
2.	Locked panning, 1:1 with touch movement, stops dead when finger lifts.

Type 1 is used for entire pages where the document is larger than the viewport. Type 2 is used for overflowing content, e.g. wide code blocks clipped to viewport width, scrolling horizontally.

It used to be that docs.rs pages were bigger than the viewport, and scrolled with type 1. Now that the entire page is in an `overflow: auto` div, Type 2 is used on everything. This means:

1.	Scrolling is slow, because no inertia. Docs pages are very long!
2.	You can get stuck in code blocks that are viewport height or larger. Because they take priority on touch events for scroll, you actually cannot scroll the page while they take over the screen. All touches are interpreted as locked horizontal pans, no matter how vertically you move your finger. Fortunately there is a small white margin on the left that allows (type 2) page scrolling, so you can get out, but it’s pretty bad.
3. You can’t use the “tap status bar to scroll to top” feature any more.

Together, this is very painful UI. I use docs.rs from my phone every day. Please use `position: fixed` for fixed position navigation bars, or don’t do them at all.

jyn514:
Hmm, maybe we should have gone with https://github.com/rust-lang/docs.rs/pull/510 after all ... cc @GuillaumeGomez @Koxiaet

cormacrelf:
FWIW the standard practice for the link-offset-with-fixed-header issue in that thread is to have an explicit, invisible, empty anchor tag with a name (maybe an id attribute instead, I’d have to check), absolutely positioned, relative to the original heading, with `top: -$header_height`. Navigation then scrolls to the invisible anchor, leaving it under the bar and the actual header visible. No JS necessary. Just an extra anchor for each navigable link.

cormacrelf:
For an example, open the devtools on this header: https://aglc4.cormacrelf.net/csl#3.1-Generally

jyn514:
[...]

PR #610, opened 2020-02-15
Revert style changes to search bar

jyn514:
Since the changes to search have caused a number of issues (#595, [images no longer showing up](https://discordapp.com/channels/442252698964721669/541978667522195476/678004143301984268)), I think it makes sense to revert and think about a better solution.

Let me know if I missed any commit, I made this by hitting the 'revert' button on https://github.com/rust-lang/docs.rs/pull/587 and then on top of that running `git revert` on the only commit in https://github.com/rust-lang/docs.rs/pull/577.

r? @GuillaumeGomez 

cc @Zexbe


GuillaumeGomez:
So the problem was the CSS... I see. I'll make a more precise one if that can help but meanwhile we can merge this so the issue is temporarily fixed. Thanks for investigating @jyn514 !

PR #611, opened 2020-02-15
Make search box more noticeable, second attempt

GuillaumeGomez:
Can you check if this works better @jyn514 please? :)


jyn514:
I'm not sure if I'm the best person to check, I thought the last PR looked fine too ... maybe @cormacrelf would be interested in reviewing?

jyn514:
I think this is OK to merge, the search box wasn't the cause of the icon problem and it looks like it wasn't the cause of 595 either.

PR #613, opened 2020-02-15
Fix svg mimes not being saved correctly

Zexbe:
For some reason, svg's started saving as "image/svg" instead of "image/svg+xml".
Essential files will need to be re-uploaded / edited for the change to work.


jyn514:
I tested locally, this fixes the problem for new crates but not existing ones built incorrectly over the weekend. We can fix the existing crates later, I think we'll need to modify the database directly.

PR #619, opened 2020-02-22
Allow users to copy dependency text.

### LINKED ISSUE #609: Add button to copy dependency text
It's a tiny enhancement feature. A small button on the side of a particular crate version that copies the text required for the `cargo.toml` file.

1. It will make it super easy to copy and paste dependencies.
2. It might reduce the chance of a one of memory slip error.

I have added an image of where the copy to clipboard button might be placed.
rust_docs

reese:
Closes #609 

This change adds a clipboard icon next to the crate name that copies the crate information in the format required for the user's TOML file. For example, a user on the page for the `regex` crate at version `1.3.0` will have `regex = "1.3.0"` copied to their clipboard.

Currently, the UX for this is fairly bare-bones. It may be useful to have some kind of notification from the UI that the copy happened, such as using a tooltip or check mark, but I opted to go with the minimal case for this PR, and other changes can always be added later.

DocsRsClipboard

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -439,14 +439,6 @@ fn load_js(file_path_str: &'static str) -> IronResult<Response> {
     Ok(response)
 }
 
-fn menu_js_handler(_: &mut Request) -> IronResult<Response> {
-    load_js(MENU_JS)
-}
-
-fn index_js_handler(_: &mut Request) -> IronResult<Response> {
-    load_js(INDEX_JS)
-}
-
 fn opensearch_xml_handler(_: &mut Request) -> IronResult<Response> {
     let mut response = Response::with((status::Ok, OPENSEARCH_XML));
     let cache = vec![CacheDirective::Public,
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -73,6 +73,7 @@ use std::sync::{Arc, Mutex};
 const STATIC_FILE_CACHE_DURATION: u64 = 60 * 60 * 24 * 30 * 12;   // 12 months
 const STYLE_CSS: &'static str = include_str!(concat!(env!("OUT_DIR"), "/style.css"));
 const MENU_JS: &'static str = include_str!(concat!(env!("OUT_DIR"), "/menu.js"));
+const INDEX_JS: &'static str = include_str!(concat!(env!("OUT_DIR"), "/index.js"));
 const OPENSEARCH_XML: &'static [u8] = include_bytes!("opensearch.xml");
 
 const DEFAULT_BIND: &str = "0.0.0.0:3000";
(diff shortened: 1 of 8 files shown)


src/web/mod.rs
```diff
@@ -429,15 +430,23 @@ fn style_css_handler(_: &mut Request) -> IronResult<Response> {
...
     Ok(response)
 }
 
+fn menu_js_handler(_: &mut Request) -> IronResult<Response> {
+    load_js(MENU_JS)
+}
+
+fn index_js_handler(_: &mut Request) -> IronResult<Response> {
+    load_js(INDEX_JS)
+}
```
jyn514: Maybe it would be easier to pass a closure to `static_resource` than to have this boilerplate?

templates/footer.hbs
```diff
@@ -1,4 +1,5 @@
 {{#if varsb.javascript_highlightjs}}<script type="text/javascript" charset="utf-8">hljs.initHighlighting();</script>{{/if}}
 <script type="text/javascript" src="/menu.js?{{cratesfyi_version_safe}}"></script>
+<script type="text/javascript" src="/index.js"></script>
```
jyn514: This will prevent the page from being cached indefinitely.
```suggestion
<script type="text/javascript" src="/index.js?{{cratesfyi_version_safe}}"></script>
```

templates/index.js
```diff
@@ -0,0 +1,21 @@
+function formatCrateName(crateTitleAndVersion) {
+    const stringParts = crateTitleAndVersion.split(" ", 2);
+    return stringParts[0] + ' = "' + stringParts[1] + '"';
```
jyn514: I think this would be a good use case for destructuring, but it's not supported in IE 11 :(

reese: Yeah, it would have been _really_ nice to use a template literal, but a lot of the JS "nice-to-haves" aren't consistently supported without adding a build step :disappointed:

templates/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+    document.body.append(temporaryInput);
+    temporaryInput.select();
+    document.execCommand("copy");
+
+    temporaryInput.remove();
+}
+
+document.getElementById("clipboard")
+    .addEventListener("click", copyTextHandler);
```
GuillaumeGomez: It doesn't fit in 100 columns?

GuillaumeGomez:
JS and CSS look good to me. :+1:

templates/index.js
reese: Oh that was left over from before I refactored some things; it should be on its own line. Thanks for the catch!

src/web/mod.rs
```diff
@@ -429,8 +430,8 @@ fn style_css_handler(_: &mut Request) -> IronResult<Response> {
     Ok(response)
 }
 
-fn menu_js_handler(_: &mut Request) -> IronResult<Response> {
-    let mut response = Response::with((status::Ok, MENU_JS));
+fn load_js(file_path_str: &'static str) -> IronResult<Response> {
```
jyn514: Is there a reason you made this `&'static` instead of just `&'a str`?

reese: It definitely can be `&'a`, I just made it `&'static` because it's being passed `const` strings. It was more a semantic choice to show that we're passing known strings as opposed to, say dynamically loading JS, but I'm also happy to make it more permissive.

jyn514: Hmm actually I like this to show that it's a constant string :)

PR #621, opened 2020-02-23
Add Windows and Mac to CI tests

XAMPPRocky:
r? @jyn514

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,10 +22,22 @@ jobs:
       - name: Install stable Rust
         run: rustup update stable && rustup default stable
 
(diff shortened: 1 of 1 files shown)


jyn514:
If you're still interested in working on this, I think you could make it work without `createuser` and `createdb` as long as you have super user privileges:

```sh
sudo -u postgres psql -c "CREATE USER cratesfyi WITH PASSWORD 'password';"
sudo -u postgres psql -c "CREATE DATABASE cratesfyi OWNER cratesfyi;"
```

Not sure how to get those privileges without sudo, I know it doesn't exist on Windows and not sure about MacOS.

PR #624, opened 2020-02-25
Hide clipboard if request is not a 2xx.

### LINKED ISSUE #622: Clipboard still shows up on 404 pages
e.g. https://docs.rs/pyo3/0.9.0-alpha.1/src/pyo3

cc @reese, are you interested in fixing this?

reese:
Resolves #622 

This change conditionally shows the clipboard only if the request was a 2xx, preventing it from showing in 404 pages.

### DIFF
--- a/src/web/page.rs
+++ b/src/web/page.rs
@@ -137,6 +137,7 @@ impl<T: ToJson> ToJson for Page<T> {
         }
 
         tree.insert("content".to_owned(), self.content.to_json());
+        tree.insert("is_successful_request".to_owned(), self.status.is_success().to_json());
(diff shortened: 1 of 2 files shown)


jyn514:
I'm not sure this is the right approach, it won't prevent it from being shown e.g. on https://docs.rs/about. Maybe instead show it only on the /crate page?

reese:
@jyn514 Oh sorry about that, I didn't know it was also being shown on other pages like that. Guess I didn't know where all that `navigation` template was used, gonna close this PR and look at this again, I think I see a better way around it.

jyn514:
I didn't notice it was on /about until this afternoon either, yeah

PR #625, opened 2020-02-25
Only show clipboard on crate/* endpoints.

### LINKED ISSUE #622: Clipboard still shows up on 404 pages

reese:
Resolves #622 

This PR sets a boolean flag on the `source`, `build`,  and `crate_details` handlers to show the clipboard. Those three handlers presently make up all of the routes under the `/crate` page prefix. This prevents the clipboard from showing up under any other resource, including the `about` page, `releases` pages, and so on.

(Note for some future refactoring: there are several of these `Page` objects that get created with the same flags. It would probably be wise to have these strings be `const` strings, and the fact that each of those same flags are used repeatedly probably implies different types of `Page`s or something like that. I didn't do any of this because it's not really related to this bug, but it's a thought nonetheless.)

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -520,6 +520,7 @@ impl ToJson for MetaData {
 #[cfg(test)]
 mod test {
     use crate::test::*;
+    use html5ever::tendril::TendrilSink;
     use web::{match_version, MatchVersion};
 
     const DEFAULT_ID: i32 = 0;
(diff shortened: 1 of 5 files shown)


jyn514:
We already have a way to tell apart the pages with a crate from those with just a title: `title`.

How about this instead? This means we don't have to add a new variable to the page.

```
diff --git a/templates/navigation.hbs b/templates/navigation.hbs
index f25d992..eefaa13 100644
--- a/templates/navigation.hbs
+++ b/templates/navigation.hbs
@@ -34,8 +34,14 @@
     {{#unless varsb.hide_package_navigation}}
     <div class="cratesfyi-package-container">
       <div class="container">
-        <h1 id="crate-title">{{#if title}}{{title}}{{else}}{{content.metadata.name}} {{content.metadata.version}}{{/if}}</h1>
-        {{#if varsb.show_clipboard}}<i class="fa fa-clipboard fa-1" id="clipboard" aria-label="Copy crate name and version information"></i>{{/if}}
+        <h1 id="crate-title">
+          {{#if title}}
+            {{title}}
+          {{else}}
+            {{content.metadata.name}} {{content.metadata.version}}
+            <i class="fa fa-clipboard fa-1" id="clipboard" aria-label="Copy crate name and version information"></i>
+          {{/if}}
+        </h1>
         <div class="description">{{#if content.metadata.description }}{{content.metadata.description}}{{else}}{{varss.description}}{{/if}}</div>
 
         {{#if ../varsb.show_package_navigation}}
```

reese:
@jyn514 Yep that's definitely nicer :sweat_smile: I had just assumed that content name was a fallback, but I definitely appreciate the suggestion! Just updated the PR :+1:

jyn514:
Do you mind adding some tests for this? The easiest way is probably to make a helper function to see whether the HTML for a clipboard exists, see https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L450 for an example.

src/web/mod.rs
```diff
@@ -548,6 +555,50 @@ mod test {
...
+                .create()
+                .unwrap();
+            let web = env.frontend();
+            assert!(!clipboard_is_present_for_path("/about", web));
+            assert!(!clipboard_is_present_for_path("/releases", web));
+            assert!(!clipboard_is_present_for_path("/", web));
+            assert!(!clipboard_is_present_for_path(
+                "/fake_crate/0.0.1/fake_crate",
+                web
+            ));
```
jyn514: Please add a test for 404 pages as well (maybe `/nonexistent` or the like).

jyn514:
Thanks for working on this!

Issue #616, opened 2020-02-16
Sandbox limits increase request for stm32ral


adamgreig:
**Crate name:** stm32ral
**Build failure link:** https://docs.rs/crate/stm32ral/0.4.1/builds

**Requested RAM limit:** 6GB
**Requested timeout:** 15min is OK

**Why your crate needs the resource increases:**

Related to #608, stm32ral is an alternative register access crate for all STM32 microcontrollers in one package. As with those crates, it documents thousands of hardware registers and so is quite resource-intensive to build. I measured the doc build for the latest version as taking 1m39s and 5432MB RAM, so a 6GB RAM limit should work OK. 8GB might be more conservative if the crate continues to grow.

On balance the #608 crates are more important as they are much more widely used, but if it would be possible to get the RAM limit increased for stm32ral too, I'd really appreciate it.

Thanks!

adamgreig:
Any chance of getting this limit raised alongside #608 now that #632 is in?

Thanks!

jyn514:
Just updated the memory limit for stm32ral to 8 GB. Could you also set `targets = []` for this crate to avoid unnecessary load on the server? Thanks again for your patience.

adamgreig:
Great, thank you! I have done so in https://github.com/adamgreig/stm32ral/commit/cd0bfa3386282d6699025daf2a746eba6a8c94ab so will be in the next release.

Issue #608, opened 2020-02-13
Sandbox limits increase request for stm32 family crates


adamgreig:
**Crate name:** stm32f1, stm32f4, stm32h7, stm32g4
**Build failure link:** [stm32f1](https://docs.rs/crate/stm32f1/0.10.0/builds) [stm32f4](https://docs.rs/crate/stm32f4/0.10.0/builds) [stm32h7](https://docs.rs/crate/stm32h7/0.10.0/builds) [stm32g4](https://docs.rs/crate/stm32g4/0.10.0)

**Requested RAM limit:** stm32f1: 4GB, stm32f4: 5GB, stm32h7: 8GB, stm32g4: 5GB
**Requested timeout:** existing 15min limit is OK

**Why your crate needs the resource increases:**

These crates all belong to the [stm32-rs](https://github.com/stm32-rs/stm32-rs) group of crates, which are "peripheral access crates" for embedded Rust projects using STM32 microcontrollers. The crates contain functions for accessing every hardware register on the relevant microcontroller, for all microcontrollers in a family (e.g. the STM32F4 family). Some families have many members or especially many hardware registers, so take a lot of memory to build.

I built all the crate docs in a local sandbox and measured memory usage, which is what the requested RAM limits above are based on:

```
stm32f0: 1m27s 2361MB (OK on docs.rs)
stm32f1: 2m33s 3356MB (Failed on docs.rs)
stm32f2: 1m32s 2345MB (OK on docs.rs)
stm32f3: 1m12s 1807MB (OK on docs.rs)
stm32f4: 3m11s 4136MB (Failed on docs.rs)
stm32f7: 2m26s 3024MB (OK on docs.rs)
stm32h7: 8m46s 7121MB (Failed on docs.rs)
stm32l0: 1m36s 2178MB (OK on docs.rs)
stm32l1: 1m15s 1878MB (OK on docs.rs)
stm32l4: 1m30s 2287MB (OK on docs.rs)
stm32g0: 1m45s 2500MB (OK on docs.rs)
stm32g4: 3m32s 4542MB (Failed on docs.rs)
```

At present we already use the docs.rs metadata field in Cargo.toml to select only a few device features for each crate. Ideally we would select every device so they are all documented, but this compromise helps reduce build time/disk space/memory required while still documenting the most popular devices. If it's possible to have slightly higher memory limits on more of the crates (e.g., 8GB for all the crates in the family), we could add more devices to be documented.

[...]

Issue #563, opened 2020-01-11
Add  support for more targets


rubberduck203:
**Crate name:** stm32f3-discovery
**Build failure link:** https://docs.rs/crate/stm32f3-discovery/0.1.2/builds/212158

**Additional details:**

Crates targeting ARM Cortex-M processors fail to build because docs.rs currently doesn't support the `thumbv*` targets.

This can be worked around by setting the default target to `x86_64-unknown-linux-gnu` per https://github.com/stm32-rs/stm32f3xx-hal/issues/46.

```toml
[package.metadata.docs.rs]
default-target = "x86_64-unknown-linux-gnu"
```

Possibly related to https://github.com/rust-lang/docs.rs/issues/343#issuecomment-525888748

Disasm:
Earlier docs were built with the explicit `--target x86_64-unknown-linux-gnu` parameter and now this parameter is not passed. For some projects like embedded ones, cargo takes target from `.cargo/config` and it happens to be different from `x86_64-unknown-linux-gnu` and not installed.

jyn514:
@Disasm We [stopped passing `--target`](https://github.com/rust-lang/docs.rs/pull/491) because [it broke proc-macros](https://github.com/rust-lang/docs.rs/issues/422), it's unrelated to #343.

Thanks for pointing out .cargo/config, I wouldn't have noticed that. Would ignoring configuration in .cargo/config solve your issue? Or does your crate sometimes depend on that config to build?

jyn514:
Also I suppose this is as good a place as any to talk about additional targets. I would like to support more targets, but it requires some significant refactors on our ends. Adapted from a chat I had on Discord:

Targets are currently a little complicated. We have 5 targets that we build which are only documented [here](https://github.com/rust-lang/docs.rs/issues/399#issuecomment-568221002). These are mostly kept in sync with [Tier 1 platforms](https://forge.rust-lang.org/release/platform-support.html).

Additionally, we unconditionally build each crate for each target. This is about to change in #532. Until we stop building each crate for each target, storage and build costs don't make it practical to add more targets.

Once #532 is merged, I think adding more targets might be interesting under the following conditions:
1. They would only be built if explicitly requested (part of #532)
2. They would have to be supported by rustup without any special effort on our part

Note that build-scripts and proc-macros [will still be broken when passed `--target`](https://github.com/rust-lang/cargo/issues/7677), although I don't expect that to be a big deal once cross-compiling is opt-in instead of opt-out.

[...]

PR #632, opened 2020-03-07
Allow crates to opt-in to building a single target

### LINKED ISSUE #627: Add opt-in option to build for one target only
Almost all of the hard work has been done in #534 and #532. This would allow merging a non-breaking version of #532, enabling cool things like #563 and possibly other issues tagged with `uses-more-storage` (#616, #608 come to mind).

The only thing that needs to change is updating the documentation and still using `TARGETS` instead of only `default-target` if `extra-targets` is unset.

jyn514:
Based off of #532, but backwards compatible.

Closes #627.

Motivation: If crates are only built on one target, I'd be much more comfortable raising limits on a case-by-case basis (e.g. #608, #616). I would also be more comfortable adding more targets (i.e. not tier 1: #563), since specifying those targets would require setting `extra-targets` and thereby opting out of all others.

~~Waiting on me to test that this actually works since we don't have unit tests for builds.~~ Done.

- [x] Can build from a clean cache (no `.rustwide-docker` directory)
- [x] Can add essential files (`build add-essential-files`)
- [x] Does not pass `--target` for proc-macros (and in general, for the default build unless `default-target` is explicitly set) (`build crate rstest 0.4.0`, testing for https://github.com/rust-lang/docs.rs/issues/422)
- [x] Builds for cross-compiled targets work
- [x] Builds for host target work
- [x] `targets` behaves as documented (I only tested `targets = ["x86_64-pc-windows-msvc"]` and unset `default-target`; others are tested with unit tests)
- [x] Passing a duplicate `default-target` still only builds the default target once

### DIFF
--- a/src/docbuilder/metadata.rs
+++ b/src/docbuilder/metadata.rs
@@ -125,7 +125,7 @@ impl Metadata {
                         .and_then(|v| v.as_bool()).unwrap_or(metadata.all_features);
                     metadata.default_target = table.get("default-target")
                         .and_then(|v| v.as_str()).map(|v| v.to_owned());
-                    metadata.targets = table.get("extra-targets").and_then(|f| f.as_array())
+                    metadata.targets = table.get("targets").and_then(|f| f.as_array())
                         .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
                     metadata.rustc_args = table.get("rustc-args").and_then(|f| f.as_array())
                         .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
@@ -191,7 +191,7 @@ mod test {
             all-features = true
             no-default-features = true
             default-target = "x86_64-unknown-linux-gnu"
-            extra-targets = [ "x86_64-apple-darwin", "x86_64-pc-windows-msvc" ]
+            targets = [ "x86_64-apple-darwin", "x86_64-pc-windows-msvc" ]
             rustc-args = [ "--example-rustc-arg" ]
             rustdoc-args = [ "--example-rustdoc-arg" ]
         "#;
@@ -248,7 +248,7 @@ mod test {
         // extra targets explicitly set to empty array
         let metadata = Metadata::from_str(r#"
             [package.metadata.docs.rs]
-            extra-targets = []
+            targets = []
         "#);
         assert!(metadata.targets.unwrap().is_empty());
     }
@@ -257,7 +257,7 @@ mod test {
         use crate::docbuilder::rustwide_builder::{HOST_TARGET, TARGETS};
 
         let mut metadata = Metadata::default();
-        // unchanged default_target, extra targets not specified
+        // unchanged default_target, targets not specified
         let (default, tier_one) = metadata.targets();
         assert_eq!(default, HOST_TARGET);
         // should be equal to TARGETS \ {HOST_TARGET}
--- a/src/docbuilder/metadata.rs
+++ b/src/docbuilder/metadata.rs
@@ -1,4 +1,3 @@
-use std::collections::HashSet;
 use std::path::Path;
 use toml::Value;
 use error::Result;
@@ -130,27 +129,34 @@ impl Metadata {
 
         metadata
     }
-    pub(super) fn select_extra_targets<'a>(&'a self, default_target: &str) -> HashSet<&'a str> {
-        use super::rustwide_builder::TARGETS;
-        // this is a breaking change, don't enable it by default
-        let build_specific = std::env::var("DOCS_RS_BUILD_ONLY_SPECIFIED_TARGETS")
-            .map(|s| s == "true").unwrap_or(false);
-        // If the env variable is set, _only_ build the specified targets
-        // If no targets are specified, only build the default target.
-        let mut extra_targets: HashSet<_> = if build_specific {
-            self.extra_targets
-                    .as_ref()
-                    .map(|v| v.iter().map(|s| s.as_str()).collect())
-                    .unwrap_or_default()
-        // Otherwise, let people opt-in to only having specific targets
-        } else if let Some(explicit_targets) = &self.extra_targets {
-            explicit_targets.iter().map(|s| s.as_str()).collect()
-        // Otherwise, keep the existing behavior
-        } else {
-            TARGETS.iter().copied().collect()
+    pub(super) fn select_extra_targets<'a: 'ret, 'b: 'ret, 'ret>(&'a self) -> (&'ret str, Vec<&'ret str>) {
+        use super::rustwide_builder::{DEFAULT_TARGET, TARGETS};
+        // Let people opt-in to only having specific targets
+        // Ideally this would use Iterator instead of Vec so I could collect to a `HashSet`,
+        // but I had trouble with `chain()` ¯\_(ツ)_/¯
+        let mut all_targets: Vec<_> = self.default_target.as_deref().into_iter().collect();
+        match &self.extra_targets {
+            Some(targets) => all_targets.extend(targets.iter().map(|s| s.as_str())),
+            None => all_targets.extend(TARGETS.iter().copied()),
         };
-        extra_targets.remove(default_target);
-        extra_targets
+
+        // default_target unset and extra_targets set to `[]`
+        let landing_page = if all_targets.is_empty() {
+            DEFAULT_TARGET
+        } else {
+            // This `swap_remove` has to come before the `sort()` to keep the ordering
+            // `swap_remove` is ok because ordering doesn't matter except for first element
+            all_targets.swap_remove(0)
+        };
+        // Remove duplicates
+        all_targets.sort();
+        all_targets.dedup();
+        // Remove landing_page so we don't build it twice.
+        // It wasn't removed during dedup because we called `swap_remove()` first.
+        if let Ok(index) = all_targets.binary_search(&landing_page) {
+            all_targets.swap_remove(index);
+        }
+        (landing_page, all_targets)
     }
 }
 
@@ -279,6 +285,5 @@ mod test {
 
         // try it with a different default target just for sanity
         assert!(metadata.select_extra_targets("i686-pc-windows-msvc").is_empty());
-
     }
 }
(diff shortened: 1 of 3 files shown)


rubberduck203:
That’s promising news. Thanks for the effort you’ve put into this!

jyn514:
Tested:

- [x] Building crates without `extra-targets` keeps the current behavior (all tier 1 targets)
- [x] Building crates with `extra-targets = []` only builds on the default target
- [x] Building crates with `extra-targets = [target]` builds on that target as well as the default target
- [x] Building crates with `extra-targets = [default]`, i.e. adding the default target to extra targets, is a no-op (same as `extra-targets = []`)

src/docbuilder/metadata.rs
```diff
@@ -43,6 +44,13 @@ pub struct Metadata {
     /// is always built on this target. You can change default target by setting this.
     pub default_target: Option<String>,
 
+    /// If you want a crate to build only for specific targets,
+    /// set `extra-targets` to the list of targets to build, in addition to `default-target`.
+    ///
+    /// If you do not set `extra_targets`, all of the tier 1 supported targets will be built.
+    /// If you set `extra_targets` to an empty array, only the default target will be built.
+    pub extra_targets: Option<Vec<String>>,
```
pietroalbini: I'd rename this just to `targets`, as it doesn't add *extra* targets, it sets the full list of targets the author wants to build.

src/docbuilder/rustwide_builder.rs
```diff
@@ -340,18 +342,41 @@ impl RustwideBuilder {
...
+                                       .unwrap_or_default();
+                        &strs
+                    // Otherwise, let people opt-in to only having specific targets
+                    } else if let Some(explicit_targets) = &metadata.extra_targets {
+                        strs = explicit_targets.iter().map(|s| s.as_str()).collect();
+                        &strs
+                    // Otherwise, keep the existing behavior
+                    } else {
+                        TARGETS
+                    };
```
pietroalbini: I would extract all of this into its own method (maybe of the `Metadata` struct?), so we can add tests for it without executing builds.

src/docbuilder/rustwide_builder.rs
```diff
@@ -340,18 +342,41 @@ impl RustwideBuilder {
...
+
                     // Then build the documentation for all the targets
-                    for target in TARGETS {
-                        if *target == res.target {
+                    for &target in targets {
+                        if target == res.target {
```
pietroalbini: I don't think this is handling someone having the following metadata:

```toml
default-target = "x86_64-pc-windows-msvc"
targets = ["x86_64-unknown-linux-gnu", "x86_64-unknown-linux-gnu"]
```

src/docbuilder/metadata.rs
```diff
@@ -111,6 +120,8 @@ impl Metadata {
                         .and_then(|v| v.as_bool()).unwrap_or(metadata.all_features);
                     metadata.default_target = table.get("default-target")
                         .and_then(|v| v.as_str()).map(|v| v.to_owned());
+                    metadata.extra_targets = table.get("extra-targets").and_then(|f| f.as_array())
+                        .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
```
pietroalbini: You don't have to do it in this PR, but it would be nice to just use `#[derive(Deserialize)]`.

pietroalbini:
Oh, also, at least for now we need to filter the targets to make sure someone doesn't use an unsupported target, and fail the build in that case.

src/docbuilder/rustwide_builder.rs
pietroalbini: If we end up moving the target selection code to a method in `Metadata`, this is a check we could move in that method as well (i.e. by filtering the default target out).

src/docbuilder/metadata.rs
jyn514: I don't think this is the case? This specifies targets to build _in addition to_ the default target. The default target will always be built even if `extra-targets` is empty.

src/docbuilder/rustwide_builder.rs
jyn514: Good call, we should remove duplicates (maybe just by using a TreeSet).

jyn514:
> Oh, also, at least for now we need to filter the targets to make sure someone doesn't use an unsupported target, and fail the build in that case.

This is our current behavior for an unsupported `default-target`:

```
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] error[E0463]: can't find crate for `std`
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr]   |
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr]   = note: the `x86_64-pc-windows-gnu` target may not be installed
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] 
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] error: aborting due to previous error
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] 
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] For more information about this error, try `rustc --explain E0463`.
2020/03/07 12:51:25 [INFO] rustwide::cmd: [stderr] error: Could not document `hexponent`.
```

Our behavior for an unsupported `extra-target` is the same, I think it doesn't need to be handled explicitly.

jyn514:
Addressed review comments.

src/docbuilder/metadata.rs
pietroalbini: It's more that it gives me the impression that those targets are *in addition* to the default list of tier 1 targets.

pietroalbini:
> Our behavior for an unsupported extra-target is the same, I think it doesn't need to be handled explicitly.

Keep in mind that we only show the build log of the default target, so users wouldn't be able to see those errors.

src/docbuilder/metadata.rs
jyn514: Ah, that makes sense ... Do you think that `targets` makes it clear that the default target will still be built no matter what? I'm worried that people will try to set the default target by setting `extra-targets = [default]`

jyn514:
> Keep in mind that we only show the build log of the default target, so users wouldn't be able to see those errors.

Hmm, that does make it harder ... I don't want to fail the default build if one of the targets isn't installed, since that would require publishing a full point release. Maybe we could show a popup on the `/crate` page that says the target doesn't exist?

jyn514:
Also I plan to make a follow-up PR _very_ shortly which will install targets dynamically so I'm not sure it's worth worrying about.

src/docbuilder/metadata.rs
pietroalbini: Honestly I don't see that as a problem. If no explicit `default-target` is set and `targets` is set we should pick the first item in `targets` as the default.

jyn514: Ok, I can implement that. I think it's better to do the expected thing than to document how to do the unexpected thing :)

Here's what I propose to do:

- If `default-target` is set, use that as the default. If it appears again in `targets`, it is ignored; all other targets are built (removing duplicates).
- If `default-target` is unset and `targets` has at least one target, use the first target as the default and build all others (removing duplicates).
- If `default-target` is unset and `targets` is explicitly set to empty (`[]`), use `DEFAULT_TARGET` (x86-unknown-linux-gnu)
- If `default-target` is unset and `targets` is not set, use `DEFAULT_TARGET` and also build all the tier 1 targets.

Does that sounds reasonable?

jyn514: I don't know ... looking back at this it might be more intuitive, but it's hard to explain. I'd rather go for the simpler version..

jyn514: Writing this a little more concisely for my sanity:

```
targets = [default_target] + targets.unwrap_or(TARGETS);
deduplicate(targets);
if targets.is_empty() { // `default_target` unset, `targets` set
    targets = [DEFAULT_TARGET];
}
landing_page = targets[0];
for target in targets {
    build(target);
}
```

jyn514: This is my code so far, I'm out of energy to work on this. Feel free to take over the PR if you have time :)

```
    pub(super) fn select_extra_targets<'a: 'ret, 'b: 'ret, 'ret>(&'a self, default_target: Option<&'b str>) -> (&'ret str, HashSet<&'ret str>) {
        use super::rustwide_builder::{DEFAULT_TARGET, TARGETS};
        // Let people opt-in to only having specific targets
        let mut all_targets = default_target
            .iter()s
            .chain(self.extra_targets.map_or(TARGETS.iter(), |v| v.iter().map(|s| s.as_str())))
            .copied();
        // default_target unset and extra_targets set to `[]`
        let landing_page = all_targets.next().unwrap_or(DEFAULT_TARGET);
        let mut extras: HashSet<_> = all_targets.collect();
        extras.remove(landing_page);
        (landing_page, extras)
    }
```

jyn514: Implemented this in the most recent commit. I still think it's more complicated than it needs to be, but I'm ok keeping the new behavior if you think it's useful.

Note that this required a significant refactor of the build internals. This hasn't been tested yet, and the PR shouldn't be merged until it has been.

jyn514:
This is ready for review.

src/docbuilder/metadata.rs
```diff
@@ -119,6 +135,41 @@ impl Metadata {
 
         metadata
     }
+    // Return (default_target, all other targets that should be built with duplicates removed)
+    pub(super) fn targets(&self) -> (&str, Vec<&str>) {
```
pietroalbini: I think the code in the function can be greatly simplified:

```rust
let default_target = self.default_target
    .or_else(|| self.targets.and_then(|targets| targets.iter().next()))
    .unwrap_or(HOST_TARGET);

let mut targets: HashSet<_> = self.targets
    .map(|targets| targets.iter().collect())
    .unwrap_or_else(|| TARGETS.iter().collect());
targets.remove(default_target);
```

pietroalbini: It'd be better to have a struct as the return type, to make it clear that one is the default target and the other is the rest of the targets.

pietroalbini:
Other than these two comments this looks good to be merged!

jyn514:
Addressed review comments.

PR #633, opened 2020-03-11
Allow non-tier-one targets to be built

### LINKED ISSUE #563: Add  support for more targets

jyn514:
Closes #563.
~~Waiting on #632.~~ Merged.

This uses rustup to install toolchains if they are not already installed. These toolchains will be removed on the next toolchain update, and will be downloaded again on request.

~~Only the last few commits are related to #563, all others are from #632.~~

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -296,6 +296,17 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
             "ALTER TABLE releases ALTER COLUMN default_target DROP NOT NULL;
              ALTER TABLE releases ALTER COLUMN default_target DROP DEFAULT",
         ),
+        migration!(
+            context,
+            // version
+            9,
+            // description
+            "Allow max number of targets to be overriden",
+            // upgrade query
+            "ALTER TABLE sandbox_overrides ADD COLUMN targets INT;",
+            // downgrade query
+            "ALTER TABLE sandbox_overrides DROP COLUMN targets;"
+        ),
     ];
 
     for migration in migrations {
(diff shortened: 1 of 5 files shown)


pietroalbini:
I didn't look at this PR too closely yet, but we'll also want to implement limits to avoid people building tons of targets. The limit should be overridable through the usual sandbox limits mechanism.

jyn514:
Ah, good idea. Does 10 sounds like a reasonable limit? We currently build 6 so it's the same order of magnitude.

pietroalbini:
Yeah that sounds fine. We'll also need to update the issue template for sandbox limit increases.

jyn514:
Added a default limit of 10 targets and updated /about and the issue template.

pietroalbini:
You didn't update the schema.

jyn514:
I need to test locally that the schema change is right, one second.

src/db/migrate.rs
```diff
@@ -296,6 +296,17 @@ fn migrate_inner(version: Option<Version>, conn: &Connection, apply_mode: ApplyM
...
+        migration!(
+            context,
+            // version
+            9,
+            // description
+            "Allow max number of targets to be overriden",
+            // upgrade query
+            "ALTER TABLE sandbox_overrides ADD COLUMN targets INT;",
+            // downgrade query
+            "ALTER TABLE sandbox_overrides DROP COLUMN targets;"
```
pietroalbini:
```suggestion
            "ALTER TABLE sandbox_overrides ADD COLUMN max_targets INT;",
            // downgrade query
            "ALTER TABLE sandbox_overrides DROP COLUMN max_targets;"
```

PR #635, opened 2020-03-13
Change layout to fix missing scrollbar highlighting (chromium text search)

### LINKED ISSUE #623: Search result positions not shown in chrome
The page content is contained in an element with the fixed css attribute. This causes the following result position indicators to not show up when searching:

cynecx:
Fixes #623.

This is more of a hack but I am not sure how a proper fix would look like.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -84,15 +84,6 @@ div.rustdoc {
             }
         }
     }
-
-    // this is actual fix for docs.rs navigation and rustdoc sidebar
-    position: fixed;
-    top: $top-navbar-height;
-    bottom: 0;
-    left: 0;
-    right: 0;
-    display: block;
-    overflow-y: auto;
 }
 
 body {
@@ -810,4 +801,9 @@ div.search-page-search-form {
 i.dependencies.normal {
   visibility: hidden;
   display: none;
+}
+
+// Fix position of theme-picker
+.theme-picker {
+	margin-top: $top-navbar-height;
 }


jyn514:
Will this bring back the issue mentioned in https://github.com/rust-lang/docs.rs/pull/510#issuecomment-562851135 ?

cynecx:
@jyn514 I can't really reproduce the anchor issue?

jyn514:
It was there a couple months ago, if it's not present in your PR that's a good sign.

r? @GuillaumeGomez

GuillaumeGomez:
I'm not sure to understand the connection between the remove code and the theme picker menu...

cynecx:
@GuillaumeGomez Without it, it seems that the theme-picker is misaligned, why? I don’t really know, hence I have mentioned that this is more like a hack and not a proper fix.

I can’t really take screenshots now because I am not anywhere near a computer, but you can try out these changes directly in your browser and you will probably see what the issue is.

GuillaumeGomez:
It's currently working for on my computer without those changes. So I'd definitely like to have screenshots. :)

cynecx:
Patch without fix for theme-picker:

Screenshot from 2020-03-14 01-02-57

With theme-picker:

Screenshot from 2020-03-14 01-04-50

GuillaumeGomez:
Ok, I confirm that your change doesn't break things on firefox and chrome. Thanks for the fix!

cynecx:
Well, I think I broke it. I assume that this went already to production because it seems that I can reproduce the mentioned anchor issue. :thinking:

GuillaumeGomez:
Did you try to empty the cache? Check if the CSS is the one from this PR to be sure it's yours too.

cynecx:
@GuillaumeGomez Yes, in fact I have disabled it xD.

GuillaumeGomez:
The new style is definitely there (tested on https://docs.rs/regex/1.3.5/regex/) and is working for me. You're sure your chromium is "fin"? :p

cynecx:
Peek 2020-03-14 18-15

Firefox has the same behavior as Chromium. :/

GuillaumeGomez:
Oh, that's what you meant! Yes, you definitely broke it. I completely forgot about the anchors...

cynecx:
How should we proceed? I mean the trade off here doesn't seem that bad compared to losing the search highlighting functionality. 

Meanwhile: I am currently trying to find a workaround the fixed position of the sticky header.

GuillaumeGomez:
No, the new behavior is worse. We need to find a way to fix the theme picker without breaking the anchors.

cynecx:
Well, I guess it's a revert then. PR is coming right up.

PR #636, opened 2020-03-14
Revert "Change layout to fix missing scrollbar highlighting (chromium text search)"

cynecx:
Reverts rust-lang/docs.rs#635

See the comments in #635.


GuillaumeGomez:
Merging then!

Issue #464, opened 2019-11-02
Feature request: rebuild documentation for old crates


tspiteri:
When crates are not updated in a long time, their style starts to drift from the latest style (or the latest style drifts for them) and consequently they do not gain the latest rustdoc features. I think it would be nice if crates which were last updated one or two years ago are rerun and the docs updated with the latest rustdoc.

This could be low priority so that it does not slow down documentation of new crates.

If a compilation fails for whatever reason, of course the old version should be retained.

jyn514:
Hmm I don't mind this idea but it seems expensive to rerun the build on all versions that have ever been published ... This would just affect style sheets and such, right? What should we do if building with a newer version of `rustdoc` leads to broken links?

jyn514:
Related: https://github.com/rust-lang/docs.rs/issues/301

jyn514:
https://github.com/rust-lang/docs.rs/issues/295 also mentioned allowing crate owners to re-queue builds for their own crates, e.g. if the build succeeds but the JavaScript throws a runtime error because of a bad nightly.

jyn514:
The reason I originally thought this wasn't feasible is because rustdoc puts out releases very frequently (either daily or every 6 weeks, depending how you count it). But nothing says we have to rebuild for every release - if we rebuilt, say every edition, that would still fix most of the rustdoc bugs while being feasible to finish all crates.

Another thing we could do is only rebuild the crates that are used the most: say the latest version and most recent release of every crate. That would bring the number of builds down from ~220k to between 40-80k builds.

This shouldn't be implemented until after #1004 to avoid very large re-upload costs (@pietroalbini estimates the current cost of reuploading every crate at $2k even if there are no changes in the documentation).

Kixiron:
I think with good culling parameters we could significantly reduce the number of crates needed to be built & uploaded
- Only the most recent release, pretty much required for feasibility
- Download/Visit thresholds. If we set cutoffs on crates that haven't been touched in years, no one really loses anything
- Don't rebuild the past few months of releases. The most recent builds will look pretty much the same, bumping a few nightlies shouldn't affect anything. It'll also let us avoid the "edition rush" of crates that are preparing for the new edition

I really want to propose some sort of documentation hashing/checksuming to attempt to find out if the docs currently uploaded differ from the ones we build and then don't delete & reupload if there's no change, but I doubt that's possible due to the unstable nature of rustdoc's output

jyn514:
[...]

PR #640, opened 2020-03-14
Add anchor-workaround for 'small-section-header' and method heading

cynecx:
See https://github.com/rust-lang/docs.rs/pull/637#issuecomment-599148480.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -809,9 +809,16 @@ i.dependencies.normal {
 }
 
 // Fix for anchors, so that they are well positioned right below the sticky header
-.section-header::before {
+// FIXME: This is a hack and should be replaced with a proper solution.
+.small-section-header::before, .section-header::before, h3[class="method"]::before {
     display: block;
     content: "";
     height: 3rem;
     margin-top: -3rem;
+}
+
+// Due to the anchor fix and sub-optimal rustdoc css, the method collapser is mispositioned.
+// FIXME: Workaround by applying a top margin.
+h3[class="method"] > a[class="collapse-toggle"] {
+    margin-top: 3rem;
 }


GuillaumeGomez:
Just tried this: https://docs.rs/regex/1.3.5/regex/struct.Matches.html#impl-Borrow%3CT%3E

Doesn't look very nice, but it's not from this change. Otherwise, this PR makes things a bit better. So I'm find with merging it. Your call @jyn514. :)

cynecx:
@GuillaumeGomez Ehrm, it kinda looks "normal" to me. Do you mind taking a screenshot for me?

GuillaumeGomez:
With this PR, it gives (on firefox):

Screenshot from 2020-03-16 18-44-34

jyn514:
That looks like it still has the anchor issue to me. The prod version running right now has the same issue.

cynecx:
I've just checked that out and it seems like this trick won't work with impl headings because the styling for `<h3>` uses `display: flex`. I honestly don't know what we can do here except use a wrapper element so `display: flex` doesn't affect the `::before`-pseudoelement.

GuillaumeGomez:
I still think that pushing all the content under the docs.rs top banner is the best solution.

jyn514:
> I still think that pushing all the content under the docs.rs top banner is the best solution.

The problem with this is that `overflow: auto` seems to cause lots of other issues. I don't know CSS very well but @cormacrelf said in https://github.com/rust-lang/docs.rs/issues/595#issue-561499035 that `fixed` causes lots of problems on mobile.

GuillaumeGomez:
If this is mobile only, then we can implement mobile-specific fixes.

cynecx:
Honestly, I feel bad :/. Introducing that PR which allowed better chromium search experience doesn't really justify the broken things right now.

> I still think that pushing all the content under the docs.rs top banner is the best solution.

Do you mean not having a sticky header at all?

jyn514:
> Honestly, I feel bad :/

Oh don't worry, the CSS has been broken long before you started making changes 😆. We might end up reverting your change though.

cynecx:
@jyn514 Yes I kinda agree, I thought this could have worked by apply various fixes here and there but this requires a more “thorough” rework/redesign (probably starting with rustdoc) I guess.

If @GuillaumeGomez agrees too, I would close this PR and revert all the changes related to this and perhaps open a new tracking issue.

GuillaumeGomez:
> Do you mean not having a sticky header at all?

Yes and no. Depends how we do it. If we make the whole content start below the header, it won't go under it.

jyn514:
> If @GuillaumeGomez agrees too, I would close this PR and revert all the changes related to this and perhaps open a new tracking issue.

Ok, please do this.

jyn514:
> @jyn514 Yes I kinda agree, I thought this could have worked by applying various fixes here and there but this requires a more “thorough” rework/redesign (probably starting with rustdoc) I guess.

Note that we cannot change the generated HTML on past documentation. Doing so would at a minimum take several months of compute time, and might also cause builds to fail that previously passed. See https://github.com/rust-lang/docs.rs/issues/464.

GuillaumeGomez:
No need to change the HTML, just overload the CSS with our own.

PR #643, opened 2020-03-17
Refactor storage

jyn514:
~~This could definitely use some more work, but I thought I'd get an initial draft up to see if this is the right approach. Thanks to @pietroalbini for doing the initial work :)~~

- Separate storage backends into `S3Backend` and `DatabaseBackend` instead of selecting each backend at runtime for each file.
- Split up `add_path_into_database` into 3 functions:
    1. `add_path_into_database` so I don't have to change a bunch of unrelated code. This just delegates to `store_all` and then turns the returned HashMap into JSON to be stored in the DB.
    2. `store_all`, which takes the folder to upload and what path to upload it to. `store_all` reads all the files in the folder into memory, then divides them into batches of size `MAX_CONCURRENT_UPLOADS`, and forwards those batches on to `store_batch`
    3. `store_batch`, which differs based on whether we're uploading to S3 or storing in the database.
- Added tests for storing and retrieving files. This tests S3 using `min.io` and Postgres using the same temporary tables we've been using
- Remove unused `move_to_s3` function
- Don't create a new `tokio::Runtime` for each batch of files

I tried to reuse code wherever possible so @Mark-Simulacrum 's hard work on simultaneous uploads is still here. ~~However, I don't yet have tests for simultaneous uploads.~~


src/db/file.rs
```diff
@@ -160,237 +31,20 @@ pub fn add_path_into_database<P: AsRef<Path>>(
...
+    let file_list = backend.store_all(conn, prefix, path.as_ref())?;
+    file_list_to_json(file_list.into_iter().collect())
 }
 
-fn file_list_to_json(file_list: Vec<(String, PathBuf)>) -> Result<Json> {
+fn file_list_to_json(file_list: Vec<(PathBuf, String)>) -> Result<Json> {
     let mut file_list_json: Vec<Json> = Vec::new();
 
     for file in file_list {
         let mut v: Vec<String> = Vec::new();
```
Kixiron, line 43: Could pre-allocate this vector
```suggestion
        let mut v: Vec<String> = Vec::with_capacity(2);
```

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+                .and_then(|l| l.try_into().ok())
+                .unwrap_or(0),
+        );
+        b.read_to_end(&mut content).unwrap();
+
+        let date_updated = parse_timespec(&res.last_modified.unwrap())?;
+
+        Ok(Blob {
+            path: path.into(),
+            mime: res.content_type.unwrap(),
```
Kixiron, line 44: Maybe mime normalization to `text/rust` and such should happen here if all downstream consumers use that

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+            date_updated,
+            content,
+        })
+    }
+
+    pub(super) fn store_batch(&self, batch: &[Blob]) -> Result<(), Error> {
+        let mut rt = tokio::runtime::Runtime::new().unwrap();
```
Kixiron, line 51: Is creating a new tokio runtime every time we store files the best way to do things? Tokio already uses a fairly global state, so that could just be reused

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+        let mut attempts = 0;
+
+        loop {
+            let mut futures = Vec::new();
+            for blob in batch {
+                futures.push(
```
Kixiron, line 57: What are the merits of this over using `tokio::spawn` and pushing the handle to the vec instead?

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+            content,
+        })
+    }
+
+    pub(super) fn store_batch(&self, batch: &[Blob]) -> Result<(), Error> {
+        let mut rt = tokio::runtime::Runtime::new().unwrap();
+        let mut attempts = 0;
+
+        loop {
+            let mut futures = Vec::new();
```
Kixiron, line 55: Can pre-allocate this
```suggestion
            let mut futures = Vec::with_capacity(batch.len());
```

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(not(test))]
+const TIME_FMT: &str = "%a, %d %b %Y %H:%M:%S %Z";
+#[cfg(test)]
+pub(crate) const TIME_FMT: &str = "%a, %d %b %Y %H:%M:%S %Z";
```
Kixiron, line 92: Couldn't this just be one thing? The only change between test and non-test is the access level, is that that large of an issue?

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+    let rows = trans.query(
+        &format!(
+            "SELECT path, mime, content FROM files WHERE content != E'in-s3' LIMIT {}",
+            n
+        ),
+        &[],
+    )?;
+    let count = rows.len();
+
+    let mut rt = ::tokio::runtime::Runtime::new().unwrap();
```
Kixiron, line 208: Same as the last tokio one, is there an advantage to making a new tokio runtime?

Kixiron, line 208: In the 2018 edition the `::` syntax isn't necessary 
```suggestion
    let mut rt = tokio::runtime::Runtime::new().unwrap();
```

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+    let mut futures = Vec::new();
```
Kixiron, line 209: Can preallocate this vec
```suggestion
    let mut futures = Vec::with_capacity(rows.len());
```

src/storage/s3.rs
```diff
@@ -0,0 +1,244 @@
...
+    for row in &rows {
+        let path: String = row.get(0);
+        let mime: String = row.get(1);
+        let content: Vec<u8> = row.get(2);
+        let path_1 = path.clone();
+        futures.push(
```
Kixiron, line 215: Same as the last tokio thing, would `tokio::spawn` be better here?

src/storage/s3.rs
jyn514: I guess it's fine to always have it `pub(crate)`, yeah.

src/storage/s3.rs
jyn514: I like having it closer where the mime is created so it's clear what's going on. If we normalized it just before uploading there would be some doubt whether code in the middle used the denormalized version.

src/storage/s3.rs
jyn514: I'm not sure. Does `spawn` launch the task when this doesn't? I don't know very much about `async`?

src/storage/s3.rs
jyn514: Without creating the runtime, how would I use `block_on`? Or would I use `await` instead?

src/storage/database.rs
```diff
@@ -0,0 +1,108 @@
...
+                    "INSERT INTO files (path, mime, content) VALUES ($1, $2, $3)",
+                    &[&blob.path, &blob.mime, &blob.content],
+                )?;
+            } else {
+                trans.query(
+                    "UPDATE files SET mime = $2, content = $3, date_updated = NOW() \
+                                WHERE path = $1",
+                    &[&blob.path, &blob.mime, &blob.content],
+                )?;
+            }
```
pietroalbini, line 54: All of this can be simplified with a single [`INSERT INTO ... ON CONFLICT SET ...`](https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT) query.

src/storage/s3.rs
```diff
@@ -0,0 +1,242 @@
...
+            .ok()
+            .map(|e| Region::Custom {
+                name: std::env::var("S3_REGION").unwrap_or_else(|_| "us-west-1".to_owned()),
+                endpoint: e,
+            })
+            .unwrap_or(Region::UsWest1),
+    ))
+}
+
+pub fn move_to_s3(conn: &Connection, n: usize) -> Result<usize, Error> {
```
pietroalbini, line 193: Is it worth keeping this code around?

jyn514: If you don't think it's useful I'll take it out, it will still be in the git history.

src/storage/s3.rs
Kixiron: Well, isn't this the creation point of the `Blob`? I may be misunderstanding the code flow though

jyn514: Hmm, I misread the code too. This is when we fetch the file from S3, where it's already had a content-type set. The content-type should be correct in S3, if not it's a) impossible to fix manually and b) very unclear how things are happenin.

Kixiron: Wouldn't S3 serve it back with the 'official' (and incorrect for what we need) mime types?

jyn514: No, we explicitly set the mimetype when uploading: https://github.com/rust-lang/docs.rs/pull/643/files#diff-c3a2a09475e9683a0e82b9a1d6f5e0b9R63

jyn514: Although in general I'm not super happy about setting an explicit content type, I think browser do a better job of mine detection than we could for source files.

src/storage/s3.rs
jyn514: This ended up being impossible because it required the task to be `&'static`, but it had the lifetime of `self.client`.

src/storage/s3.rs
jyn514: I removeded this whole function in response to https://github.com/rust-lang/docs.rs/pull/643#discussion_r408855551

src/storage/s3.rs
jyn514: I removed this whole function in response to https://github.com/rust-lang/docs.rs/pull/643#discussion_r408855551.

jyn514:
Status update: this has most of the code changes I want but is missing any tests for S3, which kind of defeats the point. It sounds like rusoto_mock is not going to be flexible enough for our needs so the next best thing is minio. We have minio set up with docker-compose already, but it needs to be tied into the test suite somehow, probably by having `test::wrapper` start a new instance of the docker images.

Given the scale of the task, I plan to make a separate PR so it's not lost in this details of the storage refactor. This should also help with testing on other platforms since developers will no longer need a local postgres database.

jyn514:
Status update to the status update: it ended up being really easy to add min.io, so I just did it here.

Note that this will require contributors to have min.io running before running tests. They can do this the same way as on CI: `docker-compose up -d s3`. This is documented in the README.

src/storage/s3.rs
```diff
@@ -0,0 +1,289 @@
...
+            }
+            Ok(())
+        })
+    }
+    // NOTE: trying to upload a file ending with `/` will behave differently in test and prod.
+    // NOTE: On s3, it will succeed and create a file called `/`.
+    // NOTE: On min.io, it will fail with 'Object name contains unsupported characters.'
+}
+
+pub(crate) static S3_BUCKET_NAME: &str = "rust-docs-rs";
```
pietroalbini, line 263: The constant should be at the top of the file.

src/storage/s3.rs
```diff
@@ -0,0 +1,289 @@
...
+        creds,
+        std::env::var("S3_ENDPOINT")
+            .ok()
+            .map(|e| Region::Custom {
+                name: std::env::var("S3_REGION").unwrap_or_else(|_| "us-west-1".to_owned()),
+                endpoint: e,
+            })
+            .unwrap_or(Region::UsWest1),
+    ))
+}
```
pietroalbini, line 289: Let's have the tests at the end.

src/storage/s3.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    use crate::storage::s3::S3Backend;
+    use rusoto_s3::{
+        CreateBucketRequest, DeleteBucketRequest, DeleteObjectRequest, ListObjectsRequest, S3,
+    };
+
+    use std::cell::RefCell;
+    use std::slice;
+
+    pub(crate) struct TestS3(RefCell<S3Backend<'static>>);
```
pietroalbini, line 116: I'd move this to `src/tests/s3.rs`.

jyn514: Moving this to the tests module requires making almost everything in `src/storage/s3.rs` `pub(crate)`. I'm ok doing that if you want me to but I think the current layout gives a better idea of what functions are called where.

src/storage/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+                file_paths_and_mimes.insert(file_path, mime.to_string());
+                Ok(Blob {
+                    path: bucket_path,
+                    mime: mime.to_string(),
+                    content,
+                    // this field is ignored by the backend
+                    date_updated: Timespec::new(0, 0),
+                })
+            })
+            .collect::<Result<Vec<_>, Error>>()?
```
jyn514, line 136: Hmm, I just realized this will load all the files into memory at once. Do we want me to try and make this lazy so it instead only loads a batch at a time?

src/storage/s3.rs
pietroalbini: Can we do a `src/storage/s3_test.rs` instead?

jyn514: I moved it to `src/storage/s3/test_s3.rs` so it had access to all of `s3.rs`, does that work?

jyn514: I ended up making it `test.rs` and renaming the inner module to `tests` to avoid a naming conflict.

Issue #645, opened 2020-03-18
Winapi's README header is rendered weirdly on docs.rs


lzutao:
Compare between: https://docs.rs/crate/winapi/0.3.8 and https://github.com/retep998/winapi-rs
![winapi-header](https://media.discordapp.net/attachments/541978667522195476/689669739898732567/unknown.png)

PR #646, opened 2020-03-18
Don't add newlines after code snippets in headers

jyn514:
Fixes the very odd behavior on https://docs.rs/crate/winapi/0.3.8

Closes https://github.com/rust-lang/docs.rs/issues/645

r? @GuillaumeGomez

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -813,6 +813,6 @@ i.dependencies.normal {
 }
 
 /* Don't put a newline after code fragments in headers */
-h3 > code {
-  display: inline;
+h3 > code, h4 > code {
+  display: inline-block;
 }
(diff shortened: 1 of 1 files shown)


templates/style.scss
```diff
@@ -811,3 +811,8 @@ i.dependencies.normal {
   visibility: hidden;
   display: none;
 }
+
+/* Don't put a newline after code fragments in headers */
+h3 > code {
+  display: inline;
```
GuillaumeGomez: Just wondering: why `inline` and not `inline-block`?

GuillaumeGomez:
Couldn't find issues so once I have an answer to my question, all good for me!

templates/style.scss
jyn514: I had to look up the difference - `inline-block` prevents newlines within the code snippet, while `inline` would allow line wrapping? In that case we should probably do inline-block.

https://stackoverflow.com/a/45061654/7669110

GuillaumeGomez:
Thanks!

PR #649, opened 2020-03-21
Add tests for limits

jyn514:
We had a queue backup this morning because the builder would panic whenever a crate had limits set (https://github.com/rust-lang/docs.rs/pull/648, https://docs.rs/releases/queue). This adds tests so that doesn't happen again.

r? @kvrhdn

### DIFF
--- a/src/docbuilder/limits.rs
+++ b/src/docbuilder/limits.rs
@@ -104,7 +104,7 @@ fn scale(value: usize, interval: usize, labels: &[&str]) -> String {
     // 2.x
     let mut value = format!("{:.1}", value);
     // 2.0 -> 2
-    if &value[value.len() - 1..] == "0" {
+    if value.ends_with(".0") {
         value.truncate(value.len() - 2);
     }
     format!("{} {}", value, chosen_label)
--- a/src/docbuilder/limits.rs
+++ b/src/docbuilder/limits.rs
@@ -3,6 +3,7 @@ use postgres::Connection;
 use std::collections::BTreeMap;
 use std::time::Duration;
 
+#[derive(Debug, PartialEq)]
 pub(crate) struct Limits {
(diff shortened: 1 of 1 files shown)


kvrhdn:
I should be able to look at this in the next couple of days @jyn514.

jyn514:
Good idea! I added a test for `scale` and noticed at the same time that it always rounded down. I changed it so now it rounds to the closest integer instead. I was also considering printing a float rounded to 1 decimal place, what do you think?

kvrhdn:
Yeah, makes sense! Having one digit after the decimal point provides a lot more precision. There is a significant difference between `1 GB` and `1,5 GB`. But at the same time, I find it difficult to reason about `1,2 hour`. So it will be a bit of a trade off 😅

jyn514:
Implemented fractional precision up to 1 decimal point. I also displayed it as an integer if the decimal was 0 (2.0 -> 2).

src/docbuilder/limits.rs
```diff
@@ -88,15 +87,102 @@ impl Limits {
...
             chosen_label = label;
             value = value / interval;
         } else {
             break;
         }
     }
+    // 2.x
+    let mut value = format!("{:.1}", value);
+    // 2.0 -> 2
+    if &value[value.len() - 1..] == "0" {
```
kvrhdn: This can be simplified with [`ends_with`](https://doc.rust-lang.org/std/primitive.str.html#method.ends_with) 🙂 

```suggestion
    if value.ends_with(".0") {
```

kvrhdn:
But overall this looks neat :+1: I love Rust's `format!`!

jyn514:
Rust is indeed wonderful! Thanks for the review, I think the new precision will be helpful :)

PR #653, opened 2020-03-24
Pass -j2 to Cargo to avoid using all threads

jyn514:
This avoids rustc using all available cores and preventing users from visiting the website (or admins from logging in).

r? @pietroalbini @Mark-Simulacrum

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -455,7 +455,7 @@ impl RustwideBuilder {
         if let Some(package_rustdoc_args) = &metadata.rustdoc_args {
             rustdoc_flags.append(&mut package_rustdoc_args.iter().map(|s| s.to_owned()).collect());
         }
-        let mut cargo_args = vec!["doc", "--lib", "--no-deps"];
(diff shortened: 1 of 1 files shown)


pietroalbini:
While this could be a stopgap to avoid the outage we recently had, it does not prevent a crate from starving our CPU resources. `-j N` is just a hint (that cargo, rustc and other tools like make choose to respect), and nothing prevents a build script from spawning as many parallel processes as it likes.

The proper fix for this is to change rustwide to allow limiting the amount of CPU resources the build can use (by passing the appropriate flags to Docker). Still, if this PR actually helps with the crates that caused the outage we should land it first and then implement the proper fix (and I approve merging in that case).

Mark-Simulacrum:
We will want this regardless, I think, with the parameter set according to what we give in Docker -- otherwise Cargo would probably detect more cores than it should, if I understand how the detection works.

jyn514:
Ok, so merge this as a stopgap soft limit and rustwide will work on implementing hard limits upstream?

Mark-Simulacrum:
I agree with merging as a stopgap, most things should be pretty well-behaved and this isn't intended as a security mitigation anyway.

One thing to note is that things are somewhat more likely to timeout as a result, but I wouldn't worry too much.

Issue #656, opened 2020-03-26
Tracking issue: docs.rs leaks memory


pietroalbini:
We started noticing that docs.rs occasionally leaks gigabytes of memory at once, causing production outages due to the excessive amount of used RAM.

![2020-03-26--16-19-04](https://user-images.githubusercontent.com/2299951/77663605-87e8d680-6f7d-11ea-94dc-6d77840f11fa.png)

At the moment we're investigating what the root cause is.

jyn514:
Possibly fixed by #666. Since it's intermittent, we won't know for a while, though.

jyn514:
We haven't had any spikes of this sort in a week, are you comfortable closing this?

![image](https://user-images.githubusercontent.com/23638587/78926099-8550ab80-7a6a-11ea-8a42-a79d90332866.png)

jyn514:
Hmm, we don't have any spikes, but we're still using 2 more GB of RAM at rest than after a server restart ... maybe it's worth keeping this open.

![image](https://user-images.githubusercontent.com/23638587/78926754-a06feb00-7a6b-11ea-82dd-f16084550f73.png)

jyn514:
Someone on hackernews suggested using processes instead of threads, which would automatically clean up resources on panics: https://news.ycombinator.com/item?id=22946237

Not sure how crucial that is at the moment, but if we start having issues again it's certainly something to consider.

Mark-Simulacrum:
That's true, but I think the main concern in my mind right now is the high baseline memory usage. I wouldn't expect a docs.rs idling (i.e. not serving requests) to need, well, really any memory (a hundred megabytes seems more than enough). It would be helpful to narrow down what is using memory in the current process architecture.

jyn514:
I'm going to close this since we haven't had trouble with leaks as such - we might be running into buffer bloat or something, but given that we're only using 3 GB out of the 15 GB available I don't think that's a big deal.

PR #660, opened 2020-03-27
Remove unused `staticfile` dependency

jyn514:
~~Should help address https://github.com/rust-lang/docs.rs/issues/656~~ Nope.

r? @pietroalbini

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,8 @@ jobs:
       - name: Prepare the test environment
         run: |
           prefix=$(pwd)/ignored/prefix
-          mkdir -p ${prefix}/public-html
+          mkdir -p ${prefix}
+          ln -s ../../static ${prefix}/public_html
           echo "::set-env name=CRATESFYI_PREFIX::${prefix}"
 
       - name: Test docs.rs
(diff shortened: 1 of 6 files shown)


Mark-Simulacrum:
We might also want to add a cache header to the static files -- to avoid requests from CloudFront -- but I suspect that can be handled later. (Previously, the staticfile crate did set ETag and Last-Modified on our end).

https://github.com/iron/staticfile/blob/0.4.0/src/static_handler.rs#L212-L214

Otherwise, this looks good to me, though I don't know what our process these days is for merging -- r=me :)

jyn514:
I think we do that already, actually: https://github.com/rust-lang/docs.rs/blob/master/src/web/file.rs#L25

> I don't know what our process these days is for merging -- r=me :)

We don't really have a process ... I don't have admin access to the repo so I can't turn on protected branches.

Mark-Simulacrum:
Please also update the Dockerfile to not touch public_html.

Mark-Simulacrum:
Looks like tests are failing, but feel free to merge whenever.

jyn514:
After testing locally, it looks like this breaks something with the favicon. I don't feel comfortable merging this until we figure out what's going on there.

Issue #658, opened 2020-03-26
Uploading large files to S3 can use an excessive amount of RAM


pietroalbini:
While investigating #656 we discovered that our current upload code uses a lot of RAM when uploading large files to S3. Unscientific tests show that the RAM usage is 10x the size of the file we're trying to upload. We should change the implementation to reduce the amount of RAM usage, or to limit how much data we upload at once.

jyn514:
I think this can be closed now, after profiling the new implementation (unscientifically) it uses 120 MB to upload a 100 MB file, which seems reasonable. Possibly using mmap would reduce that even further if you're interested in that.

pietroalbini:
Hmm, another possibility to reduce the amount of used RAM is to pass a [streaming body](https://docs.rs/rusoto_s3/0.43.0/rusoto_s3/struct.PutObjectRequest.html#structfield.body) to the request, somehow converting a stream of the file content to it.

jyn514:
I think the only way to do that without first reading the file into memory (and copying it into user space and back) is with mmap. I haven't tried to implement this yet though.

Mark-Simulacrum:
I would not try to do `mmap`. Ultimately that's not even necessarily sound to use in Rust, and regardless, our files are usually on the order of dozens of megabytes at most.

If we do want to optimize, we should avoid using the standard allocator for buffering in memory, instead reading into something like a bump allocator so that we can drop the file contents without potentially increasing fragmentation etc in the primary allocator.

But I personally don't think we should do anything here, files are tiny.

jyn514:
I'm onboard with this, the largest file we've served in the last week is only 45 MB. I think it's more important to limit the memory usage in a batch: if we have 1000 files at 10 MB each, that's way more likely to blow out memory than uploading one enormous file.

Mark-Simulacrum:
Right, yes, if we have problems around memory usage I would personally decrease parallelism rather than try to be smarter about how each file is handled.

Ultimately 99% of the files we upload are probably less than one megabyte.

PR #661, opened 2020-03-28
Reduce memory usage when uploading files by a factor of 11

jyn514:
Addresses #658. ~~A future PR could reduce usage by a further factor of 2 by removing unnecessary `clone()`s in the upload code.~~ Done.

This uses the Rust mime_guess package instead libmagic.
Most work was already done in https://github.com/rust-lang/docs.rs/pull/600.
This just sets mime_guess to build unconditionally instead of only on Windows.

This improves the memory usage of docs.rs by over a factor of 5 when
uploading files in some cases. This can be tested locally like so:

Run minio in the background:
```
docker run -p 9000:9000 -e MINIO_ACCESS_KEY=password -e MINIO_SECRET_KEY=password minio/minio server /data
```

Navigate to localhost:9000/ and add a new bucket called `rust-docs-rs`.

Now in another terminal, run docs.rs. This assumes you already have the
CRATESFYI_DATABASE_URL set up locally as described in
[developing without docker-compose](https://forge.rust-lang.org/docs-rs/no-docker-compose.html):

```
 # set up a fairly large file
mkdir -p ignored/tmp
yes | head -n $((1024 * 1024 * 50)) > ignored/tmp/100MB

git checkout master
cargo build --release
RUST_LOG=cratesfyi,info S3_ENDPOINT=http://localhost:9000 AWS_ACCESS_KEY_ID=password AWS_SECRET_ACCESS_KEY=password valgrind --tool=massif --massif-out-file=libmagic.massif.out target/release/cratesfyi database add-directory ignored/tmp/
 # this will show ~1.5 GB used, depending on your system
ms_print libmagic.massif.out | less

 # now try without libmagic
git checkout druid
cargo build --release
RUST_LOG=cratesfyi,info S3_ENDPOINT=http://localhost:9000 AWS_ACCESS_KEY_ID=password AWS_SECRET_ACCESS_KEY=password valgrind --tool=massif --massif-out-file=no_libmagic.massif.out target/release/cratesfyi database add-directory ignored/tmp/
 # this will show ~250 MB used, depending on your system
ms_print no_libmagic.massif.out | less
```

cc @Zexbe 

r? @Mark-Simulacrum

### DIFF
--- a/src/db/file.rs
+++ b/src/db/file.rs
@@ -17,8 +17,6 @@ use rusoto_s3::{S3, PutObjectRequest, GetObjectRequest, S3Client};
 use rusoto_core::region::Region;
 use rusoto_credential::DefaultCredentialsProvider;
 use std::ffi::OsStr;
-#[cfg(not(windows))]
-use magic::{Cookie, flags};
 
 const MAX_CONCURRENT_UPLOADS: usize = 1000;
 
@@ -157,12 +155,11 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
                                               prefix: &str,
                                               path: P)
                                               -> Result<Json> {
-    let trans = conn.transaction()?;
-    #[cfg(not(windows))]
-    let mime_data = load_mime_data()?;
     use std::collections::HashMap;
-    let mut file_paths_and_mimes: HashMap<PathBuf, String> = HashMap::new();
     use futures::future::Future;
+
+    let trans = conn.transaction()?;
+    let mut file_paths_and_mimes: HashMap<PathBuf, String> = HashMap::new();
 
     let mut rt = ::tokio::runtime::Runtime::new().unwrap();
 
@@ -194,10 +191,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
             #[cfg(not(windows))]
             let bucket_path = bucket_path.into_os_string().into_string().unwrap();
 
-            #[cfg(windows)]
             let mime = detect_mime(&content, &file_path)?;
-            #[cfg(not(windows))]
-            let mime = detect_mime(&content, &file_path, &mime_data)?;
 
             if let Some(client) = &client {
                 futures.push(client.put_object(PutObjectRequest {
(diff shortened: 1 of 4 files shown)


src/db/file.rs
```diff
@@ -260,20 +254,6 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
...
-}
-
-#[cfg(not(windows))]
-fn detect_mime(content: &Vec<u8>, file_path: &Path, cookie: &Cookie) -> Result<String> {
-    let mime = cookie.buffer(&content)?;
-    correct_mime(&mime, &file_path)
-}
-
-#[cfg(windows)]
 fn detect_mime(_content: &Vec<u8>, file_path: &Path) -> Result<String> {
```
Mark-Simulacrum: Since we don't need the contents anymore in any of these, could you drop that argument as well?

jyn514:
Note that this now moves from setting the content-type based on the contents to based on the file extension. I don't foresee this impacting us, but I'll run a build with the new library and make sure that everything looks about right in the browser.

jyn514:
Everything looks fine on a local build.

image

src/db/file.rs
```diff
@@ -224,7 +218,7 @@ pub fn add_path_into_database<P: AsRef<Path>>(conn: &Connection,
                 }
             }
 
-            file_paths_and_mimes.insert(file_path.clone(), mime.clone());
+            file_paths_and_mimes.insert(file_path.clone(), mime.to_owned());
```
Mark-Simulacrum: We should be able to avoid the clone here

jyn514:
After the last commit, this now only uses 129 MB locally, for a reduction of over 11x.

src/db/file.rs
jyn514: I don't think we can. `file_path` is borrowed from `currently_uploading`, which can be reused on a later iteration of the loop. I might be able to make `file_paths_and_mimes` store a &'static str but that would require some refactoring elsewhere.

jyn514: Anyway I don't think it should be a big deal since file paths aren't very long compared to the size of a file contents. It will be dropped at the end of the loop in any case.

Mark-Simulacrum: er, sorry I meant the mime type - it's potentially significant (3 pointers + heap to just two pointers).

jyn514: Still doesn't work.

```
error[E0599]: no method named `to_json` found for type `std::vec::Vec<&str>` in the current scope
   --> src/db/file.rs:286:31
    |
286 |         file_list_json.push(v.to_json());
    |                               ^^^^^^^ method not found in `std::vec::Vec<&str>`
    |
    = note: the method `to_json` exists but the following trait bounds were not satisfied:
            `[&str] : rustc_serialize::json::ToJson`
            `std::vec::Vec<&str> : rustc_serialize::json::ToJson`
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use serde_json::value::ToJson;`
```

jyn514: If we're willing to switch this from rustc_serialize to serde_json we could make it work, but I'm not sure I'm comfortable doing that without tests.

Mark-Simulacrum: Gah, okay. Thanks for trying.

Mark-Simulacrum:
I'm happy to merge if you're ready, I think this is good to go.

PR #663, opened 2020-03-30
Run `cargo update`

jyn514:
This is a prerequisite for any later changes to our dependencies.

The last time I tried this we had a production outage (#466), so we should monitor docs.rs carefully after deploying. Note that this has fewer changes than https://github.com/rust-lang/docs.rs/pull/461 since I _only_ ran `cargo update` without changing anything else.

r? @Mark-Simulacrum

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,23 +2,23 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler32"
-version = "1.0.3"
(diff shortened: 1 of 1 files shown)


jyn514:
As mentioned by @Mark-Simulacrum on discord, it would be better to do small incremental updates instead of a mass `cargo update`.

PR #664, opened 2020-03-30
Update some of our dependencies, with focus on removing duplicates

jyn514:
This has just the changes to Cargo.toml from #461, without running `cargo update`.

This reduces our number of dependencies from 418 to 412.
Additionally, we now only compile 2 versions of quote, down from 4 :tada: :sweat_smile:

r? @Mark-Simulacrum


jyn514:
Whoops, don't merge, I added `libc` by accident.

src/web/mod.rs
```diff
@@ -302,7 +302,6 @@ fn render_markdown(text: &str) -> String {
 
     let options = {
         let mut options = ComrakOptions::default();
-        options.safe = true;
```
jyn514: This option is now on by default and the field is called `unsafe_` instead.

Mark-Simulacrum:
This looks good to me, but let's merge only when we have someone monitoring for a few hours (maybe tomorrow).

PR #667, opened 2020-03-31
Update only params and env_logger

jyn514:
Follow-up to #664 updating only two of our dependencies.

The following diffs are generated with [`cargo lock list`](https://docs.rs/cargo-lock/4.0.1/cargo_lock/#list-summarize-packages-in-cargolock) and `diff -y --suppress-common-lines`. `>` means newly added, < means removed,  - means the version changed.

- [Transitive dependencies changed in 664](https://termbin.com/qx2g)
- [Transitive dependencies changed in this PR](https://termbin.com/yd7r)

r? @Mark-Simulacrum

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,20 +164,24 @@ dependencies = [
 
 [[package]]
 name = "bodyparser"
-version = "0.5.0"
+version = "0.8.0"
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
This seems (even more) safe, let's try it! Waiting for CI to pass.

PR #672, opened 2020-04-01
Benchmark HTML parsing

jyn514:
See https://github.com/rust-lang/docs.rs/pull/671, https://github.com/servo/html5ever/issues/414 for rationale.

This makes the `extract_head_and_body` function public since I couldn't
find a cfg flag that would do so only during tests.

This does not run benchmarks in CI, but makes it easy to do so locally. I plan to run benchmarks whenever dependencies are updated in the future.


Mark-Simulacrum:
Looks good to me. I guess the idea is to gather benchmarks before/after changes locally?

Feel free to merge.

jyn514:
> I guess the idea is to gather benchmarks before/after changes locally?

Right, this would give me some idea of 'is rendering faster or slower after switching to kuchiki' without having to deploy to prod and possibly have the site go down again.

PR #673, opened 2020-04-01
Prevent Cargo.toml changes without an updated Cargo.lock

pietroalbini:
In a previous PR the `Cargo.toml` file was changed without committing the updated `Cargo.lock`. This fixes the mismatch and adds a check in CI to prevent this from happening in the future.


Mark-Simulacrum:
r=me

jyn514:
Good catch, thank you!

PR #674, opened 2020-04-01
Limit CPU usage in the sandbox

pietroalbini:
This improves the previously added support for limiting the CPU usage of the builds by:

* Removing the hardcoded `-j2` and allowing the limit to be configured with an environment variabile.
* Limiting the CPU in the sandbox itself, in addition to the `-jN`.

~~Before deploying this~~ (edit: done) the following environment variable will need to be set:

```
DOCS_RS_BUILD_CPU_LIMIT=2
```

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -109,6 +109,13 @@ impl RustwideBuilder {
         })
     }
 
+    fn prepare_sandbox(&self, limits: &Limits) -> SandboxBuilder {
+        SandboxBuilder::new()
+            .cpu_limit(self.cpu_limit.map(|limit| limit as f32))
+            .memory_limit(Some(limits.memory()))
+            .enable_networking(limits.networking())
+    }
+
     pub fn update_toolchain(&mut self) -> Result<()> {
         // Ignore errors if detection fails.
         let old_version = self.detect_rustc_version().ok();
@@ -195,12 +202,8 @@ impl RustwideBuilder {
         let krate = Crate::crates_io(DUMMY_CRATE_NAME, DUMMY_CRATE_VERSION);
         krate.fetch(&self.workspace)?;
 
-        let sandbox = SandboxBuilder::new()
-            .memory_limit(Some(limits.memory()))
-            .enable_networking(limits.networking());
-
         build_dir
-            .build(&self.toolchain, &krate, sandbox)
+            .build(&self.toolchain, &krate, self.prepare_sandbox(&limits))
             .run(|build| {
                 let metadata = Metadata::from_source_dir(&build.host_source_dir())?;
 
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
You should be able to add DOCS_RS_BUILD_CPU_LIMIT=2 to the .env file on the server now, right?

Other than that though this looks good to me.

src/docbuilder/rustwide_builder.rs
```diff
@@ -93,15 +94,27 @@ impl RustwideBuilder {
             .map(|t| Cow::Owned(t))
             .unwrap_or_else(|_| Cow::Borrowed("nightly"));
 
+        let cpu_limit = std::env::var("DOCS_RS_BUILD_CPU_LIMIT")
+            .map(|limit| Some(limit.parse::<u32>().expect("invalid DOCS_RS_BUILD_CPU_LIMIT")))
+            .unwrap_or(None);
```
jyn514, line 99:
```suggestion
            .map(|limit| limit.parse::<u32>().expect("invalid DOCS_RS_BUILD_CPU_LIMIT"));
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -468,6 +473,12 @@ impl RustwideBuilder {
             cargo_args.push(target);
         };
 
+        let tmp_jobs;
+        if let Some(cpu_limit) = self.cpu_limit {
+            tmp_jobs = format!("-j{}", cpu_limit);
+            cargo_args.push(&tmp_jobs);
```
jyn514, line 479: If we're adding a hard limit in rustwide, why are we also telling cargo to slow down? Is the concern that cargo will detect the true number of CPUs and try to spawn too many threads?

src/docbuilder/rustwide_builder.rs
```diff
@@ -93,15 +94,27 @@ impl RustwideBuilder {
             .map(|t| Cow::Owned(t))
             .unwrap_or_else(|_| Cow::Borrowed("nightly"));
 
+        let cpu_limit = std::env::var("DOCS_RS_BUILD_CPU_LIMIT")
```
jyn514, line 97: I think we should consider setting a limit even if the environment variable isn't set. #cores - 1 sounds reasonable, is there a way to detect the number of CPU cores using the stdlib or would we need a crate?

jyn514: Looks like there's a crate for this: https://docs.rs/num_cpus/

src/docbuilder/rustwide_builder.rs
Mark-Simulacrum: You don't want unnecessary contention over nonexistent resources yes.

src/docbuilder/rustwide_builder.rs
pietroalbini: Hmm I'd prefer not to set a limit. If we end up splitting the build server and the web server we'll want the web server to use all cores.

pietroalbini:
Set the variable on the production server and addressed review comments.

src/docbuilder/rustwide_builder.rs
```diff
@@ -93,15 +94,26 @@ impl RustwideBuilder {
...
+            .map(|limit| limit.parse::<u32>().expect("invalid DOCS_RS_BUILD_CPU_LIMIT"));
```
jyn514, line 98: Whoops, I never actually tested the suggested change. Maybe try this?

```suggestion
        let cpu_limit = std::env::var("DOCS_RS_BUILD_CPU_LIMIT")
            .ok().map(|limit| limit.parse::<u32>().expect("invalid DOCS_RS_BUILD_CPU_LIMIT"));
```

jyn514: Just checked and this does fix the issue.

pietroalbini:
Fixed and rebased on top of the latest master (with rustfmt).

PR #675, opened 2020-04-01
Refactored the Badges crate

Kixiron:
* Updated dependencies
* Transferred to the 2018 edition of Rust
* Formatting
* Un-ignored test and gitignored the file output by it (`test.svg`)


src/web/badge/Cargo.toml
```diff
@@ -6,10 +6,8 @@ authors = ["Onur Aslan <onur@onur.im>"]
...
 repository = "https://github.com/rust-lang/docs.rs"
 documentation = "https://docs.rs/badge"
-
-[lib]
-path = "badge.rs"
+edition = "2018"
 
 [dependencies]
-base64 = "0.9.0"
-rusttype = "0.7"
```
jyn514, line 15: Please do not update dependencies in a refactor PR. You can make a separate PR for this.

src/web/badge/src/lib.rs
```diff
@@ -33,95 +26,106 @@ impl Default for BadgeOptions {
...
 
-
     pub fn to_svg_data_uri(&self) -> String {
-        format!("data:image/svg+xml;base64,{}",
-            Base64Display::standard(self.to_svg().as_bytes()))
+        format!(
+            "data:image/svg+xml;base64,{}",
+            Base64Display::with_config(
+                self.to_svg().as_bytes(),
+                Config::new(CharacterSet::UrlSafe, true)
```
jyn514, line 70: Oh this is a really good change, it would be worth having on its own. Can you  check whether this fixes https://github.com/rust-lang/docs.rs/issues/395 and if so add a test?

You can test manually like so:
```
docker-compose run web build crate zstd 0.5.1+zstd.1.4.4
```
and then navigating to http://localhost:3000/zstd/badge.svg?version=0.5.1+zstd.1.4.4.

(actually I'll just do this myself, give me a second)

src/web/badge/src/lib.rs
```diff
@@ -166,17 +168,17 @@ mod tests {
...
+
         let mut file = File::create("test.svg").unwrap();
         let options = BadgeOptions {
             subject: "build".to_owned(),
             status: "passing".to_owned(),
             ..BadgeOptions::default()
         };
         let badge = Badge::new(options).unwrap();
+
         file.write_all(badge.to_svg().as_bytes()).unwrap();
```
jyn514, line 182: I think a better idea might be to write the badge to memory instead of to disk. It would also be nice to compare it to a known good badge, but it would be fine even without.

src/web/badge/src/lib.rs
jyn514: Update: this does not fix #395. No need to make it a separate PR.

PR #676, opened 2020-04-01
Add tests for HTML parsing

jyn514:
Enables #671 

r? @Mark-Simulacrum


src/utils/html.rs
```diff
@@ -72,3 +72,38 @@ fn extract_class(node: &Handle) -> String {
...
+mod test {
+    use super::*;
+
+    fn round_trip(a: &str) -> String {
+        let parser = parse_document(RcDom::default(), ParseOpts::default());
+        stringify(parser.one(a).document)
+    }
+    fn mostly_equal(a: &str, b: &str) -> bool {
+        round_trip(a.trim()).trim() == round_trip(b.trim()).trim()
+    }
```
Mark-Simulacrum, line 86: I would rather not depend on the parser from the tests directly, just exactly compare, is there a reason you chose to do this?

jyn514: The whitespace was messing up the tests. Also if we compare exactly then it's dependent on how it's serialized by the parser, which might change in a minor version. The way the document is parsed is much less likely to change between version or between parsers.

Mark-Simulacrum: I would rather update our tests than have this, even if that means updating on a minor version bump.

PR #680, opened 2020-04-01
Updated the GitHub Actions script to include caching

Kixiron:
* Added caching to the Actions flow
  * Caches the cargo registry at `~/.cargo/registry`
  * Caches cargo binaries at `~/.cargo/bin`
  * Caches cargo git at `~/.cargo/git`
  * Caches `/target` build artifacts
  * Caches based off of the hash of the Cargo.lock, so changes that alter it will require full rebuilds
  * Will hopefully provide faster CI build/test times

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,28 @@ jobs:
         with:
           fetch-depth: 2
 
+      # Caching
+      - name: Cache Cargo Registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache Cargo Bin
+        uses: actions/cache@v1
+        with:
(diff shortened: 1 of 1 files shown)


pietroalbini:
I don't think this is effective in the long term: GitHub Actions's cache is immutable, and the same key won't be reuploaded if the files on disk change. Files in `target/` are regenerated after rustc updates, so we'll need to include the rustc version in the cache key.

You can see [how crates.io does it](https://github.com/rust-lang/crates.io/blob/9e47a175c96aada9e7a622549581e6cf8664395a/.github/workflows/ci.yml#L145-L156).

Kixiron:
Anything that speeds up builds is a plus in my books, do you prefer cargo's caching + caching cargo artifacts?

pietroalbini:
I think this PR is fine if the rustc version is added to the cache keys. Otherwise once a new stable version is released we'd lose all the performance boost of caches.

pietroalbini:
Great, thanks!

PR #681, opened 2020-04-02
Upgraded to Rust 2018 edition

Kixiron:
* Upgraded to the 2018 edition of Rust
* Fixed module imports and relative imports
* Removed `extern crate` usages
* Removed `#[macro_use]` imports where possible (`log` would be an entirely different PR in of itself)
  * Apart from `log`, all macro usages are localized


src/web/metrics.rs
```diff
@@ -2,9 +2,12 @@ use super::pool::Pool;
 use iron::headers::ContentType;
 use iron::prelude::*;
 use iron::status::Status;
-use prometheus::{Encoder, IntGauge, IntCounter, TextEncoder};
+use prometheus::{
+    opts, register_counter, register_int_counter, register_int_gauge, Encoder, IntCounter,
+    IntGauge, TextEncoder, __register_gauge,
```
jyn514, line 7: What are these imports doing? `register_int_counter` and `register_int_gauge` are being used but the others don't look necessary.

Kixiron:
I'm so sorry, that was an accidental push. In response to your question though, the 'unused' macros are actually used because prometheus's macro scoping is really bad, and they don't canonically import their own macros but opt to take macros from the use-site environment.

jyn514:
Ouch, ok as long as they're being used that's fine. If you can add back the 'Missed one' commit this will be good to merge once CI passes.

Kixiron:
Alright, added back the commit, I have a funny feeling that whenever we get around to updating prometheus this problem will be solved for us

PR #683, opened 2020-04-02
Little Refactors

Kixiron:
* Made code more consistent in formatting
* More closely held to Rust 2018 idioms 
* Readability improvements


src/docbuilder/mod.rs
```diff
@@ -56,40 +52,43 @@ impl DocBuilder {
...
-                               WHERE crates.id = releases.crate_id",
-                   &[])
-            .unwrap() {
+        for row in &conn
+            .query(
+                "SELECT name, version FROM crates, releases \
+                 WHERE crates.id = releases.crate_id",
+                &[],
+            )
+            .unwrap()
```
jyn514, line 67: Can you return this error instead of unwrapping?

src/docbuilder/rustwide_builder.rs
```diff
@@ -108,7 +108,7 @@ impl RustwideBuilder {
 
         let mut targets_to_install = TARGETS
             .iter()
-            .map(|t| t.to_string())
+            .map(|t| (*t).to_string())
```
jyn514, line 111: This dereference is redundant, probably inserted by clippy.
```suggestion
            .map(|t| t.to_string())
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -215,7 +215,7 @@ impl RustwideBuilder {
                     let file_name = if versioned {
                         format!("{}-{}.{}", segments[1], rustc_version, segments[0])
                     } else {
-                        file.to_string()
+                        (*file).to_string()
```
jyn514, line 218: Redundant dereference
```suggestion
                        file.to_string()
```

src/utils/daemon.rs
```diff
@@ -25,13 +25,16 @@ use ::{
 };
 
 pub fn start_daemon(background: bool) {
+    const CRATE_VARIABLES: [&str; 4] = [
+        "CRATESFYI_PREFIX",
+        "CRATESFYI_PREFIX",
+        "CRATESFYI_GITHUB_USERNAME",
+        "CRATESFYI_GITHUB_ACCESSTOKEN"
+    ];
```
jyn514, line 33:
```suggestion
    const CRATE_VARIABLES: [&str; 3] = [
        "CRATESFYI_PREFIX",
        "CRATESFYI_GITHUB_USERNAME",
        "CRATESFYI_GITHUB_ACCESSTOKEN"
    ];
```

src/utils/github_updater.rs
```diff
@@ -109,27 +108,29 @@ fn get_github_fields(path: &str) -> Result<GitHubFields> {
...
-                             reponame
-                         }))
+
+            Some(format!(
+                "{}/{}",
+                username,
+                if reponame.ends_with(".git") {
+                   reponame.split(".git").next().unwrap()
+                } else {
+                    reponame
```
jyn514, line 129: Can you turn this into a temporary variable so it doesn't make this expression really big?

src/utils/release_activity_updater.rs
```diff
@@ -1,42 +1,50 @@
...
 
 pub fn update_release_activity() -> Result<()> {
-
     let conn = connect_db()?;
-    let mut dates = Vec::new();
-    let mut crate_counts = Vec::new();
-    let mut failure_counts = Vec::new();
+    let mut dates = Vec::with_capacity(30);
+    let mut crate_counts = Vec::with_capacity(30);
+    let mut failure_counts = Vec::with_capacity(30);
```
jyn514, line 11: Ooh this is nice, I'm surprised clippy was this smart.

src/web/mod.rs
```diff
@@ -475,29 +475,29 @@ pub(crate) struct MetaData {
...
-        None
+        let rows = conn.query(
+        "SELECT crates.name,
+                       releases.version,
+                       releases.description,
+                       releases.target_name,
+                       releases.rustdoc_status,
+                       releases.default_target
+                FROM releases
+                INNER JOIN crates ON crates.id = releases.crate_id
```
jyn514, line 486: We probably should make this JOIN a view at some point ... but that's separate from your PR.

src/web/releases.rs
```diff
@@ -398,14 +398,16 @@ pub fn author_handler(req: &mut Request) -> IronResult<Response> {
...
     let author = ctry!(router.find("author")
-        .ok_or(IronError::new(Nope::CrateNotFound, status::NotFound)));
-
-    let (author_name, packages) = if author.starts_with("@") {
-        let mut author = author.clone().split("@");
-        get_releases_by_owner(&conn,
-                              page_number,
-                              RELEASES_IN_RELEASES,
-                              cexpect!(author.nth(1)))
+        .ok_or_else(|| IronError::new(Nope::CrateNotFound, status::NotFound)));
```
jyn514, line 401: Is this change necessary? This isn't doing anything expensive and it adds mental overhead.

src/docbuilder/rustwide_builder.rs
Kixiron: It's actually a performance thing, `&str` has a fast, specialized `to_string` implementation while `&&str` does not

src/docbuilder/rustwide_builder.rs
Kixiron: Same thing with the first one, it's more performant to dereference

src/utils/release_activity_updater.rs
Kixiron: Well my friend, that wasn't clippy

src/web/releases.rs
Kixiron: It's actually a performance thing, `ok_or` always allocates, no matter the status of what it's called on, while `ok_or_else` is lazy, which means it's only allocated for when it's actually needed

src/docbuilder/mod.rs
Kixiron: I'm not sure what the error type would be, or can I just `?`?

jyn514: Yes, `?` should work. All the error types are `failure::Error`, which implements `From` for most errors.

src/docbuilder/rustwide_builder.rs
jyn514: Oh interesting. Can you leave that as a comment and maybe a link to the documentation?

src/web/releases.rs
jyn514: Right, but there's no allocation here. `IronError` is only on the stack, as is `Nope` and `NotFound`.

Kixiron: Alright, I'll revert that

src/docbuilder/rustwide_builder.rs
Kixiron: Best I've got is the [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string), but it explains it fairly well. `&str` has it's own `ToString` impl, while `&&str` goes through `Display`, which is much slower

jyn514:
Let's come back to this after the clippy and rustfmt PRs, so I'll be able to see what actually changed instead of looking at formatting.

src/docbuilder/rustwide_builder.rs
```diff
@@ -122,8 +122,9 @@ impl RustwideBuilder {
 
         let mut targets_to_install = TARGETS
             .iter()
-            .map(|&t| t.to_string())
+            .map(|t| (*t).to_string()) // &str has a specialized ToString impl, while &&str goes through Display
             .collect::<HashSet<_>>();
```
jyn514, line 126: I liked this better before. The comment is fine but I prefer `|&t|` to the parentheses.

src/docbuilder/rustwide_builder.rs
```diff
@@ -225,7 +226,7 @@ impl RustwideBuilder {
                     let file_name = if versioned {
                         format!("{}-{}.{}", segments[1], rustc_version, segments[0])
                     } else {
-                        file.to_string()
+                        (*file).to_string() // &str has a specialized ToString impl, while &&str goes through Display
```
jyn514, line 229: I don't think this dereference is necessary, `file` is already a `&str` because of the implicit dereference in the pattern on line 223.

src/utils/daemon.rs
```diff
@@ -18,17 +18,15 @@ use time;
...
+    const CRATE_VARIABLES: [&str; 3] = [
         "CRATESFYI_PREFIX",
         "CRATESFYI_GITHUB_USERNAME",
         "CRATESFYI_GITHUB_ACCESSTOKEN",
-    ]
-    .iter()
-    {
-        if env::var(v).is_err() {
-            panic!("Environment variable {} not found", v);
-        }
```
jyn514, line 31: I prefer the way it was before, I think it was easier to read. Also I don't see the point of having a `const` since it's only used once, and that directly next to where the array is defined.

src/utils/release_activity_updater.rs
```diff
@@ -25,22 +25,25 @@ pub fn update_release_activity() -> Result<()> {
...
+
         let release_count: i64 = rows.get(0).get(0);
         let failure_count: i64 = failures_count_rows.get(0).get(0);
         let now = now();
         let date = now - Duration::days(day);
+
+        // unwrap is fine here, as our date format is always valid
         dates.push(format!("{}", date.strftime("%d %b").unwrap()));
-        // unwrap is fine here,             ~~~~~~~~~~~~^  our date format is always valid
+
```
jyn514, line 46:
```suggestion
```

jyn514:
Please update the PR description to reflect what this does, it's changed a lot since you opened it.

src/utils/daemon.rs
Kixiron: I think it's much cleaner this way because of how it removes the rather large array away from the for statement

Kixiron: The const also just gives some more guarantees that it'll be flattened in the final code

jyn514: > I think it's much cleaner this way because of how it removes the rather large array away from the for statement

Ok, can you change it to just being a `const` array instead of `unwrap_or_else`? The first time I read the unwrap_or_else I was worried it panicked eagerly instead of on error :sweat_smile: 

> The const also just gives some more guarantees that it'll be flattened in the final code

Can you expand on this? My understanding is that for any `const X = blah;`, `X` behaves exactly as if you'd typed `blah` in the source code (modulo bugs in rustc).

Kixiron: > Ok, can you change it to just being a `const` array instead of `unwrap_or_else`? The first time I read the unwrap_or_else I was worried it panicked eagerly instead of on error 😅
Sure thing, it's actually better as the if now that I look at it

> Can you expand on this? My understanding is that for any` const X = blah;`, `X` behaves exactly as if you'd typed` blah` in the source code (modulo bugs in rustc).
Yah, that's exactly what I mean, the `const` makes sure that the variable is "the same" which could possibly not happen with a `let`. Adding onto the `const` train, if ever more variables are added then we don't end up with the longest for statement in the world, we just extend the const some. It gives more clarity to the code imo, but that's a personal thing

PR #684, opened 2020-04-02
Refactor add_package.rs

Kixiron:
* Made the function signatures readable (Still need further work with splitting into separate functions)
* Changed passing integers by reference into passing them directly, pointers are larger than `i32`s
* Overall style and formatting changes


jyn514:
As discussed on discord, I think having one principled `rustfmt` PR would be better than incrementing piecemeal and mixing it with other clippy lints/style issues. I'm happy to take this change:

> Changed passing integers by reference into passing them directly, pointers are larger than i32s

but I think the rest are not the best direction for docs.rs at this time.

PR #686, opened 2020-04-02
Made cratesfyi.rs more readable

Kixiron:
* Broke the clap argument parser into a separate function
* Changed redundant `if let`s


jyn514:
As discussed on Discord, I don't think these changes are the best direction for docs.rs at this time.

PR #687, opened 2020-04-02
Refactored metadata.rs

Kixiron:
* Removed redundant `if let`s
* Minimized code duplication
* Split off get fetching of the package table into a separate function
* Overall style and formatting changes


src/docbuilder/metadata.rs
```diff
@@ -83,23 +83,26 @@ impl Metadata {
...
-        use std::io::Read;
-        let mut f = match File::open(path) {
-            Ok(f) => f,
-            Err(_) => return Metadata::default(),
+        use std::{fs::File, io::Read};
+
+        let mut file = if let Ok(file) = File::open(path) {
+            file
+        } else {
+            return Metadata::default();
```
jyn514, line 96: I think this looked better before, actually.

src/docbuilder/metadata.rs
```diff
@@ -114,56 +117,98 @@ impl Metadata {
...
     fn from_str(manifest: &str) -> Metadata {
         let mut metadata = Metadata::default();
 
-        let manifest = match manifest.parse::<Value>() {
-            Ok(m) => m,
-            Err(_) => return metadata,
+        let manifest = if let Ok(manifest) = manifest.parse::<Value>() {
+            manifest
+        } else {
+            return metadata;
```
jyn514, line 126: Ditto, I think the match looked better.

src/docbuilder/metadata.rs
```diff
@@ -114,56 +117,98 @@ impl Metadata {
...
-                    metadata.targets = table.get("targets").and_then(|f| f.as_array())
-                        .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
-                    metadata.rustc_args = table.get("rustc-args").and_then(|f| f.as_array())
-                        .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
-                    metadata.rustdoc_args = table.get("rustdoc-args").and_then(|f| f.as_array())
-                        .and_then(|f| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect());
-                }
+        if let Some(table) = fetch_manifest_tables(&manifest) {
+            let collect_into_array =
+                |f: &Vec<Value>| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect();
```
jyn514, line 131:
```suggestion
                |f: &[Value]| f.iter().map(|v| v.as_str().map(|v| v.to_owned())).collect();
```

Kixiron:
Ah, turns out using `&[Value]` doesn't work because it's a closure fed into the function, and therefore its signature has to perfectly match

src/docbuilder/metadata.rs
```diff
@@ -192,6 +198,18 @@ impl Metadata {
     }
 }
 
+fn fetch_manifest_tables<'a>(manifest: &'a Value) -> Option<&'a Map<String, Value>> {
```
jyn514, line 201: What's the reason for a separate function?

Kixiron: Makes things much cleaner because of how chained the internals are

jyn514: I'd rather not separate out functions unless it represents another level of abstraction (e.g. we expect the implementation to change) or the code is used more than once. Can you make this a temporary variable in `from_str` instead? That will make it easier to see what's going on without jumping around in the file.

Kixiron: I moved the function up to where it was used, as the entire reason I made it a function (That I totally forgot about) is that the parent function doesn't return an option, but the internals of `fetch_manifest_tables` uses *tons*, and the separate function allows using the try operator for much cleaner code, since we ultimately want to handle the `None` variant. I attempted to make it a closure, but the lifetimes didn't want to play nice

PR #689, opened 2020-04-02
Add a dependency audit to the CI flow

Kixiron:
* Adds a dependency audit to the CI
  * Dependencies audited against the [rustsec](https://rustsec.org/) database using [cargo-audit](https://github.com/RustSec/cargo-audit)
  * Runs on pushes that modify the Cargo.toml or Cargo.lock (Not needed on pushes that don't change these, will help speed things up
  * Runs on its own once per day

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,18 @@
+name: Audit
+
+on:
+  push:
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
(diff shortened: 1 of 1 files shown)


jyn514:
r? @pietroalbini

This mostly looks fine to me but I'm a little hesitant to depend on a random GitHub action. I know I have a tendency to avoid dependencies though so I think it's best to have a second opinion.

pietroalbini:
Thanks!

PR #690, opened 2020-04-02
Got rid on hard-coded host target in rustwide_builder.rs

Kixiron:
* Now set by the build.rs, its value will be the same as the `TARGET` variable set by cargo
* May have also formatted the build.rs, whoops

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,32 +1,34 @@
+use git2::Repository;
 use std::env;
-use std::path::Path;
 use std::fs::{self, File};
 use std::io::Write;
-use git2::Repository;
-
+use std::path::Path;
 
 fn main() {
+    // Set the host target
+    println!(
+        "cargo:rustc-env=CRATESFYI_HOST_TARGET={}",
+        env::var("TARGET").unwrap()
+    );
+
     write_git_version();
     compile_sass();
     copy_js();
 }
 
-
 fn write_git_version() {
     let git_hash = get_git_hash().unwrap_or("???????".to_owned());
     let build_date = time::strftime("%Y-%m-%d", &time::now_utc()).unwrap();
     let dest_path = Path::new(&env::var("OUT_DIR").unwrap()).join("git_version");
+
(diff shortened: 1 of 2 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -23,8 +23,7 @@ const DEFAULT_RUSTWIDE_WORKSPACE: &str = ".rustwide";
 
 // It is crucial that this be the same as the host that `docs.rs` is being run on.
 // Other values may cause strange and hard-to-debug errors.
-// TODO: use `TARGET` instead? I think `TARGET` is only set for build scripts, though.
-pub(super) const HOST_TARGET: &str = "x86_64-unknown-linux-gnu";
+pub(super) const HOST_TARGET: &str = env!("CRATESFYI_HOST_TARGET"); // Set in build.rs
```
jyn514, line 26: Is there a way to test this somehow?

build.rs
```diff
@@ -1,57 +1,57 @@
+use git2::Repository;
 use std::env;
-use std::path::Path;
 use std::fs::{self, File};
 use std::io::Write;
-use git2::Repository;
-
+use std::path::Path;
```
jyn514, line 5: There's no need to reorder the imports. I like dependency imports being separate from std imports anyway.

src/docbuilder/rustwide_builder.rs
Kixiron: Not that I know of, tests only have access to the variables that crates themselves do

jyn514:
This looks good but I want to wait to merge until I get a chance to test builds locally.

jyn514:
Looks good to me! Not sure how much it will help builds on windows, but it's at least one more roadblock out of the way.

Kixiron:
Even if it doesn't help with windows builds it's still a good step away from hard-coding, I'm sure this will allow for Mac support

PR #691, opened 2020-04-02
Moved off of blanket #[macro_use] import for log

Kixiron:
* Removed global `#[macro_use]` import for log macros, instead local imports are now used


Kixiron:
Checks are failing due to network errors

jyn514:
GitHub is showing this as 'cannot be rebased due to conflicts', can you look into that?

PR #685, opened 2020-04-02
Refactored web/mod.rs

Kixiron:
* Removed redundant `'static` lifetimes from constants
* Method chain formatting
* Transferred from the painful if statement to a wonderful match
* Overall style and formatting changes

PR #696, opened 2020-04-03
Fix clippy warnings and add clippy to CI

jyn514:
This is similar to #683, #684, #685, #686, #687, but only fixes clippy warnings without trying to make any style changes. I plan to make a follow-up PR running `rustfmt` shortly.

cc @Kixiron 
r? @pietroalbini


.github/workflows/ci.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       - name: Build docs.rs
         run: cargo build --locked
 
+      - name: Run clippy
+        run: cargo clippy -- -D clippy::all -D unused-imports
```
pietroalbini, line 34: I'd just do `-D warnings`.

jyn514: Hmm, it looks like `-D warnings` doesn't work: 
```
$ cargo clippy -- -D warnings
warning: constant item is never used: `DEFAULT_ID`
   --> src/web/mod.rs:516:5
    |
516 |     const DEFAULT_ID: i32 = 0;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

jyn514: How strange, it worked after a `cargo clean`. It must have been introduced by `cargo` caching warnings after `check`.

PR #697, opened 2020-04-03
Run rustfmt on the codebase

jyn514:
r? @pietroalbini 
cc @Kixiron

This should be exactly the same diff as running `cargo fmt` from master.


pietroalbini:
Could you gate CI on rustfmt in a followup PR?

Issue #699, opened 2020-04-03
Building both HOST_TARGET and a different default-target will overwrite the default target


jyn514:
Consider the following Cargo.toml:

```
[package.metadata.docs.rs]
targets = ["i686-unknown-linux-gnu", "x86_64-unknown-linux-gnu"]
```

This sets `default-target` to `"i686-unknown-linux-gnu"` and additionally builds `"x86_64-unknown-linux-gnu"`, which is ([usually](https://github.com/rust-lang/docs.rs/pull/690)) the HOST_TARGET.

Now consider how we currently handle cross-compiling the default target:
1. [Build the target and output it into `target/$target/doc`](https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L499)
2. [Move `target/$target/doc` to `target/doc`](https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L521).

Entirely separately, whenever we build `HOST_TARGET`, we output into `target/doc`. What happens if there's already something in `target/doc`? It gets overwritten! What we should do instead is
1. After building the HOST_TARGET, _instead of_ leaving it in `target/doc`, move it to `target/$HOST_TARGET/doc`.
2. After building _every_ target, move the default target to `target/doc`.

Giant thanks to @simlay for catching this bug, it would have gone unnoticed for quite a while otherwise.

PR #700, opened 2020-04-03
Distinguish between the host target and the default target

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/699

r? @Mark-Simulacrum

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -238,7 +238,7 @@ impl RustwideBuilder {
                     })?;
                 }
 
-                add_path_into_database(&conn, "", &dest)?;
+                add_path_into_database(&conn, HOST_TARGET, &dest)?;
                 conn.query(
                     "INSERT INTO config (name, value) VALUES ('rustc_version', $1) \
                      ON CONFLICT (name) DO UPDATE SET value = $1;",
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -325,8 +325,8 @@ impl RustwideBuilder {
         // Builds are somewhat of a hack because we don't control where cargo outputs the documentation.
         //
         // Cargo's logic is simple:
-        // - if building for the host target, output to `target/doc`
-        // - otherwise, we're cross compiling to `$target`, output to `target/$target/doc`.
+        // - if passed the `--target` flag, output to `target/$target/doc`.
+        // - otherwise, output to `target/doc`
         //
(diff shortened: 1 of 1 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -514,16 +539,20 @@ impl RustwideBuilder {
...
-        // If we're passed a default_target which requires a cross-compile,
-        // cargo will put the output in `target/<target>/doc`.
-        // However, if this is the default build, we don't want it there,
-        // we want it in `target/doc`.
-        if target != HOST_TARGET && is_default_target {
-            // mv target/$target/doc target/doc
+        // If we're passed the host target, cargo will put the output in `target/doc`.
+        // However, we don't want it there, we want it in `target/$HOST_TARGET/doc`.
+        // See comments to `build_package` for an explanation of why.
+        if target == HOST_TARGET {
```
Mark-Simulacrum, line 545: So Cargo actually doesn't *quite* behave this way, I think, if I follow this comment. In particular, so long as we pass `--target`, target-specific artifacts will always go into `target/$target/` I think. But host-artifacts (e.g. build scripts and proc macros) go into `target/`.

Can you make sure that this doesn't require changes to the code here? I'm not 100% confident myself; it looks right, though, I think.

Mark-Simulacrum:
Modulo my comment, and presuming some local testing has been done on the following:
 * proc macro
 * some library with a proc macro dep
 * some library with a build script
 * some library

r=me

src/docbuilder/rustwide_builder.rs
jyn514: Ah yes good point - the reason files are going into `target/doc` is because we specifically don't pass `--target` for the HOST_TARGET. I'll add that as a comment.

jyn514:
Working on building the following locally:

- [ ] proc macro (rstest 0.6.1)
- [ ] library with proc macro dependency (grex 1.0.0)
- [ ] build script (target_lexicon 0.10.0)
- [ ] library (hexponent 0.2.3)

Can you expand on why you think the build script would make a difference?

jyn514:
Good call on asking for the builds! I found that I forgot to update `add_esssential_files`.

jyn514:
Tested all the types of crates you mentioned and they work.

Mark-Simulacrum:
@jyn514 The build script likely has different handling for us and for Cargo at some level as it's older and perhaps has more quirks -- it also results in slightly different places for some of the artifacts I think (e.g. build dir is populated perhaps?).

PR #705, opened 2020-04-10
Revert "Distinguish between the host target and the default target"

jyn514:
Reverts rust-lang/docs.rs#700.

This was likely the cause of newly built docs missing Javascript files (e.g. https://docs.rs/basic-lang/0.5.0/basic/). Reverting until I can figure out the problem.

cc @pietroalbini, @Mark-Simulacrum


jyn514:
Possibly https://github.com/rust-lang/docs.rs/pull/705/commits/f3a979fb3b7a389e11500b934165119b4813ff3b was adding the essential files to the wrong URL? I don't see why that wouldn't show up locally though.

PR #706, opened 2020-04-10
Expose PostgreSQL database ports in docker-compose

repnop:
This commit exposes the PostgreSQL ports used for the test database in the docker-compose.yml and updates the README.md with the URL that can be used for running tests when set as the CRATESFYI_DATABASE_URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,13 +86,16 @@ cargo test
 ```
 
 Some tests require access to the database. To run them, set the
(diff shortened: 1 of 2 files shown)


jyn514:
Can you also add a note to the `# Running tests` section pointing to the docker-compose section?

jyn514:
Other than that this looks great, thanks for working on it!

PR #348, opened 2019-05-04
attempt to look up crates with -/_ replaced before 404ing

QuietMisdreavus:
While reading the logs today, i noticed a few 404s where people had tried to go to the URL `docs.rs/crate_name` when the real crate is called `crate-name` with a hyphen instead of an underscore. This PR adds an extra lookup in the `/:crate` route that replaces hyphens with underscores (and vice-versa) before returning the 404. So in this case, attempting to go to `/crate_name` will redirect to `/crate-name/version/target` instead.

PR #707, opened 2020-04-10
Percent decode crate names

### LINKED ISSUE #670: Try to percent-decode the crate name if it gives a 404
> I use the DuckDuckGo bangs feature a lot to search for crates on docs.rs using the `!drs cratename` bang, however it doesn't seem to work with crates that have hyphens in the name like `async-std`, for example. I get redirected to `https://docs.rs/async%2Dstd` -- which upon pasting that looks like its because it percent encodes the hyphen! I wouldn't mind putting in a PR to percent decode the parameters if that's a change everyone is okay with :+1: (edit: well, assuming that's the problem haha)

This would be great to have, but it should only redirect if the current page would give a 404 to avoid making some crates unnameable.

_Originally posted by @repnop in https://github.com/rust-lang/docs.rs/issues/603#issuecomment-606913272_

repnop:
Fixes #670

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -87,6 +87,8 @@ impl iron::Handler for RustLangRedirector {
 /// Handler called for `/:crate` and `/:crate/:version` URLs. Automatically redirects to the docs
 /// or crate details page based on whether the given crate version was successfully built.
 pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
+    use url::percent_encoding::percent_decode;
+
     fn redirect_to_doc(
         req: &Request,
         name: &str,
@@ -157,6 +159,9 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
     let router = extension!(req, Router);
     // this handler should never called without crate pattern
     let crate_name = cexpect!(router.find("crate"));
+    let crate_name = percent_decode(crate_name.as_bytes())
+        .decode_utf8()
+        .unwrap_or_else(|_| crate_name.into());
     let req_version = router.find("version");
 
     let conn = extension!(req, Pool).get();
@@ -665,5 +670,22 @@ mod test {
             )?;
             Ok(())
         })
+    }
+    #[test]
+    fn crate_name_percent_decoded_redirect() {
+        wrapper(|env| {
+            env.db()
+                .fake_release()
+                .name("fake-crate")
+                .version("0.0.1")
+                .rustdoc_file("fake_crate/index.html", b"some content")
+                .create()
+                .unwrap();
+
+            let web = env.frontend();
+            assert_redirect("/fake%2Dcrate", "/fake-crate/0.0.1/fake_crate/", web)?;
+
+            Ok(())
+        });
     }
 }


repnop:
This initial commit doesn't solve the problem everywhere. The crate name decoding needs to happen higher up for it to apply everywhere but I'm not sure where is best to put it.

src/web/rustdoc.rs
```diff
@@ -666,4 +671,21 @@ mod test {
...
+            env.db()
+                .fake_release()
+                .name("fake-crate")
+                .version("0.0.1")
+                .rustdoc_file("fake_crate/index.html", b"some content")
+                .create()
+                .unwrap();
+
+            let web = env.frontend();
+            assert_redirect("/fake%2Dcrate", "/fake-crate/0.0.1/fake_crate/", web)?;
```
jyn514, line 686: Can you also test `/crate/fake%2Dcrate`?

Also, I think this will break for `/fake%2Dcrate/0.0.1/fake_crate`, but that can be added in a follow-up PR.

jyn514:
Ah, you already knew what I was going to comment ;)

I think this is ok for an initial change, it's certainly better than what we have currently. However I think expanding this to be more general would require a redesign of `match_version`, something like what https://github.com/rust-lang/docs.rs/pull/348 did. If you're comfortable doing that, it would be great to have, but if not I'm happy with this as-is.

repnop:
I actually don't think the problem as a whole is fixed by changing `match_version` because the crate name needs URL decoded higher up otherwise the other SQL queries fail (and is why I moved the change and test from `mod.rs` into `rustdoc.rs` which contains the redirect handler). As it stands right now this would need applied to every routing function that takes the crate name as a parameter. Undoing my change to `rustdoc.rs` and moving it into `match_version` results in a test failure because it attempts to redirect to the percent-encoded name:
```
---- web::rustdoc::test::crate_name_percent_decoded_redirect stdout ----
thread 'web::rustdoc::test::crate_name_percent_decoded_redirect' panicked at '/fake%2Dcrate: 
expected redirect to /fake-crate/0.0.1/fake_crate/,
got redirect to /fake%2Dcrate/0.0.1/fake_crate/', src/test/mod.rs:65:13
```
So that makes me unsure of where all we want to apply this change to or if we want to do something more drastic involving the routing

jyn514:
Ugh, this really should be done by `iron` and not by us. Maybe try url-decoding in [`rustdoc_page`](https://github.com/rust-lang/docs.rs/blob/master/src/web/routes.rs#L221) and see if that helps? That should affect all the pages that are served from S3 (i.e. don't do their own URL-decoding).

repnop:
Unfortunately it flat out 404s on `/fake%2Dcrate` :(

jyn514:
> Unfortunately it flat out 404s on `/fake%2Dcrate` :( Adding in the decoding in all 3 places still nets us a panic:
> 
> ```
> '/crate/fake%2Dcrate: expected redirect to /fake-crate/0.0.1/fake_crate/, got redirect to /crate/fake%2Dcrate/0.0.1'
> ```

That redirect looks correct to me. `/crate/:crate` should always redirect to `/crate/:crate/:version`.

jyn514:
I'm happy with only fixing the common case instead of every URL, since this is primarily intended for fixing automated searches from duck-duck-go. Other urls still have links that work fine, so I don't see a need to decode those unless it causes problems.

PR #709, opened 2020-04-11
Replaced rustc_serialize with serde

Kixiron:
* Removed `rustc_serialize`
* Added `serde` and `serde_json`

Notes: 
* ~~This is currently broken, as [handlebars hates every fiber of my being](https://github.com/Kixiron/docs.rs/blob/7b009eb278b92f30ba0e4734ca119db8d4110620/src/web/page.rs#L120). I can't figure it out, so I'm hoping one of y'all can figure it out. I've enabled the `serde` feature for it and am feeding it a `serde` compatible struct, but it says that `Page<T>` doesn't implement `Serialize`, when it does.~~


Kixiron:
I hate everything about this, while updating I encountered a compile error in `staticfile`, so this PR is now upgrading iron *and* running `cargo update`

Kixiron:
I've followed the original json format, but I may have missed something, so whoever reviews make sure they're identical

Cargo.toml
```diff
@@ -41,12 +40,16 @@ rustwide = "0.6.0"
 tempdir = "0.3"
 mime_guess = "2"
 
+# json {de}serialization
+serde_json = "1.0.51"
+serde = { version = "1.0.106", features = ["derive"] }
```
jyn514, line 45: These exact version numbers are ignored by cargo. Please change them to the semver that cargo actually listens to:

```suggestion
serde_json = "1"
serde = { version = "1", features = ["derive"] }
```

src/web/builds.rs
```diff
@@ -7,6 +7,36 @@ use router::Router;
 use serde::{Serialize, Serializer};
 use time;
 
+/// A crate version's build
```
jyn514, line 10:
```suggestion
/// A build for a specific release of a crate
```

src/web/builds.rs
```diff
@@ -27,6 +57,58 @@ fn serde_relative_time<S: Serializer>(
     time::at(*ts).rfc3339().to_string().serialize(serializer)
 }
 
+/// The list of builds
```
jyn514, line 60:
```suggestion
/// All builds for a specific release of a crate.
///
/// In most cases, there will only be one build.
/// A release may have multiple builds if a server administrator decides to requeue a build.
/// Similarly, it may have no builds if the crate was removed from the queue before being built.
```

src/web/crate_details.rs
```diff
@@ -13,6 +13,97 @@ use time;
...
+///
+/// ### Rust:
+///
+/// ```ignore
+/// let details = CrateDetails {
+///     name: "some_crate",
+///     version: "0.0.0",
+///     description: None,
+///     authors: vec![("The robot overlords", "dalek49@docs.rs")],
+///     owners: vec![("The robot overlords", "dalek49@docs.rs")],
```
jyn514, line 26: I love these names

src/web/crate_details.rs
```diff
@@ -13,6 +13,97 @@ use time;
 
 // TODO: Add target name and versions
 
+/// The details of a Crate
```
jyn514, line 16:
```suggestion
/// The details of a Crate
///
/// These are shown both on the `/crate/:name` page and the dropdown on rustdoc pages.
```

src/web/crate_details.rs
```diff
@@ -22,11 +113,15 @@ pub struct CrateDetails {
...
     dependencies: Option<Value>,
+    // Is serialized as rendered markdown, but skipped if
+    // it's `None`
     #[serde(
         serialize_with = "serde_render_markdown",
         skip_serializing_if = "Option::is_none"
     )]
     readme: Option<String>,
+    // Is serialized as rendered markdown, but skipped if
+    // it's `None`
```
jyn514, line 124:
```suggestion
```

src/web/crate_details.rs
```diff
@@ -22,11 +113,15 @@ pub struct CrateDetails {
     owners: Vec<(String, String)>,
     authors_json: Option<Value>,
     dependencies: Option<Value>,
+    // Is serialized as rendered markdown, but skipped if
+    // it's `None`
```
jyn514, line 117: I think you can remove this comment, it's pretty clear from the annotation.
```suggestion
```

src/web/crate_details.rs
```diff
@@ -55,16 +150,39 @@ pub struct CrateDetails {
...
-    }
+    render_markdown(
+        md.as_ref()
+            .expect("If it's none serialization should have been skipped"),
+    )
+    .serialize(serializer)
 }
 
+/// A crate release
+///
```
jyn514, line 164:
```suggestion
/// A crate release
///
/// TODO: the `yanked` field is never updated, see https://github.com/rust-lang/docs.rs/pull/322
///
```

src/web/mod.rs
```diff
@@ -139,27 +139,42 @@ impl CratesfyiHandler {
...
+            debug!("Handling Request: {}", line);
+        }
+
         // try serving shared rustdoc resources first, then router, then db/static file handler
         // return 404 if none of them return Ok
         self.shared_resource_handler
             .handle(req)
-            .or_else(|e| self.router_handler.handle(req).or(Err(e)))
             .or_else(|e| {
+                debug!("Shared resource handler returned {:?}", &e);
```
jyn514, line 151: Please remove these, we have enough logging as is.

```suggestion
```

src/web/mod.rs
```diff
@@ -139,27 +139,42 @@ impl CratesfyiHandler {
...
+
+                self.router_handler.handle(req).or(Err(e))
+            })
+            .or_else(|e| {
+                debug!("Router handler returned {:?}", &e);
```
jyn514, line 156:
```suggestion
```

src/web/mod.rs
```diff
@@ -139,27 +139,42 @@ impl CratesfyiHandler {
...
+
                 // if router fails try to serve files from database first
                 self.database_file_handler.handle(req).or(Err(e))
             })
             .or_else(|e| {
+                debug!("Database file handler returned {:?}", &e);
```
jyn514, line 162:
```suggestion
```

src/web/mod.rs
```diff
@@ -139,27 +139,42 @@ impl CratesfyiHandler {
...
+
                 // and then try static handler. if all of them fails, return 404
                 self.static_handler.handle(req).or(Err(e))
             })
             .or_else(|e| {
-                let err = if let Some(err) = e.error.downcast_ref::<error::Nope>() {
+                debug!("Static handler returned {:?}", &e);
```
jyn514, line 168:
```suggestion
```

src/web/mod.rs
```diff
@@ -139,27 +139,42 @@ impl CratesfyiHandler {
...
+
+                let err = if let Some(err) = e.error.downcast::<error::Nope>() {
                     *err
-                } else if e.error.downcast_ref::<NoRoute>().is_some() {
+                } else if e.error.downcast::<NoRoute>().is_some() {
                     error::Nope::ResourceNotFound
                 } else {
                     panic!("all cratesfyi errors should be of type Nope");
                 };
+                debug!("Error finding route: {:?}", &err);
```
jyn514, line 177: 404s are already logged.

```suggestion
```

src/web/routes.rs
```diff
@@ -126,6 +126,8 @@ pub(super) fn build_routes() -> Routes {
         );
     }
 
+    log::trace!("Mounted Routes: {:?}", &routes);
```
jyn514, line 129:
```suggestion
```

src/web/routes.rs
```diff
@@ -159,13 +161,18 @@ impl Routes {
     pub(super) fn iron_router(mut self) -> Router {
         let mut router = Router::new();
         for (pattern, handler) in self.get.drain(..) {
+            debug!("Adding the route {}", &pattern);
```
jyn514, line 164:
```suggestion
```

src/web/routes.rs
```diff
@@ -159,13 +161,18 @@ impl Routes {
...
             router.get(&pattern, handler, calculate_id(&pattern));
         }
 
         // All rustdoc pages have the prefixes of other docs.rs pages blacklisted. This prevents,
         // for example, a crate named "about" from hijacking /about/0.1.0/index.html.
         let blacklist = self.page_prefixes();
+        debug!("Blacklisted patterns: {:?}", &blacklist);
```
jyn514, line 171:
```suggestion
```

src/web/routes.rs
```diff
@@ -159,13 +161,18 @@ impl Routes {
...
+
         for (pattern, handler) in self.rustdoc_get.drain(..) {
+            debug!("Adding the rustdoc route {}", &pattern);
```
jyn514, line 174:
```suggestion
```

src/web/builds.rs
```diff
@@ -7,6 +7,36 @@ use router::Router;
...
+/// ### Json
+///
+/// ```json
+/// {
+///     "id": 100,
+///     "rustc_version": "rustc 2.0.0-nightly (000000000 1970-01-01)",
+///     "cratesfyi_version": "0.0.0 (FFFFFF)",
+///     "build_status": true,
+///     "build_time": "just now",
+///     "build_time_relative": "2020-04-12T9:52:00Z",
```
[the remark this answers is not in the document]
Kixiron: You're completely correct, I swapped them, fixed in abf28f1

jyn514:
This needs to be rebased. Also, I'm not convinced that switching everything at once is the right way to go, see e.g. https://github.com/rust-lang/docs.rs/issues/459#issuecomment-574157795.

The added tests are great, would it be possible to add those in a separate PR? That would avoid blocking them on switching to serde.

Kixiron:
Update: Waiting on #743, and then am just going to rewrite this PR entirely

src/web/crate_details.rs
```diff
@@ -46,72 +46,61 @@ pub struct CrateDetails {
...
+
         if let Some(ref readme) = self.readme {
-            m.insert("readme".to_string(), render_markdown(&readme).to_json());
+            state.serialize_field("readme", &render_markdown(&readme))?;
         }
+
         if let Some(ref rustdoc) = self.rustdoc {
-            m.insert("rustdoc".to_string(), render_markdown(&rustdoc).to_json());
+            state.serialize_field("rustdoc", &render_markdown(&rustdoc))?;
         }
```
jyn514, line 70: It looks like `rustdoc` and `readme` will sometimes not be serialized. In that case, the `len` argument you pass to `serialize_struct` will be incorrect. Do you know what the purpose of that parameter is?

As an alternative, you could do the `if let Some` check ahead of time and increment/decrement the count if they are/are not present.

src/web/crate_details.rs
```diff
@@ -46,72 +46,61 @@ pub struct CrateDetails {
...
+        state.serialize_field("yanked", &self.yanked)?;
+        state.serialize_field("license", &self.license)?;
+        state.serialize_field("documentation_url", &self.documentation_url)?;
+
+        state.end()
     }
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, serde::Serialize)]
```
jyn514, line 97: nit: can you use `Serialize` instead? I think it looks nicer and serde is pretty clear from context.

src/web/crate_details.rs
```diff
@@ -587,4 +619,112 @@ mod tests {
...
+                "description": "serde does stuff",
+                "target_name": null,
+                "rustdoc_status": true,
+                "default_target": "x86_64-unknown-linux-gnu"
+            },
+            "is_library": true,
+            "doc_targets": [],
+            "license": null,
+            "documentation_url": null
+        });
```
jyn514, line 709: All this duplicate code makes me really uncomfortable, it also makes it really hard to read the diffs (both the commit diff and the changes from the previous `correct_json`). Does serde_json have some sort of record update syntax, the way rust has `struct S { field1: x, ..other_s }`?

src/web/page.rs
```diff
@@ -119,42 +111,145 @@ impl<T: ToJson> Page<T> {
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("Page", 10)?;
 
         if let Some(ref title) = self.title {
-            tree.insert("title".to_owned(), title.to_json());
+            state.serialize_field("title", title)?;
         }
```
jyn514, line 123: Same bug as above with `serialize_struct` and a hard-coded value.

src/web/page.rs
```diff
@@ -119,42 +111,145 @@ impl<T: ToJson> Page<T> {
...
+            }],
+            "varss": { "test": "works" },
+            "varsb": { "test2": true },
+            "varsi": { "test3": 1337 },
+            "rustc_resource_suffix": &*RUSTC_RESOURCE_SUFFIX,
+            "cratesfyi_version": crate::BUILD_VERSION,
+            "cratesfyi_version_safe": crate::BUILD_VERSION
+                .replace(" ", "-")
+                .replace("(", "")
+                .replace(")", ""),
```
jyn514, line 199: Can you add a test for this as an example? Maybe refactor it into a small function? I understand that `BUILD_VERSION` is not predictable from within the code, but you could hard-code the current version, `0.6.0 (44f6f29 2020-05-26)`.

jyn514: If there is a way to test whether the generated `BUILD_VERSION` is a valid URL (maybe using `Url::parse`?) that would also be great, since we use `build_version_safe` in several query parameters.

src/web/page.rs
```diff
@@ -119,42 +111,145 @@ impl<T: ToJson> Page<T> {
...
+        let alert = GlobalAlert {
+            url: "http://www.hasthelargehadroncolliderdestroyedtheworldyet.com/",
+            text: "THE WORLD IS ENDING",
+            css_class: "THE END IS NEAR",
+            fa_icon: "https://gph.is/1uOvmqR",
+        };
+
+        let correct_json = json!({
+            "url": "http://www.hasthelargehadroncolliderdestroyedtheworldyet.com/",
+            "text": "THE WORLD IS ENDING",
```
jyn514, line 248: This is clearly incorrect.
```suggestion
            text: "THE WORLD WILL SOON END",
            css_class: "THE END IS NEAR",
            fa_icon: "https://gph.is/1uOvmqR",
        };

        let correct_json = json!({
            "url": "http://www.hasthelargehadroncolliderdestroyedtheworldyet.com/",
            "text": "THE WORLD WILL SOON END",
```

src/web/releases.rs
```diff
@@ -1001,4 +1003,59 @@ mod tests {
...
+        };
+
+        let correct_json = json!({
+            "name": "serde",
+            "version": "0.0.0",
+            "description": "serde makes things other things",
+            "target_name": "x86_64-pc-windows-msvc",
+            "rustdoc_status": true,
+            "release_time": duration_to_str(time),
+            "release_time_rfc3339": time::at(time).rfc3339().to_string(),
```
jyn514, line 1027: `time::at(time)` is really confusing, can you use `now` for the local variable instead?

src/web/rustdoc.rs
```diff
@@ -30,19 +29,23 @@ struct RustdocPage {
...
+        state.serialize_field("rustdoc_body", &self.body)?;
+        state.serialize_field("rustdoc_body_class", &self.body_class)?;
+        state.serialize_field("rustdoc_full", &self.full)?;
+        state.serialize_field("rustdoc_status", &true)?;
+        state.serialize_field("name", &self.name)?;
+        state.serialize_field("version", &self.version)?;
+        state.serialize_field("description", &self.description)?;
+        state.serialize_field("crate_details", &self.crate_details)?;
+
+        state.end()
```
jyn514, line 48: This looks really similar to what the `derive` would output, the only difference is `&true`, right? I wonder if that would be better as an assert somehow.

src/web/source.rs
```diff
@@ -19,42 +21,94 @@ enum FileType {
...
+///     "files": [{
+///         "name": "main.rs",
+///         "file_type_rust_source": true,
+///     }, {
+///         "name": "lib.rs",
+///         "file_type_rust_source": true,
+///     }],
+/// }
+/// ```
+///
```
jyn514, line 80: Unless this is automatically updated on changes, I would rather just have the tests. Otherwise, it's too easy for it to get out of sync with the code. The tests are pretty self-documenting anyway.

src/web/source.rs
```diff
@@ -19,42 +21,94 @@ enum FileType {
...
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("FileList", 2)?;
+        state.serialize_field("metadata", &self.metadata)?;
 
+        let mut files = Vec::with_capacity(self.files.len());
         for file in &self.files {
-            let mut file_m: BTreeMap<String, Json> = BTreeMap::new();
-            file_m.insert("name".to_string(), file.name.to_json());
+            let mut map = HashMap::with_capacity(3);
```
jyn514, line 96: We're only serializing `name` and `file_type`.
```suggestion
            let mut map = HashMap::with_capacity(2);
```

src/web/crate_details.rs
Kixiron: This is a direct translation of what was, I was planning to cover making it sane in another PR

src/web/page.rs
Kixiron: Hard coded value?

src/web/page.rs
Kixiron: I donno, I think the first may be pretty accurate as of now

src/web/releases.rs
Kixiron: Can do, it's not ideal because `time` itself isn't ideal, but yet another PR

src/web/rustdoc.rs
Kixiron: As above, my goal was a direct translation with testing, leaving the groundwork for myself to come in and make everything sane without breaking it

src/web/source.rs
Kixiron: Oops, my bad

src/web/rustdoc.rs
jyn514: Good point, we can make things simpler in a follow-up PR.

src/web/crate_details.rs
jyn514: This is not a direct translation because the `len` parameter has no equivalent in `rustc_serialize`. We should make sure we're serializing correctly and not breaking any of `serde`'s internal invariants.

Now that I think of it, it would be great to have tests both with and without `readme` and `rustdoc`.

src/web/page.rs
jyn514: The `10` argument for `serialize_struct`.

src/web/crate_details.rs
Kixiron: Slightly misunderstood you but I've fixed it

src/web/releases.rs
```diff
@@ -1007,6 +1007,8 @@ mod tests {
     #[test]
     fn serialize_releases() {
         let time = time::get_time();
+        let now = time::at(time);
```
jyn514: I meant more like this:
```suggestion
        let now = time::at(time::get_time());
```
That way there's no shadowing of the `time` module by a `time` variable, which requires you to know that those are different namespaces (I didn't know that until I saw your code compile :P)

PR #712, opened 2020-04-12
attempt to look up crates with -/_ replaced before 404ing

### LINKED ISSUE #105: Underscore/Dash in crate name
Hi,

I think the following should be supported: `https://docs.rs/tokio_core`.

Today I have to write `https://docs.rs/tokio-core` to end up at `https://docs.rs/tokio-core/0.1.4/tokio_core/`

Nemo157:
- attempt to look up crates with -/_ replaced before 404ing
- Add tests for mismatched -/_ handling
- Normalize crate name in database and handle mixed -/_ separators

fixes #105
closes #348

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ features = [ "with-time", "with-rustc-serialize" ]
 once_cell = "1.2.0"
 kuchiki = "0.8"
 criterion = "0.3"
+rand = "0.7.3"
 
 [[bench]]
 name = "html5ever"
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -272,6 +272,29 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             // downgrade query
             "ALTER TABLE sandbox_overrides DROP COLUMN max_targets;"
         ),
+        migration!(
+            context,
+            // version
+            10,
+            // description
+            "Add function to normalize underscores in crate names",
+            // upgrade query
+            "
+                CREATE FUNCTION normalize_crate_name(VARCHAR)
+                RETURNS VARCHAR
+                AS $$
+                    SELECT LOWER(REPLACE($1, '_', '-'));
+                $$ LANGUAGE SQL;
+
+                CREATE UNIQUE INDEX crates_normalized_name_idx
+                    ON crates (normalize_crate_name(name));
+            ",
+            // downgrade query
+            "
+                DROP INDEX crates_normalized_name_idx;
+                DROP FUNCTION normalize_crate_name;
+            "
+        ),
     ];
 
     for migration in migrations {
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -4,31 +4,6 @@ use crate::error::Result as CratesfyiResult;
 use postgres::{transaction::Transaction, Connection, Error as PostgresError};
 use schemamama::{Migration, Migrator, Version};
 use schemamama_postgres::{PostgresAdapter, PostgresMigration};
-use std::borrow::Cow;
-
-#[derive(Copy, Clone)]
-enum ApplyMode {
-    Permanent,
-    #[cfg(test)]
-    Temporary,
-}
-
-#[derive(Copy, Clone)]
-struct MigrationContext {
-    apply_mode: ApplyMode,
-}
-
-impl MigrationContext {
-    fn format_query<'a>(&self, query: &'a str) -> Cow<'a, str> {
-        match self.apply_mode {
-            ApplyMode::Permanent => Cow::Borrowed(query),
-            #[cfg(test)]
-            ApplyMode::Temporary => {
-                Cow::Owned(query.replace("CREATE TABLE", "CREATE TEMPORARY TABLE"))
-            }
-        }
-    }
-}
 
 /// Creates a new PostgresMigration from upgrade and downgrade queries.
 /// Downgrade query should return database to previous state.
@@ -43,9 +18,7 @@ impl MigrationContext {
 /// ```
 macro_rules! migration {
     ($context:expr, $version:expr, $description:expr, $up:expr, $down:expr $(,)?) => {{
-        struct Amigration {
-            ctx: MigrationContext,
-        };
+        struct Amigration;
         impl Migration for Amigration {
             fn version(&self) -> Version {
                 $version
@@ -61,9 +34,7 @@ macro_rules! migration {
                     self.version(),
                     self.description()
                 );
-                transaction
-                    .batch_execute(&self.ctx.format_query($up))
-                    .map(|_| ())
+                transaction.batch_execute($up).map(|_| ())
             }
             fn down(&self, transaction: &Transaction) -> Result<(), PostgresError> {
                 info!(
@@ -71,35 +42,16 @@ macro_rules! migration {
                     self.version(),
                     self.description()
                 );
-                transaction
-                    .batch_execute(&self.ctx.format_query($down))
-                    .map(|_| ())
+                transaction.batch_execute($down).map(|_| ())
             }
         }
-        Box::new(Amigration { ctx: $context })
+        Box::new(Amigration)
     }};
 }
 
 pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<()> {
-    migrate_inner(version, conn, ApplyMode::Permanent)
-}
-
-#[cfg(test)]
-pub fn migrate_temporary(version: Option<Version>, conn: &Connection) -> CratesfyiResult<()> {
-    migrate_inner(version, conn, ApplyMode::Temporary)
-}
-
-fn migrate_inner(
-    version: Option<Version>,
-    conn: &Connection,
-    apply_mode: ApplyMode,
-) -> CratesfyiResult<()> {
-    let context = MigrationContext { apply_mode };
-
     conn.execute(
-        &context.format_query(
-            "CREATE TABLE IF NOT EXISTS database_versions (version BIGINT PRIMARY KEY);",
-        ),
+        "CREATE TABLE IF NOT EXISTS database_versions (version BIGINT PRIMARY KEY);",
         &[],
     )?;
     let adapter = PostgresAdapter::with_metadata_table(conn, "database_versions");
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -2,6 +2,7 @@ mod fakes;
 
 use crate::web::Server;
 use failure::Error;
+use log::error;
 use once_cell::unsync::OnceCell;
 use postgres::Connection;
 use reqwest::{Client, Method, RequestBuilder};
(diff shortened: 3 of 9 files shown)


src/db/migrate.rs
```diff
@@ -322,6 +323,32 @@ fn migrate_inner(
         ),
     ];
 
+    if let ApplyMode::Permanent = apply_mode {
+        migrations.push(migration!(
```
Nemo157, line 327: With the current migration setup running the tests will attempt to recreate this function and index on each test. There's probably some nicer way to handle this than just not registering the migration during testing, I'm open to suggestions.

jyn514: I'm confused why you don't want to register the function during testing? Isn't that part of what we want to test?

Nemo157: It attempts to create the function multiple times, failing because it already exists. I guess the issue with this setup now is that if you try and run the tests without migrating the database manually the function won't exist at all.

jyn514: Yes, I would like tests to be as automated as possible. Maybe you could use `CREATE OR REPLACE FUNCTION` instead?

src/db/migrate.rs
```diff
@@ -322,6 +323,32 @@ fn migrate_inner(
...
+                $$ LANGUAGE SQL;
+
+                CREATE UNIQUE INDEX crates_normalized_name_idx
+                    ON crates (normalize_crate_name(name));
+            ",
+            // downgrade query
+            "
+                DROP FUNCTION normalize_crate_name;
+                DROP INDEX crates_normalized_name_idx;
+            "
```
jyn514, line 348: I think the order of the downgrade query will have to be switched or DROP FUNCTION will give an error that the index depends on it.

src/db/migrate.rs
```diff
@@ -322,6 +323,32 @@ fn migrate_inner(
         ),
     ];
 
+    if let ApplyMode::Permanent = apply_mode {
+        migrations.push(migration!(
+            context,
+            // version
+            10,
+            // description
+            "",
```
jyn514, line 332:
```suggestion
            "Add function to normalize underscores in crate names",
```

src/web/rustdoc.rs
```diff
@@ -688,4 +700,90 @@ mod test {
...
+                "/dummy_mixed_separators/*",
+                "/dummy_mixed-separators/0.2.0/dummy_mixed_separators/",
+                web,
+            )?;
+            assert_redirect(
+                "/dummy-mixed-separators/0.1.0",
+                "/dummy_mixed-separators/0.1.0/dummy_mixed_separators/",
+                web,
+            )?;
+
```
jyn514, line 761: Can you also add a few tests for `/crate/dummy_mixed-separators`?

src/web/mod.rs
```diff
@@ -191,27 +191,45 @@ impl Handler for CratesfyiHandler {
     }
 }
 
+struct MatchVersion {
+    /// Represents the crate name that was found when attempting to load a crate release.
+    ///
+    /// `match_version` will attempt to match a provided crate name against similar crate names with
+    /// dashes (`-`) replaced with underscores (`_`) and vice versa. If
```
jyn514, line 198: Looks like this doc-comment got cut off.
```suggestion
    /// dashes (`-`) replaced with underscores (`_`) and vice versa.
```

src/web/mod.rs
```diff
@@ -238,29 +260,39 @@ fn match_version(conn: &Connection, name: &str, version: Option<&str>) -> MatchV
...
-            return MatchVersion::Exact(version.clone());
+            return Some(MatchVersion {
+                corrected_name,
+                version: MatchSemver::Exact(version.clone()),
+            });
         }
     }
 
     // Now try to match with semver
     let req_sem_ver = match VersionReq::parse(&req_version) {
```
jyn514, line 293:
```suggestion
    let req_sem_ver = VersionReq::parse(&req_version).ok()?;
```

src/web/mod.rs
```diff
@@ -272,7 +304,7 @@ fn match_version(conn: &Connection, name: &str, version: Option<&str>) -> MatchV
             // but check result just in case
             let version = match Version::parse(&version.0) {
```
jyn514, line 305:
```suggestion
            let version = Version::parse(&version.0).ok()?;
```

src/web/crate_details.rs
```diff
@@ -309,14 +309,14 @@ pub fn crate_details_handler(req: &mut Request) -> IronResult<Response> {
...
-        MatchVersion::Semver((version, _)) => {
+        Some(MatchSemver::Semver((version, _))) => {
             let url = ctry!(Url::parse(
                 &format!("{}/crate/{}/{}", redirect_base(req), name, version)[..]
             ));
 
             Ok(super::redirect(url))
         }
-        MatchVersion::None => Err(IronError::new(Nope::CrateNotFound, status::NotFound)),
+        None => Err(IronError::new(Nope::CrateNotFound, status::NotFound)),
```
jyn514, line 319: I wish `Try` were stable so we could implement it for `IronError` :(

src/web/mod.rs
```diff
@@ -238,29 +260,39 @@ fn match_version(conn: &Connection, name: &str, version: Option<&str>) -> MatchV
...
-        }
-        rows.iter().map(|row| (row.get(0), row.get(1))).collect()
+        let mut rows = rows.iter().peekable();
+
+        if let Some(row) = rows.peek() {
+            let db_name = row.get(0);
+            if db_name != name {
+                corrected_name = Some(db_name);
+            }
+        };
```
jyn514, line 276: This removes the `is_empty()` check, was that intentional? I think the behavior should be the same, but it would be best to add a test if you want to keep this change.

src/web/rustdoc.rs
Nemo157: Added in the other test which checks that `/crate/:crate` paths are _not_ affected by this change.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ router = "0.5"
 handlebars-iron = "0.22"
 params = "0.8"
 staticfile = { version = "0.4", features = [ "cache" ] }
+rand = "0.7.3"
```
jyn514, line 50: Please make this a dev-dependency. I know it doesn't make a difference in practice since it's a transitive dependency but it communicates intent more clearly.

src/db/migrate.rs
Nemo157: Changed the test setup to create a separate schema for each test run, so they're running on nearly fully independent databases.

src/test/mod.rs
```diff
@@ -138,6 +149,17 @@ impl TestDatabase {
     }
 }
 
+impl Drop for TestDatabase {
+    fn drop(&mut self) {
+        if let Err(e) = self
+            .conn()
+            .execute(&format!("DROP SCHEMA {} CASCADE;", self.schema), &[])
+        {
+            error!("failed to drop test schema {}: {}", self.schema, e);
```
jyn514, line 158: This will get lost in the test logs (docs.rs is kind of noisy). I guess a panic would fail the test for no reason though?

src/test/mod.rs
```diff
@@ -114,18 +115,28 @@ impl TestEnvironment {
...
 }
 
 impl TestDatabase {
     fn new() -> Result<Self, Error> {
-        // The temporary migration uses CREATE TEMPORARY TABLE instead of CREATE TABLE, creating
-        // fresh temporary copies of the database on top of the real one. The temporary tables are
-        // only visible to this connection, and will be deleted when it exits.
+        // A random schema name is generated and used for the current connection. This allows each
+        // test to create a fresh instance of the database to run within.
+        let schema = format!("test_schema_{}", rand::random::<u64>());
```
jyn514, line 125:
```suggestion
        let schema = format!("docs_rs_test_schema_{}", rand::random::<u64>());
```

src/web/mod.rs
Nemo157: It was like that in #348 too :shrug:, guess I could fix it.

src/web/mod.rs
Nemo157: Yes, there is very minimal extra work that would occur without the early exit, so it seems cleaner to just let it fall-through. I'll add a test.

src/test/mod.rs
Nemo157: A panic would cause the test-runner to abort if the current drop was from a panic, leaving any other test schemas still on the database server. I could make it check `std::thread::panicking` and panic if that wouldn't abort.

jyn514: Hmm good point. I think it's fine to leave it as is, I want to avoid making it unnecessarily complicated. If you see `docs_rs_test_schema` in a database you know where it's from anyway.

src/test/mod.rs
```diff
@@ -114,18 +115,28 @@ impl TestEnvironment {
...
+        // A random schema name is generated and used for the current connection. This allows each
+        // test to create a fresh instance of the database to run within.
+        let schema = format!("docs_rs_test_schema_{}", rand::random::<u64>());
+
         let conn = crate::db::connect_db()?;
-        crate::db::migrate_temporary(None, &conn)?;
+        conn.batch_execute(&format!(
+            "
+                CREATE SCHEMA {0};
+                SET search_path TO {0};
```
jyn514, line 131: I was curious whether there was a way for postgres to do this automatically - turns out there is!

```suggestion
                SET search_path TO pg_temp;
```

This says it's different for every connection, so I'm not sure if the tests will have difference schemas, but I think it's worth a shot :)

https://dba.stackexchange.com/a/76661

src/web/rustdoc.rs
```diff
@@ -688,4 +700,116 @@ mod test {
...
+                web.get("/crate/dummy_mixed_separators").send()?.status(),
+                StatusCode::NOT_FOUND
+            );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn nonexisting_crate_404s() {
```
jyn514, line 805:
```suggestion
    fn nonexistent_crate_404s() {
```

src/test/mod.rs
Nemo157: You cannot look up functions inside `pg_temp`. I spent quite a while hacking around with it to see if I could get it working, but as far as I can tell it's impossible to use functions inside it without using an explicit schema-name when calling them.

jyn514: If I add `pg_temp` as a prefix it seems to work:

```
cratesfyi=> set search_path to pg_temp;
SET
cratesfyi=>                 CREATE FUNCTION normalize_crate_name(VARCHAR)
cratesfyi->                 RETURNS VARCHAR
cratesfyi->                 AS $$
cratesfyi$>                     SELECT LOWER(REPLACE($1, '_', '-'));
cratesfyi$>                 $$ LANGUAGE SQL;
CREATE FUNCTION
cratesfyi=> select normalize_crate_name('test-one');
ERROR:  function normalize_crate_name(unknown) does not exist
LINE 1: select normalize_crate_name('test-one');
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
cratesfyi=> select pg_temp.normalize_crate_name('test-one');
 normalize_crate_name 
----------------------
 test-one
(1 row)

cratesfyi=> select pg_temp.normalize_crate_name('test_one');
 normalize_crate_name 
----------------------
 test-one
(1 row)
```

Nemo157: Yep, that's what I meant by using an explicit schema-name to call them. Which would mean having to change the sql run inside the app depending on whether it's in test mode or not.

jyn514: > Which would mean having to change the sql run inside the app depending on whether it's in test mode or not.

Ah, I see. Yeah @pietroalbini tried that a while back (I think in https://github.com/rust-lang/docs.rs/pull/556?) and it made it much more difficult to understand the code, so this looks fine.

jyn514:
This looks great! Thanks so much for working on this :)

PR #714, opened 2020-04-13
Automatic prioritization of crates

Kixiron:
* Creates the `crate_priorities` table in the database
  * Holds a [pattern](https://www.postgresql.org/docs/8.1/functions-matching.html) and a priority
  * If the requested crate name matches the pattern, the crate is given the associated priority for the build queue
* Added tests for getting priority and queuing crates
* Added docs for getting priority and queuing crates
* Can be extended in the future with quantity-based timeouts where users who submit `x` number of crates in `y` time get a temporary priority drop

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -317,27 +317,30 @@ pub fn main() {
                 priority,
             )
             .expect("Could not add crate to queue");
-        } else if let Some(matches) = matches.subcommand_matches("set") {
-            let pattern = matches
-                .value_of("PATTERN")
-                .expect("You must give a pattern to match with");
-            let priority = clap::value_t!(matches.value_of("PRIORITY"), i32)
-                .expect("You must give a priority for a crate");
-            let conn = connect_db().expect("Could not connect to the database");
+        } else if let Some(matches) = matches.subcommand_matches("default-priority") {
+            if let Some(matches) = matches.subcommand_matches("set") {
+                let pattern = matches
+                    .value_of("PATTERN")
+                    .expect("You must give a pattern to match with");
+                let priority = clap::value_t!(matches.value_of("PRIORITY"), i32)
+                    .expect("You must give a priority for a crate");
+                let conn = connect_db().expect("Could not connect to the database");
 
-            set_crate_priority(&conn, pattern, priority).expect("Could not set pattern's priority");
-        } else if let Some(matches) = matches.subcommand_matches("remove") {
-            let pattern = matches
-                .value_of("PATTERN")
-                .expect("You must give a pattern to remove");
-            let conn = connect_db().expect("Could not connect to the database");
+                set_crate_priority(&conn, pattern, priority)
+                    .expect("Could not set pattern's priority");
+            } else if let Some(matches) = matches.subcommand_matches("remove") {
+                let pattern = matches
+                    .value_of("PATTERN")
+                    .expect("You must give a pattern to remove");
+                let conn = connect_db().expect("Could not connect to the database");
 
-            if let Some(priority) =
-                remove_crate_priority(&conn, pattern).expect("Could not remove pattern's priority")
-            {
-                println!("Removed pattern with priority {}", priority);
-            } else {
-                println!("Pattern did not exist and so was not removed");
+                if let Some(priority) = remove_crate_priority(&conn, pattern)
+                    .expect("Could not remove pattern's priority")
+                {
+                    println!("Removed pattern with priority {}", priority);
+                } else {
+                    println!("Pattern did not exist and so was not removed");
+                }
             }
         }
     }
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -158,7 +158,7 @@ pub fn main() {
                     .arg(Arg::with_name("PATTERN")
                         .index(1)
                         .required(true)
-                        .help("See https://www.postgresql.org/docs/current/functions-matching.html"))))))
+                        .help("See https://www.postgresql.org/docs/current/functions-matching.html")))))
         .get_matches();
 
     if let Some(matches) = matches.subcommand_matches("build") {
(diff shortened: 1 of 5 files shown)


src/db/migrate.rs
```diff
@@ -295,6 +295,20 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
                 DROP FUNCTION normalize_crate_name;
             "
         ),
+        migration!(
+            context,
+            // version
+            11,
+            // description
+            "Allow crate priorities to be stored",
```
jyn514, line 303:
```suggestion
            "Allow crates to be given a different default priority",
```

src/db/migrate.rs
```diff
@@ -295,6 +295,20 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
...
+            // upgrade query
+            "CREATE TABLE crate_priorities (
+                pattern VARCHAR NOT NULL PRIMARY KEY,
```
jyn514, line 306: Just in case we decide to allow crate IDs in the future.

```suggestion
                pattern VARCHAR NOT NULL UNIQUE,
```

src/utils/queue.rs
```diff
@@ -3,6 +3,26 @@
 use crate::error::Result;
 use postgres::Connection;
 
+/// The default queue priority
+const DEFAULT_PRIORITY: i32 = 0;
```
jyn514, line 7: I think this is pretty clear from the variable and file name.

```suggestion
const DEFAULT_PRIORITY: i32 = 0;
```

src/utils/queue.rs
```diff
@@ -13,5 +33,169 @@ pub fn add_crate_to_queue(
...
+            "INSERT INTO crate_priorities (pattern, priority) VALUES ($1, $2)",
+            &[&pattern, &priority],
+        )?;
+
+        Ok(())
+    }
+
+    /// Remove a pattern from the priority table, returning the priority that it was associated with or `None`
+    /// if nothing was removed
+    pub fn remove_crate_priority(conn: &Connection, pattern: &str) -> Result<Option<i32>> {
```
jyn514, line 61: Maybe it would be nice to expose these in the binary by adding a `cratesfyi queue priority` subcommand? That would go in https://github.com/rust-lang/docs.rs/blob/master/src/bin/cratesfyi.rs.

Kixiron: Would you prefer I do that now or in a separate PR, I've been meaning to refactor `cratesfyi.rs`

jyn514: This should not be part of a refactor PR, I'd do that now.

src/bin/cratesfyi.rs
```diff
@@ -139,7 +139,26 @@ pub fn main() {
...
+                .subcommand(SubCommand::with_name("set")
+                    .about("Set all crates matching the given pattern to a priority level")
+                    .arg(Arg::with_name("PATTERN")
+                        .index(1)
+                        .required(true)
+                        .help("See https://www.postgresql.org/docs/current/functions-matching.html"))
+                    .arg(Arg::with_name("PRIORITY")
+                        .index(2)
+                        .required(true)
+                        .help("The priority to give crates matching PATTERN"))
```
jyn514, line 155: Without this `remove` shows up as a subcommand of `add`:
```
$ cargo run --quiet queue priority set -h
cratesfyi-queue-priority-set 
Set all crates matching the given pattern to a priority level

...

SUBCOMMANDS:
    help      Prints this message or the help of the given subcommand(s)
    remove    Remove the prioritization of crates by the given pattern
```

```suggestion
                        .help("The priority to give crates matching PATTERN")))
```

src/bin/cratesfyi.rs
```diff
@@ -298,6 +317,29 @@ pub fn main() {
...
+            let pattern = matches
+                .value_of("PATTERN")
+                .expect("You must give a pattern to remove");
+            let conn = connect_db().expect("Could not connect to the database");
+
+            if remove_crate_priority(&conn, pattern)
+                .expect("Could not remove pattern's priority")
+                .is_some()
+            {
+                println!("Removed pattern");
```
jyn514, line 339:
```suggestion
            if let Some(priority) = remove_crate_priority(&conn, pattern)
                .expect("Could not remove pattern's priority")
            {
                println!("Removed pattern with priority {}", priority);
```

src/bin/cratesfyi.rs
```diff
@@ -139,7 +139,26 @@ pub fn main() {
                     .short("p")
                     .long("priority")
                     .help("Priority of build (default: 5) (new crate builds get priority 0)")
-                    .takes_value(true))))
+                    .takes_value(true)))
+            .subcommand(SubCommand::with_name("priority")
```
pietroalbini, line 143: Nit: let's call the command "default-priority".

src/bin/cratesfyi.rs
```diff
@@ -298,6 +317,28 @@ pub fn main() {
                 priority,
             )
             .expect("Could not add crate to queue");
+        } else if let Some(matches) = matches.subcommand_matches("set") {
```
jyn514, line 320: This doesn't work - it checks for `set` in the `queue` namespace, not in the `default-priority` namespace. You'll need to do something like `if let Some(matches) = matches.subcommand_matches("default-priority") {` and then move this whole if let inside that block.

src/bin/cratesfyi.rs
```diff
@@ -298,6 +317,28 @@ pub fn main() {
...
+            let pattern = matches
+                .value_of("PATTERN")
+                .expect("You must give a pattern to match with");
+            let priority = clap::value_t!(matches.value_of("PRIORITY"), i32)
+                .expect("You must give a priority for a crate");
+            let conn = connect_db().expect("Could not connect to the database");
+
+            set_crate_priority(&conn, pattern, priority).expect("Could not set pattern's priority");
+        } else if let Some(matches) = matches.subcommand_matches("remove") {
```
jyn514, line 329: Same issue as above, this `remove` isn't in the `default-priority` namespace.

jyn514:
You can test the CLI changes by running _without_ the CRATESFYI_DATABASE_URL variable set. Currently, it does nothing; it should instead give an error.

jyn514:
Thanks for working on this!

PR #715, opened 2020-04-13
Attempt to retain current location when loading a different platform

### LINKED ISSUE #24: Platform switch should attempt to preserve current module
Being in [some module](https://docs.rs/libloading/0.3.0/x86_64-pc-windows-gnu/libloading/os/unix/index.html) and then attempting to switch to a different platform (try unknown-linux-gnu) throws you back to the root of documentation.

It would be nice if an attempt was made to not forget the module user is in when switching the platform.

A naive implementation strategy could be something along the lines of: Given an URI like `https://docs.rs/{crate}/{version}/{platform}/{...path}` try to load `https://docs.rs/{crate}/{version}/{new-platform}/{...path}` and _if_ that fails, load the crate root documentation.

Nemo157:
For each target platform, if the current location exists within that platform then generate a link directly to it. Otherwise link to the root of the crate in that platform like normal.

fixes #24

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -42,7 +42,7 @@ pub struct CrateDetails {
     github_issues: Option<i32>,
     pub(crate) metadata: MetaData,
     is_library: bool,
-    doc_targets: Option<Json>,
+    pub(crate) doc_targets: Option<Json>,
     license: Option<String>,
     documentation_url: Option<String>,
 }
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -42,7 +42,7 @@ pub struct CrateDetails {
     github_issues: Option<i32>,
     pub(crate) metadata: MetaData,
     is_library: bool,
-    doc_targets: Option<Json>,
+    pub(crate) doc_targets: Vec<String>,
     license: Option<String>,
     documentation_url: Option<String>,
 }
@@ -185,6 +185,18 @@ impl CrateDetails {
             default_target: rows.get(0).get(25),
         };
 
+        let doc_targets = {
+            let data: Json = rows.get(0).get(22);
+            data.as_array()
+                .map(|array| {
+                    array
+                        .iter()
+                        .filter_map(|item| item.as_string().map(|s| s.to_owned()))
+                        .collect()
+                })
+                .unwrap_or_else(Vec::new)
+        };
+
         let mut crate_details = CrateDetails {
             name: rows.get(0).get(2),
             version: rows.get(0).get(3),
@@ -211,7 +223,7 @@ impl CrateDetails {
             github_issues: rows.get(0).get(20),
             metadata,
             is_library: rows.get(0).get(21),
-            doc_targets: rows.get(0).get(22),
+            doc_targets,
             license: rows.get(0).get(23),
             documentation_url: rows.get(0).get(24),
         };
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -42,7 +42,7 @@ pub struct CrateDetails {
     github_issues: Option<i32>,
     pub(crate) metadata: MetaData,
     is_library: bool,
-    pub(crate) doc_targets: Option<Json>,
+    doc_targets: Option<Json>,
     license: Option<String>,
     documentation_url: Option<String>,
 }
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -381,11 +381,6 @@ pub fn target_redirect_handler(req: &mut Request) -> IronResult<Response> {
         file_path.remove(3);
         if file_path[3] == crate_details.metadata.default_target {
             file_path.remove(3);
-        } else if file_path[4] != crate_details.target_name {
-            // For non-default targets we only redirect to paths within the current crate
-            file_path.drain(4..);
-            file_path.push(&crate_details.target_name);
-            file_path.push("index.html");
         }
         if let Some(last) = file_path.last_mut() {
             if *last == "" {
@@ -1032,6 +1027,7 @@ mod test {
                 .rustdoc_file("dummy/index.html", b"some content")
                 .rustdoc_file("dummy/struct.Dummy.html", b"some other content")
                 .rustdoc_file("dummy/struct.DefaultOnly.html", b"some otter content")
+                .rustdoc_file("x86_64-pc-windows-msvc/settings.html", b"top-level items")
                 .rustdoc_file("x86_64-pc-windows-msvc/dummy/index.html", b"some content")
                 .rustdoc_file(
                     "x86_64-pc-windows-msvc/dummy/struct.Dummy.html",
@@ -1053,7 +1049,7 @@ mod test {
                 &[
                     (
                         "x86_64-pc-windows-msvc",
-                        "/dummy/0.4.0/x86_64-pc-windows-msvc/dummy/index.html",
+                        "/dummy/0.4.0/x86_64-pc-windows-msvc/settings.html",
                     ),
                     ("x86_64-unknown-linux-gnu", "/dummy/0.4.0/settings.html"),
                 ],
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -292,7 +292,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     let path_in_latest = if !is_latest_version {
         let mut latest_path = req_path.clone();
         latest_path[2] = &latest_version;
-        path_for_version(&latest_path, &crate_details.target_name, &conn)
+        path_for_version(&latest_path, &crate_details.doc_targets, &conn)
     } else {
         Default::default()
     };
@@ -302,7 +302,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         let mut inner_path = req_path.clone();
         // Drop the `rustdoc/:crate/:version[/:platform]` prefix
         inner_path.drain(..3);
-        if inner_path[0] != crate_details.target_name {
+        if inner_path.len() > 1 && crate_details.doc_targets.iter().any(|s| s == inner_path[0]) {
             inner_path.remove(0);
         }
         inner_path.join("/")
@@ -331,7 +331,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
 /// `rustdoc/crate/version[/platform]/module/[kind.name.html|index.html]`
 ///
 /// Returns a path that can be appended to `/crate/version/` to create a complete URL.
-fn path_for_version(req_path: &[&str], target_name: &str, conn: &Connection) -> String {
+fn path_for_version(req_path: &[&str], known_platforms: &[String], conn: &Connection) -> String {
     // Simple case: page exists in the latest version, so just change the version number
     if File::from_path(&conn, &req_path.join("/")).is_some() {
         // NOTE: this adds 'index.html' if it wasn't there before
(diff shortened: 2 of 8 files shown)


src/web/rustdoc.rs
```diff
@@ -301,21 +291,60 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+    // that target and generate a list of targets + paths to render.
+    if let Some(Json::Array(doc_targets)) = &crate_details.doc_targets {
+        content.platforms = doc_targets
+            .iter()
+            .filter_map(|target| {
+                if let Json::String(target) = target {
+                    // rustdoc/:crate/:version[/:current_target]/:path*
+                    let mut platform_path = req_path.clone();
+                    // rustdoc/:crate/:version/:target/:path*
+                    if platform_path[3] == crate_details.target_name {
```
jyn514, line 312: Ugh, the naming in crate_details is so confusing.
```suggestion
                    //  `target_name` is the name of the crate, not the name of the platform.
                    if platform_path[3] == crate_details.target_name {
```

src/web/rustdoc.rs
```diff
@@ -301,21 +291,60 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+                        platform_path.insert(3, target);
+                    } else {
+                        platform_path[3] = target;
+                    }
+                    if target == &crate_details.metadata.default_target {
+                        // rustdoc/:crate/:version/:path*
+                        platform_path.remove(3);
+                    }
+                    if File::from_path(&conn, &platform_path.join("/")).is_none() {
```
jyn514, line 321:
```suggestion
                    // if the current page is not present on the other platform, link to the index page for the crate instead
                    if File::from_path(&conn, &platform_path.join("/")).is_none() {
```

jyn514: Is there a reason you didn't reuse `path_for_version` for this? That would also handle searching for the page if it doesn't exist on the other platform.

That function should probably be renamed to `link_or_search` or similar, now that I think about it...

src/web/rustdoc.rs
```diff
@@ -812,4 +842,277 @@ mod test {
...
+                ),],
+            );
+
+            assert_eq!(
+                get_platform_links("/dummy/0.1.0/dummy/index.html", web)?,
+                vec![(
+                    "x86_64-unknown-linux-gnu".into(),
+                    "/dummy/0.1.0/dummy/index.html".into()
+                ),],
+            );
```
jyn514, line 898: Perhaps it would be helpful to write a test abstraction that does some of this boilerplate for you? Something like this:

```
fn test_platforms(default_target: Option<&str>, platforms: &[&str], pages: &[&str]) {
    let mut builder = db.fake_release().name("dummy").version("0.1.0");
    for page in pages {
        builder = builder.rustdoc_file(page, b"dummy content");
    }
    for platform in platforms {
        for page in pages {
            builder = builder.rustdoc_file(format!("{}/{}", platform, page), b"dummy content");
        }
        builder = builder.add_target(platform);
    }
    if let Some(default) = default_target {
        builder = builder.default_target(default);
    }
    builder.create()
}
```
That would shrink down the boilerplate, making it easier to read and write tests.

src/web/rustdoc.rs
Nemo157: I doubt that searching would be that useful, and IIRC it didn't fit this case exactly. But I think now that I understand how this is all working much better I might be able to refactor that to fit both usecases.

src/web/rustdoc.rs
Nemo157: The builder boilerplate doesn't seem too bad to me, it's the assertions that are really annoying. I can probably simplify them a _little_ with a custom assertion macro (one day `track(caller)` will be stable and we can start writing nice assertion functions).

src/web/rustdoc.rs
```diff
@@ -275,6 +263,8 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         }
     };
 
+    let req_path = req_path;
```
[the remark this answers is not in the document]
Nemo157: line 258 also modifies it. I'm tempted to try removing `req_path` from this function completely, there is probably a nicer way to handle it.

pietroalbini:
Hmm, querying S3 to check if the file exists on all built targets at every request sounds a bit expensive. Another approach we could take is to replace all target links with a URL like:

```
/crate/{name}/{version}/target-redirect/{target}/{path}
```

The new route will always issue 301 redirects to either:

* That path on the requested target, if it exists
* The root of the crate if the path doesn't exist on that target

Nemo157:
:+1: that sounds like a good way to go.

src/web/rustdoc.rs
jyn514: Hmm I feel like searching would still be useful. For example [winapi::winrt::winstring::HSTRING_UserFree64](https://docs.rs/winapi/0.3.8/winapi/winrt/winstring/fn.HSTRING_UserFree64.html) is [only present on `x86_64-pc-windows-msvc`](https://docs.rs/winapi/0.3.8/i686-pc-windows-msvc/winapi/winrt/winstring/fn.HSTRING_UserFree64.html), but if you search for it on `i686-pc-windows-msvc` [a similar function comes up](https://docs.rs/winapi/0.3.8/i686-pc-windows-msvc/winapi/winrt/winstring/fn.HSTRING_UserFree.html).

Nemo157: Ah, yeah, 32 vs 64 bit differences make sense; and I guess crates with types like `OsExt` might have them in platform specific locations but providing equivalent functionality.

src/web/rustdoc.rs
```diff
@@ -301,22 +289,37 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
     };
 
+    // The path within this crate version's rustdoc output
+    let inner_path = {
+        let mut inner_path = req_path.clone();
+        // Drop the `rustdoc/:crate/:version[/:platform]` prefix
+        inner_path.drain(..3);
+        // This could be a top-level file like rustdoc/:crate/:version/settings.html, hopefully
+        // rustdoc won't ever output files that are under directories other than the `target_name`
+        if inner_path.len() > 1 && inner_path[0] != crate_details.target_name {
```
jyn514, line 307: Can you change this to instead check against a whitelist of the targets, instead of assuming all directories are targets? @GuillaumeGomez said that rustdoc may occasionally output files that aren't under `/target_name`. You should be able to get the targets from `crate_details`.

src/web/rustdoc.rs
```diff
@@ -361,7 +364,57 @@ fn path_for_version(req_path: &[&str], target_name: &str, conn: &Connection) ->
     } else {
         req_path[3]
     };
-    format!("{}?search={}", crate_root, search_item)
+    format!("{}/?search={}", crate_root, search_item)
```
jyn514, line 367: This breaks https://github.com/rust-lang/docs.rs/issues/502.

src/web/rustdoc.rs
```diff
@@ -812,4 +863,297 @@ mod test {
...
+                .rustdoc_file(
+                    "x86_64-pc-windows-msvc/dummy/struct.WindowsOnly.html",
+                    b"some otter content",
+                )
+                .default_target("x86_64-unknown-linux-gnu")
+                .add_target("x86_64-pc-windows-msvc")
+                .create()?;
+
+            // For top-level items on non-default platforms we redirect to the target-specific doc
+            // root as the top-level items can't know which target they're for
```
Nemo157, line 1047: This is actually a bad assumption, `settings.html` is distributed for each target.

GuillaumeGomez: There is no need to have one `settings.html` per target. Per rustdoc version eventually but that's it.

src/web/rustdoc.rs
Nemo157: reverted

Nemo157:
Ok, latest change fixes redirecting to a search page under a non-default target, along with fixing the "Latest version" link on pages where you are on a non-default target and the latest version no longer has the current item, e.g. https://docs.rs/libc/0.1.5/x86_64-pc-windows-gnu/libc/funcs/extra/winsock/index.html

src/web/rustdoc.rs
jyn514: Hmm, maybe we should add it to `ESSENTIAL_FILES` with everything else that's shared between rustdoc versions?

jyn514: But either way it doesn't need to be addressed in this PR.

src/web/crate_details.rs
```diff
@@ -185,6 +185,18 @@ impl CrateDetails {
             default_target: rows.get(0).get(25),
         };
 
+        let doc_targets = {
+            let data: Json = rows.get(0).get(22);
```
jyn514, line 189: This will now panic if `doc_targets` is null, can you add a migration making it `NOT NULL`? I checked manually and there's currently no releases with a null entry.

src/web/rustdoc.rs
```diff
@@ -164,6 +161,7 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
         .unwrap_or_else(|_| crate_name.into())
         .into_owned();
     let req_version = router.find("version");
+    let mut target = router.find("target");
```
jyn514, line 164: I don't see a `:target` binding introduced in the routes. What is this meant to do?

src/web/rustdoc.rs
```diff
@@ -593,13 +652,14 @@ mod test {
             // check it searches for removed pages
             let redirect =
                 latest_version_redirect("/dummy/0.1.0/dummy/struct.will-be-deleted.html", &web)?;
-            assert_eq!(redirect, "/dummy/0.2.0/dummy?search=will-be-deleted");
+            // This must be a double redirect to deal with crates that failed to build in the
+            // latest version
+            assert_eq!(redirect, "/dummy/0.2.0/?search=will-be-deleted");
```
jyn514, line 657: Why did this redirect change?

jyn514:
Ok, so to summarize the changes I'm seeing:
- Refactor `crate_details.doc_targets` to be a `Vec<String>` instead of a JSON `Value`
- Add a `/crate/:name/:version/target-redirect/` page which redirects to the page on the new target if it exists, or searches for the page otherwise
- In the main redirect handler (`/:name/:version/*`), keep the target if it exists (I left a comment about this)
- Set an `inner_path` variable in the template so that it knows to redirect to `/crate/:name/:version/target-redirect/:inner_path`
- Change `path_for_version` to check against a whitelist of platforms instead of assuming anything that's not `target_name` is a platform.
- Clean up some test code/formatting

src/web/rustdoc.rs
Nemo157: This handles this route:

https://github.com/rust-lang/docs.rs/blob/554bb0e7f40d22db3728d42e019898c4a084ba97/src/web/routes.rs#L113-L116

Which is the one hit by non-default-target search paths.

src/web/rustdoc.rs
Nemo157: The extra `target_name` here is useless information that isn't needed for the redirect. https://github.com/rust-lang/docs.rs/pull/715/files#diff-db933256db133927163e75c554ffdfb6R197 will strip it, but it can just be not sent anyway.

src/web/rustdoc.rs
Nemo157: (The lack of checking this parameter is why non-default-target searches didn't work).

jyn514: Got it, thank you!

src/web/rustdoc.rs
```diff
@@ -188,16 +186,20 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
         (rows.get(0).get(0), rows.get(0).get(1))
     };
 
+    if target == Some("index.html") || target == Some(&target_name) {
+        target = None;
+    }
+
     if has_docs {
-        redirect_to_doc(req, &crate_name, &version, &target_name)
+        redirect_to_doc(req, &crate_name, &version, target, &target_name)
```
jyn514, line 202: At some point we should probably change this `target_name` to `crate_name` or similar ... but that can wait for another PR.

Nemo157: Maybe `library_name` since non-library-target docs aren't built. I was thinking about it a little and target is actually the correct nomenclature for this: [_Cargo Targets_](https://doc.rust-lang.org/nightly/cargo/reference/cargo-targets.html), it's just unfortunate that there is overlap between platform-target and library-target.

Nemo157:
Oops, turns out sometimes the list of platforms changes per version, so my fix to the non-default platforms wasn't 100%, "latest version" on https://docs.rs/libc/0.1.5/x86_64-pc-windows-gnu/libc/funcs/extra/winsock/index.html will redirect eventually to a 404 page, but at least it's working on https://docs.rs/libc/0.1.5/x86_64-pc-windows-msvc/libc/funcs/extra/winsock/index.html now.

jyn514:
I think it's enough of an edge case we can deal with it later, I'll open an issue.

Issue #156, opened 2017-09-20
Trailing whitespace lead to no matches


Boscop:
On mobile devices, auto complete often inserts a trailing space after selecting a word.
When searching for "cratename " on docs.rs, it doesn't find the crate named "cratename" which is frustrating...
It would be helpful if the search could be made more forgiving by e.g. trimming the trailing whitespace.

bluss:
I ran across this bug today. The message that is given for no matches is "The requested resource does not exist", which is also not quite clear, it gives the impression that there was an error, not just that no results matched the search.

[Test url for search 'ndarray '](https://docs.rs/releases/search?query=ndarray+)

jyn514:
Fixed in #721

Issue #13, opened 2016-06-30
Better search


ticki:
Upgrade the searching algorithm to something more flexible, like [Eudex](https://github.com/ticki/eudex).

onur:
Thanks for suggestion.

Did you see any issue in current search? I am using Postgres' [full text search](https://www.postgresql.org/docs/9.4/static/textsearch.html). It is fast and I believe I only need to adjust weights for ranking results better.

mjkillough:
It would be nice if exact matches appeared at the top. If I search for "serde" it is the 15th entry in the results list, which means it can take a little while to find.

ticki:
@mjkillough https://github.com/onur/docs.rs/issues/144

Kixiron:
Fixed by #721, crate searching is now well-ordered and more accurate

PR #721, opened 2020-04-17
Improve search functions

### LINKED ISSUE #489: docs.rs search is ordered randomly for non-exact matches

Kixiron:
Closes #489, #156 and #13

* Use fuzzy searching, pattern matching and description text searching to get better search matches
* Fuzzy search gives better results and gives accurate rankings
* Description searching allows you to search for concepts as well as crates, so `serialization` will also pull up results for `serde` (With any better name matches before it)
* Also trimmed the string, so the leading/trailing whitespace thing ain't a thing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,10 +58,6 @@ jobs:
           sleep 5
           # Make sure the database is actually working
           psql "${CRATESFYI_DATABASE_URL}"
-          psql "${CRATESFYI_DATABASE_URL}" -c "CREATE EXTENSION IF NOT EXISTS fuzzystrsearch"
-
-      - name: Build docs.rs
-        run: cargo build --locked
 
       - name: Run rustfmt
         run: cargo fmt -- --check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           sleep 5
           # Make sure the database is actually working
           psql "${CRATESFYI_DATABASE_URL}"
-          sudo -u postgres psql -c "CREATE EXTENSION IF NOT EXISTS fuzzystrsearch"
+          psql "${CRATESFYI_DATABASE_URL}" -c "CREATE EXTENSION IF NOT EXISTS fuzzystrsearch"
 
       - name: Build docs.rs
         run: cargo build --locked
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,4 @@
 ---
-
 on: [push, pull_request]
 name: CI
 
@@ -59,6 +58,10 @@ jobs:
           sleep 5
           # Make sure the database is actually working
           psql "${CRATESFYI_DATABASE_URL}"
+          sudo -u postgres psql -c "CREATE EXTENSION IF NOT EXISTS fuzzystrsearch"
+
+      - name: Build docs.rs
+        run: cargo build --locked
 
       - name: Run rustfmt
         run: cargo fmt -- --check
@@ -79,7 +82,6 @@ jobs:
     name: Docker
     runs-on: ubuntu-latest
     steps:
-
       - uses: actions/checkout@master
         with:
           fetch-depth: 2
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -330,19 +330,11 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             // version
             13,
             // description
-            "Add string searching",
+            "Add fuzzy string searching",
             // upgrade query
-            "DO $$ BEGIN
-                IF (SELECT COUNT(*) FROM pg_extension WHERE extname = 'fuzzystrmatch') = 0 THEN
-                    CREATE EXTENSION fuzzystrmatch;
-                END IF;
-            END $$;",
+            "CREATE EXTENSION IF NOT EXISTS fuzzystrmatch",
             // downgrade query
-            "DO $$ BEGIN
-                IF (SELECT COUNT(*) FROM pg_extension WHERE extname = 'fuzzystrmatch') > 0 THEN
-                    DROP EXTENSION fuzzystrmatch;
-                END IF;
-            END $$;",
+            "DROP EXTENSION IF EXISTS fuzzystrmatch;",
         ),
     ];
 
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -325,6 +325,25 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
                 ALTER TABLE releases ALTER COLUMN doc_targets DROP NOT NULL;
             "
         ),
+        migration!(
+            context,
+            // version
+            13,
+            // description
+            "Add string searching",
+            // upgrade query
+            "DO $$ BEGIN
+                IF (SELECT COUNT(*) FROM pg_extension WHERE extname = 'fuzzystrmatch') = 0 THEN
+                    CREATE EXTENSION fuzzystrmatch;
+                END IF;
+            END $$;",
+            // downgrade query
+            "DO $$ BEGIN
+                IF (SELECT COUNT(*) FROM pg_extension WHERE extname = 'fuzzystrmatch') > 0 THEN
+                    DROP EXTENSION fuzzystrmatch;
+                END IF;
+            END $$;",
+        ),
     ];
 
     for migration in migrations {
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -325,17 +325,6 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
                 ALTER TABLE releases ALTER COLUMN doc_targets DROP NOT NULL;
             "
         ),
-        migration!(
-            context,
-            // version
-            13,
-            // description
-            "Add fuzzy string searching",
-            // upgrade query
-            "CREATE EXTENSION IF NOT EXISTS fuzzystrmatch",
-            // downgrade query
-            "DROP EXTENSION IF EXISTS fuzzystrmatch;",
-        ),
     ];
 
     for migration in migrations {
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
     (author_name, packages)
 }
 
+/// Get the search results for a search query
+///
+/// Retrieves crates which names have a levenshtein distance of less than or equal to 3,
+/// crates who fit into or otherwise are made up of the query or crates who's descriptions
+/// match the search query.
+///
+/// * `query`: The query string, unfiltered
+/// * `page`: The page of results to show (1-indexed)
+/// * `limit`: The number of results to return
+///
+/// Returns `None` if no results are found and `Some` with the total number of results and the
+/// currently requested results
+///
 fn get_search_results(
     conn: &Connection,
-    query: &str,
+    mut query: &str,
     page: i64,
     limit: i64,
 ) -> Option<(i64, Vec<Release>)> {
+    query = query.trim();
+    let split_query = query.replace(' ', " & ");
     let offset = (page - 1) * limit;
-    let mut packages = Vec::new();
 
-    let rows = match conn.query(
-        "SELECT crates.name,
-                                    releases.version,
-                                    releases.description,
-                                    releases.target_name,
-                                    releases.release_time,
-                                    releases.rustdoc_status,
-                                    ts_rank_cd(crates.content, to_tsquery($1)) AS rank
-                                 FROM crates
-                                 INNER JOIN releases ON crates.latest_version_id = releases.id
-                                 WHERE crates.name LIKE concat('%', $1, '%')
-                                    OR crates.content @@ to_tsquery($1)
-                                 ORDER BY crates.name = $1 DESC,
-                                    crates.name LIKE concat('%', $1, '%') DESC,
-                                    rank DESC
-                                 LIMIT $2 OFFSET $3",
-        &[&query, &limit, &offset],
-    ) {
-        Ok(r) => r,
-        Err(_) => return None,
-    };
+    let rows = conn
+        .query(
+            "SELECT crates.name,
+                MAX(releases.version) AS version,
+                MAX(releases.description) AS description,
+                MAX(releases.target_name) AS target_name,
+                MAX(releases.release_time) AS release_time,
+                -- Cast the boolean into an integer and then cast it into a boolean.
+                -- Posgres moves in mysterious ways, don't question it
+                CAST(MAX(releases.rustdoc_status::integer) AS boolean) as rustdoc_status,
+                crates.github_stars,
+                SUM(releases.downloads) AS downloads,
 
-    for row in &rows {
-        let package = Release {
+                -- The levenshtein distance between the search query and the crate's name
+                levenshtein_less_equal($1, crates.name, 3) as distance,
+                -- The similarity of the tokens of the search vs the tokens of `crates.content`.
+                -- The `32` normalizes the number by using `rank / (rank + 1)`
+                ts_rank_cd(crates.content, to_tsquery($2), 32)  as content_rank
+            FROM releases INNER JOIN crates on releases.crate_id = crates.id
+
+            -- Filter crates that haven't been built and crates that have been yanked
+            WHERE releases.rustdoc_status = true
+                AND releases.yanked = false
+                AND (
+                    -- Crates names that match the query sandwiched between wildcards will pass
+                    crates.name ILIKE CONCAT('%', $1, '%')
+                    -- Crate names with which the levenshtein distance is closer or equal to 3 will pass
+                    OR levenshtein_less_equal($1, crates.name, 3) <= 3
+                    -- Crates where their content matches the query will pass
+                    OR to_tsquery($2) @@ crates.content
+                )
+            GROUP BY crates.id
+            -- Ordering is prioritized by how closely the query matches the name, how closely the
+            -- query matches the description finally how many downloads the crate has
+            ORDER BY distance DESC,
+                content_rank DESC,
+                SUM(downloads) DESC
+            -- Allows pagination
+            LIMIT $3 OFFSET $4",
+            &[&query, &split_query, &limit, &offset],
+        )
+        .ok()?;
+
+    let packages: Vec<Release> = rows
+        .into_iter()
+        .map(|row| Release {
             name: row.get(0),
             version: row.get(1),
             description: row.get(2),
             target_name: row.get(3),
             release_time: row.get(4),
             rustdoc_status: row.get(5),
-            ..Release::default()
-        };
-
-        packages.push(package);
-    }
+            stars: row.get(6),
+        })
+        .collect();
 
     if !packages.is_empty() {
-        // get count of total results
+        // Get the total number of results that the query matches
         let rows = conn
             .query(
-                "SELECT COUNT(*) FROM crates WHERE content @@ to_tsquery($1)",
-                &[&query],
+                "SELECT COUNT(*) FROM crates
+                WHERE crates.name ILIKE CONCAT('%', CAST($1 AS TEXT), '%')
+                    OR levenshtein_less_equal(CAST($1 AS TEXT), crates.name, 3) <= 3
+                    OR crates.content @@ to_tsquery(CAST($2 AS TEXT))",
+                &[&(query as &str), &(&split_query as &str)],
             )
             .unwrap();
 
@@ -570,17 +607,20 @@ pub fn search_handler(req: &mut Request) -> IronResult<Response> {
             }
         }
 
-        let search_query = query.replace(" ", " & ");
-        #[allow(clippy::or_fun_call)]
-        get_search_results(&conn, &search_query, 1, RELEASES_IN_RELEASES)
-            .ok_or_else(|| IronError::new(Nope::NoResults, status::NotFound))
-            .and_then(|(_, results)| {
-                // FIXME: There is no pagination
-                Page::new(results)
-                    .set("search_query", &query)
-                    .title(&format!("Search results for '{}'", query))
-                    .to_resp("releases")
-            })
+        if let Some((_, results)) = get_search_results(&conn, &query, 1, RELEASES_IN_RELEASES) {
+            // FIXME: There is no pagination
+            Page::new(results)
+                .set("search_query", &query)
+                .title(&format!("Search results for '{}'", query))
+                .to_resp("releases")
+        } else {
+            // Return an empty page with an error message and an intact query so that
+            // the user can edit it
+            Page::new("".to_string())
+                .set("search_query", &query)
+                .title(&format!("No results found for '{}'", query))
+                .to_resp("releases")
+        }
     } else {
         Err(IronError::new(Nope::NoResults, status::NotFound))
     }
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -265,7 +265,7 @@ fn get_releases_by_owner(
 ///
 fn get_search_results(
     conn: &Connection,
-    query: &str,
+    mut query: &str,
     page: i64,
     limit: i64,
 ) -> (i64, Vec<Release>) {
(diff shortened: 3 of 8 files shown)


src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
+///
 fn get_search_results(
     conn: &Connection,
-    query: &str,
+    mut query: &str,
     page: i64,
     limit: i64,
 ) -> Option<(i64, Vec<Release>)> {
+    query = query.trim();
+    let split_query = query.replace(' ', " & ");
```
jyn514, line 267: What does this do?

src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
             .query(
-                "SELECT COUNT(*) FROM crates WHERE content @@ to_tsquery($1)",
-                &[&query],
+                "SELECT COUNT(*) FROM crates
+                WHERE crates.name ILIKE CONCAT('%', CAST($1 AS TEXT), '%')
+                    OR levenshtein_less_equal(CAST($1 AS TEXT), crates.name, 3) <= 3
+                    OR crates.content @@ to_tsquery(CAST($2 AS TEXT))",
+                &[&(query as &str), &(&split_query as &str)],
             )
             .unwrap();
```
jyn514, line 336: This doesn't look right - it's redoing the query, but it already knows how many rows there are: it's `packages.len()`. Can you use that instead?

src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
+            "SELECT crates.name,
+                MAX(releases.version) AS version,
+                MAX(releases.description) AS description,
+                MAX(releases.target_name) AS target_name,
+                MAX(releases.release_time) AS release_time,
+                -- Cast the boolean into an integer and then cast it into a boolean.
+                -- Posgres moves in mysterious ways, don't question it
+                CAST(MAX(releases.rustdoc_status::integer) AS boolean) as rustdoc_status,
+                crates.github_stars,
+                SUM(releases.downloads) AS downloads,
```
jyn514, line 281:
```suggestion
                crates.downloads_total as downloads,
```

src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
+                    crates.name ILIKE CONCAT('%', $1, '%')
+                    -- Crate names with which the levenshtein distance is closer or equal to 3 will pass
+                    OR levenshtein_less_equal($1, crates.name, 3) <= 3
+                    -- Crates where their content matches the query will pass
+                    OR to_tsquery($2) @@ crates.content
+                )
+            GROUP BY crates.id
+            -- Ordering is prioritized by how closely the query matches the name, how closely the
+            -- query matches the description finally how many downloads the crate has
+            ORDER BY distance DESC,
```
jyn514, line 304:
```suggestion
            -- NOTE: this means that exact matches will be shown first
            ORDER BY distance DESC,
```

src/web/releases.rs
Kixiron: Ah, the first one trims leading/trailing whitespace and the second makes a query that's ready to be made into a `ts_query`

src/web/releases.rs
Kixiron: It's not, actually. This gets the *total* number of matching crates, not the number of returned crates. This is for pagination (eventually), e.g. `Showing 50 out of 500`

src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
+/// crates who fit into or otherwise are made up of the query or crates who's descriptions
+/// match the search query.
+///
+/// * `query`: The query string, unfiltered
+/// * `page`: The page of results to show (1-indexed)
+/// * `limit`: The number of results to return
+///
+/// Returns `None` if no results are found and `Some` with the total number of results and the
+/// currently requested results
+///
```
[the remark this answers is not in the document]
Kixiron: Well, in my opinion it's more ergonomic from a user perspective, and it also allows easier error handling within the function itself

src/db/migrate.rs
```diff
@@ -325,6 +325,25 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
...
+            // version
+            13,
+            // description
+            "Add string searching",
+            // upgrade query
+            "DO $$ BEGIN
+                IF (SELECT COUNT(*) FROM pg_extension WHERE extname = 'fuzzystrmatch') = 0 THEN
+                    CREATE EXTENSION fuzzystrmatch;
+                END IF;
+            END $$;",
```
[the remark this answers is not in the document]
Kixiron: After all my work learning about anonymous functions in postgres :(

src/web/releases.rs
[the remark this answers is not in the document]
jyn514: My concern is there are now two ways to represent 'no results': None and (0, vec![]). I'd rather there only be one.

src/web/releases.rs
jyn514: Ok, in that case this should be updated to match the other query and take `rustdoc_status` and `yanked` into account. Maybe we could turn that into a postgres function?

Nemo157: You can select `COUNT(*) OVER()` to get the total count ignoring limit in the first query.

src/web/releases.rs
Nemo157: Shouldn't it be `ASC`? The lower the distance the closer to the query it is.

src/web/releases.rs
```diff
@@ -244,57 +244,94 @@ fn get_releases_by_owner(
...
-        &[&query, &limit, &offset],
-    ) {
-        Ok(r) => r,
-        Err(_) => return None,
-    };
+    let rows = conn
+        .query(
+            "SELECT crates.name,
+                MAX(releases.version) AS version,
+                MAX(releases.description) AS description,
```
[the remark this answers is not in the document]
Nemo157: You can add a `WHERE` clause like

```sql
releases.id = (
  SELECT releases.id
  FROM releases
  WHERE
    releases.crate_id = crates.id
    AND releases.rustdoc_status
    AND NOT releases.yanked
  ORDER BY releases.version DESC
  LIMIT 1
)
```

and then add `releases.id` to the `GROUP BY` since there'll only be a distinct row. (That also takes care of filtering out crates that don't have successful builds since it'll return NULL and not match any rows).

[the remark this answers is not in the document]
Nemo157: Oh, you'll also have to change `SUM(downloads)` to `downloads_total` since there won't be all the releases to sum over anymore.

src/web/releases.rs
jyn514: Let's use `plainto_tsquery` instead which removes the need to normalize spaces and also gets rid of words like 'the'.

https://www.postgresql.org/docs/9.5/textsearch-controls.html

src/web/releases.rs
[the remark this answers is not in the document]
hakib: Hey, you can use the array thingy to get the latest value like this:

```sql
MAX(ARRAY([EXTRACT('EPOCH' FROM releases.release_time)::text, releases.target_name]))[1] AS target_name,
```

But that's really a nasty hack. Looks like what you really need is another table with the latest release for each crate:

```sql

SELECT 
    crates.name,
    latest_release.version AS version,
    latest_release.description AS description,
    latest_release.target_name AS target_name,
    latest_release.release_time AS release_time,
    latest_release.rustdoc_status AS rustdoc_status
    crates.github_stars,
    SUM(releases.downloads) AS downloads,
    -- The levenshtein distance between the search query and the crate's name
    levenshtein_less_equal($1, crates.name, 3) as distance,
    -- The similarity of the tokens of the search vs the tokens of `crates.content`.
    -- The `32` normalizes the number by using `rank / (rank + 1)`
    ts_rank_cd(crates.content, to_tsquery($2), 32)  as content_rank
FROM 
    crates 
    INNER JOIN releases.crate_id = crates.id
    INNER JOIN (
        SELECT DISTINCT ON (crate_id)
            crate_id,
            version,
            description,
            target_name,
            release_time,
            rustdoc_status
        FROM 
            releases
        ORDER BY
            create_id,
            release_time DESC
    ) AS latest_release ON latest_release.crate_id = crates.id
```
I haven't tested any of this and i'm not familiar with the tables. Also, it looks like you are missing a group by.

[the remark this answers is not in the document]
jyn514: I think that's a great idea! Shouldn't the inner query have `LIMIT 1` though since we only want the latest?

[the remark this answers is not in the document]
hakib: Notice the DISTINCT ON in the latest_release subquery. It will return only the first row for each crate. The order is determined by the order by, so it will return the latest. 

You can an example here:
https://hakibenita.com/the-many-faces-of-distinct-in-postgre-sql#distinct-on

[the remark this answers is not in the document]
Kixiron: Well, all modifications included, I've got this query
```sql
SELECT
    crates.name,
    latest_release.version AS version,
    latest_release.description AS description,
    latest_release.target_name AS target_name,
    latest_release.release_time AS release_time,
    latest_release.rustdoc_status AS rustdoc_status,
    crates.github_stars,
    SUM(releases.downloads) AS downloads,
    -- Get the total number of results, disregarding the limit
    COUNT(*) OVER() as total,

    -- The levenshtein distance between the search query and the crate's name
    levenshtein_less_equal(CAST($1 AS TEXT), CAST(crates.name AS TEXT), 3) as distance,
    -- The similarity of the tokens of the search vs the tokens of `crates.content`.
    -- The `32` normalizes the number by using `rank / (rank + 1)`
    ts_rank_cd(crates.content, plainto_tsquery($1), 32)  as content_rank
FROM
    crates
    INNER JOIN releases.crate_id = crates.id
    INNER JOIN (
        SELECT DISTINCT ON (crate_id)
            crate_id,
            version,
            description,
            target_name,
            release_time,
            rustdoc_status
        FROM
            releases
        ORDER BY
            create_id,
            release_time DESC
    ) AS latest_release ON latest_release.crate_id = crates.id

-- Filter crates that haven't been built and crates that have been yanked and
-- crates that don't match the query closely enough
WHERE
    latest_release.rustdoc_status
    AND NOT latest_release.yanked
    AND (
        -- Crates names that match the query sandwiched between wildcards will pass
        crates.name ILIKE CONCAT('%', $1, '%')
        -- Crate names with which the levenshtein distance is closer or equal to 3 will pass
        OR levenshtein_less_equal($1, crates.name, 3) <= 3
        -- Crates where their content matches the query will pass
        OR plainto_tsquery($1) @@ crates.content
    )

GROUP BY crates.id, releases.id

-- Ordering is prioritized by how closely the query matches the name, how closely the
-- query matches the description finally how many downloads the crate has
ORDER BY
    distance ASC,
    content_rank DESC,
    downloads_total DESC

-- Allows pagination
LIMIT $2 OFFSET $3
```
but it seems that the `INNER JOIN releases.crate_id = crates.id` is a syntax error?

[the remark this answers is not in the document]
jyn514:
```sql
INNER JOIN releases ON releases.crate_id = crates.id
```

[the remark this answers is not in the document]
hakib: You are also missing `latest_release.yanked`. add `yanked` to the `latest_release` subquery.

Kixiron:
#708 brings up a good point, does the actual date a crate was published matter when we have semver?

jyn514:
I don't know how to sort by semver in SQL. I think we should wait on that until later, we're making a big change already.

src/db/migrate.rs
```diff
@@ -325,6 +325,17 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
                 ALTER TABLE releases ALTER COLUMN doc_targets DROP NOT NULL;
             "
         ),
+        migration!(
+            context,
+            // version
+            13,
+            // description
+            "Add string searching",
```
jyn514, line 333:
```suggestion
            "Add fuzzy string searching",
```

src/web/releases.rs
```diff
@@ -676,3 +692,96 @@ pub fn build_queue_handler(req: &mut Request) -> IronResult<Response> {
...
+            let (num_results, results) = get_search_results(&db.conn(), "foo", 1, 100);
+            let mut results = results.into_iter();
+
+            assert_eq!(num_results, 4);
+
+            let expected = ["foo", "fo0", "bar-foo", "foo-bar"];
+            for expected in expected.iter() {
+                assert_eq!(expected, &results.next().unwrap().name);
+            }
+            assert!(results.collect::<Vec<_>>().is_empty());
```
jyn514, line 739:
```suggestion
            assert_eq!(results.count(), 0);
```

src/web/releases.rs
```diff
@@ -676,3 +692,96 @@ pub fn build_queue_handler(req: &mut Request) -> IronResult<Response> {
...
+        let near_matches = ["Regex", "rEgex", "reGex", "regEx", "regeX"];
+
+        for name in near_matches.iter() {
+            wrapper(|env| {
+                let db = env.db();
+                db.fake_release().name(name).version("0.0.0").create()?;
+
+                non_exact(&db)?;
+
+                let (num_results, results) = get_search_results(&db.conn(), "foo", 1, 100);
```
jyn514, line 775: I don't understand what this is testing. Shouldn't there be no results for "foo" because you haven't released any? Or does this depend on `database_search` running first?

src/web/releases.rs
```diff
@@ -254,64 +262,89 @@ fn get_releases_by_owner(
...
             -- Allows pagination
-            LIMIT $3 OFFSET $4",
-            &[&query, &split_query, &limit, &offset],
-        )
-        .ok()?;
+            LIMIT $2 OFFSET $3";
 
+    let rows = if let Ok(rows) = conn
+        .query(statement, &[&query, &limit, &offset])
+        .map_err(|err| dbg!(err))
```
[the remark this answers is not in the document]
Kixiron: Sorry, it's for debugging currently

src/web/releases.rs
Kixiron: That is a typo, my bad

src/web/releases.rs
```diff
@@ -254,64 +262,89 @@ fn get_releases_by_owner(
...
+    {
+        rows
+    } else {
+        return (0, Vec::new());
+    };
+
+    let total_results: i64 = rows.iter().next().map(|row| row.get(8)).unwrap_or_default();
```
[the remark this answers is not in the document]
Nemo157: No, every row gets the same total value, you should just take the first.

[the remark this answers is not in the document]
Nemo157: _Buuut_ you'll need to use a peekable iterator to read it from the first row without dropping it from the results.

src/web/releases.rs
```diff
@@ -273,75 +273,61 @@ fn get_search_results(
...
-            LIMIT $2 OFFSET $3";
+    let statement = "
+        SELECT
+            crates.name,
+            releases.version,
+            releases.description,
+            releases.target_name,
+            releases.release_time,
+            releases.rustdoc_status,
+            SUM(crates.github_stars),
```
jyn514, line 284: This doesn't look quite right, we only want the latest stars ... All the entries should be the same anyway since I don't think the stars are linked to a specific release.

src/web/releases.rs
```diff
@@ -273,75 +273,61 @@ fn get_search_results(
...
+            COUNT(*) OVER() as total
```
jyn514, line 285:
```suggestion
             -- Get the total number of results, disregarding the limit 
            COUNT(*) OVER() as total
```

src/web/releases.rs
```diff
@@ -273,75 +273,61 @@ fn get_search_results(
...
+                -- Filter unbuilt/failing builds and yanked releases
+                WHERE
+                    releases.crate_id = crates.id
+                    AND releases.rustdoc_status
+                    AND NOT releases.yanked
+                ORDER BY releases.release_time DESC
+                LIMIT 1
+            )
+            -- Only select releases/crates that pass our criteria:
+            --      - Levenshtein distance if the name and query is greater than three
```
jyn514, line 302:
```suggestion
            --      - Levenshtein distance between the name and query is less than three
```

src/web/releases.rs
```diff
@@ -351,7 +337,7 @@ fn get_search_results(
             target_name: row.get(3),
             release_time: row.get(4),
             rustdoc_status: row.get(5),
-            stars: row.get(6),
+            stars: row.get::<_, i64>(6) as i32,
```
jyn514, line 340: Why did this change? Do we store an i64 in postgres?

src/web/releases.rs
```diff
@@ -273,75 +273,61 @@ fn get_search_results(
...
+            AND (
+                levenshtein_less_equal($1, crates.name, 3) <= 3
+                OR crates.name ILIKE CONCAT('%', $1, '%')
+                OR plainto_tsquery($1) @@ to_tsvector(releases.description)
+            )
+        GROUP BY crates.id, releases.id
+        -- Order by the levenshtein distance of the name, the text search ranking of the description
+        -- and finally the number of downloads
+        ORDER BY
+            levenshtein_less_equal($1, crates.name, 3) ASC,
```
jyn514, line 314: I think this should also take into account the `ILIKE` so that e.g. `regex-syntax` will come before a crate with an unrelated name but related description.

src/web/releases.rs
```diff
@@ -770,15 +756,145 @@ mod tests {
...
+            assert_eq!(results.next().unwrap().name, "something_fantastical");
+            assert_eq!(
+                results.next().unwrap().name,
+                "something_completely_unrelated"
+            );
+            assert_eq!(results.count(), 0);
+
+            Ok(())
+        })
+    }
```
jyn514, line 899: Can you add tests for
- ordering by release date
- fuzzy name search taking precedence over description search
- unrelated crates not being returned
- github stars (optional)

jyn514:
Looks like you need to remove the migration to get the tests to pass: https://github.com/rust-lang/docs.rs/pull/721/checks?check_run_id=606111571#step:14:80

src/web/releases.rs
Kixiron: Yes, postgres has an `INT` while `stars` is an `i32`, without this there's a runtime error

src/web/releases.rs
Kixiron: Do you know of a way to do that?

jyn514: Same thing you're selecting by:
```suggestion
            levenshtein_less_equal($1, crates.name, 3) ASC,
            crates.name ILIKE CONCAT('%', $1, '%'),
```

src/web/releases.rs
Kixiron: I stand corrected, I was lied to

src/web/releases.rs
Kixiron: Didn't do github stars, as I don't sort by them

.github/workflows/ci.yml
```diff
@@ -56,6 +55,7 @@ jobs:
           sudo -u postgres createuser $(whoami) -w
           sudo -u postgres createdb $(whoami) -O $(whoami)
           echo "::set-env name=CRATESFYI_DATABASE_URL::postgresql://$(whoami)@%2Fvar%2Frun%2Fpostgresql/$(whoami)"
+          sudo -u postgres psql -c "CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;"
```
Nemo157, line 58: You probably need to specify `-d $(whoami)` here, otherwise it's going to create the extension on the `postgres` database.

jyn514: `-d cratesfyi`, not the current user.

Nemo157: The `createdb` line above uses `$(whoami)`, and the database url points to it.

.github/workflows/ci.yml
```diff
@@ -52,9 +52,9 @@ jobs:
         run: |
           sudo apt-get update && DEBIAN_FRONTEND=noninteractive sudo apt-get install -y postgresql
           sudo systemctl start postgresql
-          sudo -u postgres createuser cratesfyi -w
-          sudo -u postgres createdb cratesfyi -O cratesfyi
-          echo "::set-env name=CRATESFYI_DATABASE_URL::postgresql://cratesfyi@%2Fvar%2Frun%2Fpostgresql/cratesfyi"
+          sudo -u postgres createuser $(whoami) -w
+          sudo -u postgres createdb cratesfyi -O $(whoami)
+          echo "::set-env name=CRATESFYI_DATABASE_URL::postgresql://$(whoami)@%2Fvar%2Frun%2Fpostgresql/cratesfyi"
           sudo -u postgres psql -d cratesfyi -c "CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;"
```
jyn514, line 58:
```suggestion
          sudo -u postgres psql -d $(whoami) -c "CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;"
```

jyn514: What what why did the test pass D:

.github/workflows/ci.yml
Kixiron: Thank you so much!

.github/workflows/ci.yml
Kixiron: The database name doesn't have to be the user's name

Nemo157: The name of the database was changed to `cratesfyi` to make it a bit more consistent with the docker setup, but the user used to login was left as `$(whoami)` because it's using the local machine's user accounts to login.

src/web/releases.rs
```diff
@@ -244,64 +252,106 @@ fn get_releases_by_owner(
...
+                ORDER BY releases.release_time DESC
+                LIMIT 1
+            )
+            -- Only select releases/crates that pass our criteria:
+            --      - Levenshtein distance between the name and query is acceptable
+            --      - The query sandwiched between wildcards matches the crate's name
+            --      - The query matches the release's description
+            AND (
+                -- Turn the levenshtein distance into a percentage using `distance / max(query.len(), crates.name.len())`
+                -- this percentage is normalized and allows us to imperatively compare the 'sameness' of different names
```
jyn514, line 308:
```suggestion
                -- this percentage is normalized and allows us to empirically compare the 'sameness' of different names
```

src/web/releases.rs
```diff
@@ -244,64 +252,106 @@ fn get_releases_by_owner(
...
+                (((CAST(GREATEST(char_length($1), char_length(crates.name)) AS DOUBLE PRECISION) - CAST(levenshtein(CAST(crates.name AS TEXT), CAST($1 AS TEXT)) AS DOUBLE PRECISION))
```
jyn514, line 309: As discussed on discord, the casts make this hard to read and the logic is pretty confusing. I would rather do something like `levenshtein(crates.name, $1) < ROUND(LOG(2, char_length($1)))`.

Some sample cutoffs:

| input_length | cutoff |
|--------------|--------|
| 2            | 1      |
| 5            | 2      |
| 10           | 3      |
| 25           | 5      |

src/web/releases.rs
```diff
@@ -244,64 +252,106 @@ fn get_releases_by_owner(
...
+        GROUP BY crates.id, releases.id
+        -- Order by the levenshtein distance of the name, the text search ranking of the description
+        -- and finally the number of downloads
+        ORDER BY
+            -- Order the levenshtein matches by their literal distance, so that `fo` matches `foo` more closely than `fooo`,
+            -- because their normalized distances will be the same
+            levenshtein(CAST(crates.name AS TEXT), CAST($1 AS TEXT)) ASC,
+            crates.name ILIKE CONCAT('%', $1, '%'),
+            ts_rank_cd(to_tsvector(releases.description), plainto_tsquery($1), 32) DESC,
+            releases.downloads DESC
```
jyn514, line 323: Can you add a test for ordering by downloads?

jyn514:
You need to rebase over the changes to CI. Additionally, you may need to write an extra (small) dockerfile to add the fuzzy-search extension to postgres, see https://stackoverflow.com/a/54630526/7669110 for an example of what that would look like.

Past that, the current blocker is the performance of the query. The old query took less than half a second, this one takes almost 3. We've been working on this in Discord.

Kixiron:
Got a much-improved query in, had to remove description searching, as with it the query took 700-800ms, while without it the query takes a consistent 300ms

jyn514:
:tada: :tada: :tada:

PR #724, opened 2020-04-20
Test macOS & Windows platforms and Beta & Nightly channels

XAMPPRocky:
supercedes  / closes #621 

r? @jyn514

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,9 @@ jobs:
         run: choco install postgresql
       - name: Install PostgreSQL (macOS)
         if: contains(matrix.os, 'macos')
-        run: brew install postgresql
+        run: brew services start postgresql
       - name: Install PostgreSQL (Linux)
-        if: contains(matrix.os, 'linux')
+        if: contains(matrix.os, 'ubuntu')
         run: |
           sudo apt-get update && DEBIAN_FRONTEND=noninteractive sudo apt-get install -y postgresql
           sudo systemctl start postgresql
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,18 +9,21 @@ env:
...
+      matrix:
+        os: [ubuntu, macos, windows]
+        channel: [stable, beta, nightly]
 
     steps:
       - uses: actions/checkout@master
         with:
           fetch-depth: 2
 
       - name: Install stable Rust
```
jyn514, line 25:
```suggestion
      - name: Install Rust
```

.github/workflows/postgres.yml
```diff
@@ -0,0 +1,35 @@
...
+    - master
+
+jobs:
+  container-job:
+    runs-on: ubuntu-latest
+    
+    # runs all of the steps inside the specified container rather than on the VM host.  
+    # Because of this the network configuration changes from host based network to a container network.
+    container:
+      image:  node:10.16-jessie
```
jyn514, line 18: Why node? I don't think we use any javascript on the backend.

.github/workflows/postgres.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    services:
+      postgres:
+        image: postgres:10.8
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
```
jyn514, line 26: Would it be easier to just set the user/password/database in the image directly, instead of creating both a primary and secondary user?

XAMPPRocky:
I'm going to close this, as getting postgresql to actually work as a service on macOS and Windows is too much for me to investigate and work on currently.

jyn514:
Ok, thanks for taking a look!

PR #731, opened 2020-04-25
Use a docker image for testing on CI

jyn514:
This makes it easier to add custom extensions to the database (e.g. for #721). It also makes sure that we're testing the same things that contributors are running. Finally, it makes it easier to test on different platforms (i.e. #621).

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -7,6 +7,7 @@ pub use self::delete_crate::delete_crate;
 pub use self::file::{add_path_into_database, move_to_s3};
 pub use self::migrate::migrate;
 
+use failure::Fail;
 use postgres::error::Error;
 use postgres::{Connection, TlsMode};
 use std::env;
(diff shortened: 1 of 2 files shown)


src/db/mod.rs
```diff
@@ -18,11 +18,10 @@ pub(crate) mod file;
 mod migrate;
 
 /// Connects to database
-pub fn connect_db() -> Result<Connection, Error> {
+pub fn connect_db() -> Result<Connection, failure::Error> {
     // FIXME: unwrap might not be the best here
```
Nemo157, line 22: This comment is outdated now.

src/db/mod.rs
```diff
@@ -18,11 +18,10 @@ pub(crate) mod file;
...
-    let db_url = env::var("CRATESFYI_DATABASE_URL")
-        .expect("CRATESFYI_DATABASE_URL environment variable is not exists");
```
Nemo157, line 24: It'd be good to keep this message via `.context` or something, [`VarError::NotPresent`](https://doc.rust-lang.org/nightly/std/env/enum.VarError.html#variant.NotPresent) doesn't carry the variable name or anything useful.

PR #734, opened 2020-04-26
[WIP] Test on more platforms and channels

jyn514:
Essentially the same as https://github.com/rust-lang/docs.rs/pull/724, but much simpler now that postgres is containerized. Also, this only runs `cargo check` on channels besides `stable`.

This is currently failing because `docker-compose` isn't installed on windows and mac, not sure how to do that.

r? @XAMPPRocky


XAMPPRocky:
> This is currently failing because docker-compose isn't installed on windows and mac, not sure how to do that.

Yeah, you can't this is point I reached in the old PR. You can't run containers on the macOS or Windows runners. You can see it in this note in the documentation. https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers

PR #736, opened 2020-04-28
rustdoc_html_server_handler match crate name with -/_ replacement

Folyd:
I just made some minimal changes to fix this. Hope this PR is ok to be merged.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -875,17 +875,16 @@ mod test {
             assert_success("/dummy-dash/0.1.0/dummy_dash/index.html", web)?;
             assert_success("/crate/dummy_mixed-separators", web)?;
 
-            assert_eq!(
-                web.get("/dummy_dash/0.1.0/dummy_dash/index.html")
-                    .send()?
-                    .status(),
-                StatusCode::NOT_FOUND
-            );
-
-            assert_eq!(
-                web.get("/crate/dummy_mixed_separators").send()?.status(),
-                StatusCode::NOT_FOUND
-            );
+            assert_redirect(
+                "/dummy_dash/0.1.0/dummy_dash/index.html",
+                "/dummy-dash/0.1.0/dummy_dash/index.html",
+                web,
+            )?;
+            assert_redirect(
+                "/crate/dummy_mixed_separators",
+                "/crate/dummy_mixed-separators",
+                web,
+            )?;
 
             Ok(())
         })
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -210,7 +210,7 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
 /// also crate-specific.
 pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     let router = extension!(req, Router);
-    let mut name = router.find("crate").unwrap_or("").to_string();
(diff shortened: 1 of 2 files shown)


jyn514:
Thank you for the PR! Unfortunately this does not look like the right approach. See for example https://github.com/rust-lang/docs.rs/pull/736/checks?check_run_id=626374777#step:13:1756, which shows an infinite redirect for `/dummy/0.1.0/dummy/` - the problem is that you changed `/:crate/:version/:target/` to use the redirect handler instead of the HTML handler, but `/dummy/0.1.0/dummy/` is a valid page that doesn't need a redirect. Try removing that change and see if it helps.

As a side note, let me know if you're having trouble running the tests locally and I can try to help. I find discord is better for those conversations than github: https://discord.gg/f7mTXPW

src/web/rustdoc.rs
```diff
@@ -223,20 +223,34 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
-        _ => return Err(IronError::new(Nope::ResourceNotFound, status::NotFound)),
+        None => return Err(IronError::new(Nope::ResourceNotFound, status::NotFound)),
     };
 
+    if let Some(name) = corrected_name {
+        let url = ctry!(Url::parse(
+            &format!("{}/{}/{}/{}", base, name, version, req_path.join("/"))[..]
+        ));
+        return Ok(super::redirect(url));
+    }
```
jyn514, line 252: Hmm, it might be worth making a helper function for this, I think we use it in a couple other places as well.

jyn514:
Looks good to me modulo comment :) Thanks for the hard work on this!

src/web/rustdoc.rs
```diff
@@ -223,6 +222,13 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         req_path.remove(0);
     }
 
+    let redirect = |name: &str, vers: &str, path: &str| -> IronResult<Response> {
```
jyn514, line 225: nit: I think this would be better if it took a slice and called `.join()` here. It's not a blocker though.
```suggestion
    let redirect = |name: &str, vers: &str, path: &[str]| -> IronResult<Response> {
```

src/web/rustdoc.rs
```diff
@@ -223,20 +222,41 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+    let redirect = |name: &str, vers: &str, path: &[&str]| -> IronResult<Response> {
+        let url = ctry!(Url::parse(
+            &format!(
+                "{}/{}/{}/{}",
+                redirect_base(req),
+                name,
+                vers,
+                path.join("/")
+            )[..]
+        ));
```
jyn514, line 234: This looks pretty ugly to me ... I assume that rustfmt wanted it this way? Can you try something like this?
```suggestion
        let redirect_path = format!("{}/{}/{}/{}", redirect_base(req), name, vers, path.join("/"));
        let url = ctry!(Url::parse(&redirect_path[..]));
```

(sorry for the continuing nits)

Folyd: Yes, this is a cargo fmt result, otherwise, we couldn't pass the `cargo fmt` step in CI. 😂

jyn514: Hmm I guess it's fine, it's not unreadable, just ugly.

jyn514:
Thank you again for working on this!

PR #738, opened 2020-04-30
Added more metrics

Kixiron:
* Added the response time metric to show how long requests are taking
    * Shows when the server is flagging or struggling
* Added the traffic metric to show what routes are being visited
    * Shows what routes need to be further maintained
    * Shows peak and low traffic times so that maintenance can occur on downtimes

<details>
<summary>New metrics output</summary>

```
# HELP docsrs_failed_builds Number of builds that generated a compile error
# TYPE docsrs_failed_builds counter
docsrs_failed_builds 0
# HELP docsrs_failed_crates_count Number of crates that failed to build
# TYPE docsrs_failed_crates_count gauge
docsrs_failed_crates_count 0
# HELP docsrs_non_library_builds Number of builds that did not complete due to not being a library
# TYPE docsrs_non_library_builds counter
docsrs_non_library_builds 0
# HELP docsrs_queued_crates_count Number of crates in the build queue
# TYPE docsrs_queued_crates_count gauge
docsrs_queued_crates_count 0
# HELP docsrs_response_time The response times of various docs.rs routes
# TYPE docsrs_response_time histogram
docsrs_response_time_bucket{route="not_found",le="0.005"} 0
docsrs_response_time_bucket{route="not_found",le="0.01"} 0
docsrs_response_time_bucket{route="not_found",le="0.025"} 2
docsrs_response_time_bucket{route="not_found",le="0.05"} 2
docsrs_response_time_bucket{route="not_found",le="0.1"} 2
docsrs_response_time_bucket{route="not_found",le="0.25"} 2
docsrs_response_time_bucket{route="not_found",le="0.5"} 2
docsrs_response_time_bucket{route="not_found",le="1"} 2
docsrs_response_time_bucket{route="not_found",le="2.5"} 2
docsrs_response_time_bucket{route="not_found",le="5"} 2
docsrs_response_time_bucket{route="not_found",le="10"} 2
docsrs_response_time_bucket{route="not_found",le="+Inf"} 2
docsrs_response_time_sum{route="not_found"} 0.043141413
docsrs_response_time_count{route="not_found"} 2
docsrs_response_time_bucket{route="router",le="0.005"} 29
docsrs_response_time_bucket{route="router",le="0.01"} 30
docsrs_response_time_bucket{route="router",le="0.025"} 30
docsrs_response_time_bucket{route="router",le="0.05"} 31
docsrs_response_time_bucket{route="router",le="0.1"} 31
docsrs_response_time_bucket{route="router",le="0.25"} 35
docsrs_response_time_bucket{route="router",le="0.5"} 35
docsrs_response_time_bucket{route="router",le="1"} 35
docsrs_response_time_bucket{route="router",le="2.5"} 35
docsrs_response_time_bucket{route="router",le="5"} 35
docsrs_response_time_bucket{route="router",le="10"} 35
docsrs_response_time_bucket{route="router",le="+Inf"} 35
docsrs_response_time_sum{route="router"} 0.9536012539999996
docsrs_response_time_count{route="router"} 35
docsrs_response_time_bucket{route="statics",le="0.005"} 0
docsrs_response_time_bucket{route="statics",le="0.01"} 5
docsrs_response_time_bucket{route="statics",le="0.025"} 5
docsrs_response_time_bucket{route="statics",le="0.05"} 5
docsrs_response_time_bucket{route="statics",le="0.1"} 5
docsrs_response_time_bucket{route="statics",le="0.25"} 5
docsrs_response_time_bucket{route="statics",le="0.5"} 5
docsrs_response_time_bucket{route="statics",le="1"} 5
docsrs_response_time_bucket{route="statics",le="2.5"} 5
docsrs_response_time_bucket{route="statics",le="5"} 5
docsrs_response_time_bucket{route="statics",le="10"} 5
docsrs_response_time_bucket{route="statics",le="+Inf"} 5
docsrs_response_time_sum{route="statics"} 0.040871595999999996
docsrs_response_time_count{route="statics"} 5
# HELP docsrs_routes_visited The traffic of various docs.rs routes
# TYPE docsrs_routes_visited counter
docsrs_routes_visited{route="not_found"} 2
docsrs_routes_visited{route="router"} 35
docsrs_routes_visited{route="statics"} 5
# HELP docsrs_successful_builds Number of builds that successfully generated docs
# TYPE docsrs_successful_builds counter
docsrs_successful_builds 0
# HELP docsrs_total_builds Number of crates built
# TYPE docsrs_total_builds counter
docsrs_total_builds 0
# HELP docsrs_uploaded_files_total Number of files uploaded to S3 or stored in the database
# TYPE docsrs_uploaded_files_total counter
docsrs_uploaded_files_total 0
```

</details>

Route names aren't as accurate as I'd like and I'm not entirely sure what do do with tracking what comes from `CratesfyiHandler.pool_factory` accurately, but the base is there. Any more suggestions for additional metrics would also be helpful.
@pietroalbini Mentioned that he wanted a to have the rss feed be included in the new metrics, but I couldn't find where that happens.
The histogram (Used by `docsrs_response_time`) seems to be giving weird output to me, but it could also be correct for prometheus


src/web/metrics.rs
```diff
@@ -4,45 +4,67 @@ use iron::prelude::*;
 use iron::status::Status;
 use prometheus::{
     opts, register_counter, register_int_counter, register_int_gauge, Encoder, IntCounter,
-    IntGauge, TextEncoder, __register_gauge,
+    IntGauge, TextEncoder, __register_gauge, register_int_counter_vec, IntCounterVec,
+    __register_counter_vec, histogram_opts, register_histogram_vec, HistogramVec,
```
jyn514, line 8: I think I commented on this in another PR and you said prometheus isn't very careful about it's imports in macros - do you know if that would be fixed in the latest release? We have 0.7 and the latest is 0.8.

src/web/mod.rs
```diff
@@ -141,18 +141,35 @@ impl CratesfyiHandler {
...
         // try serving shared rustdoc resources first, then router, then db/static file handler
         // return 404 if none of them return Ok
         self.shared_resource_handler
             .handle(req)
-            .or_else(|e| self.router_handler.handle(req).or(Err(e)))
+            .or_else(|e| {
+                let router = self.router_handler.handle(req).or(Err(e));
+
+                response_time.route("router");
+                response_route.route("router");
```
jyn514, line 155: This looks a little odd to me ... could you instead make `response_time` a field of `response_route` and have `response_route` call `response_time.route()` internally? That would prevent people from calling one but not the other by accident.

src/web/mod.rs
```diff
@@ -141,18 +141,35 @@ impl CratesfyiHandler {
...
+                router
+            })
             .or_else(|e| {
                 // if router fails try to serve files from database first
-                self.database_file_handler.handle(req).or(Err(e))
+                let rustdoc = self.database_file_handler.handle(req).or(Err(e));
+
+                response_time.route("docs");
+                response_route.route("docs");
```
jyn514, line 163: Just having 'docs' isn't super useful, could you use `DebugPath(&req.url()).to_string()` here instead? If you want to have the metadata that it's a doc path maybe you could add that as a separate counter.

src/web/mod.rs
```diff
@@ -141,18 +141,35 @@ impl CratesfyiHandler {
...
+                rustdoc
             })
             .or_else(|e| {
                 // and then try static handler. if all of them fails, return 404
-                self.static_handler.handle(req).or(Err(e))
+                let static_file = self.static_handler.handle(req).or(Err(e));
+
+                response_time.route("statics");
+                response_route.route("statics");
```
jyn514, line 171: Ditto `DebugPath`

src/web/mod.rs
```diff
@@ -141,18 +141,35 @@ impl CratesfyiHandler {
...
+
         // try serving shared rustdoc resources first, then router, then db/static file handler
         // return 404 if none of them return Ok
         self.shared_resource_handler
             .handle(req)
-            .or_else(|e| self.router_handler.handle(req).or(Err(e)))
+            .or_else(|e| {
+                let router = self.router_handler.handle(req).or(Err(e));
+
+                response_time.route("router");
```
jyn514, line 154: Ditto `DebugPath`

jyn514:
The old output for `/about/metrics` looks like this:
```
# HELP docsrs_failed_builds Number of builds that generated a compile error
# TYPE docsrs_failed_builds counter
docsrs_failed_builds 0
# HELP docsrs_failed_crates_count Number of crates that failed to build
# TYPE docsrs_failed_crates_count gauge
docsrs_failed_crates_count 0
# HELP docsrs_non_library_builds Number of builds that did not complete due to not being a library
# TYPE docsrs_non_library_builds counter
docsrs_non_library_builds 0
# HELP docsrs_queued_crates_count Number of crates in the build queue
# TYPE docsrs_queued_crates_count gauge
docsrs_queued_crates_count 1
# HELP docsrs_successful_builds Number of builds that successfully generated docs
# TYPE docsrs_successful_builds counter
docsrs_successful_builds 0
# HELP docsrs_total_builds Number of crates built
# TYPE docsrs_total_builds counter
docsrs_total_builds 0
# HELP docsrs_uploaded_files_total Number of files uploaded to S3 or stored in the database
# TYPE docsrs_uploaded_files_total counter
docsrs_uploaded_files_total 0
```

src/web/mod.rs
Nemo157: You don't want to split the metrics up too finely, aggregating all paths that result in returning a file from S3 makes sense to me. One way to improve this could be to actually inspect the response and subset these like `docs (found)`, `docs (redirect)` and `docs (404)` so that these individual categories can be tracked (e.g. in case a change makes calculating redirects much more expensive).

src/web/mod.rs
Kixiron: I was considering making one struct that contained both sub-structs for this very reason, but I don't know if I should do that or just combine them outright because I can see situations where only one may apply, thoughts?

jyn514: I don't know when we'd want to know the route but not the response time. Having the response time separate sounds useful, but if we don't have a use case yet then I'd rather combine them for now.

src/web/metrics.rs
Kixiron: From looking at the source, it looks like some of the internal macros have proper pathing, but the user-facing ones (that we use) do not

Kixiron:
As requested by @jyn514, I figured out how to test locally, so the two graphs added are these
Requests per minute
Response times

jyn514:
>  @pietroalbini Mentioned that he wanted a to have the rss feed be included in the new metrics, but I couldn't find where that happens.

`/releases/feed`: https://github.com/rust-lang/docs.rs/blob/master/src/web/routes.rs#L26, https://github.com/rust-lang/docs.rs/blob/master/src/web/releases.rs#L315

Kixiron:
Weird things discovered during testing
* Requesting `robots.txt` produces a response with `None` as the `Status`
* `/crate/:name/:version/` redirects to `/crate/:name/:version`
* `/releases/recent` is a 404, while `/releases/recent/1` isn't, maybe it should redirect to `/releases/recent/1`?
* `/releases/search` is only a valid route when parameters are supplied, should it redirect somewhere instead of giving a 404?
* Stars can't be added within tests, so `/releases/stars` and `/releases/stars/:page` can't be tested
* Release activity isn't added within tests, so `/releases/activity` can't be tested
* Above with `/releases/failures` and `/releases/failures/:page` (But funnily enough, not `/releases/recent-failures` or `/releases/recent-failures/:page`)
* Badges also don't work in tests

I think these tests are really good and very needed, but as they're implemented now they are strictly single-threaded, which isn't a problem for CI (which tests with `--test-threads 1`) but is a problem for devs, as testing takes quite a while

src/web/metrics.rs
```diff
@@ -100,12 +100,12 @@ fn duration_to_seconds(d: Duration) -> f64 {
 }
 
 #[derive(Debug, Clone)]
-pub struct ResponseTimer {
+pub struct ResponseRecorder {
     start_time: Instant,
-    route: Option<&'static str>,
+    route: Option<String>,
```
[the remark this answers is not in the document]
Kixiron: `String` is needed, as `route` is a formatted `String`

src/web/mod.rs
```diff
@@ -204,12 +201,66 @@ impl Handler for CratesfyiHandler {
...
+            [_, _] | [_, _, _] => metrics.route(msg!(response.as_ref(), "crate docs")),
+            [_, _, _, file] if file.ends_with(".html") => {
+                metrics.route(msg!(response.as_ref(), "crate docs"));
+            }
+
+            [route] if routes::DOC_RUST_LANG_ORG_REDIRECTS.contains(route) => {
+                metrics.route(msg!(response.as_ref(), "rust-lang redirect"));
+            }
+
+            _ => metrics.route(msg!(response.as_ref(), "crate docs")),
```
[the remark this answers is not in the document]
Kixiron: This is kind of a hard thing, as doc routes can be pretty much any route. This *may* work in catching only docs, but I'm not entirely sure

```rust
[_] | [_, _] | [_, _, "settings.html"] | [_, _, "all.html"] | [_, _, _] => { /* Docs */ }
[_, _, file] if file.ends_with(".html") => { /* Docs */ }

_ => { /* Unknown */ }
```

src/web/mod.rs
```diff
@@ -188,7 +208,80 @@ impl Handler for CratesfyiHandler {
                 }
 
                 Self::chain(&self.pool_factory, err).handle(req)
-            })
+            });
+
+        let path: &[&str] = &req.url.path();
+        match path {
```
pietroalbini, line 214: Why did you choose this approach?

I'm not convinced this is the way to go, as it requires maintenance every time we add a route if we want it to be tracked. I'm thinking the recorder struct could be changed into an iron handler:

```rust
struct RequestRecorder {
    handler: Box<dyn Handler>,
    route_name: String,
}

impl Handler for RequestRecorder {
    fn handle(&self, request: &Request) -> IronResult<Response> {
        let start = Instant::now();
        let result = self.handler.handle(request);
        let elapsed = start.elapsed();

        // Record the request and the elapsed time

        result
    }
}
```

Then, in `src/web/routes.rs`, the methods to add static, internal and rustdoc routes are changed to wrap the handlers in `RequestRecorder`. We'd also get the route name for free if we assume "static" and "rustdoc" for static and rustdoc routes, while using the `pattern` as the route name for internal routes.

pietroalbini:
Great, thanks!

PR #739, opened 2020-05-01
Track when crates are yanked and handle yanked status better

### LINKED ISSUE #112: Hide yanked crate versions.

Nemo157:
- mark releases as yanked when a yank is detected
- show yanked status on crate details page
- don't redirect to a yanked version if it's the latest
- add test to load handlebars templates
- Don't show latest version redirect to yanked versions
- Allow loading yanked docs via exact version number
- Show yanked status in nav on rustdoc pages
- Show yanked status on crate details version list
- Show yanked status in version drop down

    
I skipped 7a57350e90b5884ec1a322731ab7b40422ec828c from #322 to avoid any conflicts with #721, it may be useful to look into rebasing that too later.
    
fixes #112, #396
closes #322

Screenshots:

nav bar when viewing latest version and all versions are yanked:
image

nav bar when there's at least one non-yanked version, or viewing non-latest version of a fully yanked crate:
image

crate details page:
image

crate details version list:
image

crate details nav drop-down:
image

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -527,14 +527,10 @@ mod tests {
             db.fake_release().name("foo").version("0.0.3").create()?;
             db.fake_release().name("foo").version("0.0.2").create()?;
 
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.3").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
+            for version in &["0.0.1", "0.0.2", "0.0.3"] {
+                let details = CrateDetails::new(&db.conn(), "foo", version).unwrap();
+                assert_eq!(details.latest_release().version, "0.0.3");
+            }
 
             Ok(())
         })
@@ -553,14 +549,10 @@ mod tests {
                 .create()?;
             db.fake_release().name("foo").version("0.0.2").create()?;
 
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.2");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.2");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.3").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.2");
+            for version in &["0.0.1", "0.0.2", "0.0.3"] {
+                let details = CrateDetails::new(&db.conn(), "foo", version).unwrap();
+                assert_eq!(details.latest_release().version, "0.0.2");
+            }
 
             Ok(())
         })
@@ -587,14 +579,10 @@ mod tests {
                 .cratesio_data_yanked(true)
                 .create()?;
 
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
-
-            let details = CrateDetails::new(&db.conn(), "foo", "0.0.3").unwrap();
-            assert_eq!(details.latest_release().version, "0.0.3");
+            for version in &["0.0.1", "0.0.2", "0.0.3"] {
+                let details = CrateDetails::new(&db.conn(), "foo", version).unwrap();
+                assert_eq!(details.latest_release().version, "0.0.3");
+            }
 
             Ok(())
         })
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -375,13 +375,13 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.0.4")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release()
                 .name("foo")
                 .version("0.0.5")
                 .build_result_successful(false)
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
 
             assert_last_successful_build_equals(&db, "foo", "0.0.1", None)?;
@@ -411,7 +411,7 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.0.3")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
 
             assert_last_successful_build_equals(&db, "foo", "0.0.1", None)?;
@@ -435,7 +435,7 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.0.3")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release().name("foo").version("0.0.4").create()?;
 
@@ -465,7 +465,7 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.2.0")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release()
                 .name("foo")
@@ -545,7 +545,7 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.0.3")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release().name("foo").version("0.0.2").create()?;
 
@@ -566,17 +566,17 @@ mod tests {
             db.fake_release()
                 .name("foo")
                 .version("0.0.1")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release()
                 .name("foo")
                 .version("0.0.3")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
             db.fake_release()
                 .name("foo")
                 .version("0.0.2")
-                .cratesio_data_yanked(true)
+                .yanked(true)
                 .create()?;
 
             for version in &["0.0.1", "0.0.2", "0.0.3"] {
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -275,10 +275,15 @@ impl CrateDetails {
         Some(crate_details)
     }
 
-    /// Returns the version of the latest release of this crate.
+    /// Returns the version of the latest non-yanked release of this crate (or latest yanked if
+    /// they are all yanked).
     pub fn latest_version(&self) -> &str {
-        // releases will always contain at least one element
-        &self.releases[0].version
+        let release = self
+            .releases
+            .iter()
+            .find(|release| !release.yanked)
+            .unwrap_or(&self.releases[0]);
+        &release.version
     }
 }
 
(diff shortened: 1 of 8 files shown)


jyn514:
cc @QuietMisdreavus  :)

src/web/crate_details.rs
```diff
@@ -180,11 +183,11 @@ impl CrateDetails {
             description: rows.get(0).get(4),
             rustdoc_status: rows.get(0).get(11),
             target_name: rows.get(0).get(16),
-            default_target: rows.get(0).get(25),
+            default_target: rows.get(0).get(26),
```
jyn514, line 187: Oh boy, we have some technical debt stacking up. Once this PR is in I think we should start think about getting things by name instead of by index: https://docs.rs/postgres/0.15.2/postgres/rows/struct.Row.html#method.get

src/web/crate_details.rs
```diff
@@ -271,10 +275,20 @@ impl CrateDetails {
...
-        &self.releases[0].version
+        let release = self
+            .releases
+            .iter()
+            .find(|release| !release.yanked)
+            .unwrap_or_else(|| {
+                self.releases
+                    .iter()
+                    .next()
+                    .expect("there is always at least one release")
```
jyn514, line 290: Does this work without running into borrowck issues?
```suggestion
                &self.releases[0]
```

src/web/crate_details.rs
```diff
@@ -531,4 +546,64 @@ mod tests {
...
+            db.fake_release().name("foo").version("0.0.2").create()?;
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
+            assert_eq!(details.latest_version(), "0.0.2");
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
+            assert_eq!(details.latest_version(), "0.0.2");
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.3").unwrap();
+            assert_eq!(details.latest_version(), "0.0.2");
```
jyn514, line 570:
```suggestion
            for &version in &["0.0.1", "0.0.2", "0.0.3"] {
                let details = CrateDetails::new(&db.conn(), "foo", version).unwrap();
                assert_eq!(details.latest_version(), "0.0.2");
            }
```

src/web/crate_details.rs
```diff
@@ -531,4 +546,64 @@ mod tests {
...
+    #[test]
+    fn test_latest_version_ignores_yanked() {
+        crate::test::wrapper(|env| {
+            let db = env.db();
+
+            db.fake_release().name("foo").version("0.0.1").create()?;
+            db.fake_release()
+                .name("foo")
+                .version("0.0.3")
+                .cratesio_data_yanked(true)
```
jyn514, line 559: I think this should be named `yanked()` instead to better fit the other functions, are you interested in doing that here? If not I can make a follow-up PR.

src/web/crate_details.rs
```diff
@@ -531,4 +546,64 @@ mod tests {
...
+                .create()?;
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
+            assert_eq!(details.latest_version(), "0.0.3");
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.2").unwrap();
+            assert_eq!(details.latest_version(), "0.0.3");
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.3").unwrap();
+            assert_eq!(details.latest_version(), "0.0.3");
```
jyn514, line 604:
```suggestion
            for &version in &["0.0.1", "0.0.2", "0.0.3"] {
                let details = CrateDetails::new(&db.conn(), "foo", version).unwrap();
                assert_eq!(details.latest_version(), "0.0.3");
            }
```

src/web/mod.rs
```diff
@@ -314,11 +319,11 @@ fn match_version(conn: &Connection, name: &str, version: Option<&str>) -> Option
...
-    if req_version == "*" && !versions_sem.is_empty() {
-        return Some(MatchVersion {
+    // return first non-yanked version if requested version is '*'
+    if req_version == "*" {
+        return versions_sem.first().map(|v| MatchVersion {
             corrected_name,
-            version: MatchSemver::Semver((versions_sem[0].0.to_string(), versions_sem[0].1)),
+            version: MatchSemver::Semver((v.0.to_string(), v.1)),
         });
     }
```
jyn514, line 328: Can you add a test that this returns a 404 if you request `/:crate` for a crate with only yanked versions?

src/web/rustdoc.rs
```diff
@@ -549,9 +552,15 @@ mod test {
...
+            Ok(Some(link))
+        } else {
+            Ok(None)
         }
-        panic!("no redirect found for {}", path);
+    }
+
+    fn latest_version_redirect(path: &str, web: &TestFrontend) -> Result<String, failure::Error> {
+        try_latest_version_redirect(path, web)
+            .and_then(|v| v.ok_or_else(|| failure::format_err!("no redirect found for {}", path)))
```
jyn514, line 563: Instead of making `try_latest` a separate function, could you put this error in the `Ok(None)` branch above?

src/web/rustdoc.rs
```diff
@@ -762,6 +771,118 @@ mod test {
...
+    #[test]
+    fn redirect_latest_does_not_go_to_yanked_versions() {
+        wrapper(|env| {
+            let db = env.db();
+            db.fake_release()
+                .name("dummy")
+                .version("0.1.0")
+                .rustdoc_file("dummy/index.html", b"lah")
+                .create()
+                .unwrap();
```
jyn514, line 783:
```suggestion
                .create()?;
```
This doesn't actually affect the test, but it makes it easier to read IMO

src/web/rustdoc.rs
```diff
@@ -762,6 +771,118 @@ mod test {
...
+            db.fake_release()
+                .name("dummy")
+                .version("0.2.0")
+                .rustdoc_file("dummy/index.html", b"lah")
+                .create()
+                .unwrap();
```
jyn514, line 789:
```suggestion
                .create()?;
```

src/web/rustdoc.rs
```diff
@@ -762,6 +771,118 @@ mod test {
...
+            db.fake_release()
+                .name("dummy")
+                .version("0.2.1")
+                .rustdoc_file("dummy/index.html", b"lah")
+                .cratesio_data_yanked(true)
+                .create()
+                .unwrap();
```
jyn514, line 796:
```suggestion
                .create()?;
```

src/web/rustdoc.rs
```diff
@@ -762,6 +771,118 @@ mod test {
...
+
+    #[test]
+    fn yanked_release_shows_warning_in_nav() {
+        fn has_yanked_warning(path: &str, web: &TestFrontend) -> Result<bool, failure::Error> {
+            use html5ever::tendril::TendrilSink;
+            assert_success(path, web)?;
+            let data = web.get(path).send()?.text()?;
+            Ok(kuchiki::parse_html()
+                .one(data)
+                .select("form ul li .warn")
```
jyn514, line 854: `get_platform_links` is pretty similar to this, do you think it's worth factoring out a `html_select()` function?

templates/crate_details.hbs
```diff
@@ -44,10 +44,14 @@
               <ul class="pure-menu-list">
                 {{#each releases}}
                 <li class="pure-menu-item">
-                  {{#if this.build_status}}
-                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link">{{this.version}}</a>
+                  {{#if this.yanked}}
+                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link warn" title="{{../name}}-{{this.version}} is yanked"><i class="fa fa-fw fa-warning"></i> {{this.version}}</a>
                   {{else}}
```
jyn514, line 49: Why is this an else? If it failed to build I feel like we should show both warnings.

templates/navigation_rustdoc.hbs
```diff
@@ -78,10 +78,14 @@
                           <ul class="pure-menu-list">
                             {{#each releases}}
                             <li class="pure-menu-item">
-                              {{#if this.build_status}}
-                              <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link">{{this.version}}</a>
+                              {{#if this.yanked}}
+                              <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link warn" title="{{../name}}-{{this.version}} is yanked"><i class="fa fa-fw fa-warning"></i> {{this.version}}</a>
                               {{else}}
```
jyn514, line 83: Ditto, I think this should be in addition, not instead of.

src/web/mod.rs
```diff
@@ -778,4 +783,9 @@ mod test {
             Ok(())
         });
     }
+
+    #[test]
+    fn test_templates_are_valid() {
+        handlebars_engine().expect("Failed to load handlebar templates");
+    }
```
Kixiron, line 790: Great idea to test that the templates are valid

templates/crate_details.hbs
```diff
@@ -44,10 +44,14 @@
               <ul class="pure-menu-list">
                 {{#each releases}}
                 <li class="pure-menu-item">
-                  {{#if this.build_status}}
-                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link">{{this.version}}</a>
+                  {{#if this.yanked}}
+                  <a href="/crate/{{../name}}/{{this.version}}" class="pure-menu-link warn" title="{{../name}}-{{this.version}} is yanked"><i class="fa fa-fw fa-warning"></i> {{this.version}}</a>
```
Kixiron, line 48: I can't help but thinking that this kind of crate-wide popup is something that could used for a lot of different things, it should probably get modularization out into a component sometime to prevent an eventual chain of `#if`s that all do essentially the same things

src/web/crate_details.rs
Nemo157: It should

src/web/crate_details.rs
Nemo157: I was _so close_ to making that change, then thought I'd avoid extraneous changes 😄

src/web/rustdoc.rs
Nemo157: This was from when I had a test that asserted that there would be no redirect, so separating the lack of redirect from an error finding a redirect was useful.

templates/crate_details.hbs
Nemo157: Double ⚠️ symbols, or just concatenate the hover messages?

src/web/rustdoc.rs
```diff
@@ -320,15 +320,23 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
-        path_for_version(&latest_path, &crate_details.doc_targets, &conn)
+        format!(
+            "/{}/{}/{}",
+            name,
+            latest_version,
+            path_for_version(&latest_path, &crate_details.doc_targets, &conn)
+        )
     } else {
-        Default::default()
+        format!("/crate/{}/{}", name, latest_version)
```
jyn514, line 339: Just so I understand - the reason for the change is so that the link to the latest doesn't give a 404 if the build doesn't succeed? As far as I know this should have worked the same way as you have here before, just with an extra redirect.

Nemo157: Yep, this is just removing one extra redirect step, and making the behaviour match what the test thought it was testing 😄

templates/crate_details.hbs
jyn514: I think double :warning: is fine.

templates/crate_details.hbs
jyn514: Agreed, can you make an issue for that?

Nemo157: Yeah, there's a bunch of redundancy in the current templates that probably should be pulled out into components, but I'm not sure exactly how to set that up with handlebars here.

jyn514: https://github.com/rust-lang/docs.rs/issues/740

jyn514:
Whoops, I misread the changes. You addressed both the comments I linked.

jyn514:
Looks good once nits have been addressed.

PR #741, opened 2020-05-03
Put back tab index on top crate search input

### LINKED ISSUE #118: Hide the search crate text field by default
Hi!

Some web plugins/web browsers have a shortcut to focus the first text field in a page, which is what I use in Rust documentations to start a search within the crate.
However, on docs.rs, there's another "Find crate" text field which is before the documentation search field which forbids me to use this shortcut.

So, it would be nice to hide the "Find crate" text field by default, requiring to click a button or hover to show the text field.
Another solution that would make it after the other text field would also fix this issue.

If you need help to implement that, I can contribute.

Thanks.

GuillaumeGomez:
Fixes #118

### DIFF
--- a/templates/navigation_rustdoc.hbs
+++ b/templates/navigation_rustdoc.hbs
@@ -5,7 +5,8 @@
             {{#unless varsb.show_search_form}}
             <div id="search-input-nav">
               <label for="nav-search"><i class="fa fa-fw fa-search"></i></label>
-              <input id="nav-search" name="query" type="text" aria-label="Find crate by search query" placeholder="Find crate"{{#if varss.search_query}} value="{{varss.search_query}}"{{/if}}>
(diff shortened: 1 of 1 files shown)


antoyo:
Can you add a comment to explain why we do this so that is does not get removed again in the future?

GuillaumeGomez:
You mean in the HTML?

antoyo:
It's a template, so it could be a template comment.

GuillaumeGomez:
Yes, my bad for the incomplete answer. That's what I meant. ;)

GuillaumeGomez:
Updated!

jyn514:
This one fell through the cracks, sorry. Thanks for working on this!

jyn514:
@antoyo I just deployed the change, do you mind visiting the site to see whether the problem is fixed?

antoyo:
@jyn514: it works fine. Thanks.

PR #742, opened 2020-05-03
Refactored rustdoc_html_server_handler

Kixiron:
Commented and refactored the `rustdoc_html_server_handler` function

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -319,18 +319,30 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         body_class.push_str(" container-rustdoc");
     }
 
+    let latest_release = crate_details.latest_release();
+
     // Get the latest version of the crate
-    let latest_version = crate_details.latest_release().version.to_owned();
+    let latest_version = latest_release.version.to_owned();
     let is_latest_version = latest_version == version;
 
+    // If the requested crate version is the most recent, use it to build the url
+    let latest_path = if is_latest_version {
+        format!("/{}/{}", name, latest_version)
+
     // If the requested version is not the latest, then find the path of the latest version for the `Go to latest` link
-    let path_in_latest = if !is_latest_version {
+    } else if latest_release.build_status {
         // Replace the version of the old path with the latest version
         let mut latest_path = req_path.clone();
         latest_path[2] = &latest_version;
-        path_for_version(&latest_path, &crate_details.doc_targets, &conn)
+
+        format!(
+            "/{}/{}/{}",
+            name,
+            latest_version,
+            path_for_version(&latest_path, &crate_details.doc_targets, &conn)
+        )
     } else {
-        String::new()
+        format!("/crate/{}/{}", name, latest_version)
     };
 
     // The path within this crate version's rustdoc output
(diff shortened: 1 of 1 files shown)


src/web/rustdoc.rs
```diff
@@ -209,141 +209,158 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
     let latest_version = crate_details.latest_version().to_owned();
     let is_latest_version = latest_version == version;
 
+    // If the requested version is not the latest, then find the path of the latest version for easy redirection
     let path_in_latest = if !is_latest_version {
+        // Replace the version of the old path with the latest version
         let mut latest_path = req_path.clone();
         latest_path[2] = &latest_version;
+
+        // Retrieve the path in the database of the latest crate version
```
jyn514, line 331: I think this is a _little_ overboard with the comments :P. If you want to improve the comments on `path_for_version` though, that would be much appreciated. The point of that function is to search for the current page if it doesn't exist in the latest version.
```suggestion

```

src/web/rustdoc.rs
```diff
@@ -209,141 +209,158 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
     }
-    content.body_class = body_class;
-
-    content.full = file_content;
 
+    // Get the latest version of the crate
     let latest_version = crate_details.latest_version().to_owned();
     let is_latest_version = latest_version == version;
 
+    // If the requested version is not the latest, then find the path of the latest version for easy redirection
```
jyn514, line 325: That's not actually what this does. If this isn't the latest version, then it shows the little 'Go to latest version' link, but it does _not_ redirect you automatically.

Kixiron: Yes, that was my intention with the `for easy redirection`, it makes redirection easier for the user

src/web/rustdoc.rs
```diff
@@ -209,141 +209,158 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
             }
         }
-        None => return Err(IronError::new(Nope::ResourceNotFound, status::NotFound)),
+    } else {
+        // Return a 404, as a crate by that name and version doesn't exist
+        return Err(IronError::new(Nope::ResourceNotFound, status::NotFound));
     };
 
-    if let Some(name) = corrected_name {
-        return redirect(&name, &version, &req_path);
```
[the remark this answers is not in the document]
Kixiron: `if let` is generally more idiomatic, plus clippy screamed at me iirc

src/web/rustdoc.rs
```diff
@@ -209,141 +209,157 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
-    content.head = head;
-    content.body = body;
 
+    // Add the `rustdoc` classes to the html body
     if body_class.is_empty() {
         body_class = "rustdoc container-rustdoc".to_string();
     } else {
-        // rustdoc adds its own "rustdoc" class to the body
+         // rustdoc adds its own "rustdoc" class to the body 
         body_class.push_str(" container-rustdoc");
```
jyn514, line 319: I really need to preview my suggestions before hitting submit.

```suggestion
        // rustdoc adds its own "rustdoc" class to the body 
        body_class.push_str(" container-rustdoc");
```

src/web/rustdoc.rs
jyn514: I get that, but I think it would be more clear if you mentioned 'Go to latest' or a link rather than saying redirection.

PR #743, opened 2020-05-03
Added tests for json serialization

Kixiron:


Kixiron:
#709 works thanks to this, so this PR is no longer needed

jyn514:
I agree, I think #709 covers this.

PR #744, opened 2020-05-04
Use ON CONFLICT DO UPDATE

### LINKED ISSUE #719: Use ON CONFLICT DO UPDATE instead of rewriting it manually
@pietroalbini pointed out in https://github.com/rust-lang/docs.rs/pull/643#discussion_r408852995 that code that looks like

```rust
            // check if file already exists in database
            let rows = self
                .conn
                .query("SELECT COUNT(*) FROM files WHERE path = $1", &[&blob.path])?;

            if rows.get(0).get::<usize, i64>(0) == 0 {
                trans.query(
                    "INSERT INTO files (path, mime, content) VALUES ($1, $2, $3)",
                    &[&blob.path, &blob.mime, &blob.content],
                )?;
            } else {
                trans.query(
                    "UPDATE files SET mime = $2, content = $3, date_updated = NOW() \
                                WHERE path = $1",
                    &[&blob.path, &blob.mime, &blob.content],
                )?;
            }
```

can be simplified to

```rust
            trans.query(
                "INSERT INTO files (path, mime, content)
                 VALUES ($1, $2, $3)
                 ON CONFLICT (path) DO UPDATE
                    SET mime = EXCLUDED.mime, content = EXCLUDED.content",
                &[&blob.path, &blob.mime, &blob.content],
            )?;
```

It would be great to do this for other similar patterns in docs.rs:

- [`add_package`](https://github.com/rust-lang/docs.rs/blob/master/src/db/add_package.rs#L53)
- [`update_release_activity`](https://github.com/rust-lang/docs.rs/blob/master/src/utils/release_activity_updater.rs#L60) - this would also remove the postgres errors on startup!

Kixiron:
Closes #719

Use the postgres `ON CONFLICT` function to update tables instead of manually writing the queries and error handling out

### DIFF
--- a/src/utils/github_updater.rs
+++ b/src/utils/github_updater.rs
@@ -21,15 +21,15 @@ pub fn github_updater() -> Result<()> {
     //       always the same across all versions of a crate
     for row in &conn.query(
         "SELECT DISTINCT ON (crates.name)
-                                        crates.name,
-                                        crates.id,
-                                        releases.repository_url
-                                 FROM crates
-                                 INNER JOIN releases ON releases.crate_id = crates.id
-                                 WHERE releases.repository_url ~ '^https*://github.com' AND
-                                       (crates.github_last_update < NOW() - INTERVAL '1 day' OR
-                                        crates.github_last_update IS NULL)
-                                 ORDER BY crates.name, releases.release_time DESC",
+                crates.name,
+                crates.id,
+                releases.repository_url
+         FROM crates
+         INNER JOIN releases ON releases.crate_id = crates.id
+         WHERE releases.repository_url ~ '^https*://github.com' AND
+               (crates.github_last_update < NOW() - INTERVAL '1 day' OR
+                crates.github_last_update IS NULL)
+         ORDER BY crates.name, releases.release_time DESC",
         &[],
     )? {
         let crate_name: String = row.get(0);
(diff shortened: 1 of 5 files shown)


src/utils/github_updater.rs
```diff
@@ -21,15 +21,15 @@ pub fn github_updater() -> Result<()> {
...
-                                 WHERE releases.repository_url ~ '^https*://github.com' AND
-                                       (crates.github_last_update < NOW() - INTERVAL '1 day' OR
-                                        crates.github_last_update IS NULL)
-                                 ORDER BY crates.name, releases.release_time DESC",
+                crates.name,
+                crates.id,
+                releases.repository_url
+         FROM crates
+         INNER JOIN releases ON releases.crate_id = crates.id
+         WHERE releases.repository_url ~ '^https*://github.com' AND
```
jyn514, line 29: This regex doesn't look right ... it matches `https*` but instead it should match `https?`. Currently it will accept `httpsssssssssssssssss` when it shouldn't.

src/utils/github_updater.rs
```diff
@@ -42,10 +42,10 @@ pub fn github_updater() -> Result<()> {
...
                 conn.execute(
                     "UPDATE crates
-                              SET github_description = $1,
-                                  github_stars = $2, github_forks = $3,
-                                  github_issues = $4, github_last_commit = $5,
-                                  github_last_update = NOW() WHERE id = $6",
+                     SET github_description = $1,
+                         github_stars = $2, github_forks = $3,
+                         github_issues = $4, github_last_commit = $5,
+                         github_last_update = NOW() WHERE id = $6",
```
jyn514, line 48:
```suggestion
                     SET github_description = $1,
                         github_stars = $2, github_forks = $3,
                         github_issues = $4, github_last_commit = $5,
                         github_last_update = NOW()
                     WHERE id = $6",
```

src/web/crate_details.rs
```diff
@@ -113,34 +113,34 @@ impl CrateDetails {
...
+                crates.github_forks,
+                crates.github_issues,
+                releases.is_library,
+                releases.doc_targets,
+                releases.license,
+                releases.documentation_url,
+                releases.default_target
+         FROM releases
+         INNER JOIN crates ON releases.crate_id = crates.id
+         WHERE crates.name = $1 AND releases.version = $2;";
```
jyn514, line 143: Why this change?

Kixiron: It uses a little less screen real estate

PR #746, opened 2020-05-04
Added some fault-tolerance to the build queue

### LINKED ISSUE #628: Crates that are fetched but not added to the index because of a crash will be lost
Secondary problem: apparently crates that are fetched but not added to the index will be lost and not build. Relevant code: https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/queue.rs#L13

_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/issues/626#issuecomment-591989434_

`crates-index-diff` just released a change in 5.1.0 that will let us peek without consuming the changes :tada: :tada: . This will give us much better durability in the event of a crash, I just need to implement using the new API.

Kixiron:
Closes #628

Stops losing crates that are being added to the build queue when a crash occurs. Attempted to make `add_crate_to_queue` batched, but couldn't figure out how to with the current db driver


Nemo157:
I just noticed the

```rust
add_crate_to_queue(&conn, &krate.name, &krate.version, priority).ok();
debug!("{}-{} added into build queue", krate.name, krate.version);
```

hidden in there, maybe this should be changed to 

```rust
match add_crate_to_queue(&conn, &krate.name, &krate.version, priority) {
  Ok(_) => debug!("{}-{} added into build queue", krate.name, krate.version),
  Err(err) => error!("failed adding {}-{} into build queue: {}", krate.name, krate.version, err),
}
```

Kixiron:
I realized that we could also add a prometheus hook for this, if it's something we think is important enough

jyn514:
> I realized that we could also add a prometheus hook for this, if it's something we think is important enough

 I don't this is necessary, we already track the number of crates queued and we can look at the logs or /releases/queue if we want to know the exact crates.

src/docbuilder/queue.rs
```diff
@@ -29,13 +28,15 @@ impl DocBuilder {
...
-            add_count += 1;
         }
 
         index.set_last_seen_reference(oid)?;
 
-        Ok(add_count)
+        // Get the size of the queue
+        let query = conn.query("SELECT COUNT(*) as size FROM queue", &[])?;
+        let queue_size = query.iter().next().unwrap().get::<_, i64>("size") as usize;
+
```
jyn514, line 38: This doesn't look right, it returns the _total_ number of crates in the queue, not the number added.

Kixiron: Yes, I changed it because that's what the comment [here](https://github.com/rust-lang/docs.rs/blob/822026308ff8250fa3f20ee374fc2d839d9c56de/src/docbuilder/queue.rs#L12) says it's supposed to do

jyn514: Are you sure the comment matches up with how the function is used? I don't think a behavior change like this belongs in the same PR.

pietroalbini:
Why did you change the permissions of the docker entrypoint?

Kixiron:
Fixed, something with Windows messed up

jyn514:
Thanks for working on this!

PR #747, opened 2020-05-05
RFC - Migrate to asynchronous code using Hyper

Kixiron:
[Rendered](https://github.com/rust-lang/docs.rs/blob/e9be96daee7a8fed37b41e1063658da3a3ed5dbe/rfcs/migrate-to-async-hyper.md)

### DIFF
--- a/rfcs/migrate-to-async-hyper.md
+++ b/rfcs/migrate-to-async-hyper.md
@@ -1,0 +1,55 @@
+# Migrating docs.rs to asynchronous code and hyper
+
+1. Update `tokio`, `futures` and `rusoto_{s3, core, credential}` to their latest versions (I think Joshua started on this, I'd be happy to help with finishing it)
+2. Make `main` an `async fn` with `#[tokio::main]` to start the tokio runtime, add the `hyper` dependency and migrate the side services to async (By side services I mean the ones in the daemon)
+
+Making the side services async sounds like a larger thing than it actually is, e.g.
+
+```rust
+thread::spawn(move || loop {
+    thread::sleep(Duration::from_secs(60 * 60 * 6));
+    if let Err(e) = github_updater() {
+        error!("Failed to update github fields: {}", e);
+    }
+})
+```
+
+Would become
+
+```rust
+task::spawn(async move {
+    let mut interval = time::interval(Duration::from_secs(60 * 60 * 6));
+    loop {
+        interval.tick().await;
+        if let Err(e) = github_updater().await {
+            error!("Failed to update github fields: {}", e);
+        }
+    }
+})
+```
+
+This actually will do a lot for the server's performance, as the many threads that were doing nothing but sleeping can now be used for useful work. Additionally, for the `release activity updater` this is a good change, as instead un-sleeping every minute and checking if it's `23:55`, we can use tokio's `interval_at` to make it automatically wake once at `23:55` and execute.
+Additionally, all the services are quite small and consist of basically http requests, so migrating can be literally as simple as adding awaits in some cases.
+
+3. Now, at this point the next step would be to migrate from `iron` to `hyper`, and there's a few ways to go about it
+    1. Move all routing to `hyper`, separate the handlers out from being within `iron` code and forwarding requests with `spawn_blocking`
+        1. Requires some glue code, but adding that (in an inert state where it isn't used) can be done in separate prs
+        2. Will be slightly slower until full migration is complete, as we'll have to translations from hyper requests to iron ones and from `IronResult` into the hyper equivalent
+        3. Looking at the code, this will actually be made much easier by the fact that we already have a custom wrapper around managing routes, and we can use that to our advantage
+    2. Use nginx to have two 'separate' servers on different ports that cross-forward things they can't handle
+        1. Messy and kind of impractical, not to mention difficult to develop locally
+        2. Probably the worst option
+    3. Do it all at once
+        1. Has the greatest code churn, but rips off the band-aid, so to speak
+        2. The easiest and quickest option as far as developers are concerned, takes much more review time
+    4. Have iron forward any requests it doesn't know how to handle to hyper and slowly move over handlers to hyper
+        1. Very similar to option 1 except flipped, as instead of going hyper -> iron we go iron -> hyper
+        2. Like option 1 it requires some initial setup/glue code that can be separated out into PRs
+4. Here's where the database migration would happen, and again there are options
+    1. Switch to `tokio-postgres`
+        1. Virtually no change in actual code past changing imports and adding `.await`s
+        2. Still requires *eventually* migrating to diesel
+    2. Incrementally migrate to `diesel`/`tokio-diesel`
+        1. The initial switch can be relatively small and can just involve switching migrations to diesel's format, making the diesel schemas and changing all queries to use `diesel::sql_query` (Allows using raw sql for queries, this lets us have a smaller change in code)
+        2. (Not blocking for anything else, can be much smaller and more numerous PRs) Move existing queries to use actual diesel query building
+        3. In the end I think this would be the better route, as it'd "force" us to actually migrate the database to the better option, as well as removing quite a few dependencies, namely `schemamam`, `r2d2` and `postgres`. Moving from those (postgres in particular) would also allow us to move other dependencies to better/newer options, such as changing the deprecated `time` crate to `chrono`. This would also force us to think a little more critically about how the database is structured and help with starting thinking about how to improve it (as there's some significant tech debt there, but that's a separate thing entirely)
--- a/rfcs/migrate-to-async-hyper.md
+++ b/rfcs/migrate-to-async-hyper.md
@@ -6,7 +6,7 @@
 
 2. Make `main` an `async fn` with `#[tokio::main]` to start the tokio runtime, add the `hyper` dependency and migrate the side services to async (By side services I mean the ones in the daemon)
 
-3. Move all routing to `hyper`, separate the handlers out from being within `iron` code and forward requests with `spawn_blocking`
+3. Add a catch-all handler that uses `hyper`, so that any requests not picked up by `iron` are handled by hyper
    1. Routes should be migrated to asynchronous code from least used to most used
 
 4. Switch from `postgres` to `tokio-postgres` for database interaction
--- a/rfcs/migrate-to-async-hyper.md
+++ b/rfcs/migrate-to-async-hyper.md
@@ -15,7 +15,7 @@
 
 ### Why async
 
-A web server is inherently io bound, which is the type of task that asynchronous execution excels at. If all goes well, the server will run faster and with potentially less resource usage than it currently has.
+A web server is inherently io bound, which is the type of task that asynchronous execution excels at. If all goes well, the server will run faster and with potentially less resource usage than it currently has due to using a single runtime to execute all operations.
 Some of our current dependencies already use async (such as `rusoto`), and we could also potentially see improvements from them executing in an actually asynchronous environment.
 Additionally, updating things allows us to both gain newer (and generally faster, safer and better supported) dependencies, as well as letting us give a critical eye to our current infrastructure to see what's good and what's bad.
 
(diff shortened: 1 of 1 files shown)


rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,55 @@
...
+3. Now, at this point the next step would be to migrate from `iron` to `hyper`, and there's a few ways to go about it
+    1. Move all routing to `hyper`, separate the handlers out from being within `iron` code and forwarding requests with `spawn_blocking`
+        1. Requires some glue code, but adding that (in an inert state where it isn't used) can be done in separate prs
+        2. Will be slightly slower until full migration is complete, as we'll have to translations from hyper requests to iron ones and from `IronResult` into the hyper equivalent
+        3. Looking at the code, this will actually be made much easier by the fact that we already have a custom wrapper around managing routes, and we can use that to our advantage
+    2. Use nginx to have two 'separate' servers on different ports that cross-forward things they can't handle
+        1. Messy and kind of impractical, not to mention difficult to develop locally
+        2. Probably the worst option
+    3. Do it all at once
+        1. Has the greatest code churn, but rips off the band-aid, so to speak
```
pietroalbini, line 43: Another downside is, if this migration doesn't turn to be viable much more development and review time will be wasted compared to just migrating parts of the code first.

For example, crates.io has a custom web framework backed by [a C++ web framework](https://github.com/civetweb/civetweb): when they tried to migrate the underlying framework to Hyper they were forced to roll back due to excessive resource usage. This was a while ago and shouldn't be taken as a "don't switch to Hyper", but still, we *might* have to revert this.

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,55 @@
...
+        2. Still requires *eventually* migrating to diesel
+    2. Incrementally migrate to `diesel`/`tokio-diesel`
+        1. The initial switch can be relatively small and can just involve switching migrations to diesel's format, making the diesel schemas and changing all queries to use `diesel::sql_query` (Allows using raw sql for queries, this lets us have a smaller change in code)
+        2. (Not blocking for anything else, can be much smaller and more numerous PRs) Move existing queries to use actual diesel query building
+        3. In the end I think this would be the better route, as it'd "force" us to actually migrate the database to the better option, as well as removing quite a few dependencies, namely `schemamam`, `r2d2` and `postgres`. Moving from those (postgres in particular) would also allow us to move other dependencies to better/newer options, such as changing the deprecated `time` crate to `chrono`. This would also force us to think a little more critically about how the database is structured and help with starting thinking about how to improve it (as there's some significant tech debt there, but that's a separate thing entirely)
```
pietroalbini, line 55: We don't have to switch away from `postgres` to get rid of `time`: [there is a Cargo feature in `postgres`](https://docs.rs/postgres/0.17.3/postgres/#features) to enable chrono support.

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,55 @@
...
+    2. Use nginx to have two 'separate' servers on different ports that cross-forward things they can't handle
+        1. Messy and kind of impractical, not to mention difficult to develop locally
+        2. Probably the worst option
+    3. Do it all at once
+        1. Has the greatest code churn, but rips off the band-aid, so to speak
+        2. The easiest and quickest option as far as developers are concerned, takes much more review time
+    4. Have iron forward any requests it doesn't know how to handle to hyper and slowly move over handlers to hyper
+        1. Very similar to option 1 except flipped, as instead of going hyper -> iron we go iron -> hyper
+        2. Like option 1 it requires some initial setup/glue code that can be separated out into PRs
+4. Here's where the database migration would happen, and again there are options
```
pietroalbini, line 48: I don't think this RFC needs to propose switching from `postgres` to `diesel`: if the goal is to just make the whole application async we can always switch to `tokio-postgres`, and think about the switch later.

Kixiron: I agree, I just thought it was a good thing to keep in mind/start thinking about, I should have made that more clear

rfcs/migrate-to-async-hyper.md
Kixiron: Yes, that is a definite concern. Hopefully though there won't be any adverse side effects to switching over

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Why async
+
+A web server is inherently io bound, which is the type of task that asynchronous execution excels at. If all goes well, the server will run faster and with potentially less resource usage than it currently has.
+Some of our current dependencies already use async (such as `rusoto`), and we could also potentially see improvements from them executing in an actually asynchronous environment.
+Additionally, updating things allows us to both gain newer (and generally faster, safer and better supported) dependencies, as well as letting us give a critical eye to our current infrastructure to see what's good and what's bad.
+
+### The approach
+
+1. Updating all async-related dependencies gives a foothold to start the update process. Updating all of the dependencies at once is required because their versioning is intertwined
```
Nemo157, line 24: This could potentially be avoided by using `tokio-compat`'s runtime (this supports running Tokio 0.1 libraries on top of the Tokio 0.2 runtime).

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,47 @@
+# Migrating docs.rs to asynchronous code and hyper
+
+## Approach
+
+1. Update `tokio`, `futures` and `rusoto_{s3, core, credential}` to their latest versions
+
+2. Make `main` an `async fn` with `#[tokio::main]` to start the tokio runtime, add the `hyper` dependency and migrate the side services to async (By side services I mean the ones in the daemon)
+
+3. Move all routing to `hyper`, separate the handlers out from being within `iron` code and forward requests with `spawn_blocking`
+   1. Routes should be migrated to asynchronous code from least used to most used
```
Nemo157, line 10: Rather than moving all routing to `hyper` at once, the `hyper` router could just have a catch-all final route that forwards into the `iron` router.

rfcs/migrate-to-async-hyper.md
Kixiron: I considered that, but it's a relatively easy change and it eliminates the future problem of having multiple versions of things spinning around. It's easier in the long run to just get it over and done in my opinion

rfcs/migrate-to-async-hyper.md
Kixiron: I also considered that, but I think that moving routing would be the easiest option, as the way we have things set up makes it so that every route is a potential crate, so forwarding would be very difficult to implement well. As mentioned, we already have custom routing code in place, so moving over shouldn't be to large of a hassle

jyn514: > I also considered that, but I think that moving routing would be the easiest option, as the way we have things set up makes it so that every route is a potential crate, so forwarding would be very difficult to implement well.

I don't see why this is the case? If we recognize the route, send it to hyper, otherwise send it to iron. If hyper recognizes the route as a crate it will serve the request, otherwise it won't recognize it and will forward to iron. We just have to make sure we don't add new top-level routes in hyper.

Kixiron: The other reason that that's not an ideal solution is that it breaks the flow of async, so we'll have to do a lot more work to get hyper's responses back onto the runtime. Having hyper determine things also makes it easier to migrate routes, as it only affects the route code, so multiple routes can be migrated in separate PRs without conflicting. Using hyper to forward requests is just much easier of a solution from my point of view

Nemo157: `iron` and `hyper` are both `fn(Request) -> Response` based aren't they? So you'd just have a final handler like

```rust
get("*", |request| async move {
  tokio::blocking(|| {
    let request = map_to_iron(request);
    let response = iron_router.handle(request);
    map_to_hyper(response)
  }.await
});
```

Kixiron: Oh, I think I understand what y'all are proposing, I guess that's a viable solution

Kixiron:
Waiting on review from the following members
[ ] @pietroalbini 
[ ] @jyn514 
[ ] @Nemo157 
[ ] @GuillaumeGomez
[x] @Kixiron

jyn514:
Please update the RFC to address the comments, including the comments you marked as resolved.

Nemo157:
One other thought I just had is that it might be good to avoid having the actual site code rely directly on `hyper` by using something like [`http-service`](https://docs.rs/http-service/0.5.0/http_service/) (though there are ecosystem issues that make me hesitant to directly recommend `http-service`). If there are performance issues with `hyper` then that decoupling would make moving to another service host easier.

Kixiron:
I think pretty much anything we use will make for a slightly painful experience because most are fairly heavily opinionated and very different. Additionally, `hyper` has a pretty good relationship with the `tower` webstack, so if need be transferring to that probably wouldn't be too bad

Nemo157:
That was meant to be the point of `http-service`, it’s a (relatively) unopinionated Interface between an opinionated web framework and the underlying server; so that you can swap web servers while using the same framework (or proposed lack of framework) on top.

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,47 @@
+# Migrating docs.rs to asynchronous code and hyper
+
+## Approach
+
+1. Update `tokio`, `futures` and `rusoto_{s3, core, credential}` to their latest versions
+
+2. Make `main` an `async fn` with `#[tokio::main]` to start the tokio runtime, add the `hyper` dependency and migrate the side services to async (By side services I mean the ones in the daemon)
```
jyn514, line 7: Why should main be async? It seems to me the benefit is coming from web requests/io be async, but there's very little io in main. It spawns worker threads if it needs to do anything heavy.

rfcs/migrate-to-async-hyper.md
jyn514: Please update the RFC to reflect this alternative.

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,47 @@
...
+   1. Routes should be migrated to asynchronous code from least used to most used
+
+4. Switch from `postgres` to `tokio-postgres` for database interaction
+
+## Rationale
+
+### Why async
+
+A web server is inherently io bound, which is the type of task that asynchronous execution excels at. If all goes well, the server will run faster and with potentially less resource usage than it currently has.
+Some of our current dependencies already use async (such as `rusoto`), and we could also potentially see improvements from them executing in an actually asynchronous environment.
```
jyn514, line 19: It sounds like you're talking about a single global runtime so we don't have to create a new one for each crate upload. Can you say that explicitly?

rfcs/migrate-to-async-hyper.md
```diff
@@ -0,0 +1,47 @@
...
+Additionally, updating things allows us to both gain newer (and generally faster, safer and better supported) dependencies, as well as letting us give a critical eye to our current infrastructure to see what's good and what's bad.
```
jyn514, line 20: nit: we could update without completely switching over to async

rfcs/migrate-to-async-hyper.md
Kixiron: It's a transitive thing, async functions call async functions. The idea is that all spawned tasks are part of the same runtime which makes the most efficient use of the system.

rfcs/migrate-to-async-hyper.md
Kixiron: Yes, I meant that as an aside bonus that will come through along with the transition, not necessarily exclusively because of it

rfcs/migrate-to-async-hyper.md
jyn514: I don't see why it needs to be async to call async functions, we already use `spawn_blocking` in a few places. My concern is this adds overhead for interactive usage since we have to start the whole tokio runtime before processing CLI args.

Nemo157: Testing in debug mode with hello world the overhead of `tokio::main` is <4ms on some not great hardware.

PR #750, opened 2020-05-06
Disable outline on focused rustdoc element to workaround safari performance issue

### LINKED ISSUE #725: Low FPS on Safari 13.1
I noticed a low FPS when opening the docs on Safari 13.1. This problem does not happen when the page is not focused and does not happen on Chrome. I do not use any extension.

I tested the page by compiling the source code from this repo and configuring it to create the docs for the crate postgres.

I have recorded two videos showing the problem, being the first one a screen recording (less noticiable) and the second one I recorded my screen from a iPhone. I recommend you to watch the one recorded from the iPhone since the low FPS is more visible.

The low FPS does not happen if the page is not focused, I show this in the second video. To simulate this just open the page and move to another app.

https://www.youtube.com/watch?v=zhvuoTpoz2c&feature=youtu.be
https://www.youtube.com/watch?v=Y7Sy7rsCfsM&feature=youtu.be

Nemo157:
Tested on Safari 13.1 and Safari Tech Preview 13.2, fixes scrolling issues on the rustdoc portion of the page when it's focused (as happens automatically whenever the page is navigated to) and stops the page from jumping around when switched to.

Tested on some pages in production via userstyle injection as well, to see that it works with pages built with older rustdoc.

fixes #725

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -294,6 +294,10 @@ div.nav-container-rustdoc {
     z-index: 999;
 }
 
+.rustdoc:focus {
+    outline: unset;
+}
+
 div.landing {
     text-align: center;
     padding-top: 30px;


jyn514:
r? @GuillaumeGomez 

cc https://github.com/rust-lang/docs.rs/issues/595 which may be related.

GuillaumeGomez:
I don't have any apple device so I can only trust you on this one. Let's merge then...

PR #762, opened 2020-05-18
Remove local usage of deprecated tempdir

Nemo157:
see #761


jyn514:
We still use `tempdir` because of `params`, so I think it makes more sense to wait until we switch to hyper. Otherwise we'll compile both tempdir and tempfile.

```
$ cargo tree -i -p tempdir
tempdir v0.3.7
├── multipart v0.13.6
│   └── params v0.8.0
│       └── cratesfyi v0.6.0 (/home/joshua/src/docs.rs)
└── params v0.8.0 (*)
```

Nemo157:
We already use `tempfile` too (not at my checkout to check the tree).

jyn514:
You're quite right, we use it through Rustwide.

PR #758, opened 2020-05-14
Update asynchronous dependencies

Kixiron:
Step one of #747

* Updated `tokio` and added the `rt-threaded` feature
* Updated `futures`
* Updated `rusoto_s3`, `rusoto_core` and `rusoto_credential`

PR #768, opened 2020-05-22
Revert 758 Update async dependencies

jyn514:
This caused a production outage, then a second outage after the server restarted.

cc @kixiron


jyn514:
For posterity: #758

PR #769, opened 2020-05-23
Don't require the db connection pool to sit at maximum when idle

Nemo157:


src/db/mod.rs
```diff
@@ -28,19 +28,27 @@ pub fn connect_db() -> Result<Connection, failure::Error> {
...
         })
         .unwrap_or(90);
 
+    let min_pool_idle = env::var("DOCSRS_MIN_POOL_IDLE")
+        .map(|s| {
+            s.parse::<u32>()
+                .expect("DOCSRS_MIN_POOL_IDLE must be an integer")
+        })
+        .unwrap_or(10);
+
```
Kixiron, line 44: Maybe this should be kept an `Option` and then later passed to `.min_idle` as an `Option<u32>` so that prod keeps the same behavior it's been having

src/web/metrics.rs
```diff
@@ -77,6 +77,12 @@ lazy_static::lazy_static! {
         "The number of currently used database connections"
     )
     .unwrap();
+
+    pub static ref TOTAL_DB_CONNECTIONS: IntGauge = register_int_gauge!(
+        "docsrs_db_total_connections",
+        "The number of currently active + idle database connections"
+    )
+    .unwrap();
```
Kixiron, line 85: This is being expounded on in #770

src/db/mod.rs
Nemo157: I don't think having the idle set to the max as a default makes sense, you never really want the connections sitting around idle if they're never used, with the default 10 minute timeout before idle connections expire you essentially have the pool sit at the maximum used in the last 10 minutes + this value extra connections.

jyn514:
For posterity: the reason this change is necessary is so it's possible to run tests while the web server is running.

PR #770, opened 2020-05-23
More metrics

Kixiron:
* `docsrs_used_db_connections` Used database connections
* `docsrs_idle_db_connections` Idle database connections
* `docsrs_open_file_descriptors` Number of opened file descriptors
* `docsrs_running_threads` Number of threads used by docs.rs

### DIFF
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -104,6 +104,7 @@ lazy_static::lazy_static! {
 }
 
 pub fn metrics_handler(req: &mut Request) -> IronResult<Response> {
+    // Database calls are scoped in order to minimize the time a db connection is locked
     {
         let pool = extension!(req, Pool);
         let conn = pool.get()?;
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -78,9 +78,15 @@ lazy_static::lazy_static! {
     )
     .unwrap();
 
-    pub static ref CONCURRENT_DB_CONNECTIONS: IntGauge = register_int_gauge!(
-        "docsrs_db_connections",
-        "The number of currently used database connections"
+    pub static ref USED_DB_CONNECTIONS: IntGauge = register_int_gauge!(
+        "docsrs_used_db_connections",
+        "The number of used database connections"
+    )
+    .unwrap();
+
+    pub static ref IDLE_DB_CONNECTIONS: IntGauge = register_int_gauge!(
+        "docsrs_idle_db_connections",
+        "The number of idle database connections"
     )
     .unwrap();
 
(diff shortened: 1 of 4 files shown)


jyn514:
Is the metric for number of connections still needed now that #769 is merged?

src/web/metrics.rs
```diff
@@ -72,29 +78,64 @@ lazy_static::lazy_static! {
...
+        "docsrs_idle_db_connections",
+        "The number of idle database connections"
+    )
+    .unwrap();
+
+    pub static ref OPEN_FILE_DESCRIPTORS: IntGauge = register_int_gauge!(
+        "docsrs_open_file_descriptors",
+        "The number of currently opened file descriptors"
+    )
+    .unwrap();
```
jyn514, line 97: What's the reason for monitoring this in the application directly instead of through the machine it's running on? It seems a little unreliable to try and measure it ourselves, and going through procfs can just as easily be done on the host.

src/web/metrics.rs
```diff
@@ -72,29 +78,64 @@ lazy_static::lazy_static! {
...
-    );
-    FAILED_CRATES_COUNT.set(
-        ctry!(conn.query("SELECT COUNT(*) FROM queue WHERE attempt >= 5;", &[]))
-            .get(0)
-            .get(0),
-    );
-
-    let pool = extension!(req, Pool);
-    CONCURRENT_DB_CONNECTIONS.set(pool.connections() as i64);
+    {
```
jyn514, line 107: This new scope doesn't seem necessary.

Kixiron:
Yes, that’s why it was added. Seeing the number of database connections is still very relevant

src/web/metrics.rs
Kixiron: It minimizes the time that a db connection is being used, which is something we should be doing more often

src/web/metrics.rs
Kixiron: I looked into having `node_exporter` do it, but it only gave system wide statistics. See discord for more discussion

src/web/metrics.rs
```diff
@@ -72,29 +78,64 @@ lazy_static::lazy_static! {
     )
     .unwrap();
 
-    pub static ref CONCURRENT_DB_CONNECTIONS: IntGauge = register_int_gauge!(
-        "docsrs_db_connections",
-        "The number of currently used database connections"
+    pub static ref USED_DB_CONNECTIONS: IntGauge = register_int_gauge!(
+        "docsrs_used_db_connections",
```
pietroalbini, line 82: Let's not rename metrics if possible.

Kixiron: It’s renamed because it does something differently now, only reporting the number of currently used connections instead of the total number

src/web/metrics.rs
jyn514: Gotcha, can you add a comment to that effect?

jyn514:
LGTM if @pietroalbini approves.

PR #773, opened 2020-05-26
Reduce crates.io references

Nemo157:
Preparation for supporting non-crates.io registries, stop referring to stuff that isn't crates.io specific as crates.io. Also moved stuff accessing the index API out of the database code to integrate it with future index related code.

(I still haven't figured out how to handle `CRATESFYI_PREFIX` nicely, but that's independent of these changes).

### DIFF
--- a/src/index/api.rs
+++ b/src/index/api.rs
@@ -1,0 +1,162 @@
+use std::io::Read;
+
+use crate::{error::Result, utils::MetadataPackage};
+
+use failure::err_msg;
+use reqwest::{header::ACCEPT, Client};
+use rustc_serialize::json::Json;
+use time::Timespec;
+
+pub(crate) struct RegistryCrateData {
+    pub(crate) release_time: Timespec,
+    pub(crate) yanked: bool,
+    pub(crate) downloads: i32,
+    pub(crate) owners: Vec<CrateOwner>,
+}
+
+pub(crate) struct CrateOwner {
+    pub(crate) avatar: String,
+    pub(crate) email: String,
+    pub(crate) login: String,
+    pub(crate) name: String,
+}
+
+impl RegistryCrateData {
+    pub(crate) fn get_from_network(pkg: &MetadataPackage) -> Result<Self> {
+        let (release_time, yanked, downloads) = get_release_time_yanked_downloads(pkg)?;
+        let owners = get_owners(pkg)?;
+
+        Ok(Self {
+            release_time,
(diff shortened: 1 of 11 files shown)


src/index/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod api;
```
pietroalbini, line 1:
```suggestion
pub(crate) mod api;
```

src/index/api.rs
```diff
@@ -0,0 +1,162 @@
...
+            yanked,
+            downloads,
+            owners,
+        })
+    }
+}
+
+/// Get release_time, yanked and downloads from the registry's API
+fn get_release_time_yanked_downloads(pkg: &MetadataPackage) -> Result<(time::Timespec, bool, i32)> {
+    let url = format!("https://crates.io/api/v1/crates/{}/versions", pkg.name);
```
pietroalbini, line 40: This URL won't work if, for example, we point the registry to staging.

Each index has a `config.json` file [has an optional `api` key](https://rust-lang.github.io/rfcs/2141-alternative-registries.html#registry-index-format-specification) that contains the URL to a crates.io-like API (see [crates.io's `config.json`](https://github.com/rust-lang/crates.io-index/blob/master/config.json) for an example). This struct should get the URL from the index config and use it as the prefix for network requests. If the `api` key is missing we should probably abort.

src/index/api.rs
```diff
@@ -0,0 +1,162 @@
...
+        release_time.unwrap_or_else(time::get_time),
+        yanked.unwrap_or(false),
+        downloads.unwrap_or(0),
+    ))
+}
+
+/// Fetch owners from the registry's API
+fn get_owners(pkg: &MetadataPackage) -> Result<Vec<CrateOwner>> {
+    // owners available in: https://crates.io/api/v1/crates/rand/owners
+    let owners_url = format!("https://crates.io/api/v1/crates/{}/owners", pkg.name);
```
pietroalbini, line 107: Same as the previous comment.

src/index/api.rs
Nemo157: Yep, I have another branch that handles all that, I just want to get it in piece by piece to avoid having a massive PR (and I still need to look into the `rustwide` changes necessary).

pietroalbini: Sounds good!

pietroalbini:
Thanks!

PR #774, opened 2020-05-26
Added max pool size metric

Kixiron:


jyn514:
r? @pietroalbini 

This looks good to me, though. Context from discussions on discord:
> pietro: a docsrs_max_db_connections metric would be useful so we can setup an alert for 75% used database connections (I don't want to hardcode 90 connections in the monitoring). The current idle+used is not 90.

pietroalbini:
Thanks!

PR #779, opened 2020-05-27
RFC: rework index updates

pietroalbini:
[Rendered](https://github.com/pietroalbini/docs.rs/blob/rfc-rework-index-updates/rfcs/rework-index-updates.md)

### DIFF
--- a/rfcs/rework-index-updates.md
+++ b/rfcs/rework-index-updates.md
@@ -1,0 +1,60 @@
+# Summary
+
+This RFC proposes moving away from the current model of fetching new releases
+to build, moving from running [crates-index-diff] in a thread to using webhooks
+and the [crates-index] crate.
+
+# Motivation
+
+While the current approach has worked well for us so far, it has some problems:
+
+* Running the update in a cronjob every 2 minutes is wasteful, as there is
+  often a greater delay between two publishes.
+* Running the update in a crobjob every 2 minutes adds delay to getting the
+  documentation built if the queue is empty, as the release might potentially
+  have to wait those extra two minutes.
+* The approach doesn't scale, if we want to move to a setup where there is more
+  than a single frontend server we'd have to elect which server runs the fetch.
+* The way crates-index-diff stores its state (a branch in the local repo) is
+  fragile, as it might become out of sync causing the loss of a publish.
+* The way crates-index-diff stores its state makes it hard to move the server
+  installation, as the index repository needs to be moved as well.
+
+# Proposal
+
+We configure the `crates.io-index` repository to send a webhook to a new
+endpoint, `/_/index-webhook`, which starts a index sync in the background. The
+payload of the webhook is ignored, but the webhook signature is validated if a
+secret key is provided to the application through an environment variable.
+
+When an index synchronization starts, the [crates-index] crate is used to load
+in memory a list of all crates, their versions and whether each version is
+yanked. Then, the full list of releases and queued crates is fetched from the
+database, and it's compared with the contents of the index. Finally, idempotent
+queries are sent to the database to update its state (queueing crates and
+changing the yanked status) where needed.
+
+# Rationale of the proposal
+
+This proposal removes the cronjob and implements realtime updates of the index,
+which does not have to happen on a specific machine if we ever move to multiple
+frontend servers.
+
+This proposal also works if multiple index synchronizations start at the
+same time (for example, if two requests are received at the same time) without
+having to implement a job queue: since all the updates to the database are
+idempotent multiple syncs at the same time would not affect each other
+(provided we structure the SQL queries the right way). A single mutex on each
+host to lock `git fetch`es on the index might be needed though.
+
+# Alternatives
+
+We could implement only the webhook or the index synchronization, keeping the
+old code for the part we don't replace. While it would improve the status quo,
+it wouldn't address all the problems noted in the motivation.
+
+We could also do nothing: while the current system is not perfect it works
+without much trouble.
+
+[crates-index-diff]: https://crates.io/crates/crates-index-diff
+[crates-index]: https://crates.io/crates/crates-index
--- a/rfcs/rework-index-updates.md
+++ b/rfcs/rework-index-updates.md
@@ -1,20 +1,17 @@
 # Summary
 
 This RFC proposes moving away from the current model of fetching new releases
-to build, moving from running [crates-index-diff] in a thread to using webhooks
-and the [crates-index] crate.
+to build, moving from using a timer to receiving webhooks.
 
 # Motivation
 
 While the current approach has worked well for us so far, it has some problems:
 
-* Running the update in a cronjob every 2 minutes is wasteful, as there is
+* Running the update in a timer every 2 minutes is wasteful, as there is
   often a greater delay between two publishes.
-* Running the update in a crobjob every 2 minutes adds delay to getting the
+* Running the update in a timer every 2 minutes adds delay to getting the
   documentation built if the queue is empty, as the release might potentially
   have to wait those extra two minutes.
-* The approach doesn't scale, if we want to move to a setup where there is more
-  than a single frontend server we'd have to elect which server runs the fetch.
 * The way crates-index-diff stores its state (a branch in the local repo) is
   fragile, as it might become out of sync causing the loss of a publish.
 * The way crates-index-diff stores its state makes it hard to move the server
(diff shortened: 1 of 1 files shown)


rfcs/rework-index-updates.md
```diff
@@ -0,0 +1,60 @@
...
+endpoint, `/_/index-webhook`, which starts a index sync in the background. The
+payload of the webhook is ignored, but the webhook signature is validated if a
+secret key is provided to the application through an environment variable.
+
+When an index synchronization starts, the [crates-index] crate is used to load
+in memory a list of all crates, their versions and whether each version is
+yanked. Then, the full list of releases and queued crates is fetched from the
+database, and it's compared with the contents of the index. Finally, idempotent
+queries are sent to the database to update its state (queueing crates and
+changing the yanked status) where needed.
```
jyn514, line 35: This sounds really expensive. We currently have `238677` crates in the database and just running `select name, version, yanked from crates inner join releases on crates.id = releases.crate_id` takes a full second in production. Comparing it to the index sounds even slower. I would rather have the webhook contain the metadata of the crate that's been uploaded and separate the webhook from synchronization. Synchronization is still a good idea, I just don't think it should happen on every release (maybe every week or so?).

Possible alternatives to full synchronization:
- The database is the one true source of knowledge and if we crash while processing a webhook, the crate will be lost until we do another full synchronization. I don't expect crashes to be common, so as long as we have _any_ synchronization I think this is reasonable.
- The webhook has some sort of two-way communication and resends the payload until it gets a confirmation that it was received/committed
- The above, but the confirmation is 'docs.rs closed the HTTP connection' and we make sure we write the payload to disk before we do so.

rfcs/rework-index-updates.md
```diff
@@ -0,0 +1,60 @@
...
+
+This RFC proposes moving away from the current model of fetching new releases
+to build, moving from running [crates-index-diff] in a thread to using webhooks
+and the [crates-index] crate.
+
+# Motivation
+
+While the current approach has worked well for us so far, it has some problems:
+
+* Running the update in a cronjob every 2 minutes is wasteful, as there is
```
jyn514, line 11: nit: it's not actually a cronjob, we spawn a thread every minute.

https://github.com/rust-lang/docs.rs/blob/d6baf144a72edca646240507707b9853e39696a4/src/utils/daemon.rs#L64

rfcs/rework-index-updates.md
```diff
@@ -0,0 +1,60 @@
...
+  fragile, as it might become out of sync causing the loss of a publish.
+* The way crates-index-diff stores its state makes it hard to move the server
+  installation, as the index repository needs to be moved as well.
+
+# Proposal
+
+We configure the `crates.io-index` repository to send a webhook to a new
+endpoint, `/_/index-webhook`, which starts a index sync in the background. The
+payload of the webhook is ignored, but the webhook signature is validated if a
+secret key is provided to the application through an environment variable.
```
jyn514, line 28: Not sure I quite follow the bit about secret keys, is this like a public/private key or a symmetric key? Also, I assume that if no secret key is provided, it fails open so you can still update the index while developing locally?

Kixiron: It's a thing with github, it shouldn't really affect local development. You give GH a key, they hash the content of their `POST` with that key and then you verify that the key used for hashing matches the one you hold

rfcs/rework-index-updates.md
```diff
@@ -0,0 +1,60 @@
...
+
+While the current approach has worked well for us so far, it has some problems:
+
+* Running the update in a cronjob every 2 minutes is wasteful, as there is
+  often a greater delay between two publishes.
+* Running the update in a crobjob every 2 minutes adds delay to getting the
+  documentation built if the queue is empty, as the release might potentially
+  have to wait those extra two minutes.
+* The approach doesn't scale, if we want to move to a setup where there is more
+  than a single frontend server we'd have to elect which server runs the fetch.
```
Nemo157, line 17: Not necessarily, it seems like it'd be relatively easy to make it so that two services racing to add a crate to the queue only add it once (in fact, that's mentioned as being part of the implementation of the proposed method as well).

rfcs/rework-index-updates.md
```diff
@@ -0,0 +1,60 @@
...
+* The way crates-index-diff stores its state (a branch in the local repo) is
+  fragile, as it might become out of sync causing the loss of a publish.
+* The way crates-index-diff stores its state makes it hard to move the server
+  installation, as the index repository needs to be moved as well.
```
Nemo157, line 21: Given #766 being implemented, I think it's fine to discard the state and start the new server from current `HEAD` when initialized. If the new server starts up while the old server is still running so that the old server queues stuff up until the `HEAD` of the new server then there will be no missed crates, otherwise a consistency check can be run afterwards to queue missed crates.

rfcs/rework-index-updates.md
pietroalbini: Yep, that's to avoid people spamming that URL manually.

rfcs/rework-index-updates.md
pietroalbini: > Comparing it to the index sounds even slower.

On my machine, collecting all the crates in the index takes only 2 seconds [(source)](https://gist.github.com/pietroalbini/6e069454498d5028dc19d5ea4491138f). Considering how rarely crates are published I think this is not so bad, and I feel like the advantages outweight the performance issues.

> The database is the one true source of knowledge and if we crash while processing a webhook, the crate will be lost until we do another full synchronization. I don't expect crashes to be common, so as long as we have any synchronization I think this is reasonable.

This is the status quo, right?

> I would rather have the webhook contain the metadata of the crate that's been uploaded and separate the webhook from synchronization.
> * The webhook has some sort of two-way communication and resends the payload until it gets a confirmation that it was received/committed
> * The above, but the confirmation is 'docs.rs closed the HTTP connection' and we make sure we write the payload to disk before we do so.

We don't control GitHub webhooks, and we can't really do this.

rfcs/rework-index-updates.md
pietroalbini: That's what happens when I write those late at night.

rfcs/rework-index-updates.md
pietroalbini: If we end up running docs.rs on ECS (which is a thing I'd *love* to see) starting a new server from scratch might happen every day or even more often (every time we deploy and every time ECS feels like doing that), so we could actually miss a lot of crates while we wait for synchronization.

rfcs/rework-index-updates.md
pietroalbini: Hmm, that's true, even though races would happen every time with the current solution instead of just when two crates are published at the same time with this RFC.

rfcs/rework-index-updates.md
Nemo157: Ok. Another alternative would be for us to store an extra copy of the last seen state in the database ourselves, and use that to initialize a new server.

rfcs/rework-index-updates.md
Nemo157: If we go for the hybrid "use webhooks to trigger the diff" approach then I feel like we get best of both worlds, minimal updates being performed, and minimal work to do each update.

rfcs/rework-index-updates.md
jyn514: > On my machine, collecting all the crates in the index takes only 2 seconds (source). Considering how rarely crates are published I think this is not so bad, and I feel like the advantages outweight the performance issues.

Hmm, in the common case that doesn't seem outrageous, but I think it will slow down the server a lot when we get 300 crates from the same project published in a row. I'm also not convinced that the advantages are that much greater than running a full synchronization on a timer occasionally - do we really need to make sure we build _every_ crate _immediately_ after publication, even if the site just went down?

> > The database is the one true source of knowledge and if we crash while processing a webhook, the crate will be lost until we do another full synchronization. I don't expect crashes to be common, so as long as we have any synchronization I think this is reasonable.

> This is the status quo, right?

With two major differences:
1. Updates are push instead of poll, so we no longer run `git fetch` on the index every minute
2. We have full synchronization on a timer, so that crates will no longer be lost if we miss an update somehow.

> We don't control GitHub webhooks, and we can't really do this.

That's a shame, but that makes sense.

rfcs/rework-index-updates.md
jyn514: I like the idea of state being stored in the database instead of on the host. I think it would make it easier to have multiple build servers eventually: instead of each server having to have it's own fully synchronized index, each build server is stateless and builds crates when it gets a (sufficiently authenticated) webhook. Then, we only need one version of the index that's synchronized on the database, and if we find during synchronization that we missed a crate, we can just send out a webhook ourselves.

pietroalbini:
Pushed an update addressing the previous comments, and changing the approach.

Kixiron:
Is it overkill to manually update the index every `x` hours if no update is received from crates.io?

jyn514:
> Is it overkill to manually update the index every `x` hours if no update is received from crates.io?

I was thinking instead we could do a full synchronization every x hours/days, which avoids our index getting out of sync with reality.

Issue #379, opened 2019-07-18
Compress documentation uploaded to S3


pietroalbini:
At the moment we don't compress the documentation uploaded to S3, wasting a lot of space. While money for S3 isn't an issue at the moment, avoiding compression could hurt docs.rs's sustainability in the future.

~~I ran some very rough benchmarks on `reqwest 0.9.3`, compressing each `.html` file separately:~~

<details><summary>Bad benchmark</summary>

| Algorithm | Size | Compression time | Decompression time | Options |
| --- | --- | --- | --- | --- |
| *Plaintext* | 33.9 MB | - | - | - |
| Gzip | 12.0 MB | 3.2s | 3.4s | `-9` (best) |
| Gzip | 12.8 MB | 2.5s | 3.4s | `-1` (fast) |
| Zstd | 11.7 MB | 7.8s | 2.4s | `-19` (best) |
| Zstd | 12.5 MB | 2.5s | 2.4s | `-1` (fast) |
| Brotli | 11.5 MB | 5.5s | 2.3s | `-9` (best) |
| Brotli | 13.0 MB | 2.3s | 2.3s | `-0` (fast) |

</details>

~~Looking at the numbers, on average if we compress the uploaded docs we're going to save 63% of storage space, which is great from a sustainability point of view. I think we should compress all the uploaded docs going forward, and try to compress (part of) the initial import as well.~~

~~For the algoritm choice, I'd say we can go with `gzip`: there isn't much difference between the resulting sizes and the compression time delta between gzip's fast and best modes is the smallest. We can compress the initial import with `-1` to speed it up, and all the new crates with `-9`.~~

cc @Mark-Simulacrum @QuietMisdreavus 

<details><summary>Benchmark method</summary>

Installed compression tools on Ubuntu 18.04 LTS:

```
$ sudo apt install gzip brotli zstd
```

Downloaded locally the reqwest documentation:

```
$ aws s3 cp --recursive s3://rust-docs-rs/rustdoc/reqwest/0.9.3/ .
```

Compressed every `.html` file with `find`:

```
$ time find <dir> -name "*.html" -exec <command> {} \;
```

</details>

Mark-Simulacrum:
Could you populate decompression times as well? In some sense those are considerably more important for us to decide whether this is viable. Compression times for all of those are not great though :/

GuillaumeGomez:
[...]

PR #780, opened 2020-05-28
Add compression for uploaded documentation

jyn514:
This needs some tests, but otherwise I think it should be fine as is, it was super simple after #643 :)

This uses `zstd 5`, but it's very easy to change the compression now and reasonably easy to change it after it's merged (we just have to change `compression` from a boolean into an enum and continue supporting `zstd`).

This compresses transparently, so that calling `backend.get()` automatically decompresses files as needed. If the files were not compressed before uploading, they are not decompressed when downloading.

Closes https://github.com/rust-lang/docs.rs/issues/379

cc @namibj
r? @pietroalbini

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -348,25 +348,21 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             "Add compression",
             // upgrade query
             "
-            CREATE TABLE compression (
-                id SERIAL PRIMARY KEY,
-                name VARCHAR(100)
-            );
-            INSERT INTO compression (name) VALUES ('zstd');
-            -- NULL indicates the file was not compressed
-            ALTER TABLE files ADD COLUMN compression INT REFERENCES compression(id);
+            -- NULL indicates the file was not compressed.
+            -- There is no meaning assigned to the compression id in the database itself,
+            -- it is instead interpreted by the application.
+            ALTER TABLE files ADD COLUMN compression INT;
             -- many to many table between releases and compression
             -- stores the set of all compression algorithms used in the release files
             CREATE TABLE compression_rels (
                 release INT NOT NULL REFERENCES releases(id),
-                algorithm INT REFERENCES compression(id),
+                algorithm INT,
                 -- make sure we don't store duplicates by accident
                 UNIQUE(release, algorithm)
             );",
             // downgrade query
             "DROP TABLE compression_rels;
-             ALTER TABLE files DROP COLUMN compression;
-             DROP TABLE compression;",
+             ALTER TABLE files DROP COLUMN compression;"
         ),
     ];
 
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -365,6 +365,7 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             );",
             // downgrade query
             "DROP TABLE compression_rels;
+             ALTER TABLE files DROP COLUMN compression;
              DROP TABLE compression;",
         ),
     ];
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -340,6 +340,17 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
              ADD COLUMN content tsvector,
              ADD COLUMN versions JSON DEFAULT '[]';"
         ),
+        migration!(
+            context,
+            // version
+            14,
+            // description
+            "Add a field for compression",
+            // upgrade query
+            "ALTER TABLE files ADD COLUMN compressed BOOLEAN NOT NULL DEFAULT false;",
+            // downgrade query
+            "ALTER TABLE files DROP COLUMN compressed;",
+        ),
     ];
 
     for migration in migrations {
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -345,11 +345,23 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             // version
             14,
             // description
-            "Add a field for compression",
+            "Add compression",
             // upgrade query
-            "ALTER TABLE files ADD COLUMN compression VARCHAR;",
+            "
+            CREATE TABLE compression (
+                id SERIAL PRIMARY KEY,
+                name VARCHAR(100)
+            );
+            INSERT INTO compression (name) VALUES ('zstd');
+            -- many to many table between releases and compression
+            -- stores the set of all compression algorithms used in the release files
+            CREATE TABLE compression_rels (
+                release INT NOT NULL REFERENCES releases(id),
+                algorithm INT NOT NULL REFERENCES compression(id)
+            );",
             // downgrade query
-            "ALTER TABLE files DROP COLUMN compression;",
+            "DROP TABLE compression_rels;
+             DROP TABLE compression;",
         ),
     ];
 
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -353,11 +353,15 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
                 name VARCHAR(100)
             );
             INSERT INTO compression (name) VALUES ('zstd');
+            -- NULL indicates the file was not compressed
+            ALTER TABLE files ADD COLUMN compression INT REFERENCES compression(id);
             -- many to many table between releases and compression
             -- stores the set of all compression algorithms used in the release files
             CREATE TABLE compression_rels (
                 release INT NOT NULL REFERENCES releases(id),
-                algorithm INT NOT NULL REFERENCES compression(id)
+                algorithm INT REFERENCES compression(id),
+                -- make sure we don't store duplicates by accident
+                UNIQUE(release, algorithm)
             );",
             // downgrade query
             "DROP TABLE compression_rels;
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -347,9 +347,9 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
             // description
             "Add a field for compression",
             // upgrade query
-            "ALTER TABLE files ADD COLUMN compressed BOOLEAN NOT NULL DEFAULT false;",
+            "ALTER TABLE files ADD COLUMN compression VARCHAR;",
             // downgrade query
-            "ALTER TABLE files DROP COLUMN compressed;",
+            "ALTER TABLE files DROP COLUMN compression;",
         ),
     ];
 
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -312,6 +312,21 @@ mod test {
     }
 
     #[test]
+    fn test_compression() {
+        let orig = "fn main() {}";
+        let compressed = compress(orig.as_bytes()).unwrap();
+        let blob = Blob {
+            mime: "text/rust".into(),
+            content: compressed.clone(),
+            path: "main.rs".into(),
+            date_updated: Timespec::new(42, 0),
+            compressed: true,
+        };
+        test_roundtrip(std::slice::from_ref(&blob));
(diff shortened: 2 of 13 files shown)


jyn514:
Oh, I just realized this has the wrong content-types on s3: It uses the original content types instead of `zstd` or something like that. I'll add the original content type as a field in the metadata.

jyn514:
Actually that seems like a lot of work for no benefit, I don't know that we'll ever actually use the `zstd` encoding for anything.

jyn514:
Some benchmarks:
```
Benchmarking compress regex html: Collecting 100 samples in estimated 7.5520 s (10k i                                                                                     compress regex html     time:   [676.22 us 725.12 us 779.81 us]
Found 16 outliers among 100 measurements (16.00%)
  16 (16.00%) high severe

Benchmarking decompress regex html: Collecting 100 samples in estimated 5.1476 s (91k                                                                                     decompress regex html   time:   [55.181 us 57.623 us 60.467 us]
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```

src/db/migrate.rs
```diff
@@ -325,6 +325,17 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
...
+            context,
+            // version
+            13,
+            // description
+            "Add a field for compression",
+            // upgrade query
+            "ALTER TABLE files ADD COLUMN compressed BOOLEAN NOT NULL DEFAULT false;",
+            // downgrade query
+            "ALTER TABLE files DROP COLUMN compressed;",
+        ),
```
Kixiron, line 338: I still think that storing this in the s3 header is the way to go

src/storage/s3.rs
```diff
@@ -51,19 +51,25 @@ impl<'a> S3Backend<'a> {
...
     }
 
     pub(super) fn store_batch(&mut self, batch: &[Blob]) -> Result<(), Error> {
         use futures::stream::FuturesUnordered;
         use futures::stream::Stream;
+        use std::collections::HashMap;
+
         let mut attempts = 0;
+        let mut compressed = HashMap::with_capacity(1);
+        compressed.insert("compressed".into(), "true".into());
```
Nemo157, line 72: Rather than just compressed or not, this should store the encoding used, so that it’s possible to move to different encodings in the future.

Nemo157: In fact, it might make a lot of sense to use the `Content-Encoding` header instead if that’s something that S3 retains.

src/db/migrate.rs
pietroalbini: For S3 yes, this is just for the database storage backend.

pietroalbini:
The implementation looks good to me!

I agree with @Nemo157 that we should store `{algorithm}` instead of just `compressed`, and that we should use S3's native `Content-Encoding` to store it instead of a custom metadata.

Also, based on https://github.com/rust-lang/docs.rs/issues/379#issuecomment-512834913 the two options I was considering were `zstd 9` or `brotli 5` leaning torwards brotli as it uses 10% less disk space in that benchmark. You're using `zstd 5` in the PR.

src/storage/s3.rs
jyn514: `Content-Encoding` is a great idea, I didn't know it existed!

pietroalbini: The `rusoto_s3` field is this: https://docs.rs/rusoto_s3/0.43.0/rusoto_s3/struct.PutObjectRequest.html#structfield.content_encoding

Nemo157:
The other option that @namibj was investigating was zstd with a custom dictionary optimized for rustdoc output, I can probably adapt the linked benchmarking script to test that as well with a dictionary created from the docs I have.

src/storage/s3.rs
jyn514: Added in https://github.com/rust-lang/docs.rs/pull/780/commits/f13a08783b97d6af93e6603e1a8298d1168af09f

jyn514:
Oops, the benchmarks before got cut off. Here are some taken with a desktop instead of a laptop:
```
compress regex html     time:   [715.28 us 715.74 us 716.27 us]                                
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

decompress regex html   time:   [37.410 us 37.438 us 37.470 us]                                   
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe
```

Nemo157:
Another idea might be to store a set of all the encodings used for a particular release into the `release` table, so we can easily lookup basic stats on compression usage without needing to go to S3. (A set to allow for different dictionaries based on filetype to be recorded).

jyn514:
> store a set of all the encodings used for a particular release into the release table

This sounds like we'd only use it for metrics? I'm not opposed in principle but storing the same data in two places risks it getting out of sync.

Kixiron:
Metrics don’t need to be a source of complexity, we can just as easily have an `IntCounter` that we increment with the total bytes saved by each build (plus additional if we’d like)

Nemo157:
It's not really for metrics like what goes in grafana, more for future visibility into usage distribution. Especially with if we migrate dictionaries in the future knowing how many releases use which dictionaries could be useful to determine whether it seems feasible to do something like migrate all instances of a certain dictionary to a new one to remove it from the application.

Given that releases are basically read-only currently I don't think there's much chance of this data getting out of sync.

Kixiron:
How would that interact with us deleting docs in the future (if we ever decided to do that)?

jyn514:
> How would that interact with us deleting docs in the future (if we ever decided to do that)?

No effect, we'd just delete the rows out of the database as well.

jyn514:
> Another idea might be to store a set of all the encodings used for a particular release into the release table, so we can easily lookup basic stats on compression usage without needing to go to S3. (A set to allow for different dictionaries based on filetype to be recorded).

@Nemo157 a possible issue with this is it prevents using different compression algorithms for different files within a release. Does that sound worth it? Otherwise we'd have to store metadata for each file individually in the database which sort of the defeats the point of an S3 bucket.

Nemo157:
That's why I mention a set of encodings, we don't need to know exactly which files use which encoding, just that having this set of encodings is enough to access all files in the release (and if we want to know in detail then we would have to query the metadata on all the files).

jyn514:
I added the set of compression algorithms used as a many-to-many-table in the database, with a unique constraint so we don't end up with duplicates by accident.

I'm open to suggestions for more tests but otherwise I think I addressed all the comments.

jyn514:
As to using a custom zstd dictionary: it sounds like the rust `zstd` crate needs some fairly major changes to make that performant, I don't want to delay having any compression at all on that. If we want to use brotli instead for the storage that sounds reasonable but otherwise I think `zstd 9` is fine.

src/db/migrate.rs
```diff
@@ -340,6 +340,33 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
...
+            -- stores the set of all compression algorithms used in the release files
+            CREATE TABLE compression_rels (
+                release INT NOT NULL REFERENCES releases(id),
+                algorithm INT REFERENCES compression(id),
+                -- make sure we don't store duplicates by accident
+                UNIQUE(release, algorithm)
+            );",
+            // downgrade query
+            "DROP TABLE compression_rels;
+             DROP TABLE compression;",
```
Nemo157, line 368: This seems like it would fail given `files.compression` still exists.

Nemo157:
I think it definitely makes sense to start with `zstd` now and add custom dictionary support later. Going with brotli now and moving to zstd custom dictionary later would require keeping both libraries around indefinitely.

Doing it in two steps also gives us the chance to see any performance impacts of each step.

One last thought I had was whether there should be a config var to disable compression, so if there are any issues in production that can just be toggled rather than having to do a rollback.

src/db/migrate.rs
jyn514: Good catch, fixed.

jyn514:
> One last thought I had was whether there should be a config var to disable compression, so if there are any issues in production that can just be toggled rather than having to do a rollback.

It doesn't take very long to revert, about 5 minutes. In general I like the idea of having feature gates instead of "it's compiled in or it's not" but I'm a little wary of toggling features with environment variables, I'd rather come up with something more principled.

src/storage/mod.rs
```diff
@@ -6,20 +6,47 @@ pub(crate) use self::s3::S3Backend;
...
+    }
+}
+
+impl fmt::Display for CompressionAlgorithm {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            CompressionAlgorithm::Zstd => write!(f, "zstd"),
+        }
+    }
+}
```
pietroalbini, line 41: I'm not convinced this is the best way to represent the compression algorithm.

A thing I usually strive for when creating this kind of code is making sure there is only a single place where a new variant needs to be added: this way a new contributor can just add the algorithm to the enum and follow the compiler errors to implement full support for it. This advantage is not present here, as the contributor also needs to add the algorithm name in the database through a migration, which is not intuitive nor documented anywhere.

We *could* implement some code that synchronizes the `compression` table when the application starts, but I think we can implement this more cleanly. If we remove the string representation and manually set each ID in the discriminant ([with the help of a macro](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b4f6bc42c2c6caff6377b5ec9f282318)) we'd remove a lot of complexity and make it easier for new contributors to add new variants.

jyn514: The macro is not necessary, you can use `algorithm as i32` instead. Other than that I agree, storing the id would be better than storing a string.

pietroalbini: While that works, there is no builtin conversion from `i32` to `CompressionAlgorithm`.

jyn514: For that, is the duplication so bad? I think `impl TryFrom<i32> for CompressionAlgorithm` would work without obscuring what's happening: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=acde692246a13cddc897fa0e1927d004

jyn514: If you're worried that doesn't give a compile error if you forget to update the `match`, we could use https://crates.io/crates/num_enum.

pietroalbini: Yeah my worry is it doesn't give a compilation error.

jyn514: I added a test which will fail to compile if you don't add an exhaustive match for the new variant, hopefully that's close enough. I'm not sure what this should look like for S3, if we just have a number for the content-encoding it's not super descriptive, but if it's a string we're back to the same issue of needing to update the code in many places.

Nemo157: You could use `strum` to generate the `Display` and `FromStr` impls (though that doesn’t help with syncing with the database).

jyn514: As another alternative I could go with 'update code in many places' and add a test that `FromStr` is exhaustive.

pietroalbini: I think it's fine for S3 to use the string, but I think going with a custom macro is still the best option, especially considered how many (duplicated) impls we have.

jyn514: Ok, I went with the macro-based approach for extensibility. We can always change it later if need be.

jyn514:
Let me know if there are any more changes that need to be made, I think I addressed all the review comments.

jyn514:
I just found that this will prevent us from updating builds if we rebuild a crate:

```
thread 'main' panicked at 'Building documentation failed: Error(Db(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState("23505"), message: "duplicate key value violates unique constraint \"compression_rels_release_algorithm_key\"", detail: Some("Key (release, algorithm)=(1, 0) already exists."), hint: None, position: None, where_: None, schema: Some("public"), table: Some("compression_rels"), column: None, datatype: None, constraint: Some("compression_rels_release_algorithm_key"), file: Some("nbtinsert.c"), line: Some(434), routine: Some("_bt_check_unique") }))', src/bin/cratesfyi.rs:321:21
```

I think I need to add an `ON CONFLICT DO UPDATE` clause.

Issue #782, opened 2020-05-29
Feed page 404


dimlev:
Hi,

The atom feed at https://docs.rs/releases/feed returns a 404 error.

Nemo157:
Seems related to #709 based on local bisection, cc @Kixiron

PR #783, opened 2020-05-29
Fixed release feed bug

Kixiron:
Fixes the 404 with viewing the release feed, #782


Kixiron:
Closing in favor of #784

PR #784, opened 2020-05-29
Fix 404 on /releases/feed

jyn514:
Handlebars removed the `obj[0]` notation in 0.25: https://github.com/sunng87/handlebars-rust/commit/779df9868ebb713300bab3744137a5cb58c529a3#diff-e0934bb6f70a4f8c3baf22a773c14ec7L35

Closes https://github.com/rust-lang/docs.rs/issues/782

r? @Kixiron 

This also uncovered that we handle failure pretty badly in `impl Handler for CratesfyiHandler`, we always return the first error encountered even if it was a 404, ignoring all future errors. This should be fixed, but I want to get this fix in first.

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -670,7 +670,7 @@ pub fn build_queue_handler(req: &mut Request) -> IronResult<Response> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test::wrapper;
+    use crate::test::{assert_success, wrapper};
     use serde_json::json;
 
     #[test]
@@ -1059,5 +1059,13 @@ mod tests {
         });
 
         assert_eq!(correct_json, serde_json::to_value(&release).unwrap());
+    }
+
+    #[test]
+    fn release_feed() {
+        wrapper(|env| {
+            let web = env.frontend();
+            assert_success("/releases/feed", web)
+        })
     }
 }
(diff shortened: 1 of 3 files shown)


jyn514:
Whoops, let me remove the `dbg`

templates/releases_feed.hbs
```diff
@@ -7,7 +7,7 @@
 <link href="https://pubsubhubbub.appspot.com" rel="hub" />
 <link href="https://pubsubhubbub.superfeedr.com" rel="hub" />
 <id>urn:docs-rs:{{cratesfyi_version_safe}}</id>
-<updated>{{content[0].release_time_rfc3339}}</updated>
+<updated>{{content.[0].release_time_rfc3339}}</updated>
```
Kixiron, line 10:
```suggestion
<updated>{{content.0.release_time_rfc3339}}</updated>
```
I'm not sure that `.[0]` is entirely correct syntax

src/web/releases.rs
```diff
@@ -1060,4 +1060,12 @@ mod tests {
 
         assert_eq!(correct_json, serde_json::to_value(&release).unwrap());
     }
+
+    #[test]
+    fn release_feed() {
+        wrapper(|env| {
+            let web = env.frontend();
+            assert_success("/releases/feed", web)
```
pietroalbini, line 1068: Can we also add tests for the content by adding fake releases? This can be done in a followup PR though.

templates/releases_feed.hbs
jyn514: It works correctly, why would `.0` be better?

Kixiron: If it works then it's up to you what you want to do, I've got no idea if one is objectively better or not

src/web/releases.rs
jyn514: Yes, I'll add that in a follow-up.

PR #792, opened 2020-05-30
Updated rustwide to 0.7.1

Kixiron:
As discussed on discord, this is part of an attempt to semi-incrementally re-implement step 1 of [the async RFC](https://github.com/rust-lang/docs.rs/blob/master/rfcs/migrate-to-async-hyper.md) now that the (hopefully) root cause of the crash was solved with `git gc`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,11 @@ version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "base64"
+version = "0.12.1"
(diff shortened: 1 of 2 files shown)


jyn514:
Some more things mentioned:
- This means we will depend on both tokio 0.1 and 0.2, but since we plan to follow it up shortly with the rest of #758, that seems reasonable.
- Since Rustwide uses tokio much less heavily, it is unlikely this will have a performance impact (or at least, not nearly as bad as #758)

PR #793, opened 2020-05-30
Integrate env-logger with the test runners stdout capturing

Nemo157:


Nemo157:
Primary advantage of this is you don't have to see log output from passing tests (unless you pass `--nocapture`), instead only failing tests get shown.

Issue #728, opened 2020-04-23
docs.rs failed to build argmin-0.2.6


stefan-k:
**Crate name:** argmin
**Build failure link:** https://docs.rs/crate/argmin/0.2.6/builds/206318

**Additional details:** 
Hi,

First of all thanks for providing this amazing service! 
It was just pointed out to me that the docs of the [argmin](https://crates.io/crates/argmin) crate were not built successfully. It worked for 0.2.4, but fails since 0.2.5. I have to admit that I'm a bit lost and don't know where to start looking for solutions.
Since the error originates in a dev-dependency, I was wondering if it is possible for me to deactivate dev-dependencies during the docs build. 

This is the relevant part of the build log:

```
[INFO] [stderr] error: failed to run custom build command for `openblas-src v0.7.0`
[INFO] [stderr] 
[INFO] [stderr] Caused by:
[INFO] [stderr]   process didn't exit successfully: `/opt/rustwide/target/debug/build/openblas-src-13fdc4d8599cd912/build-script-build` (exit code: 101)
[INFO] [stderr] --- stdout
[INFO] [stderr] Running: `"cp" "cp" "-R" "source" "source_x86_64-unknown-linux-gnu_tmp"`
[INFO] [stderr] 
[INFO] [stderr] --- stderr
[INFO] [stderr] cp: cannot create directory 'source_x86_64-unknown-linux-gnu_tmp': Read-only file system
[INFO] [stderr] thread 'main' panicked at 'Failed: `"cp" "cp" "-R" "source" "source_x86_64-unknown-linux-gnu_tmp"` (exit code: 1)', /opt/rustwide/cargo-home/registry/src/github.com-1ecc6299db9ec823/openblas-src-0.7.0/build.rs:68:17
[INFO] [stderr] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Any help is highly appreciated! Thanks! :)

jyn514:
Hmm, this is interesting, `cargo +nightly doc --lib --no-deps` doesn't compile `openblas-sys` when I run it locally, but it does in a docs.rs instance.

jyn514:
Aha, the issue is that even though `openblas-sys` is a dev-dependency on master, it is a normal dependency on `argmin 0.2.6`.

stefan-k:
[...]

PR #802, opened 2020-06-02
Only chain on 404 errors, log and return others as 500

### LINKED ISSUE #801: docs.rs reports 404s for internal server errors
Originally found in https://github.com/rust-lang/docs.rs/pull/784.

If we encounter an error in `impl Handler for CratesfyiHandler`, we always return the first error even if it was a 404, ignoring all future errors. This hides panics/errors and makes things very hard to debug (I have a suspicion https://github.com/rust-lang/docs.rs/issues/800 is because of an internal error being hidden by this bug).

Relevant code: https://github.com/rust-lang/docs.rs/blob/9bd032f0ef07b217450cee3fda4906086de9cc93/src/web/mod.rs#L140

Nemo157:
Results in logs like

```
web_1  | 2020/06/02 08:13:07 [ERROR] cratesfyi::web: internal server error: Error rendering "releases_feed" line 10, col 10: Helper not defined: "content"
```

and a page like

image

when I reapply the bug in #728 

Fixes #801

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -10,6 +10,7 @@ pub enum Nope {
     ResourceNotFound,
     CrateNotFound,
     NoResults,
+    InternalServerError,
 }
 
 impl fmt::Display for Nope {
@@ -18,6 +19,7 @@ impl fmt::Display for Nope {
             Nope::ResourceNotFound => "Requested resource not found",
             Nope::CrateNotFound => "Requested crate not found",
             Nope::NoResults => "Search yielded no results",
+            Nope::InternalServerError => "Internal server error",
         })
     }
 }
@@ -58,6 +60,13 @@ impl Handler for Nope {
                         .title("No results given for empty search query")
                         .to_resp("releases")
                 }
+            }
+            Nope::InternalServerError => {
+                // something went wrong, details should have been logged
+                Page::new("internal server error".to_owned())
+                    .set_status(status::InternalServerError)
+                    .title("Internal server error")
+                    .to_resp("error")
             }
         }
     }
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -138,19 +138,24 @@ impl CratesfyiHandler {
 
 impl Handler for CratesfyiHandler {
     fn handle(&self, req: &mut Request) -> IronResult<Response> {
+        fn if_404(
+            e: IronError,
+            handle: impl FnOnce() -> IronResult<Response>,
+        ) -> IronResult<Response> {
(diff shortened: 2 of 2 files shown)


src/web/mod.rs
```diff
@@ -142,22 +142,37 @@ impl Handler for CratesfyiHandler {
...
         self.shared_resource_handler
             .handle(req)
-            .or_else(|e| self.router_handler.handle(req).or(Err(e)))
+            .or_else(|e| {
+                if e.response.status == Some(status::NotFound) {
+                    self.router_handler.handle(req)
+                } else {
+                    Err(e)
+                }
+            })
```
Nemo157, line 151: Suggestions on ways to rewrite this would be appreciated, seems like it needs some sort of conditional `or_else_if(|e| e.response.status == Some(status::NotFound), |e| self.router_handler.handle(req))`.

src/web/error.rs
```diff
@@ -59,6 +61,13 @@ impl Handler for Nope {
                         .to_resp("releases")
                 }
             }
+            Nope::InternalServerError => {
+                // something went wrong, details should have been logged
+                Page::new("internal server error".to_owned())
+                    .set_status(status::InternalServerError)
+                    .title("Internal server error")
+                    .to_resp("error")
```
jyn514, line 69: Do we want to link to the GitHub page so people can make an issue?

src/web/mod.rs
jyn514: I think a helper `handle_if_not_404` closure would be nice.

src/web/error.rs
Nemo157: Might be a good followup issue to do once the template changes are in, all the error pages could do with a bit of a sprucing up.

src/web/mod.rs
jyn514: Something like
```rust
let handle_if_404 = |e, handler| if e.response.status == Some(status::NotFound) {
    handler(req)
} else {
    Err(e)
};
```

src/web/error.rs
pietroalbini: I don't think asking to open an issue on GitHub this way is much beneficial for us: without a backtrace it's hard to then debug those issues.

A thing we could do is generate a random UUID, print it in the same line as the error and show it in the error page, so when people open the issue we can simply do a `journalctl -u docs.rs | grep UUID` to get the exact error the user encountered.

jyn514: The UUID sounds useful but it can go in a follow-up, for now I want to get the bug fix in.

PR #804, opened 2020-06-02
[Refactor] Removed the search index updater

Kixiron:
* Removed the `update-search-index` command and service
* Updating is now relegated to `add_package_into_database`, so changes are reflected instantly
* Removed the unneeded `content` and `versions` columns from the `crates` table
* Made `CrateDetails::new()` use strings for getting query rows instead of indices for readability

Unanswered Questions:
* [`CrateDetails.rustdoc`](https://github.com/rust-lang/docs.rs/compare/master...Kixiron:update-refactor?expand=1#diff-07b7024fe121ba3b5acf483a50bf3000L217) gets its data from [`crates.description_long`](https://github.com/rust-lang/docs.rs/compare/master...Kixiron:update-refactor?expand=1#diff-07b7024fe121ba3b5acf483a50bf3000L126), maybe a rename is in order?

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -27,7 +27,6 @@ while true; do
 done
 set -e
 
-cratesfyi database update-search-index
 cratesfyi database update-release-activity
 
 if ! [ -d "${CRATESFYI_PREFIX}/crates.io-index/.git" ]; then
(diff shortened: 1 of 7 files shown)


jyn514:
> CrateDetails.rustdoc gets its data from crates.description_long, maybe a rename is in order?

Sounds reasonable, I think `rustdoc` should be changed to `description_long`. Let's do that in a separate PR.

Kixiron:
I think that when tera goes through we can actually get away with just storing the entire description and just truncating through templates

PR #805, opened 2020-06-02
Tera groundwork

Kixiron:
* Added the `tera` and `arc-swap` dependencies
* Made `page.rs` into a module and added `templates.rs`, which holds the tera templates and all data that is relevant to them
* Added custom tera filters and functions in `templates.rs`
* Added the `--reload` option for `start-web-server` that activates template reloading


src/bin/cratesfyi.rs
```diff
@@ -52,6 +52,10 @@ enum CommandLine {
     StartWebServer {
         #[structopt(name = "SOCKET_ADDR", default_value = "0.0.0.0:3000")]
         socket_addr: String,
+
+        /// Reload templates when they're changed
+        #[structopt(long = "reload")]
```
pietroalbini, line 57: Nit: maybe let's rename the flag to `--reload-templates`, otherwise it might seem it reloads the whole application.

src/web/page/mod.rs
```diff
@@ -268,3 +263,47 @@ mod tests {
...
+// --- Tera ---
+
+mod templates;
+
+pub(crate) use templates::TemplateData;
+
+lazy_static::lazy_static! {
+    /// Holds all data relevant to templating
+    pub(crate) static ref TEMPLATE_DATA: TemplateData = TemplateData::new().expect("Failed to load template data");
+}
```
pietroalbini, line 276: Can we move all of this at the top of the file, and merge it with the existing imports/mod declarations/statics? It's weird having to find them in the middle of the file.

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    pub fn start_template_reloading() {
+        use std::{sync::Arc, thread, time::Duration};
+
+        thread::spawn(|| loop {
+            match load_templates() {
+                Ok(templates) => {
+                    log::info!("Reloaded templates");
+                    TEMPLATE_DATA.templates.swap(Arc::new(templates));
+                    thread::sleep(Duration::from_secs(10));
```
pietroalbini, line 48: So, the whole reload logic is just reloading everything every 10 seconds? I'm not sure that's the best approach, especially because it takes less than 10 seconds to restart the docs.rs application if it's running outside of docker, so this just makes the experience worse.

I'd expect template reloading to watch for changes in the templates directory, and reload once something is changed. I understand that's more complex to implement, but we can always back it out of this PR and land it in a followup.

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Used to initialize a `TemplateData` instance in a `lazy_static`.
+    /// Loading tera takes a second, so it's important that this is done before any
+    /// requests start coming in
+    pub fn poke(&self) -> Result<()> {
+        Ok(())
+    }
+}
+
+// TODO: Is there a reason this isn't fatal? If the rustc version is incorrect (Or "???" as used by default), then
+//       all pages will be served *really* weird because they'll lack all CSS
```
pietroalbini, line 68: The only reason why this would fail to fetch the rustc version is if it's the first time you start docs.rs, before running `add-essential-files`. We can discuss (probably outside this PR) if that's actually the behavior we want.

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+pub(crate) struct TemplateData {
+    /// The actual templates, stored in an `ArcSwap` so that they're hot-swappable
+    // TODO: Conditional compilation so it's not always wrapped, the `ArcSwap` is unneeded overhead for prod
+    pub templates: ArcSwap<Tera>,
+    /// The current global alert, serialized into a json value
+    global_alert: Value,
+    /// The version of docs.rs, serialized into a json value
+    docsrs_version: Value,
+    /// The current resource suffix of rustc, serialized into a json value
+    resource_suffix: Value,
```
pietroalbini, line 18: It'd be helpful to explain in a comment *why* those values are serialized into a `Value`.

When I started reviewing this PR I wondered by those were encoded as value instead of just using `#[derive(Serialize)]`, but then I noticed you exposed those values as Tera functions (which is quite clever!), and indeed it makes sense those are pre-serialized. Explaining it would aid understanding the code.

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    // TODO: This will fail as long as there are `.hbs` files in `templates/`
+    // #[test]
+    // fn test_templates_are_valid() {
+    //     let tera = load_templates().unwrap();
+    //     tera.check_macro_files().unwrap();
+    // }
```
pietroalbini, line 190: An approach here could be to temporarily store the templates in `templates-tera/`, which we'd move to `templates/` after the whole migration is finished.

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        });
+    }
+
+    /// Used to initialize a `TemplateData` instance in a `lazy_static`.
+    /// Loading tera takes a second, so it's important that this is done before any
+    /// requests start coming in
+    pub fn poke(&self) -> Result<()> {
+        Ok(())
+    }
```
jyn514, line 64: You don't need a method to do this, you only have to dereference the lazy_static: `let _: &TemplateData = *page::TEMPLATE_DATA;`

src/web/page/templates.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    Ok(Value::String(fmt))
+}
+
+/// Print a tera value to stdout
+fn dbg(value: &Value, _args: &HashMap<String, Value>) -> TeraResult<Value> {
+    println!("{:?}", value);
+
+    Ok(value.clone())
+}
```
jyn514, line 168: ~~There is a built in macro in the standard library for this: https://doc.rust-lang.org/std/macro.dbg.html~~ Never mind, this is for debugging within the template, not within rust code.

Kixiron: Indeed, it's a utility for within templates, makes debugging easier (I made a PR to tera itself with it, but it's still stewing)

src/web/page/templates.rs
Kixiron: Do you want me to remove the comment?

src/web/page/templates.rs
Kixiron: It's definitely not ideal in any sense, but that was my last-ditch effort. I tried to use `notify` (which is supposed to be a cross-platform file/dir watcher), but it just never worked for me. No matter what I did, the changes weren't recognized and so things never reloaded. I can try again to get notify to work, but if all else fails this isn't the *worst* solution ever, since reloading is a development thing

src/web/page/mod.rs
Kixiron: Well, it's easier for me personally to keep them separate like this (hence the comment above them all) since that'll be all that's left in the file once I'm done. Everything above the comment will be removed since `Page` and all the stuff that goes with it gets replaced

Kixiron:
I made it good, the version I had previously tried was broken

src/web/page/templates.rs
```diff
@@ -42,19 +42,31 @@ impl TemplateData {
...
+
+        let (tx, rx) = channel();
+        let mut watcher = watcher(tx, Duration::from_secs(2)).unwrap();
+
+        watcher
+            .watch("tera-templates", RecursiveMode::Recursive)
+            .unwrap();
+
+        thread::spawn(move || {
+            let _watcher = watcher;
```
pietroalbini, line 60: Can you add a comment explaining why you moved the `watcher` inside the thread?

src/web/page/templates.rs
```diff
@@ -42,19 +42,31 @@ impl TemplateData {
...
-                }
+        use notify::{watcher, RecursiveMode, Watcher};
+        use std::{
+            sync::{mpsc::channel, Arc},
+            thread,
+            time::Duration,
+        };
+
+        let (tx, rx) = channel();
+        let mut watcher = watcher(tx, Duration::from_secs(2)).unwrap();
```
pietroalbini, line 53: I had to look in the documentation to see what the duration was, a comment would be nice.

Issue #238, opened 2018-09-24
Document a binary?


e-matteson:
My [crate](https://github.com/e-matteson/keypad) has a macro that defines structs, and I want the docs to include an example of one of those generated structs. My solution was to write a binary, `src/bin/example.rs`, that invokes the macro. 

This works great when I run `cargo doc`. I see `example` listed in the crates sidebar, and it contains docs for the generated struct. However, it doesn't show up in docs.rs. 

I tried adding this to `Cargo.toml`, but it didn't help:

```
[[bin]]
name = "example"
doc = true
```

Is there some magical argument I can pass using `[package.metadata.docs.rs]` that will make it document the binary? Any other suggestions?

e-matteson:
I worked around this issue the same way the bitflags crate does: adding an [example_generated.rs](https://github.com/bitflags/bitflags/blob/master/src/example_generated.rs) module that is only compiled when an `example_generated` feature is used, and telling `docs.rs` to use that feature:
 
```
[package.metadata.docs.rs]
features = [ "example_generated" ]
```

I'll leave this issue open for now even though my specific problem is solved, since it would still be useful to document whether `docs.rs` can handle binaries. But feel free to close it if that's more appropriate.

pickfire:
@jyn514 Do we need to get an RFC for this to allow documentation for binary?

jyn514:
@pickfire no, this does not need an RFC. To get this implemented, we need to 

- Stop [passing `--lib`](https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L467) to cargo
- Stop [filtering out binary targets](https://github.com/rust-lang/docs.rs/blob/master/src/utils/cargo_metadata.rs#L85)
- Decide what to do if there are multiple binaries. I think if one has `default-run = true` we should build that and otherwise fall back to building the library.
- Decide what to do if there is both a binary and a library. I think in that case we should only build the library.

Possibly we could also make the behavior configurable, although I'd rather avoid that in a first draft of the feature.

jyn514:
I'm happy to mentor or take PRs for this but I won't have time to work on it myself in the near future.

pickfire:
Now I am wondering what to document in a binary. I think it is possible to put what needs to be documented in lib.rs.

But looking at the source, I think it would be good if it shows the binary and examples probably?

Nemo157:
Unless you are filtering them out explicitly Cargo will include the example and binaries into the packaged crate.

[...]

Issue #170, opened 2017-12-22
/crate/:crate/latest/builds.json strips the .json suffix


suhr:
This may be useful for external badge services like [shields.io](https://shields.io/).

See also: https://github.com/badges/shields/issues/1382

paulmelnikow:
Hi! Just wanted to mention again that a JSON endpoint with build info would be really helpful! I'd be happy to help design a schema based on the other endpoints you have, if it would be helpful!

jyn514:
@paulmelnikow What are you looking for from the JSON API? What info do you need that builds.json doesn't currently provide? Do you want to be able to get the info from the latest version? https://docs.rs/crate/cargo/latest/builds.json doesn't currently work (it strips the JSON suffix), but I consider that a bug and I'd be happy to fix it.

jyn514:
For context, builds.json is a mostly undocumented feature I wish we didn't have, we introduced it so crates.io knows whether it can link to docs.rs (i.e. whether the build succeeded). I don't mind expanding that endpoint to be more useful, but I don't want to introduce other endpoints if possible.

calebcartwright:
> Do you want to be able to get the info from the latest version? https://docs.rs/crate/cargo/latest/builds.json doesn't currently work (it strips the JSON suffix), but I consider that a bug and I'd be happy to fix it.

Putting my Shields hat on... I do think this would suffice for the use case. IIUC the only option available today would require a badge be pinned to a specific version which would be of limited use, whereas being able to have a badge that dynamically reflected the build status for the latest version of a crate seems like it'd be useful for folks.

PR #809, opened 2020-06-03
Docs.rs API

Kixiron:
Adds an api for [shields.io](https://shields.io) to query docs.rs with.

Users can call `/api/v1/badges?crate=<crate>` for the latest version of a crate and `/api/v1/badges?crate=<crate>&version=<version>` for a specific version.
The api returns the crate name, release version, the url of the crate (e.g. `https://docs.rs/crate/rcc/0.0.0`) and the crate's build status, which can be one of the following:
- `built` fully built and available
- `failed` the build failed in some way
- `queued` the release is in the build queue
- `yanked` the release has been yanked
- `not_library` the release isn't a library (Not 100% sure about this one since #238 is a thing)

The api is versioned so that we have future-compatibility. I'm not entirely sure how to handle the old badge system, ideally we'd be able to remove it entirely and forward requests, but that's not possible until shields.io has a working thing, which isn't possible until *we* have a working thing

### DIFF
--- a/src/web/api/badges.rs
+++ b/src/web/api/badges.rs
@@ -1,0 +1,337 @@
+use super::{ApiErrorV1, Pool};
+use iron::Plugin;
+use iron::{headers::ContentType, status, IronResult, Request, Response};
+use params::{Params, Value};
+use serde::{Deserialize, Serialize};
+
+/// The json data of a crate release
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct BadgeV1 {
+    /// The name of the crate
+    name: String,
+    /// The version of the release
+    version: String,
+    /// The url of the crate
+    docsrs_url: String,
+    /// The crate's status
+    build_status: BuildStatus,
+}
+
+/// The status of a crate release
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename = "snake_case")]
+enum BuildStatus {
+    /// The release successfully built
+    Built,
+    /// The release failed to built
+    Failed,
+    /// The release is queued to build
+    Queued,
+    /// The release was yanked
+    Yanked,
+    /// The release is not a library
+    NotLibrary,
+}
+
+/// The badge api, expects `?crate=<string>` and an optional `version=<string>`, if the version is not
+/// provided then the latest release will be used
+pub fn badge_handler_v1(req: &mut Request) -> IronResult<Response> {
+    let params = ctry!(req.get::<Params>());
+    let conn = extension!(req, Pool).get()?;
+
+    let krate = match params.find(&["crate"]) {
+        Some(Value::String(krate)) => krate,
+        Some(..) => return ApiErrorV1::new("Crate must be a string").into_response(),
+        None => return ApiErrorV1::new("A crate was not supplied").into_response(),
+    };
+    let version = match params.find(&["version"]) {
+        Some(Value::String(version)) => Some(version),
+        Some(..) => return ApiErrorV1::new("Version must be a string").into_response(),
+        None => None,
+    };
+
+    // If a version was provided, then attempt to find the release for that crate & version
+    let (rows, version) = if let Some(version) = version {
+        let query = "
+            SELECT is_library, rustdoc_status, build_status, yanked
+            FROM releases
+            WHERE crate_id IN (
+                SELECT id from crates
+                WHERE name = $1 AND version = $2
+            )";
+        let rows = ctry!(conn.query(query, &[&krate, &version]));
+
+        (rows, version.to_owned())
+
+    // If no version was provided, find the latest release and use that
+    } else {
+        let query = "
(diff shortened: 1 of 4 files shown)


src/web/routes.rs
```diff
@@ -93,6 +93,7 @@ pub(super) fn build_routes() -> Routes {
         "/crate/:name/:version/target-redirect/*",
         super::rustdoc::target_redirect_handler,
     );
+    routes.internal_page("/api/v1/badges", super::api::badge_handler_v1);
```
pietroalbini, line 96: We shouldn't add more reserved prefixes.

I propose all new top-level routes should be prefixed by `-`, for example:

```
/-/api/v1/badges
```

src/web/api/badges.rs
```diff
@@ -0,0 +1,337 @@
...
+                .name("saltwater")
+                .version("1.4.0")
+                .create()?;
+
+            let mut resp = web.get("/api/v1/badges?crate=saltwater").send()?;
+
+            assert!(resp.status().is_success());
+            assert_eq!(
+                resp.json::<BadgeV1>()?,
+                BadgeV1 {
```
pietroalbini, line 159: When doing tests on the API I'd match with `serde_json::Value` (generating the expected result with the `json!` macro), so we can guarantee the result will be the same regardless of bugs or changes in the serialization code.

src/web/routes.rs
Kixiron: What does the `-` do, I'm not familiar with it

pietroalbini: It's just a name that can't be registered on crates.io.

leos:
It would be useful to be able to get a list of the versions available of the docs of any given crate 
- basically the equivalent of: https://docs.readthedocs.io/en/latest/api/v3.html#versions-listing

In general, the RTD api is probably a good starting point for what would be useful on docs.rs.

Kixiron:
I think I've got a decent api design that should be plenty good for the foreseeable future

* `/-/api/v1/crates/{crate_name}/versions` See all released versions of a crate, paginated with the `limit` and `offset` parameters
```jsonc
{
    "total": 25, // Total number of releases for this crate
    "next": "https://docs.rs/-/api/v1/crates/{crate_name}/versions/?limit=10&offset=20",     // The next page of releases or `null`
    "previous": "https://docs.rs/-/api/v1/crates/{crate_name}/versions/?limit=10&offset=10", // The previous page of releases or `null`
    "latest": "https://docs.rs/-/api/v1/crates/{crate_name}/versions/{latest_version}",      // The latest release of the crate
    "releases": [
        {
            "version": "{release_version}",
            "url": "https://docs.rs/-/api/v1/crates/{crate_name}/versions/{release_version}",
            "status": "{build_status}", // One of "built", "queued" or "failed"
            "yanked": false
        }
    ]
}
```

* `/-/api/v1/crates/{crate_name}/versions/{version}` Information on one specific version of a crate
```jsonc
{
    "version": "{version}",
    "crate_name": "{crate_name}",
    "project_name": "{project_name}",
    "status": "{build_status}", // One of "built", "queued" or "failed"
    "yanked": false,
    "build_time": "2020-06-06T23:31:13+00:00", // RFC 3339 timestamp or `null` if the crate failed to build
    "urls": {
        "documentation": "https://docs.rs/{project_name}/{version}/{crate_name}", // Will be `null` if the crate failed to build
        "crate": "https://docs.rs/crate/{crate_name}/{version}",
        "builds_html": "https://docs.rs/crate/{crate_name}/{version}/builds",
        "builds": "https://docs.rs/crate/{crate_name}/{version}/builds.json",
        "source": "https://docs.rs/crate/{crate_name}/{version}/source" // Will be `null` if the crate failed to build
    },
    "dependencies": [
        "https://docs.rs/-/api/v1/crates/{dep_name}/versions/{dep_version}"
    ]
}
```

src/web/api/badges.rs
```diff
@@ -0,0 +1,337 @@
+use super::{ApiErrorV1, Pool};
+use iron::Plugin;
+use iron::{headers::ContentType, status, IronResult, Request, Response};
+use params::{Params, Value};
+use serde::{Deserialize, Serialize};
+
+/// The json data of a crate release
```
teymour-aldridge, line 7:
```suggestion
/// The JSON data of a crate release
```

src/web/api/badges.rs
```diff
@@ -0,0 +1,337 @@
...
+    Failed,
+    /// The release is queued to build
+    Queued,
+    /// The release was yanked
+    Yanked,
+    /// The release is not a library
+    NotLibrary,
+}
+
+/// The badge api, expects `?crate=<string>` and an optional `version=<string>`, if the version is not
```
teymour-aldridge, line 36:
```suggestion
/// The badge API expects `?crate=<string>` and an optional `version=<string>` – if the version is not
```

jyn514:
As per https://github.com/rust-lang/docs.rs/issues/170#issuecomment-764837701, I would prefer we fixed the `latest` bug rather than introducing a completely new API.

Issue #718, opened 2020-04-15
Separate normal and deprioritized crates in the queue size metric


pietroalbini:
Now that we implemented automatic deprioritization for large projects (#714) we should extend the metrics exposed by docs.rs to our monitoring system to account for that.

I'd love for the `docsrs_queued_crates_count` metric to be split into two:

* `docsrs_queued_crates_count` with the queued crates that have a neutral or positive priority
* `docsrs_deprioritized_queued_crates_count` with the queued crates that have a negative priority

This would allow us to define alerting rules such as "alert when there are more than 50 non-prioritized crates in the queue, or more than 300 deprioritized crates".

Kixiron:
I believe I fixed this in #774

PR #810, opened 2020-06-04
metrics: fix docsrs_prioritized_crates_count

pietroalbini:
The metric was including failed crates, and didn't consider that the priority is stored with the wrong sign in the database (a positive priority means a negative priority, for whatever reason).

r? @Kixiron 
fixes https://github.com/rust-lang/docs.rs/issues/718

### DIFF
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -124,9 +124,12 @@ pub fn metrics_handler(req: &mut Request) -> IronResult<Response> {
                 .get(0),
         );
         PRIORITIZED_CRATES_COUNT.set(
-            ctry!(conn.query("SELECT COUNT(*) FROM queue WHERE attempts < 5 AND priority <= 0;", &[]))
-                .get(0)
-                .get(0),
+            ctry!(conn.query(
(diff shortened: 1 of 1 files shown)


src/web/metrics.rs
```diff
@@ -124,7 +124,7 @@ pub fn metrics_handler(req: &mut Request) -> IronResult<Response> {
                 .get(0),
         );
         PRIORITIZED_CRATES_COUNT.set(
-            ctry!(conn.query("SELECT COUNT(*) FROM queue WHERE priority >= 0;", &[]))
+            ctry!(conn.query("SELECT COUNT(*) FROM queue WHERE attempts < 5 AND priority <= 0;", &[]))
                 .get(0)
                 .get(0),
```
jyn514, line 129: This should fix the `rustfmt` test failure:
```suggestion
            ctry!(conn.query(
                "SELECT COUNT(*) FROM queue WHERE attempts < 5 AND priority <= 0;",
                &[]
            ))
            .get(0)
            .get(0),
```

PR #812, opened 2020-06-04
Implemented the Sitemap and About page in Tera

Kixiron:
* Added the `WebPage` trait, which is the interface to rendering templates
* Added all the basic tera components, these will be reused by virtually every template from now on (`base.html`, `macros.html`, `global_alert.html`, `package_navigation.html`, `topbar.html`)
* Implemented `/about` in Tera
* Implemented `/sitemap.xml` in Tera


src/web/page/templates.rs
```diff
@@ -100,7 +100,7 @@ fn load_rustc_resource_suffix() -> Result<String> {
 }
 
 pub(super) fn load_templates() -> TeraResult<Tera> {
-    let mut tera = Tera::new("templates-tera/**/*")?;
+    let mut tera = Tera::new("tera-templates/**/*")?;
```
jyn514, line 103: Do we use the subdirectories?
```suggestion
    let mut tera = Tera::new("tera-templates/*")?;
```

src/web/source.rs
```diff
@@ -29,7 +29,7 @@ struct File {
 }
 
 /// A list of source files
-struct FileList {
+pub(crate) struct FileList {
```
jyn514, line 32: Why did this change?

tera-templates/base.html
```diff
@@ -0,0 +1,43 @@
+{%- import "macros.html" as macros -%}
```
jyn514, line 1: These don't seem to be used in this template, only in templates that inherit from base. I'd rather import it explicitly there so it's clear that it's something we wrote and not part of tera.
```suggestion
```

tera-templates/base.html
```diff
@@ -0,0 +1,43 @@
...
+{%- import "header/package_navigation.html" as navigation -%}
```
jyn514, line 2: This doesn't seem to be used at all?
```suggestion
```

tera-templates/docsrs/Cargo.toml.example
```diff
@@ -0,0 +1,44 @@
+{# The example Cargo.toml used in about.html #}
+[package]
+name = "test"
+
+[package.metadata.docs.rs]
+
+# Features to pass to Cargo (default: none)
```
jyn514, line 7: Hmm, at one point @pietroalbini suggested changing this to `[]` but it never happened. Do you mind doing it now?
```suggestion
# Features to pass to Cargo (default: [])
```

tera-templates/docsrs/Cargo.toml.example
```diff
@@ -0,0 +1,44 @@
...
+#
+# Set this to `[]` to only build the default target.
+#
+# If `default-target` is unset, the first element of `targets` is treated as the default target.
+# Otherwise, these `targets` are built in addition to the default target.
+# If both `default-target` and `targets` are unset,
+#   all tier-one targets will be built and `x86_64-unknown-linux-gnu` will be used as the default target.
+targets = [ "x86_64-apple-darwin", "x86_64-pc-windows-msvc" ]
+
+# Additional `RUSTFLAGS` to set (default: none)
```
jyn514, line 40:
```suggestion
# Additional `RUSTFLAGS` to set (default: [])
```

tera-templates/docsrs/Cargo.toml.example
```diff
@@ -0,0 +1,44 @@
...
+rustc-args = [ "--example-rustc-arg" ]
+
+# Additional `RUSTDOCFLAGS` to set (default: none)
```
jyn514, line 43:
```suggestion
# Additional `RUSTDOCFLAGS` to set (default: [])
```

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
+{% extends "base.html" -%}
+
+{%- block title -%} About Docs.rs {%- endblock title -%}
+
+{%- block body -%}
+    {%- set docsrs_repo = "https://github.com/rust-lang/docs.rs" -%}
```
jyn514, line 6: Good idea :)

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+        </p>
+
+        <h4>Global sandbox limits</h4>
+
+        <p>
+            All the builds on docs.rs are executed inside a sandbox with limited
+            resources. The current limits are the following:
+        </p>
+
+        {{ macros::crate_limits(limits=limits) }}
```
jyn514, line 50: `crate_limits` seems like it would only be used here, why is it a macro?

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+        <p>
+            If a build fails because it hit one of those limits please
+            <a href="{{ docsrs_repo }}/issues/new/choose">open an issue</a>
+            to get them increased for your crate.
+        </p>
+
+        <h4>Redirections</h4>
+
+        <p>
+            Docs.rs is using semver to parse URLs. You can use this feature to access
```
jyn514, line 61:
```suggestion
            Docs.rs uses semver to parse URLs. You can use this feature to access
```

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+        <h4 id="metadata"><a href="#metadata">Metadata for custom builds</a></h4>
+
+        <p>
+            You can customize docs.rs builds by defining <code>[package.metadata.docs.rs]</code>
+            table in your crates' <code>Cargo.toml</code>.
+        </p>
+
+        <p>The available configuration flags you can customize are:</p>
+
+        {# Include the example file #}
```
jyn514, line 166: I don't think this comment is necessary, it's pretty clear what's happening.
```suggestion
```

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+        <pre><code>{%- include "docsrs/Cargo.toml.example" -%}</code></pre>
+
+        <h4>Test crate documentation build locally</h4>
+        {%- set build_subcommand = docsrs_repo ~ "/blob/master/README.md#build-subcommand" -%}
+        <p>
+            The <a href="{{ build_subcommand }}">docs.rs README</a> describes how to build an
```
jyn514, line 172: Why make a variable that's only used once?
```suggestion
        <p>
            The <a href="docsrs_repo ~ "/blob/master/README.md#build-subcommand"">docs.rs README</a> describes how to build an
```

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+            unpublished crate's documentation locally using the same build nvironment as the build agent.
```
jyn514, line 173:
```suggestion
            unpublished crate's documentation locally using the same build environment as the build agent.
```

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,195 @@
...
+        </p>
+
+        <h4>Version</h4>
+        <p>Currently running Docs.rs version is: <strong>{{ docsrs_version() }}</strong></p>
+
+        <h4>Contact</h4>
+        {%- set governance_link = "https://www.rust-lang.org/governance/teams/dev-tools#docs-rs" -%}
+        <p>
+            Docs.rs is run and maintained by the <a href="{{ governance_link }}">Docs.rs team</a>.
```
jyn514, line 182: Again, why make a variable that's only used once?

tera-templates/macros.html
```diff
@@ -0,0 +1,77 @@
...
+        hljs.initHighlighting();
+    </script>
+{% endmacro highlight_js %}
+
+{# Makes the appropriate CSS imports for highlighting #}
+{% macro highlight_css() %}
+    {# Load the highlighting theme css #}
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/styles/github.min.css"
+        type="text/css" media="all" />
+{% endmacro highlight_css %}
```
jyn514, line 36: Is this used in every page? If so, we should consider making it part of the default `{% block css %}` instead having it as a macro.

tera-templates/macros.html
```diff
@@ -0,0 +1,77 @@
...
+    {# Load the script for each provided language #}
+    {%- for language in languages -%}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/languages/{{ language }}.min.js"
+        type="text/javascript" charset="utf-8"></script>
+    {%- endfor -%}
+
+    {# Activate highlighting #}
+    <script type="text/javascript" charset="utf-8">
+        hljs.initHighlighting();
+    </script>
```
jyn514, line 28: Eventually it would be nice to get rid of inline scripts ... but that can be a separate PR.

tera-templates/header/global_alert.html
```diff
@@ -0,0 +1,12 @@
+{# Get the current global alert #}
+{%- set global_alert = global_alert() -%}
```
jyn514, line 2: What does this do? Cache the global alert so it's only calculated once? Is calculating it expensive?

tera-templates/header/topbar.html
```diff
@@ -0,0 +1,97 @@
...
+                        placeholder="Find crate" {%- if search_query %} value="{{ search_query }}" {%- endif %}>
+                </div>
+
+                {# The top-left logo and name #}
+                <a href="/" class="pure-menu-heading pure-menu-link" aria-label="Docs.rs">
+                    <i class="fa fa-cubes fa-fw"></i> Docs.rs
+                </a>
+
+                <ul class="pure-menu-list">
+                    {# The releases button #}
```
jyn514, line 27: This comment seems pretty redundant with the title of the link.
```suggestion
```

tera-templates/header/topbar.html
```diff
@@ -0,0 +1,97 @@
...
+                            Releases
+                        </a>
+                    </li>
+
+                    {# The Rust dropdown menu #}
+                    <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
+                        <a href="#" class="pure-menu-link">Rust</a>
+
+                        <ul class="pure-menu-children">
+                            {# The Rust main site #}
```
jyn514, line 39: Ditto, all of these comments are redundant with the information below.

tera-templates/macros.html
```diff
@@ -0,0 +1,77 @@
+{# Constructs a title based on the given crate name and version #}
+{% macro doc_title(name, version) %}
```
jyn514, line 2: This macro does not appear to be used. Can we remove it?

tera-templates/macros.html
```diff
@@ -0,0 +1,77 @@
...
+    {%- else -%}
+        Docs.rs
+    {%- endif -%}
+{% endmacro doc_title %}
+
+{#
+    Makes the appropriate JS imports for highlighting
+    * `languages` An array of strings where each is a valid highlight.js language
+#}
+{% macro highlight_js(languages) %}
```
jyn514, line 14: Nice use of function parameters, I didn't know tera could do that :)

src/web/page/pages.rs
```diff
@@ -0,0 +1,64 @@
...
+/// The sitemap, corresponds with `templates/sitemap.xml`
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub(crate) struct SitemapXml {
+    /// The releases to be displayed on the sitemap
+    pub releases: Vec<(String, String)>,
+}
+
+impl WebPage for SitemapXml {
+    fn template() -> &'static str {
+        "docsrs/sitemap.xml"
```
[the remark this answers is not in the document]
Kixiron: It selects the template to be rendered

src/web/page/templates.rs
Kixiron: Yes, in later stages we do

tera-templates/base.html
Kixiron: They're used throughout pretty much every template, plus macros require namespacing (`macros::whavever()`), so there's no confusion with tera

tera-templates/docsrs/about.html
Kixiron: It's also used on the build pages, it was much easier to pull into a component

tera-templates/docsrs/about.html
Kixiron: For this one it was because it's super duper long

tera-templates/macros.html
Kixiron: It's only used on some pages, hence the macro

tera-templates/macros.html
Kixiron: It's used in later stages

tera-templates/macros.html
Kixiron: Yah, the only downside is that you cannot supply arrays as default arguments, I would have liked `highlight_js(languages=["ini", "rust"])` much better

src/web/page/pages.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn content_type() -> ContentType {
+        ContentType::html()
+    }
+}
+
+/// The sitemap, corresponds with `templates/sitemap.xml`
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub(crate) struct SitemapXml {
+    /// The releases to be displayed on the sitemap
+    pub releases: Vec<(String, String)>,
```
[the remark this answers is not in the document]
Kixiron: It's `(name, rfc 3339 timestamp)`

src/web/source.rs
Kixiron: That actually gets deleted later on, it was supposed to get removed in a later stage but I guess it slipped through

tera-templates/header/global_alert.html
Kixiron: It's not super expensive, but it's used ~5 times in this template, and every time `global_alert()` is called it clones a json value, but it's just not necessary to

tera-templates/header/topbar.html
Kixiron: Agreed, I'm actually trying to figure out a good way to automate this section so it's not just a copy+paste

tera-templates/base.html
Kixiron: It is used, just in later stages

tera-templates/docsrs/about.html
Kixiron: It just looked nicer to me, since it allowed the flow of the paragraph to be broken up less

src/web/page/pages.rs
```diff
@@ -7,10 +7,10 @@ use tera::Context;
 /// The central trait that rendering pages revolves around, it handles selecting and rendering the template
 pub trait WebPage: Serialize + Sized {
     /// Turn the current instance into a `Response`, ready to be served
-    // TODO: The potential for caching similar pages is here due to render taking `&Context`
+    // TODO: We could cache similar pages using the `&Context`
     fn into_response(self) -> IronResult<Response> {
         let ctx = Context::from_serialize(&self).unwrap();
```
jyn514, line 12: test

src/web/page/pages.rs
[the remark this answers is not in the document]
jyn514: Please add that as a comment.

src/web/page/pages.rs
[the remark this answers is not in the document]
jyn514: Right, but why is it needed? Where is it called?

src/web/source.rs
jyn514: Ok, can you change it back?

tera-templates/base.html
jyn514: Ok, please remove it until it is used. Otherwise I can't review it because I don't know the purpose - I don't have any context, and in the next PR I won't think to review it because it's already been merged.

src/web/page/pages.rs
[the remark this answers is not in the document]
jyn514: Please add that as a comment.

tera-templates/macros.html
jyn514: Please remove it until it is used (by same rationale as above).

src/web/page/pages.rs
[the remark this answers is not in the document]
Kixiron: It's called as part of the `.into_response()` function to get the template that the given struct needs in order to be correctly rendered

https://github.com/rust-lang/docs.rs/blob/2d0156b52daa74ec4daba132a78cb6736f82ae1e/src/web/page/pages.rs#L14

src/web/page/pages.rs
```diff
@@ -0,0 +1,64 @@
+use super::TEMPLATE_DATA;
+use crate::docbuilder::Limits;
+use iron::{headers::ContentType, response::Response, status::Status, IronResult};
+use serde::Serialize;
+use tera::Context;
+
+/// The central trait that rendering pages revolves around, it handles selecting and rendering the template
+pub trait WebPage: Serialize + Sized {
```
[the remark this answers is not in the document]
Nemo157: It's not actually on a trait definition, only in generic parameters; it's needed because `into_response` takes `self` by value so needs to know that it's `Sized`.

src/web/page/pages.rs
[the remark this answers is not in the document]
Nemo157: Does it ever need to be dynamic? Could it be `const TEMPLATE: &'static str;` instead? (Given that it's an `&'static str` I assume it's not going to be dynamic, since that would require leaking). (I would say the same about the content type, but given it uses `vec![]` I assume it has no `const` constructors).

src/web/page/pages.rs
```diff
@@ -0,0 +1,64 @@
...
+        Status::Ok
+    }
+
+    /// The content type that the template should be served with, defaults to html
+    fn content_type() -> ContentType {
+        ContentType::html()
+    }
+}
+
+/// The sitemap, corresponds with `templates/sitemap.xml`
```
pietroalbini, line 37: I'd prefer if those structs were defined with the rest of that page's content (for example, this struct should be in `src/web/sitemap.rs`.

src/web/page/pages.rs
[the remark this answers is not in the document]
Kixiron: It could be an associated static of some sort, but this does leave the flexibility for us to do whatever we want in the future

src/web/page/pages.rs
Kixiron: I had it like that originally, but it was a lot easier to reason about and cleaner to have all the stuff that only applies to rendering all grouped together

Kixiron:
# Screenshots

<details>
<summary>About</summary>
<br>



[Current](https://docs.rs/about)

</details>

<details>
<summary>Sitemap</summary>
<br>

```xml
<urlset>
    <script/>
    <url>
        <loc>https://docs.rs/wayland-window</loc>
        <lastmod>2017-09-21T12:44:46+00:00</lastmod>
    </url>
</urlset>
```

[Current](https://docs.rs/sitemap.xml)

</details>

src/web/page/pages.rs
pietroalbini: Do you think it's cleaner when tweaking with the rendering of the whole website or that it's cleaner when working on a single page?

src/web/page/pages.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub(crate) struct About {
+    pub rustc_version: Option<String>,
+    pub limits: Limits,
+}
+
+impl WebPage for About {
```
pietroalbini, line 68: Thinking about these impls, could a small macro be more clean?

```rust
impl_webpage!(About, "docsrs/about.html");
impl_webpage!(NotFound, "docsrs/about.html", status = NotFound);
impl_webpage!(SitemapXml, "docsrs/sitemap.xml", content_type = ContentType(Mime(TopLevel::Application, SubLevel::Xml, vec![])));
```

I'm not fully sold on using the macro myself, but it might be an interesting idea.

src/web/page/pages.rs
Kixiron: Both in my opinion, it means all the cognitively similar things are located in the same place

src/web/page/pages.rs
Kixiron: Ohh, I really like that

Kixiron: It has been done, 10/10 idea

src/web/page/pages.rs
pietroalbini: Hmm, I slightly prefer having everything related to a page in that page's file. @jyn514 @Nemo157  what's your opinion here?

jyn514: I agree with having everything in the same file, that lets you see what custom variables have been added for that page without having to switch between files.

Nemo157:
Should this be deleting some unused handlebars templates?

Kixiron:
I was just going to leave the templates until the end to avoid any potential issues with them being used and to allow reviews to see the old and new ones

src/web/page/pages.rs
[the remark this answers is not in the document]
jyn514: > this does leave the flexibility for us to do whatever we want in the future

It's not flexible enough to be useful because it returns a `&'static str` (as mentioned by Nemo), and I don't think there's any point in trying to generalize the functionality without any idea of what changes we'd make. So I agree it should be an associated constant.

pietroalbini:
Looks good to me! I'd remove the outdated Handlebars templates though.

src/web/page/web_page.rs
```diff
@@ -0,0 +1,57 @@
...
+use tera::Context;
+
+#[macro_export]
+macro_rules! impl_webpage {
+    ($page:ty = $template:expr $(, status = $status:expr)? $(, content_type = $content_type:expr)? $(,)?) => {
+        impl $crate::web::page::WebPage for $page {
+            const TEMPLATE: &'static str = $template;
+
+            $(
+                fn get_status(&self) -> ::iron::status::Status {
```
jyn514, line 13: `&self` doesn't look necessary here.

src/web/page/web_page.rs
```diff
@@ -0,0 +1,57 @@
...
+                    $status
+                }
+            )?
+
+            $(
+                fn content_type() -> ::iron::headers::ContentType {
+                    $content_type
+                }
+            )?
```
jyn514, line 22: Both these functions look like they could be associated constants instead, which would decrease the boilerplate a lot and possibly get rid of the need for a macro.

src/web/sitemap.rs
```diff
@@ -31,36 +49,42 @@ pub fn sitemap_handler(req: &mut Request) -> IronResult<Response> {
...
+pub fn about_handler(req: &mut Request) -> IronResult<Response> {
     let conn = extension!(req, Pool).get()?;
     let res = ctry!(conn.query("SELECT value FROM config WHERE name = 'rustc_version'", &[]));
 
+    let mut rustc_version = None;
     if let Some(row) = res.iter().next() {
         if let Some(Ok::<Value, _>(res)) = row.get_opt(0) {
             if let Some(vers) = res.as_str() {
-                content.insert("rustc_version".to_string(), Value::String(vers.to_string()));
+                rustc_version = Some(vers.to_owned());
```
jyn514, line 80: This copy doesn't look necessary.
```suggestion
            if let Some(Value::String(vers)) = res {
                rustc_version = Some(vers);
```

tera-templates/docsrs/Cargo.toml.example
```diff
@@ -0,0 +1,44 @@
+{# The example Cargo.toml used in about.html #}
```
jyn514, line 1: Why do all the templates have `docsrs/` as a leading prefix? We know they must be for docs.rs because that's the repository they're in.

Kixiron: They’re templates that pertain to the core of the site

src/web/page/web_page.rs
Kixiron: They can’t be, status can be gotten from the template struct itself and is, while not all content types are statically constructible

src/web/page/web_page.rs
Kixiron: It is, the status can be gotten from the template itself

tera-templates/docsrs/Cargo.toml.example
jyn514: As opposed to what? Aren't all the templates part of docs.rs?

src/web/page/web_page.rs
jyn514: What do you mean? When is this used?

src/web/page/web_page.rs
jyn514: > not all content types are statically constructible

This seems strange to me, can you give an example?

src/web/page/web_page.rs
Kixiron: For error pages

tera-templates/docsrs/about.html
```diff
@@ -0,0 +1,194 @@
...
+            <a href="https://www.rust-lang.org/">Rust Programming Language</a>.
+        </p>
+
+        <p>
+            Docs.rs automatically builds crates' documentation released on
+            <a href="https://crates.io/">crates.io</a>
+            using the nightly release of the Rust compiler.
+
+            {%- if rustc_version -%}
+            The current version of the Rust compiler in use is <code>{{ rustc_version }}</code>.
```
jyn514, line 23:
```suggestion
            The current version of the Rust compiler in use is <code>{{ rustc_version }}</code>.
              Builds can take a while depending how many crates are in <a href="/releases/queue">the queue</a>.
```
#838

src/web/page/web_page.rs
pietroalbini: I imagine the content type of rustdoc pages is not only HTML, but also CSS and JS.

Kixiron: [`ContentType::html()`](https://docs.rs/iron/0.6.1/iron/headers/struct.ContentType.html#method.html) isn’t a `const fn`

tera-templates/docsrs/Cargo.toml.example
Kixiron: As opposed to user controlled (so to speak) and content pages, these are internal pages

src/web/page/web_page.rs
jyn514: I'm still not quite following but we can revisit this when you add the error pages and it's more clear why it's necessary.

tera-templates/docsrs/Cargo.toml.example
jyn514: I don't think `docsrs` is a good name then. If you want to emphasize that they're internal pages you could use `internal/` or `core` as a prefix but I don't think they need a prefix at all really.

jyn514:
🎉 🎉 🎉

jyn514:
🎉 🎉 🎉

tera-templates/core/about.html
```diff
@@ -0,0 +1,195 @@
...
+            <a href="https://www.rust-lang.org/">Rust Programming Language</a>.
+        </p>
+
+        <p>
+            Docs.rs automatically builds crates' documentation released on
+            <a href="https://crates.io/">crates.io</a>
+            using the nightly release of the Rust compiler.
+
+            {%- if rustc_version -%}
+            The current version of the Rust compiler in use is <code>{{ rustc_version }}</code>.
```
jyn514, line 23: It looks like tera cuts out all this whitespace, so it looks like `compiler.The`. Do you know how to fix that? https://docs.rs/about

Nemo157: It's the `-` on the above tag, they say to suppress all whitespace on that side of the tag.

Kixiron: Fixed in #840

Issue #767, opened 2020-05-22
Configurable index location


Nemo157:
To support a staging server and other users running their own docs.rs instance against their own registry the index repository used should be configurable.

Nemo157:
Of course this is a bit more involved than I thought at first.

As well as changing the index this will need to:

1. Change the docbuilder to pull the crate to be built from the alternative registry.
2. Use the `api` key from the index to determine which API to hit for additional details, and deal with registries that do not support this API.

jyn514:
Relevant code:
- https://github.com/rust-lang/docs.rs/blob/8fb139b/src/docbuilder/rustwide_builder.rs#L203
- https://docs.rs/rustwide/0.7.1/rustwide/struct.Crate.html#method.crates_io
- https://docs.rs/crate/rustwide/0.7.1/source/src/crates/cratesio.rs

This will require changes to rustwide.

Nemo157:
I've got ~everything except the rustwide changes done in https://github.com/Nemo157/docs.rs/tree/alternative-registry

repi:
This would be great to have! 

We currently use a private crate registry on cloudsmith.io for crates that are private or not fully ready yet to be open sourced. https://github.com/EmbarkStudios/rust-ecosystem/issues/47

jyn514:
> This would be great to have!
> 
> We currently use a private crate registry on cloudsmith.io for crates that are private or not fully ready yet to be open sourced. [EmbarkStudios/rust-ecosystem#47](https://github.com/EmbarkStudios/rust-ecosystem/issues/47)

Note that this would still not allow using crates from a private registry on docs.rs. Rather, this allows you to host your own docs.rs instance which uses a private registry.

Nemo157:
Another point to deal with here: authentication. Probably just needs documenting that the user account that docs.rs is running under must have a git credential helper setup correctly (I assume this works with libgit2 correctly? In my own alternative registry experiments I've had to use the `net.git-fetch-with-cli` option because libgit2 doesn't support the "dumb" http transport, but cloudsmith don't mention having to enable this for basic auth).

---

@repi from looking through your linked issues I see that cloudsmith support at least the subset of the crates.io API required for publishing. Do you know if there's any documentation on what other parts of the API they support? I can only find documentation on their own API, not on the package manager specific APIs. (We pull some other details from the crates.io API for each build).

---

[...]

PR #813, opened 2020-06-04
Load registry config to determine which API to use

Nemo157:
Another step on the way towards #767, removes hardcoded `https://crates.io/api` urls and determines what to use based on the config in the registry index. The API is optional for a registry, so if configured to use a registry without one it will just insert empty data when building crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,8 @@ log = "0.4"
 regex = "1"
 structopt = "0.3"
 crates-index-diff = "7"
-reqwest = { version = "0.10.6", features = ["blocking"] } # TODO: Remove blocking when async is ready
-semver = "0.9"
+reqwest = { version = "0.10.6", features = ["blocking", "json"] } # TODO: Remove blocking when async is ready
+semver = { version = "0.9", features = ["serde"] }
 slug = "=0.1.1"
 env_logger = "0.7"
 r2d2 = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ rustwide = "0.7.1"
 mime_guess = "2"
 dotenv = "0.15"
 zstd = "0.5"
+git2 = { version = "0.13.6", default-features = false }
 
 # Data serialization and deserialization
 serde = { version = "1.0", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,9 @@ slug = "=0.1.1"
 env_logger = "0.7"
 r2d2 = "0.8"
 r2d2_postgres = "0.14"
-url = "1.4"
+# iron needs url@1, but it reexports it as iron::url, so we can start using
+# url@2 for other usecases
+url = { version = "2.1.1", features = ["serde"] }
 badge = { path = "src/web/badge" }
 failure = "0.1.3"
 comrak = { version = "0.3", default-features = false }
--- a/src/index/api.rs
+++ b/src/index/api.rs
@@ -15,8 +15,9 @@ const APP_USER_AGENT: &str = concat!(
     include_str!(concat!(env!("OUT_DIR"), "/git_version"))
 );
 
-pub(crate) struct Api<'a> {
-    api_base: &'a Url,
+pub(crate) struct Api {
+    api_base: Url,
+    client: reqwest::blocking::Client,
 }
 
 pub(crate) struct RegistryCrateData {
@@ -33,24 +34,23 @@ pub(crate) struct CrateOwner {
     pub(crate) name: String,
 }
 
-fn client() -> Result<reqwest::blocking::Client> {
-    let headers = vec![
-        (USER_AGENT, HeaderValue::from_static(APP_USER_AGENT)),
-        (ACCEPT, HeaderValue::from_static("application/json")),
-    ]
-    .into_iter()
-    .collect();
+impl Api {
+    pub(super) fn new(api_base: &Url) -> Result<Self> {
+        let headers = vec![
+            (USER_AGENT, HeaderValue::from_static(APP_USER_AGENT)),
+            (ACCEPT, HeaderValue::from_static("application/json")),
+        ]
+        .into_iter()
+        .collect();
 
-    let client = reqwest::blocking::Client::builder()
-        .default_headers(headers)
-        .build()?;
+        let client = reqwest::blocking::Client::builder()
+            .default_headers(headers)
+            .build()?;
 
-    Ok(client)
-}
-
-impl<'a> Api<'a> {
-    pub(super) fn new(api_base: &'a Url) -> Self {
-        Self { api_base }
+        Ok(Self {
+            api_base: api_base.clone(),
+            client,
+        })
     }
 
     pub(crate) fn get_crate_data(&self, name: &str, version: &str) -> Result<RegistryCrateData> {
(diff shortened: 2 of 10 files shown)


Nemo157:
(I suggest hiding whitespace changes while viewing since the `api` code all got indented one more level)

Cargo.toml
```diff
@@ -19,7 +19,9 @@ slug = "=0.1.1"
 env_logger = "0.7"
 r2d2 = "0.8"
 r2d2_postgres = "0.14"
-url = "1.4"
+# iron needs url@1, but it reexports it as iron::url, so we can start using
+# url@2 for other usecases
+url = { version = "2.1.1", features = ["serde"] }
```
jyn514, line 24: Why add a second version of url2?

Cargo.toml
```diff
@@ -38,6 +40,7 @@ lazy_static = "1.0.0"
 rustwide = "0.7.1"
 mime_guess = "2"
 dotenv = "0.15"
+git2 = { version = "0.13.6", default-features = false }
```
jyn514, line 43: We already have git2 through `crates-index-diff`, why add it again here?

src/index/api.rs
```diff
@@ -20,137 +27,160 @@ pub(crate) struct CrateOwner {
...
+    fn get_release_time_yanked_downloads(
+        &self,
+        name: &str,
+        version: &str,
+    ) -> Result<(DateTime<Utc>, bool, i32)> {
+        let url = {
+            let mut url = self.api_base.clone();
+            url.path_segments_mut()
+                .map_err(|()| err_msg("Invalid API url"))?
+                .extend(&["api", "v1", "crates", name, "versions"]);
```
jyn514, line 58: The crates.io API isn't stabilized, see https://rust-lang.github.io/rfcs/2141-alternative-registries.html#registry-index-format-specification. I'm not sure how we handle an error from this function - will it abort the build altogether? I think the right thing to do is default to dummy data, the same way we do if there's no API available.

src/index/api.rs
```diff
@@ -20,137 +27,160 @@ pub(crate) struct CrateOwner {
...
-                    email: email.to_string(),
-                    login: login.to_string(),
-                    name: name.to_string(),
+    /// Fetch owners from the registry's API
+    fn get_owners(&self, name: &str) -> Result<Vec<CrateOwner>> {
+        let url = {
+            let mut url = self.api_base.clone();
+            url.path_segments_mut()
+                .map_err(|()| err_msg("Invalid API url"))?
+                .extend(&["api", "v1", "crates", name, "owners"]);
```
jyn514, line 131: Same as above, do we know this path is present in all registries?

src/index/mod.rs
```diff
@@ -1 +1,68 @@
...
+
+pub(crate) struct Index {
+    diff: crates_index_diff::Index,
+    path: PathBuf,
+    config: IndexConfig,
+}
+
+#[derive(serde::Deserialize, Clone)]
+#[serde(rename_all = "kebab-case")]
+struct IndexConfig {
```
jyn514, line 19: This is clever! `serde` is so useful :D

Cargo.toml
Nemo157: I assume that's "why _not_ add"?

All existing use of `url` was directly related to `iron`, and they happen to reexport it; so it seemed less confusing to use `iron::url` and `url` instead of `url` and `url2`.

Cargo.toml
Nemo157: We use it directly in `load_config`.

src/index/api.rs
Nemo157: The API is actually publicly documented at <https://doc.rust-lang.org/stable/cargo/reference/registries.html#web-api> with no mention of stability. I don't recall ever seeing any RFCs related to it, so I'm not sure if it's intended to be stable or not.

I also didn't actually check whether the endpoints we use are documented there 😅 I'll take a look at that tomorrow.

Nemo157: Actually, I took a look now. This path (`/versions`) isn't specified at all, the other path we use (`/owners`) is.

I think falling back to dummy data on failure makes sense, probably collecting as much as possible and just defaulting the rest.

jyn514:
Built without problems (although I found some other regressions while playing around).

Cargo.toml
jyn514: You can `use crates_index_diff::git2`, they re-export it. That prevents the versions getting out of sync and accidentally bundling two.

Cargo.toml
jyn514: Right, but why use version 2 of url at all?

Cargo.toml
Nemo157: Huh, what's with all these re-exports, that's not something I even thought to look for, I only noticed `iron::url` did a re-export since it was already used that way in some locations.

The versions can't get out of sync since cargo will refuse to include two semver compatible versions in the build, but I'll change to use it.

Cargo.toml
Nemo157: `serde 1.*` compatibility, `url 1.*` is stuck back on `serde >=0.6.1, <0.9`

src/index/mod.rs
Kixiron: Serde is great, I totally recommend reading over all their [attributes](https://serde.rs/attributes.html). Every time I look at it I discover something new

Cargo.toml
pietroalbini: > You can use crates_index_diff::git2, they re-export it. That prevents the versions getting out of sync and accidentally bundling two.

I don't think this is a good approach, if docs.rs depends on git2 it should do so explicitly.

pietroalbini:
I'll take a look at this PR on Monday.

src/index/api.rs
pietroalbini: Yeah that seems a good approach to me.

pietroalbini:
This looks good to me once we change the PR to set default values for any data it can't retrieve. Also, we should avoid creating multiple HTTP clients, ideally by sharing a single one even between builds.

src/index/api.rs
```diff
@@ -26,147 +33,132 @@ pub(crate) struct CrateOwner {
...
+        self.api_base
+            .clone()
+            .ok_or_else(|| err_msg("index is missing an api base url"))
+    }
+
+    pub(crate) fn get_crate_data(&self, name: &str, version: &str) -> RegistryCrateData {
+        let (release_time, yanked, downloads) = self
+            .get_release_time_yanked_downloads(name, version)
+            .unwrap_or_else(|err| {
+                info!("Failed to get crate data for {}-{}: {}", name, version, err);
```
jyn514, line 62:
```suggestion
                warn!("Failed to get crate data for {}-{}: {}", name, version, err);
```

src/index/api.rs
```diff
@@ -26,147 +33,132 @@ pub(crate) struct CrateOwner {
...
+                (Utc::now(), false, 0)
+            });
+
+        let owners = self.get_owners(name).unwrap_or_else(|err| {
+            info!("Failed to get owners for {}-{}: {}", name, version, err);
```
jyn514, line 67:
```suggestion
            warn!("Failed to get owners for {}-{}: {}", name, version, err);
```

src/index/api.rs
Nemo157: I wasn't sure about whether this should be a warning given that not all registries might support it, but just 2 warning logs per build seems fine.

jyn514:
Thanks for working on this! What more do we have left until we can use staging.crates.io?

PR #815, opened 2020-06-05
Updated Reqwest

Kixiron:
Updates reqwest to the latest version, trying to narrow down the FD issues


jyn514:
Previously: #814

pietroalbini:
This merge was reverted as it caused an outage.

PR #819, opened 2020-06-06
Set application specific user agent for crates.io

Nemo157:
r? @pietroalbini

### DIFF
--- a/src/index/api.rs
+++ b/src/index/api.rs
@@ -1,10 +1,16 @@
 use crate::{error::Result, utils::MetadataPackage};
 use chrono::{DateTime, Utc};
 use failure::err_msg;
-use reqwest::{header::ACCEPT, Client};
+use reqwest::header::{HeaderValue, ACCEPT, USER_AGENT};
 use semver::Version;
 use serde_json::Value;
 use std::io::Read;
+
+const APP_USER_AGENT: &str = concat!(
+    env!("CARGO_PKG_NAME"),
+    " ",
+    include_str!(concat!(env!("OUT_DIR"), "/git_version"))
+);
 
 pub(crate) struct RegistryCrateData {
     pub(crate) release_time: DateTime<Utc>,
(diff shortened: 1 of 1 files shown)


src/index/api.rs
```diff
@@ -103,11 +123,7 @@ fn get_release_time_yanked_downloads(pkg: &MetadataPackage) -> Result<(DateTime<
 fn get_owners(pkg: &MetadataPackage) -> Result<Vec<CrateOwner>> {
     // owners available in: https://crates.io/api/v1/crates/rand/owners
     let owners_url = format!("https://crates.io/api/v1/crates/{}/owners", pkg.name);
-    let client = Client::new();
-    let mut res = client
-        .get(&owners_url[..])
-        .header(ACCEPT, "application/json")
-        .send()?;
+    let mut res = client()?.get(&owners_url[..]).send()?;
```
pietroalbini, line 126: Can we reuse the same client?

src/index/api.rs
```diff
@@ -1,11 +1,19 @@
...
-use reqwest::{header::ACCEPT, Client};
+use reqwest::header::{HeaderValue, ACCEPT, USER_AGENT};
 use semver::Version;
 use serde_json::Value;
 use std::io::Read;
 
+const APP_USER_AGENT: &str = concat!(
+    env!("CARGO_PKG_NAME"),
+    "/",
+    env!("CARGO_PKG_VERSION"),
```
pietroalbini, line 12: I don't think including the version is useful, we never bump it and it doesn't really have a meaning nowadays.

Kixiron: Using a user agent from an env var is probably better since it'll mean that only *our* instance of docs.rs has the agent instead of every single build (So we can be uniquely identified)

Nemo157: That's not the purpose of a user-agent, it's to know what software is making the request. The IP from the crates.io logs can be used for checking the instance.

src/index/api.rs
Nemo157: Much more easily done as part of #813.

pietroalbini: Sure.

PR #823, opened 2020-06-06
Print a backtrace for crates which fail to build

### LINKED ISSUE #821: The backtrace is not shown when an error happens during a build
When an error happens during a build started by the daemon, the error is captured and logged with a custom handler. This handler doesn't show a backtrace even though it's present, making the debugging of errors happening during the build harder.

jyn514:
Closes #821 

r? @pietroalbini


jyn514:
I had trouble testing this because building from the queue requires running the daemon, and the daemon will try to add every crate ever published to the queue.

PR #828, opened 2020-06-08
Updated Reqwest

Kixiron:
Updated reqwest to its latest version, removes ~20 dependencies and brings us a little closer to async


jyn514:
I built a crate locally (twice in a row) and everything looks fine. Let's try this again.

jyn514:
For posterity: this is the follow-up to https://github.com/rust-lang/docs.rs/pull/815.

Issue #800, opened 2020-06-01
Source tab isn't shown for postgres 0.15.2


Kixiron:
For [`postgres 0.15.2`](https://docs.rs/postgres/0.15.2/postgres/) the source tab is displayed, but clicking on the [link](https://docs.rs/crate/postgres/0.15.2/source/) gives a 404

jyn514:
As a workaround you could use the rustdoc sources: https://docs.rs/postgres/0.15.2/src/postgres/lib.rs.html#1-1396

Kixiron:
It unfortunately doesn't show the Cargo.toml (that I know of)

Nemo157:
The source files definitely exist on S3 according to the last inventory, so somethings going wrong when attempting to load them

```
inventory/data/5b665bba-71d3-4c73-81d6-ec663ea81ec0.csv:"rust-docs-rs","sources/postgres/0.15.2/Cargo.toml","2846","c8ae098975318cbb129c3da561945e0c",""
```

PR #692, opened 2020-04-02
Worked on vector-related performance

Kixiron:
* Pre-allocated vectors wherever possible in order to reduce the total amount of allocations preformed

PR #830, opened 2020-06-11
Don't return 404 if .cargo-ok file exists

jyn514:
This was a regression introduced in https://github.com/rust-lang/docs.rs/pull/692.

Closes https://github.com/rust-lang/docs.rs/issues/800.

r? @Kixiron


src/web/source.rs
```diff
@@ -261,6 +261,7 @@ pub fn source_browser_handler(req: &mut Request) -> IronResult<Response> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::test::*;
```
Kixiron, line 264: Glob imports aren't the greatest, they obfuscate where things are coming from and make hidden dependencies on module layouts
```suggestion
    use crate::test::wrapper;
```

jyn514: Hmm, this has been the standard for all the tests (see e.g. https://github.com/rust-lang/docs.rs/blob/master/src/web/mod.rs#L580). The rationale is I want to make it as easy to add tests as possible without having to stop in the middle to add an import.

jyn514: Although I do see your point since this imported more than the wrapper:
```suggestion
    use crate::test::{wrapper, assert_success};
```

Kixiron: Whoops, didn't see the `assert_success`. I just really don't like them, plus there's [others](https://github.com/rust-lang/docs.rs/blob/master/src/web/metrics.rs#L199) [that](https://github.com/rust-lang/docs.rs/blob/master/src/web/releases.rs#L684) [don't](https://github.com/rust-lang/docs.rs/blob/master/src/web/file.rs#L69)

jyn514: Ok, I removed the glob import.

PR #834, opened 2020-06-12
Limit the size of served files

pietroalbini:
This PR limits how big the files we serve from docs.rs are, to avoid OOMs. The current limit is 5MB, but that can be changed by tweaking a constant in the source code.

This PR adds limits when downloding the files from S3 and the database, and when decompressing previously downloaded files. For now the user will see a "resource not found" error page, but a dedicated page can be created in the future.

This PR also refactors compression a bit to better enable testing. It's best to review the PR commit-by-commit.

r? @jyn514

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -13,8 +13,7 @@ pub fn main() -> Result<(), Error> {
     let _ = dotenv::dotenv();
     logger_init();
 
-    CommandLine::from_args().handle_args()?;
-    Ok(())
+    CommandLine::from_args().handle_args()
 }
 
 fn logger_init() {
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -1,17 +1,20 @@
 use std::env;
 use std::path::PathBuf;
+use std::sync::Arc;
 
 use cratesfyi::db::{self, add_path_into_database, connect_db};
 use cratesfyi::utils::{add_crate_to_queue, remove_crate_priority, set_crate_priority};
 use cratesfyi::Server;
 use cratesfyi::{DocBuilder, DocBuilderOptions, RustwideBuilder};
+use failure::Error;
 use structopt::StructOpt;
 
-pub fn main() {
+pub fn main() -> Result<(), Error> {
     let _ = dotenv::dotenv();
     logger_init();
 
-    CommandLine::from_args().handle_args();
+    CommandLine::from_args().handle_args()?;
+    Ok(())
 }
 
 fn logger_init() {
@@ -79,19 +82,23 @@ enum CommandLine {
 }
 
 impl CommandLine {
-    pub fn handle_args(self) {
+    pub fn handle_args(self) -> Result<(), Error> {
+        let config = Arc::new(cratesfyi::Config::from_env()?);
+
         match self {
             Self::Build(build) => build.handle_args(),
             Self::StartWebServer {
                 socket_addr,
                 reload_templates,
             } => {
-                Server::start(Some(&socket_addr), reload_templates);
+                Server::start(Some(&socket_addr), reload_templates, config);
             }
-            Self::Daemon { foreground } => cratesfyi::utils::start_daemon(!foreground),
+            Self::Daemon { foreground } => cratesfyi::utils::start_daemon(!foreground, config),
             Self::Database { subcommand } => subcommand.handle_args(),
             Self::Queue { subcommand } => subcommand.handle_args(),
         }
+
+        Ok(())
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,37 @@
+use failure::{bail, Error, Fail, ResultExt};
+use std::env::VarError;
+use std::str::FromStr;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct Config {
+    pub(crate) max_file_size: usize,
+    pub(crate) max_file_size_html: usize,
+}
+
+impl Config {
+    pub fn from_env() -> Result<Self, Error> {
+        Ok(Self {
+            max_file_size: env("DOCSRS_MAX_FILE_SIZE", 50 * 1024 * 1024)?,
+            max_file_size_html: env("DOCSRS_MAX_FILE_SIZE_HTML", 5 * 1024 * 1024)?,
+        })
+    }
+}
+
+impl iron::typemap::Key for Config {
+    type Value = Arc<Config>;
+}
+
+fn env<T>(var: &str, default: T) -> Result<T, Error>
+where
+    T: FromStr,
+    T::Err: Fail,
+{
+    match std::env::var(var) {
+        Ok(content) => Ok(content
+            .parse::<T>()
+            .with_context(|_| format!("failed to parse configuration variable {}", var))?),
+        Err(VarError::NotPresent) => Ok(default),
+        Err(VarError::NotUnicode(_)) => bail!("configuration variable {} is not UTF-8", var),
+    }
+}
--- a/src/storage/database.rs
+++ b/src/storage/database.rs
@@ -40,7 +40,7 @@ impl<'a> DatabaseBackend<'a> {
         } else {
             let row = rows.get(0);
 
-            if row.get::<_, bool>("is_too_big") {
+            if row.get("is_too_big") {
                 return Err(std::io::Error::new(
                     std::io::ErrorKind::Other,
                     crate::error::SizeLimitReached,
--- a/src/storage/database.rs
+++ b/src/storage/database.rs
@@ -16,18 +16,37 @@ impl<'a> DatabaseBackend<'a> {
         Self { conn }
     }
 
-    pub(super) fn get(&self, path: &str) -> Result<Blob, Error> {
+    pub(super) fn get(&self, path: &str, max_size: usize) -> Result<Blob, Error> {
         use std::convert::TryInto;
 
+        // The maximum size for a BYTEA (the type used for `content`) is 1GB, so this cast is safe:
+        // https://www.postgresql.org/message-id/162867790712200946i7ba8eb92v908ac595c0c35aee%40mail.gmail.com
+        let max_size = max_size.min(std::i32::MAX as usize) as i32;
+
+        // The size limit is checked at the database level, to avoid receiving data altogether if
+        // the limit is exceeded.
         let rows = self.conn.query(
             "SELECT path, mime, date_updated, content, compression
              FROM files
-             WHERE path = $1;",
-            &[&path],
+             WHERE path = $1 AND LENGTH(content) <= $2;",
+            &[&path, &(max_size)],
         )?;
 
         if rows.is_empty() {
-            Err(PathNotFoundError.into())
+            // This second query distinguishes between a path not found error and a size limit
+            // reached error, as the above query returns no result in either cases.
+            if self
+                .conn
+                .query("SELECT 0 FROM files WHERE path = $1;", &[&path])?
+                .is_empty()
+            {
+                Err(PathNotFoundError.into())
+            } else {
+                Err(
+                    std::io::Error::new(std::io::ErrorKind::Other, crate::error::SizeLimitReached)
+                        .into(),
+                )
+            }
         } else {
             let row = rows.get(0);
             let compression = row.get::<_, Option<i32>>("compression").map(|i| {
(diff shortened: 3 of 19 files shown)


src/utils/sized_buffer.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    pub(crate) fn reserve(&mut self, amount: usize) {
+        if self.inner.len() + amount > self.limit {
+            self.inner.reserve_exact(self.limit - self.inner.len());
+        } else {
+            self.inner.reserve_exact(amount);
+        }
+    }
```
Kixiron, line 22: Maybe this should return the amount that it ended up reserving so that we can pivot off of that as needed, allowing us to maybe less-fatally operate by retrying writes after the buffer's flushed

src/storage/mod.rs
```diff
@@ -324,12 +324,12 @@ mod test {
...
             );
 
-            let file = backend.get("rustdoc/Cargo.toml").unwrap();
+            let file = backend.get("rustdoc/Cargo.toml", std::usize::MAX).unwrap();
             assert_eq!(file.content, b"data");
             assert_eq!(file.mime, "text/toml");
             assert_eq!(file.path, "rustdoc/Cargo.toml");
 
-            let file = backend.get("rustdoc/src/main.rs").unwrap();
+            let file = backend.get("rustdoc/src/main.rs", std::usize::MAX).unwrap();
```
Kixiron, line 349: I think these tests could be improved by setting the limits to under, over and exactly at the length of the content we give, that way we can ensure that the buffering works with actual files as well

src/web/file.rs
```diff
@@ -5,12 +5,14 @@ use crate::{db, error::Result};
...
 
+const MAX_FILE_SIZE: usize = 5 * 1024 * 1024; // 5MB
+
 pub(crate) struct File(pub(crate) db::file::Blob);
 
 impl File {
     /// Gets file from database
     pub fn from_path(conn: &Connection, path: &str) -> Result<File> {
-        Ok(File(db::file::get_path(conn, path)?))
+        Ok(File(db::file::get_path(conn, path, MAX_FILE_SIZE)?))
```
Kixiron, line 15: Maybe we should use an environmental variable so we can adjust the limit on the fly instead of having to redeploy

src/storage/mod.rs
```diff
@@ -163,7 +169,8 @@ impl<'a> Storage<'a> {
                     .map(|file| (file_path, file))
             })
             .map(|(file_path, file)| -> Result<_, Error> {
-                let (content, alg) = compress(file)?;
+                let alg = DEFAULT_COMPRESSION;
+                let content = compress(file, alg)?;
```
Kixiron, line 173: What's the purpose of setting a variable to `DEFAULT_COMPRESSION` instead of using it directly?

src/storage/mod.rs
```diff
@@ -14,6 +14,7 @@ use std::io::Read;
 use std::path::{Path, PathBuf};
 
 const MAX_CONCURRENT_UPLOADS: usize = 1000;
+const DEFAULT_COMPRESSION: CompressionAlgorithm = CompressionAlgorithm::Zstd;
```
Kixiron, line 17: These could also be good targets for an environmental variable of some sort, we don't have multiple algos atm but in the future we might, and adjusting concurrent uploads would also be nice

src/storage/mod.rs
```diff
@@ -14,6 +14,7 @@ use std::io::Read;
 use std::path::{Path, PathBuf};
 
 const MAX_CONCURRENT_UPLOADS: usize = 1000;
```
Kixiron, line 16: Unrelated, but it occurred to me that we could potentially benefit from dynamically selecting the number of concurrent uploads based on the current conditions, e.g. decrease the number of uploads if we're under heavier load

src/storage/mod.rs
```diff
@@ -341,7 +349,8 @@ mod test {
     fn test_batched_uploads() {
         let uploads: Vec<_> = (0..=MAX_CONCURRENT_UPLOADS + 1)
             .map(|i| {
-                let (content, alg) = compress("fn main() {}".as_bytes()).unwrap();
+                let alg = DEFAULT_COMPRESSION;
+                let content = compress("fn main() {}".as_bytes(), alg).unwrap();
```
Kixiron, line 362: Again it seems strange to set a variable to `DEFAULT_COMPRESSION`

src/storage/mod.rs
```diff
@@ -369,16 +378,20 @@ mod test {
...
+            let blob = Blob {
+                mime: "text/rust".into(),
+                content: data.clone(),
+                path: "main.rs".into(),
+                date_updated: Utc::now(),
+                compression: Some(*alg),
+            };
+            test_roundtrip(std::slice::from_ref(&blob));
+            assert_eq!(decompress(data.as_slice(), *alg).unwrap(), orig.as_bytes());
+        }
```
Kixiron: This could also be a good place to make sure the buffering works (ensuring it works for compressed files as well as db ones)

src/storage/mod.rs
```diff
@@ -390,7 +399,27 @@ mod test {
...
+            let data = compress(orginal, *alg).unwrap();
+
+            // Ensure decompressing a file over the limit returns a SizeLimitReached error.
+            let err = decompress(data.as_slice(), *alg, 1000).unwrap_err();
+            assert!(err
+                .downcast_ref::<std::io::Error>()
+                .and_then(|io| io.get_ref())
+                .and_then(|err| err.downcast_ref::<crate::error::SizeLimitReached>())
+                .is_some());
         }
```
Kixiron: Testing edge cases here as well would be good, like undersized and files that are on the buffer limit

src/storage/s3.rs
```diff
@@ -190,6 +191,47 @@ pub(crate) mod tests {
         });
     }
 
+    #[test]
+    fn test_get_too_big() {
```
Kixiron, line 195: Buffer testing here as well?

src/storage/database.rs
```diff
@@ -16,18 +16,37 @@ impl<'a> DatabaseBackend<'a> {
         Self { conn }
     }
 
-    pub(super) fn get(&self, path: &str) -> Result<Blob, Error> {
+    pub(super) fn get(&self, path: &str, max_size: usize) -> Result<Blob, Error> {
         use std::convert::TryInto;
 
+        // The maximum size for a BYTEA (the type used for `content`) is 1GB, so this cast is safe:
+        // https://www.postgresql.org/message-id/162867790712200946i7ba8eb92v908ac595c0c35aee%40mail.gmail.com
+        let max_size = max_size.min(std::i32::MAX as usize) as i32;
```
Kixiron, line 24: Maybe add a debug assertion here to make sure that postgres behaves as expected

src/storage/database.rs
```diff
@@ -91,22 +110,69 @@ mod tests {
...
             // Test that other files are not returned
             assert!(backend
-                .get("dir/bar.txt")
+                .get("dir/bar.txt", std::usize::MAX)
                 .unwrap_err()
                 .downcast_ref::<PathNotFoundError>()
                 .is_some());
             assert!(backend
-                .get("foo.txt")
+                .get("foo.txt", std::usize::MAX)
```
Kixiron, line 123: The usages of `usize::MAX` seem strange to me, it seems like testing with a sane-ish number is better and more realistic than this, which is essentially not having a limit at all

src/storage/database.rs
```diff
@@ -91,22 +110,69 @@ mod tests {
...
                 .unwrap_err()
                 .downcast_ref::<PathNotFoundError>()
                 .is_some());
 
             Ok(())
         });
     }
+
+    #[test]
+    fn test_get_too_big() {
```
Kixiron, line 133: I know I've said it a lot, but maybe more buffering tests here?

src/storage/database.rs
pietroalbini: Hmm, I don't see a way of asserting what the postgres storage limit is. Do you have any idea that doesn't involve writing 1GB of data in the database?

src/storage/mod.rs
pietroalbini: I put `DEFAULT_COMPRESSION` in a variable as that makes changing the function to accept the algorithm as a parameter easier in the future. I'm not too attached to that though, I can surely change this.

src/storage/mod.rs
pietroalbini: See the other comment.

src/storage/mod.rs
pietroalbini: Hmm, these tests are not testing the size limits, they're just the older tests I had to change due to the new argument.

src/storage/s3.rs
pietroalbini: What do you mean?

src/utils/sized_buffer.rs
pietroalbini: Do you see this being needed anywhere in the current PR? If future code needs it, the return value can be added in a future PR.

src/storage/database.rs
pietroalbini: These tests are not testing the size limit, so there is no point in setting a lower limit.

src/storage/database.rs
pietroalbini: What do you mean?

src/storage/mod.rs
pietroalbini: I don't think that's required for the current PR. Feel free to send a future PR though.

src/web/file.rs
pietroalbini: Hmm, I don't think an environment variable is strictly needed, and implementing it correctly might be complex (I'd prefer not to read and parse it every single request).

src/storage/mod.rs
pietroalbini: Hmm? This is not testing buffering.

src/storage/mod.rs
pietroalbini: Good point.

src/storage/mod.rs
Nemo157: With custom dictionaries the compression will probably be chosen on a per-file basis, so not easily configurable other than by changing code.

Nemo157:
LGTM

src/storage/mod.rs
Kixiron: Buffering is still involved within the pipeline though, isn't it?

src/storage/database.rs
Kixiron: I meant asserting that the size we're given matches what we believe will happen, e.g `assert!(max_size <= i32::max_value() as usize);`

src/storage/s3.rs
Kixiron: Well, the sized buffer is used indirectly by the things being tested, and so making sure that the sized buffer preforms as expected in a real scenario seems logical to me

src/storage/mod.rs
pietroalbini: Yeah, but the fact buffering works is already tested elsewhere. We don't need to add such tests every time we access the database.

src/storage/database.rs
pietroalbini: Well, the line we're commenting on already ensures that... Also, the tests will catch it if for whatever reason it stops working (I needed to add this line *because* tests were failing).

pietroalbini:
Addressed all the review comments, and changed the maximum size to 50MB for non-HTML files.

I think this is ready to go, I'll follow up with a PR that does a couple refactorings and then adds a test for `File::from_path`. One last check @Kixiron @jyn514?

pietroalbini:
Nevermind, also added the final test.

src/storage/database.rs
```diff
@@ -26,29 +26,28 @@ impl<'a> DatabaseBackend<'a> {
...
-                Err(
-                    std::io::Error::new(std::io::ErrorKind::Other, crate::error::SizeLimitReached)
-                        .into(),
-                )
-            }
+            Err(PathNotFoundError.into())
         } else {
             let row = rows.get(0);
+
+            if row.get::<_, bool>("is_too_big") {
```
jyn514, line 43: nit: I think rust can infer the type here.
```suggestion
            if row.get("is_too_big") {
```

src/storage/mod.rs
```diff
@@ -369,16 +387,60 @@ mod test {
...
+        const MAX_SIZE: usize = 1024;
+
+        let small = &[b'A'; MAX_SIZE / 2] as &[u8];
+        let exact = &[b'A'; MAX_SIZE] as &[u8];
+        let big = &[b'A'; MAX_SIZE * 2] as &[u8];
+
+        for alg in CompressionAlgorithm::AVAILABLE {
+            let compressed_small = compress(small, *alg).unwrap();
+            let compressed_exact = compress(exact, *alg).unwrap();
+            let compressed_big = compress(big, *alg).unwrap();
```
jyn514, line 420: nit: you don't need all this `*alg`:
```suggestion
        for &alg in CompressionAlgorithm::AVAILABLE {
            let compressed_small = compress(small, alg).unwrap();
            let compressed_exact = compress(exact, alg).unwrap();
            let compressed_big = compress(big, alg).unwrap();
```

src/bin/cratesfyi.rs
```diff
@@ -1,17 +1,20 @@
...
 use structopt::StructOpt;
 
-pub fn main() {
+pub fn main() -> Result<(), Error> {
     let _ = dotenv::dotenv();
     logger_init();
 
-    CommandLine::from_args().handle_args();
+    CommandLine::from_args().handle_args()?;
+    Ok(())
```
jyn514, line 17: nit: I think this looks nicer.
```suggestion
    CommandLine::from_args().handle_args()
```

src/config.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct Config {
+    pub(crate) max_file_size: usize,
+    pub(crate) max_file_size_html: usize,
+}
+
+impl Config {
+    pub fn from_env() -> Result<Self, Error> {
```
jyn514, line 13: The idea is that over time we'll move more of the environment variables here, instead of being spread out over the codebase? I like it, it makes things much easier to test as well.

jyn514:
Looks good to me modulo nits.

src/config.rs
pietroalbini: Yep, that's the goal!

pietroalbini:
@jyn514 addressed the nits.

I also updated how tests can override configuration values to this:

```rust
env.override_config(|config| {
   config.key = value;
});
```

Using the closure allows to change the default configuration for tests in the future without having to update the base config in each test that overrides it.

src/config.rs
Kixiron: I completely support that, having a concrete storage of env vars in one place would be fantastic

jyn514:
Looks great, thanks!

PR #838, opened 2020-06-17
Note that builds aren't immediate

jyn514:
This (hopefully) avoids people being surprised when their docs aren't there, and also links them to the page that has a progress bar.


jyn514:
Before:
image

After:
image

PR #839, opened 2020-06-18
Deal with nullability of owner data

Nemo157:
Most of the owner fields can be `null`, depending on what data was available from the Github API when crates.io scraped it. The old implementation turned any null field into an empty string, which still seems like the best approach.


jyn514:
For reference: this is a follow-up to https://github.com/rust-lang/docs.rs/pull/813 that fixes some bugs I found while testing locally.

PR #841, opened 2020-06-18
Remove TEMPLATE_DATA static

pietroalbini:
This PR removes the `TEMPLATE_DATA` static from the codebase, switching to a struct stored in the request extensions. This is a requirement for future refactorings. The PR can be reviewed commit-by-commit.

r? @Kixiron

### DIFF
--- a/src/web/page/templates.rs
+++ b/src/web/page/templates.rs
@@ -1,11 +1,16 @@
 use crate::error::Result;
 use arc_swap::ArcSwap;
 use chrono::{DateTime, Utc};
+use notify::{watcher, RecursiveMode, Watcher};
 use serde_json::Value;
 use std::collections::HashMap;
+use std::sync::{mpsc::channel, Arc};
+use std::thread;
+use std::time::Duration;
 use tera::{Result as TeraResult, Tera};
 
 /// Holds all data relevant to templating
+#[derive(Debug)]
 pub(crate) struct TemplateData {
     /// The actual templates, stored in an `ArcSwap` so that they're hot-swappable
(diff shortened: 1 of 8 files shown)


src/web/page/templates.rs
```diff
@@ -99,13 +82,28 @@ fn load_rustc_resource_suffix() -> Result<String> {
...
+        Value::String(crate::BUILD_VERSION.into()),
+    );
+    ReturnValue::add_function_to(
+        &mut tera,
+        "rustc_resource_suffix",
+        Value::String(load_rustc_resource_suffix().unwrap_or_else(|err| {
+            log::error!("Failed to load rustc resource suffix: {:?}", err);
+            String::from("???")
+        })),
+    );
```
Kixiron, line 108: Maybe add back the comments that were on the struct so the function's meaning is clear(er)

src/web/page/templates.rs
```diff
@@ -115,28 +113,23 @@ pub(super) fn load_templates() -> TeraResult<Tera> {
...
 
-/// Returns an `Option<GlobalAlert>` in json form for templates
-fn global_alert(args: &HashMap<String, Value>) -> TeraResult<Value> {
-    debug_assert!(args.is_empty(), "global_alert takes no args");
-
-    Ok(TEMPLATE_DATA.global_alert.clone())
+/// Simple function that returns the pre-defined value.
+struct ReturnValue {
+    name: &'static str,
+    value: Value,
```
Kixiron, line 121: Good idea on this!

Issue #822, opened 2020-06-06
There are no tests for a full build


pietroalbini:
We don't currently execute any kind of test to ensure a full build actually works. Such tests would've caught the root cause of today's outage.

pietroalbini:
I'll work on implementing those tests next week.

jyn514:
As a short-term thing, I think it would be fairly easy to add unit tests that we're using the crates.io API properly.

jyn514:
This was fixed by https://github.com/rust-lang/docs.rs/pull/1309.

PR #843, opened 2020-06-21
Use a single connection pool across the whole application

pietroalbini:
This PR changes how database connections are handled in the docs.rs application:

* Some code has been shuffled around, moving `web::InjectExtensions` to its own file and `web::pool::Pool` to `db::pool::Pool`.
* The configuration for the database is now stored in `Config` instead of being loaded from the environment each time a connection is created.
* The application is changed to create a single instance of `Pool`, which is created in the `cratesfyi` CLI and passed around. One result of this is that the frontend and the builds will use the same pool, instead of having the builds create standalone connections when needed.
* `db::connect_db()` is removed, forcing every connection to go through the database pool.

This PR is part of my refactorings to make buildtests feasible, and is best reviewed commit by commit. Unfortunately the last commit is quite big, but there was no way to split it.


Nemo157:
How certain are we that the pool is fork-safe? (Given that we initialise it and pass it off to `start_daemon` which may then `fork` the process).

pietroalbini:
Good point. I'd be tempted to just remove the forking code: we run docs.rs on production with systemd, and tbh everyone should just do that.

Nemo157:
+1, I think self-backgrounding is an antipattern that can be dropped. Last I asked about it @jyn514 said it was used in production still, so might need a reconfiguration there to stop using it.

pietroalbini:
Opened https://github.com/rust-lang/docs.rs/pull/844 to remove forking.

jyn514:
> How certain are we that the pool is fork-safe? (Given that we initialise it and pass it off to start_daemon which may then fork the process).

Technically all IO is unsafe unless it's been explicitly whitelisted:

>        *  After  a  fork()  in  a multithreaded program, the child can safely call
          only async-signal-safe functions (see signal-safety(7)) until such  time
          as it calls execve(2).

src/docbuilder/queue.rs
```diff
@@ -77,32 +76,42 @@ impl DocBuilder {
...
+            }
+
+            let id: i32 = query.get(0).get(0);
+            let name: String = query.get(0).get(1);
+            let version: String = query.get(0).get(2);
 
-        let id: i32 = query.get(0).get(0);
-        let name: String = query.get(0).get(1);
-        let version: String = query.get(0).get(2);
+            (id, name, version)
```
jyn514, line 102:
```suggestion
            let next_crate = query.get(0);
            (next_crate.get("id"), next_crate.get("name"), next_crate.get("version"))
```
This might need fiddling with the `let (id, name, version)` at the top for type inferrence.

src/utils/daemon.rs
```diff
@@ -220,38 +223,44 @@ pub fn start_daemon(background: bool, config: Arc<Config>) -> Result<(), Error>
...
-
-            if now.hour() == 23 && now.minute() == 55 {
-                info!("Updating release activity");
-                if let Err(e) = update_release_activity() {
-                    error!("Failed to update release activity: {}", e);
-                }
-            }
-        })
-        .unwrap();
+    let cloned_db = db.clone();
```
jyn514, line 226: Why do each of these threads need a separate `cloned_db`?

src/utils/daemon.rs
```diff
@@ -220,38 +223,44 @@ pub fn start_daemon(background: bool, config: Arc<Config>) -> Result<(), Error>
...
+    cron("release activity updater", 60, move || {
```
jyn514, line 227: Let's avoid having an implicit 'seconds' here, can you change `cron` to take `Duration` instead of `u64`?

src/utils/daemon.rs
```diff
@@ -220,38 +223,44 @@ pub fn start_daemon(background: bool, config: Arc<Config>) -> Result<(), Error>
...
+fn cron<F>(name: &'static str, interval: u64, exec: F) -> Result<(), Error>
+where
+    F: Fn() -> Result<(), Error> + Send + 'static,
+{
+    thread::Builder::new()
+        .name(name.into())
+        .spawn(move || loop {
+            thread::sleep(Duration::from_secs(interval));
+            if let Err(err) = exec() {
+                error!("failed to run cron '{}': {:?}", name, err);
```
jyn514, line 261:
```suggestion
                error!("failed to run scheduled task '{}': {:?}", name, err);
```

src/web/mod.rs
```diff
@@ -401,11 +397,13 @@ impl Server {
         conn: Arc<Mutex<Connection>>,
         config: Arc<Config>,
     ) -> Result<Self, Error> {
+        let templates = TemplateData::new(&conn.lock().unwrap())?;
```
jyn514, line 400: We should probably move away from using `Arc<Mutex>` since it means that any test that panics will cause other tests to panic. In the meantime though:
```suggestion
        let templates = TemplateData::new(&conn.lock().expect("failed to access shared database connection (did another test panic?)"))?;
```

src/web/page/handlebars.rs
```diff
@@ -16,7 +16,14 @@ lazy_static::lazy_static! {
 }
 
 fn load_rustc_resource_suffix() -> Result<String, failure::Error> {
-    let conn = crate::db::connect_db()?;
+    // New instances of the configuration or the connection pool shouldn't be created inside the
+    // application, but we're removing handlebars so this is not going to be a problem in the long
+    // term. To avoid wasting resources, the pool is hardcoded to only keep one connection alive.
```
jyn514, line 21: Why not pass in the connection as a parameter?

src/db/pool.rs
```diff
@@ -15,20 +15,49 @@ pub(crate) enum Pool {
...
+
+        let min_pool_idle = env::var("DOCSRS_MIN_POOL_IDLE")
+            .map(|s| {
+                s.parse::<u32>()
+                    .expect("DOCSRS_MIN_POOL_IDLE must be an integer")
+            })
+            .unwrap_or(10);
+
+        let manager = r2d2_postgres::PostgresConnectionManager::new(
+            &db_url[..],
```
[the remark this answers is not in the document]
jyn514: No longer relevant since this was deleted in a later commit.

Nemo157:
> Technically all IO is unsafe unless it's been explicitly whitelisted:

I was more thinking that if the pool spawns a background thread to deal with closing old connections that thread won't make it across the `fork`.

src/utils/daemon.rs
pietroalbini: Because that instance of the pool is moved in each closure.

src/web/mod.rs
pietroalbini: Hmm? Each test has its own instance of `TestDatabase`, with its own mutexed connection.

src/web/page/handlebars.rs
pietroalbini: It's not an easy change, since that function is called from the `lazy_static!`. I'd need to make something like #841, but I don't want to bother with it since @Kixiron will hopefully remove handlebars soon (:heart:).

src/web/utils/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod extensions;
+
+pub(crate) use self::extensions::InjectExtensions;
```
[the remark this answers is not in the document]
pietroalbini: I wanted to move it to a module since at the moment `web` contains all the routes, and having that merged with the other files could be confusing. I'm open to change it though.

pietroalbini:
Replied to or addressed all review comments.

src/web/page/handlebars.rs
jyn514: Got it, I agree that the change isn't worth the effort then.

src/utils/daemon.rs
jyn514: Oh I see, this is https://github.com/rust-lang/rfcs/issues/2407. No need to change anything.

src/web/mod.rs
jyn514: Why do failures in one test cause others to fail then?

```
thread 'web::test::yanked_crates_are_not_considered' panicked at 'failed to lock the connection: "PoisonError { inner: .. }"', src/test/mod.rs:187:9
```

src/web/utils/mod.rs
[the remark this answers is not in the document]
jyn514: I thought only `web/routes.rs` had routes?

jyn514:
Marking as P-high since it blocks #822.

src/web/mod.rs
pietroalbini: Because we try to access the connection during `Drop`, and the Mutex will be poisoned if the test panicked.

pietroalbini:
Rebased on top of the latest master and fixed the last concern in the last commit.

r? @jyn514

PR #844, opened 2020-06-21
Remove daemonization code from docs.rs

pietroalbini:
This commit removes the forking code in the docs.rs daemon, forcing it to always run in the background. The rationale for the change is that forking adds complexity and sets extra constraint for development, while it's not needed anymore on modern infrastructure.

The --foreground flag is kept in the binary for compatibility, and it will only print a warning. Once this PR is deployed in production for a while, the flag can be removed.

r? @jyn514

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -62,7 +62,7 @@ enum CommandLine {
 
     /// Starts cratesfyi daemon
     Daemon {
-        /// Run the server in the foreground instead of detaching a child
+        /// Deprecated. Run the server in the foreground instead of detaching a child
         #[structopt(name = "FOREGROUND", short = "f", long = "foreground")]
         foreground: bool,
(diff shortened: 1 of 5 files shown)


jyn514:
Looks great to me, the `fork()` and then doing IO always made uncomfortable anyway (technically it's UB by the C standard if you have threads and don't call `execv`).

src/bin/cratesfyi.rs
```diff
@@ -93,7 +93,11 @@ impl CommandLine {
                 Server::start(Some(&socket_addr), reload_templates, config)?;
             }
             Self::Daemon { foreground } => {
```
jyn514, line 95: Can you also update the doc-comment on `foreground` to note that it's deprecated?

pietroalbini: Done.

PR #846, opened 2020-06-22
Add metrics for the rustdoc rendering times

pietroalbini:
I have a feeling the next refactoring I'm going to make (using a shared `Storage` instance for the whole application) is going to improve performance of rustdoc pages, so this adds some code to expose a metric that breaks down that endpoint's rendering time.

That endpoint is the one I think we should care about the most, so this is also going to be helpful to drive other performance improvements in the future.

### DIFF
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -72,6 +72,13 @@ lazy_static::lazy_static! {
     )
     .unwrap();
 
+    pub static ref RUSTDOC_RENDERING_TIMES: HistogramVec = register_histogram_vec!(
+        "docsrs_rustdoc_rendering_time",
+        "The time it takes to render a rustdoc page",
+        &["step"]
+    )
+    .unwrap();
+
     pub static ref FAILED_DB_CONNECTIONS: IntCounter = register_int_counter!(
         "docsrs_failed_db_connections",
         "Number of attempted and failed connections to the database"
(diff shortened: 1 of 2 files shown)


src/web/metrics.rs
```diff
@@ -197,6 +204,47 @@ impl iron::Handler for RequestRecorder {
...
+    pub(crate) fn step(&mut self, step: &'static str) {
+        self.record_current();
+        self.current = Some(RenderingTime {
+            start: Instant::now(),
+            step,
+        });
+    }
+
+    fn record_current(&mut self) {
+        if let Some(current) = self.current.take() {
```
jyn514, line 234: This doesn't count the time to record the metric, is that intentional? If not you could make it (mostly) atomic like this:
```suggestion
        if let Some(current) = self.current.replace(Instant::now()) {
```
and then removing `current = ...` in `step()`.

pietroalbini: Is that really needed? The thing I want to get out of this metrics is how much it takes to do database queries, fetch files from S3 and parse the HTML, and the time it takes to record metrics won't influence that.

pietroalbini:
> Can you post some screenshots of what this looks like in grafana?

I don't have a (working) Grafana setup locally.

src/web/metrics.rs
jyn514: Ok, if it's not valuable we don't need to include it.

jyn514:
Going to make a separate PR for a `prometheus` instance in docker-compose. This PR looks fine.

PR #847, opened 2020-06-22
Don't keep the index repository open long term

Nemo157:
After deploying #813 there was a big change in the number of used file descriptors. This is likely a result of <https://github.com/libgit2/libgit2/issues/2758> given that #813 changed behaviour to keep the index repository open instead of just opening it when needed.

Given how long that `libgit2` issue and the PR fixing it have been open, it seems unlikely to be fixed soon. It would probably be ok to have all the packfiles open all the time, except based on [this comment](https://github.com/libgit2/libgit2/issues/2758#issuecomment-582610956) when running a concurrent `git gc` the garbage collected packfiles will not be closed, so there will be a constant slow fd leak until the service is restarted (which you can see in the graph below).

image

This PR changes behaviour back to match the old setup, only opening the repository when needed, and sticks some more context on related errors just in case.

### DIFF
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -4,11 +4,11 @@ use url::Url;
 
 use self::api::Api;
 use crate::error::Result;
+use failure::ResultExt;
 
 pub(crate) mod api;
 
 pub(crate) struct Index {
-    diff: crates_index_diff::Index,
     path: PathBuf,
     api: Api,
 }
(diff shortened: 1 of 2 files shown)


src/index/mod.rs
```diff
@@ -41,14 +41,17 @@ fn load_config(repo: &git2::Repository) -> Result<IndexConfig> {
 impl Index {
     pub(crate) fn new(path: impl AsRef<Path>) -> Result<Self> {
         let path = path.as_ref().to_owned();
-        let diff = crates_index_diff::Index::from_path_or_cloned(&path)?;
-        let config = load_config(diff.repository())?;
-        let api = Api::new(config.api)?;
-        Ok(Self { diff, path, api })
+        let diff = crates_index_diff::Index::from_path_or_cloned(&path)
+            .context("initialising registry index repository")?;
```
jyn514, line 45:
```suggestion
        // This initializes the repository, then closes it afterwards to avoid leaking file descriptors.
        // See https://github.com/rust-lang/docs.rs/pull/847
        let diff = crates_index_diff::Index::from_path_or_cloned(&path)
            .context("initialising registry index repository")?;
```

PR #850, opened 2020-06-22
Add prometheus instance to docker-compose

jyn514:
This allows viewing metrics while developing locally.

r? @Kixiron


jyn514:
I know @Nemo157 wanted this off by default, but the time to start up is negligible compared to s3 or the database (~2 seconds).

docker/prometheus.yml
```diff
@@ -0,0 +1,14 @@
...
+  # external systems (federation, remote storage, Alertmanager).
+  external_labels:
+    monitor: 'docs.rs-monitor'
+
+scrape_configs:
+  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  - job_name: 'docs.rs'
+    metrics_path: '/about/metrics'
+    static_configs:
+      - targets: ['web:3000']
```
Kixiron, line 14: I don't know if `web` works on Docker for Windows, but `host.docker.internal` works on all platforms if you add
```yaml
extra_hosts:
  - "host.docker.internal:host-gateway"
```
to the compose file

docker-compose.yml
```diff
@@ -44,6 +48,12 @@ services:
         environment:
             MINIO_ACCESS_KEY: cratesfyi
             MINIO_SECRET_KEY: secret_key
+    prometheus:
+        build:
+            context: ./docker
+            dockerfile: ./Dockerfile-prometheus
+        ports:
+            - "9090:9090"
```
Kixiron, line 56: Making the prometheus service optional can be done by [making an extra file](https://stackoverflow.com/a/45682432). While that's not ideal in any sense, it's the best that can be done until this [open but closed](https://github.com/docker/compose/issues/1896) issue gets through

jyn514: Sure, but now we have to document and remember that there are two different compose files. I'm not sure the maintenance is worth it since the startup time is so low.

docker/prometheus.yml
jyn514: This is a docker-compose feature, not docker. Do you mind testing this on windows so we know if it works? I'd be very surprised if not.

Kixiron: Yep, works on windows

Issue #856, opened 2020-06-28
/target-redirect panics if not given a trailing slash


jyn514:
For example, `docs.rs/crate/log/0.4.8/target-redirect/x86_64-apple-darwin`.

Relevant code: https://github.com/rust-lang/docs.rs/blob/301acacdb1abd37ec1fa18e673e83375c67249cc/src/web/rustdoc.rs#L442

Instead of panicking we should give up on searching and redirect to `/:platform` instead of `/:platform?search=$search_item`.

PR #857, opened 2020-06-28
Fix panic on target-redirect with no trailing slash

jyn514:
- Remove use of unwrap
- Add regression test

Closes https://github.com/rust-lang/docs.rs/issues/856

r? @Nemo157

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -477,10 +477,8 @@ pub fn target_redirect_handler(req: &mut Request) -> IronResult<Response> {
         if file_path[3] == crate_details.metadata.default_target {
             file_path.remove(3);
         }
-        if let Some(last) = file_path.last_mut() {
-            if *last == "" {
-                *last = "index.html";
-            }
+        if let Some(last @ &mut "index.html") = file_path.last_mut() {
+            *last = "index.html";
         }
         file_path
     };
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -428,20 +428,25 @@ fn path_for_version(
         // NOTE: this adds 'index.html' if it wasn't there before
         return req_path[3..].join("/");
     }
+    // check if req_path[3] is the platform choice or the name of the crate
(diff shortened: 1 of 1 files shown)


src/web/rustdoc.rs
```diff
@@ -1397,6 +1396,20 @@ mod test {
         })
     }
 
+    #[test]
+    // regression test for https://github.com/rust-lang/docs.rs/issues/856
+    fn test_no_trailing_slash() {
+        crate::test::wrapper(|env| {
+            let db = env.db();
+            db.fake_release().name("dummy").version("0.1.0").create()?;
```
Nemo157, line 1404: It'd be good to have a non-default-platform test in here as well.

jyn514: Good idea for the test, this indeed redirects to `/dummy/` for `/target-redirect/x86_64-apple-darwin`.

However the solution is not clear to me. My first idea was to treat the last part of the path as a platform if it was missing a `.`, but then we'll get 404s for `/target-redirect/platform-that-doesnt-exist`. I think the current behavior is better since no one should be redirected here without a trailing slash, they'd have to type the URL in manually.

jyn514: Oh hold on that's a separate issue altogether. I'll add a fix for that too.

jyn514:
This is ready for re-review.

src/web/rustdoc.rs
```diff
@@ -477,10 +477,8 @@ pub fn target_redirect_handler(req: &mut Request) -> IronResult<Response> {
         if file_path[3] == crate_details.metadata.default_target {
             file_path.remove(3);
         }
-        if let Some(last) = file_path.last_mut() {
-            if *last == "" {
-                *last = "index.html";
-            }
+        if let Some(last @ &mut "index.html") = file_path.last_mut() {
```
Nemo157, line 480:
```suggestion
        if let Some(last @ &mut "") = file_path.last_mut() {
```

PR #861, opened 2020-06-28
Improve executing tests

pietroalbini:
This PR makes two changes that will help executing tests:

* Moves tests to use a downsized connection pool instead of an `Arc<Mutex<Connection>>`
* Replaces Tera's template loading code with a custom function, making template loading instantaneous (it took 4 seconds on my NVMe before). See the comment in the code for why that was happening.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,8 @@ staticfile = { version = "0.4", features = [ "cache" ] }
 tempfile = "3.1.0"
 
 # Templating
-tera = { version = "1.3.0", features = ["builtins"] }
+tera = { version = "1.3.1", features = ["builtins"] }
+walkdir = "2"
 
 # Template hot-reloading
 arc-swap = "0.4.6"
@@ -67,16 +68,16 @@ notify = "4.0.15"
 chrono = { version = "0.4.11", features = ["serde"] }
 time = "0.1" # TODO: Remove once `iron` is removed
 
+[dependencies.postgres]
+version = "0.15"
+features = ["with-chrono", "with-serde_json"]
+
 [target.'cfg(not(windows))'.dependencies]
 # Process information
 procfs = "0.7"
 
 [target.'cfg(windows)'.dependencies]
 path-slash = "0.1.1"
-
-[dependencies.postgres]
-version = "0.15"
-features = ["with-chrono", "with-serde_json"]
 
 [dev-dependencies]
 once_cell = "1.2.0"
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1,5 +1,6 @@
 mod fakes;
 
+use crate::db::{Pool, PoolConnection};
 use crate::storage::s3::TestS3;
 use crate::web::Server;
 use crate::Config;
@@ -11,10 +12,7 @@ use reqwest::{
     blocking::{Client, RequestBuilder},
     Method,
 };
-use std::{
-    panic,
-    sync::{Arc, Mutex, MutexGuard},
-};
+use std::{panic, sync::Arc};
 
 pub(crate) fn wrapper(f: impl FnOnce(&TestEnvironment) -> Result<(), Error>) {
     let _ = dotenv::dotenv();
@@ -123,7 +121,13 @@ impl TestEnvironment {
     }
 
     fn base_config(&self) -> Config {
-        Config::from_env().expect("failed to get base config")
+        let mut config = Config::from_env().expect("failed to get base config");
+
+        // Use less connections for each test compared to production.
+        config.max_pool_size = 2;
+        config.min_pool_idle = 0;
+
+        config
     }
 
     pub(crate) fn override_config(&self, f: impl FnOnce(&mut Config)) {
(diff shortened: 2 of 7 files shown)


src/web/page/templates.rs
```diff
@@ -84,7 +89,16 @@ fn load_rustc_resource_suffix(conn: &Connection) -> Result<String> {
...
 pub(super) fn load_templates(conn: &Connection) -> Result<Tera> {
-    let mut tera = Tera::new("tera-templates/**/*")?;
+    // This uses a custom function to find the templates in the filesystem instead of Tera's
+    // builtin way (passing a glob expression to Tera::new), speeding up the startup of the
+    // application and running the tests.
+    //
+    // The problem with Tera's template loading code is, it walks all the files in the current
+    // directory and matches them against the provided glob expression. Unfortunately this means
+    // Tera will walk all the rustwide workspaces, the git repository and a bunch of other
+    // unrelated data, slowing down the search a lot.
```
jyn514, line 99: I'm going to try and get this upstreamed to globwalk. This workaround is fine in the meantime though.
```suggestion
    // The problem with Tera's template loading code is, it walks all the files in the current
    // directory and matches them against the provided glob expression. Unfortunately this means
    // Tera will walk all the rustwide workspaces, the git repository and a bunch of other
    // unrelated data, slowing down the search a lot.
    //
    // TODO: remove this when https://github.com/Gilnaa/globwalk/issues/29 is fixed
```

Cargo.toml
```diff
@@ -57,7 +57,8 @@ staticfile = { version = "0.4", features = [ "cache" ] }
 tempfile = "3.1.0"
 
 # Templating
-tera = { version = "1.3.0", features = ["builtins"] }
+tera = { version = "1.3.1", features = ["builtins"] }
```
jyn514, line 60:
```suggestion
tera = { version = "1", features = ["builtins"] }
```
Cargo ignores everything but semver, this makes it look like tera was updated when it wasn't (well, it was updated in the PR but only because you ran `cargo update -p tera`)

src/test/mod.rs
```diff
@@ -123,7 +121,13 @@ impl TestEnvironment {
     }
 
     fn base_config(&self) -> Config {
-        Config::from_env().expect("failed to get base config")
+        let mut config = Config::from_env().expect("failed to get base config");
+
+        // Use less connections for each test compared to production.
+        config.max_pool_size = 2;
```
jyn514, line 127: 2 seems a little small, don't we want as many connections as we have cores so that tests can run in parallel?
```suggestion
        // Use less connections for each test compared to production.
        config.max_pool_size = 32;
```

Cargo.toml
pietroalbini: No, we need at least 1.3.1, as [some generics parameters we need were introduced in that release](https://github.com/Keats/tera/commit/f7835a2c7e2a4b0c25947b70aea3d9327f54eb29). Dunno why it was done in a patch release.

src/test/mod.rs
pietroalbini: That's the connection pool for each test, not the connection pool for the whole `cargo test`.

Cargo.toml
jyn514: Oh weird. Please say that in Cargo.toml then:
```suggestion
tera = { version = ">=1.3.1", features = ["builtins"] }
```

Kixiron: I know it's a total style thing, but I think that having a "fully qualified" versions is just better style since it allows programmers to find the correct version of documentation much quicker and is overall more specific than a blanket "give me a `1`". Additionally those two have different meanings, since `1` is asking for tera version `>=1.0.0, <2.0.0` while `1.3.1` is asking for tera version `>=1.3.1, <2.0.0`. Information is lost, and while that in practice has no effect I still think it's not good

pietroalbini: @jyn514 isn't that equivalent?

jyn514: They are not equivalent. `"1.3.1"` has exactly the same meaning as `"1"`.

pietroalbini: @jyn514 no, [according to the docs](https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) `1.3.1` is equivalent to `^1.3.1`, which is equivalent to `>=1.3.1, <2.0.0`. Using just `>=1.3.1` would allow to depend on Tera `2.0.0`, which is not what we want.

jyn514: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

Kixiron: No, unadorned versions are [interpreted as careted](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

jyn514: I missed that `^0.1.12` has an implicit `>=0.1.12` bound. So this is fine.

PR #863, opened 2020-06-29
Fix queue failure, refactor it and add a bunch of tests

pietroalbini:
This PR fixes the documentation builder failing to get the next crate out of the queue, refactors the whole queue handling out of `DocBuilder` to allow easier tests, unifies all the queue handling into the new `BuildQueue` and puts the number of attempts in the configuration. Along with the changes, tests were added for everything.

r? @jyn514


src/web/releases.rs
```diff
@@ -1079,4 +1066,48 @@ mod tests {
...
+                .expect("missing list items")
+                .collect::<Vec<_>>();
+
+            assert_eq!(items.len(), 3);
+            let expected = [
+                ("bar", "0.1.0", Some(10)),
+                ("foo", "1.0.0", None),
+                ("baz", "0.0.1", Some(-10)),
+            ];
+            for (i, li) in items.iter().enumerate() {
```
Nemo157, line 1098:
```suggestion
            for (li, expected) in items.iter().zip(expected.iter()) {
```

src/queue.rs
```diff
@@ -0,0 +1,320 @@
...
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn test_failed_count() {
+        const MAX_ATTEMPTS: u16 = 3;
+        crate::test::wrapper(|env| {
+            env.override_config(|config| {
+                config.build_attempts = 3;
```
Nemo157, line 251:
```suggestion
                config.build_attempts = MAX_ATTEMPTS;
```

src/docbuilder/queue.rs
```diff
@@ -63,74 +66,22 @@ impl DocBuilder {
...
-                 LIMIT 1",
-                &[],
-            )?;
-
-            if query.is_empty() {
-                // nothing in the queue; bail
-                return Ok(false);
-            }
+        let mut processed = false;
+        let queue = self.build_queue.clone();
```
Nemo157, line 77: Is this clone necessary?

src/queue.rs
```diff
@@ -0,0 +1,320 @@
...
+                version: row.get("version"),
+                priority: row.get("priority"),
+            })
+            .collect())
+    }
+
+    pub(crate) fn process_next_crate(
+        &self,
+        f: impl FnOnce(&QueuedCrate) -> Result<()>,
+    ) -> Result<()> {
```
Nemo157, line 84:
```suggestion
    pub(crate) fn process_next_crate<R>(
        &self,
        f: impl FnOnce(&QueuedCrate) -> Result<Option<R>>,
    ) -> Result<R> {
```

seems like it would simplify usage in a few places that use `let mut` to communicate back from the closure.

src/docbuilder/queue.rs
pietroalbini: Yes, because the closure captures `self`.

src/queue.rs
pietroalbini: Hmm, this change would be tricky without refactoring how the daemon starts builds, because `processed` needs to be `true` both if the function returns `Ok(_)` (the build passed or failed) or `Err(_)` (there was an error updating the queue after the build). I'm not sure it's worth doing it right now.

pietroalbini:
@Nemo157 addressed or responded to all review comments.

src/queue.rs
Nemo157: Hmm, since it `?`s the `process_next_crate` call won’t it skip the usage of `processed` on any errors?

src/bin/cratesfyi.rs
```diff
@@ -476,5 +490,51 @@ impl BlacklistSubcommand {
...
+            .get_or_try_init::<_, Error>(|| Ok(Arc::new(Config::from_env()?)))?
+            .clone())
+    }
+
+    fn pool(&self) -> Result<Pool, Error> {
+        Ok(self
+            .pool
+            .get_or_try_init::<_, Error>(|| Ok(Pool::new(&*self.config()?)?))?
+            .clone())
+    }
```
jyn514, line 533: This function is called a lot. Would it be better to remove the once_cell in favor of constructing the pool eagerly? That way it doesn't have to wait for a mutex lock each time.

I guess it only needs to lock the mutex when initializing for the first time, so it wouldn't be a problem in prod, but it would make tests even more painful than currently since they all run in parallel.

src/queue.rs
```diff
@@ -0,0 +1,320 @@
...
+        let queued = self.queued_crates()?;
+        let to_process = match queued.get(0) {
+            Some(krate) => krate,
+            None => return Ok(()),
+        };
+
+        match f(&to_process) {
+            Ok(()) => {
+                conn.execute("DELETE FROM queue WHERE id = $1;", &[&to_process.id])?;
+                crate::web::metrics::TOTAL_BUILDS.inc();
```
jyn514, line 96: We used to increment `TOTAL_BUILDS` regardless of whether the build succeeded or not. Was this change intentional?

src/utils/queue.rs
```diff
@@ -164,37 +149,4 @@ mod tests {
...
-                let row = query.iter().next().unwrap();
-
-                assert_eq!(&row.get::<_, String>(0), name);
-                assert_eq!(&row.get::<_, String>(1), version);
-                assert_eq!(row.get::<_, i32>(2), *priority);
-            }
-
-            Ok(())
-        })
-    }
```
jyn514, line 199: Where did this test go?

jyn514:
Oh wait, I can't run a local build because that doesn't go through the queue 🙃 hopefully your tests exercise everything.

Can you also rebase over master so the Docker failure will go away?

Nemo157:
`docker-compose run web queue add` should work I think? (Or is the queue builder also not enabled)

jyn514:
> docker-compose run web queue add should work I think? (Or is the queue builder also not enabled)

The only way to enable the queue builder is by running the daemon, which tries to build all 300,000 crates off of crates.io. Just downloading that number of crates takes several hours and it's a real waste of bandwidth.

It would be great to refactor the builder and the daemon to be separate at some point.

src/bin/cratesfyi.rs
pietroalbini: `once_cell` is not a Mutex. Also, I don't see how this would be a problem with tests: tests don't use this struct, and they don't share any state or Mutex except for metrics and the handlebars rustc version.

src/utils/queue.rs
pietroalbini: It was split in the `test_add_and_process_crates` and `test_queued_crates`.

src/queue.rs
pietroalbini: Fixed.

pietroalbini:
Addressed the feedback and moved the code from `src/queue.rs` to `src/build_queue.rs`.

src/bin/cratesfyi.rs
jyn514: https://docs.rs/crate/once_cell/1.4.0/source/src/imp_pl.rs

It uses a mutex when initializing. So if many threads try to initialize at once they might fight over the lock.

Since this doesn't run in tests though it shouldn't be an issue as we call `pool()` pretty early.

PR #849, opened 2020-06-22
Moved all templates that use Release or CrateDetails to Tera

Kixiron:
This is unfortunately a really large PR and I apologize, but there wasn't really any other choice since they're all interdependent on how `Release` and `CrateDetails` serialize. I didn't delete any handlebars templates since at this stage all the inter-dependence is really fuzzy and hard to discern, so I think it's just not worth the risk of deleting an entire segment of the site. It should only be 2 or three more PRs after this, so it shouldn't be a huge deal to keep them around for a bit
This is best taken by commit, each commit deals with one single template (which can amount to multiple pages in some cases).

## [Commit 1](https://github.com/rust-lang/docs.rs/commit/96acd9ad74f9d9bf71ca4639c0f84f2a82027f6a)

<details>

### Before

*Formatted for readability*

```xml
<?xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
    <title>Docs.rs</title>
    <subtitle>Recent Rust crates</subtitle>

    <link href="https://docs.rs/releases/feed" rel="self" />
    <link href="https://docs.rs/" />
    <link href="https://pubsubhubbub.appspot.com" rel="hub" />
    <link href="https://pubsubhubbub.superfeedr.com" rel="hub" />

    <id>urn:docs-rs:0.6.0-8b5fb70-2020-06-22</id>
    <updated>2020-06-22T21:37:04.673103+00:00</updated>

    <entry>
        <title>dyn-fmt-0.2.0</title>
        <link href="https://docs.rs/dyn-fmt/0.2.0/dyn_fmt/" />
        <id>urn:docs-rs:dyn-fmt:0.2.0</id>
        <updated>2020-06-22T21:37:04.673103+00:00</updated>
        <summary>Provides dynamic string format.</summary>
        <author>
            <name>docs.rs</name>
        </author>
    </entry>

</feed>
```

### After

*Pre-formatted*

```xml
<?xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
    <title>Docs.rs</title>
    <subtitle>Recent Rust crates</subtitle>

    <link href="https://docs.rs/releases/feed" rel="self" />
    <link href="https://docs.rs/" />
    <link href="https://pubsubhubbub.appspot.com" rel="hub" />
    <link href="https://pubsubhubbub.superfeedr.com" rel="hub" />

[...]

PR #865, opened 2020-06-29
Removed all usages of lazy_static and replaced them with once_cell

Kixiron:
* Replaced `lazy_static` with `once_cell` (using the `parking_lot` feature for smaller, faster and non-poisoning locks)
* ~~Tweaked `load_rustc_resource_suffix`, it now substitutes a dummy rustc version for tests when one isn't found. This fixed a lot of intermittent (and essentially false) test failures for me, I can remove it if it's uneeded~~

### DIFF
--- a/src/web/metrics.rs
+++ b/src/web/metrics.rs
@@ -3,6 +3,7 @@ use crate::BuildQueue;
 use iron::headers::ContentType;
 use iron::prelude::*;
 use iron::status::Status;
+use once_cell::sync::Lazy;
 use prometheus::{
     opts, register_counter, register_int_counter, register_int_gauge, Encoder, IntCounter,
     IntGauge, TextEncoder, __register_gauge, register_int_counter_vec, IntCounterVec,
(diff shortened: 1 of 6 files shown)


pietroalbini:
> This fixed a lot of intermittent (and essentially false) test failures for me, I can remove it if it's uneeded

Those are not test failures: when that function fails the error message is logged with `log::error!()`, and the execution proceeds as normal. If you disable output capturing you'll see that message in every test using the frontend. It's just that the test is having another error, so you also see the log.

Kixiron:
It was causing a *lot* of errors for me (as opposed to being silent), 29 of them to be exact. Without the patch I made I get lots of this, whereas after all tests pass
```
---- web::rustdoc::test::test_no_trailing_slash stdout ----
[2020-06-29T19:14:51Z ERROR cratesfyi::web::page::templates] Failed to load rustc resource suffix: ErrorMessage { msg: "missing rustc version" }
thread 'web::rustdoc::test::test_no_trailing_slash' panicked at '/crate/dummy/0.1.0/target-redirect/x86_64-apple-darwin: expected redirect to /dummy/0.1.0/dummy/, got 500 Internal Server Error', src\test\mod.rs:79:13

---- web::source::tests::cargo_ok_not_skipped stdout ----
[2020-06-29T19:14:51Z ERROR cratesfyi::web::page::templates] Failed to load rustc resource suffix: ErrorMessage { msg: "missing rustc version" }
thread 'web::source::tests::cargo_ok_not_skipped' panicked at 'failed to GET /crate/fake/0.1.0/source/: 500 Internal Server Error', src\test\mod.rs:44:5

---- web::rustdoc::test::yanked_release_shows_warning_in_nav stdout ----
[2020-06-29T19:14:51Z ERROR cratesfyi::web::page::templates] Failed to load rustc resource suffix: ErrorMessage { msg: "missing rustc version" }
thread 'web::rustdoc::test::yanked_release_shows_warning_in_nav' panicked at 'failed to GET /dummy/0.1.0/dummy/: 500 Internal Server Error', src\test\mod.rs:44:5

---- web::test::test_index_returns_success stdout ----
[2020-06-29T19:14:55Z ERROR cratesfyi::web::page::templates] Failed to load rustc resource suffix: ErrorMessage { msg: "missing rustc version" }
thread 'web::test::test_index_returns_success' panicked at 'assertion failed: web.get("/").send()?.status().is_success()', src\web\mod.rs:637:13

---- web::test::test_show_clipboard_for_crate_pages stdout ----
[2020-06-29T19:14:55Z ERROR cratesfyi::web::page::templates] Failed to load rustc resource suffix: ErrorMessage { msg: "missing rustc version" }
thread 'web::test::test_show_clipboard_for_crate_pages' panicked at 'assertion failed: clipboard_is_present_for_path("/crate/fake_crate/0.0.1", web)', src\web\mod.rs:653:13
```

Kixiron:
And the docker failure here is due to #864, it's fixed in #849

pietroalbini:
@Kixiron hmm? The tests you mention correctly pass for me on master.

pietroalbini:
Those tests also work on this PR with the `#[cfg(test)]` removed.

Kixiron:
I removed the `cfg`

src/web/metrics.rs
```diff
@@ -2,119 +2,149 @@ use crate::db::Pool;
...
         "The number of currently opened file descriptors"
     )
-    .unwrap();
+    .unwrap()
+});
 
-    pub static ref CURRENTLY_RUNNING_THREADS: IntGauge = register_int_gauge!(
+#[cfg(not(windows))]
+pub static CURRENTLY_RUNNING_THREADS: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!(
```
jyn514, line 142: Why was this cfg added?

src/web/metrics.rs
```diff
@@ -257,9 +288,8 @@ mod tests {
     };
 
     static ROUTES_VISITED: AtomicUsize = AtomicUsize::new(0);
-    lazy_static::lazy_static! {
-        static ref RESPONSE_TIMES: Mutex<HashMap<String, usize>> = Mutex::new(HashMap::new());
-    }
+    static RESPONSE_TIMES: Lazy<Mutex<HashMap<String, usize>>> =
+        Lazy::new(|| Mutex::new(HashMap::new()));
```
jyn514, line 292: Seems a shame to put a mutex inside a mutex ... don't we have to worry about deadlocks for something like that?

jyn514: (This shouldn't block the PR though since it was a pre-existing issue)

Kixiron: These can't deadlock since the writes don't depend on each other

src/web/metrics.rs
Kixiron: Whoops, thought I answered this yesterday! The cfg was added because after the generated code from `lazy_static` was removed it was revealed that on windows those two metrics aren't used (and therefore threw warnings), which makes sense since the driving code uses `libc`. If you want I can add (in a separate PR) the # of threads for windows, but since file descriptors don't exist on windows I can't add that metric.

src/web/metrics.rs
jyn514: Ah right, because Lazy only uses a mutex before it's been initialized, and the inner mutex doesn't exist until initialization finished. Got it.
```suggestion
        // can't deadlock: Lazy only uses a mutex before it's been initialized, and the inner mutex doesn't exist until initialization finished
        Lazy::new(|| Mutex::new(HashMap::new()));
```

src/web/metrics.rs
jyn514: Hmm I see. Since we don't run windows in prod I'm ok with not having this metric.

src/web/metrics.rs
Kixiron: I’m a little confused as to why a comment is needed since the two locks don’t interact with each other . `Lazy` has the semantics of `lazy_static` so anyone familiar with either of the libs wouldn’t even second guess this since it’s a super common use case (with mutexes not being const constructible)

PR #870, opened 2020-07-01
Update badge

jyn514:
- Remove outdated `stb_truetype` crate
- Parse the font once instead of on every request (using once_cell)

Closes https://github.com/rust-lang/docs.rs/issues/869

r? @Kixiron


jyn514:
Example badge

src/web/badge/src/lib.rs
```diff
@@ -34,12 +35,11 @@ pub struct Badge {
...
-
-        // this should never fail in practice
-        let font = collection
-            .into_font()
-            .map_err(|_| "Failed to load font data".to_owned())?;
+        static FONT: OnceCell<Font> = OnceCell::new();
+        // Font is an `Arc` and therefore cheap to clone
+        let font = FONT.get_or_init(|| {
+            Font::try_from_bytes(FONT_DATA).expect("Failed to parse FONT_DATA")
+        }).clone();
```
Kixiron, line 42: I'd rather use a `Lazy`
```suggestion
static FONT: Lazy<Font> = Lazy::new(|| {
    Font::try_from_bytes(FONT_DATA).expect("Failed to parse FONT_DATA")
});
let font = FONT.clone();
```
I don't like the clone and kinda feel like that can be removed, but that's a separate issue

jyn514: Ooh, lazy is a great idea.

The clone here is really cheap because all the font variants are `Arc`s: https://docs.rs/rusttype/0.9.2/rusttype/enum.Font.html

PR #786, opened 2020-05-29
[WIP] Migrate to Tera for templating

Kixiron:
This is an attempt to both remove handlebars and add a type-safe interface to rendering templates while simultaneously making the templates more easy to modify and read.
Template reloading is also added via the very sketchy method of a thread that reloads the templates on 10s intervals, better solutions would be appreciated. (I tried to use `notify`, but it didn't work on my system even though it's supposed to)

Note: The templates are messy at times (currently), as I've pasted the navigation header into each and edited it to suit the current page, but after everything's done I'll squash them all together into something better
Note: The diff is massive right now because I moved the old `templates` folder, on the actual release the old templates will be deleted entirely

## Pages

- [x] `/` => `home_page.html`
- [x] `/about` => `about.html`
- [x] `/releases`, `/releases/failures`, `/releases/recent-failures`, `/releases/stars` => `releases.html`
- [x] `/releases/feed` => `release_feed.xml`
- [x] `/releases/queue` => `release_queue.html`
- [x] `/crate/:name` => `crate_details.html`
- [x] `/crate/:name/:version/builds` => `builds.html`
- [x] `/releases/activity` => `release_activity.html`
- [x] `/about/metrics` => Maybe N/A?
- [x] `/releases/:author`
- [x] `/crate/source`
- [x] rustdoc pages

## Possible Extensions

* Graph rendering can be done entirely in Tera as shown in [Drawing SVG Graphs with Rust](https://cetra3.github.io/blog/drawing-svg-graphs-rust/) and could be an alternative to Highcharts.js. This has the advantage of even less JavaScript, which makes it more accessible to users and gives us finer-grained control over our graphs, along with being self-sufficient (not relying on another service for part of our site) and not having the highcharts logo that currently exists there
* Add a footer with relevant info like the other Rust sites do
* Update `highlight.js` to the most recent version

Closes #777 and #740

PR #873, opened 2020-07-01
Don't munge window.location.hash to fix scroll positions

### LINKED ISSUE #764: Going back from source requires three clicks
See https://a.pomf.cat/qyfvmb.webm

antifuchs:
This fixes #764 and #510 by changing rustdoc pages' CSS layout to work together with the `position:fixed` nav bar such that keyboard focus and anchor-ID jumping work.
(Edited a great deal from when I first submitted this)

The technique used here relies on `scroll-padding-top` in the CSS, which is supported everywhere except Safari (neither iOS nor macOS). To fix Safari, we use a JS script to adjust the scroll offset on initial visits to the page and on internal jumping.

With this, the following things work (tested on Firefox, Chrome, mobile&desktop Safari):
* Back/forward navigation only requires one single button press on non-firefox browsers
* links to jump anchors show the whole item in the viewport automatically
* input focus is set correctly on JS sessions, so that keyboard scrolling keeps working.

### DIFF
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -92,6 +92,18 @@ body {
     margin: 0;
     // Since top navbar is fixed, we need to add padding to the body content.
     padding-top: $top-navbar-height;
+    // The scroll padding on the <body> is necessary for Chrome
+    // browsers for now (see
+    // https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/
+    // for an explanation)
+    scroll-padding-top: $top-navbar-height;
+}
+
+html {
+    // Offset anchor jump targets down by this much, so they don't
+    // overlap the top navigation bar (not supported on Desktop/Mobile
+    // Safari yet):
+    scroll-padding-top: $top-navbar-height;
 }
 
 // this is a super nasty override for help dialog in rustdocs
--- a/templates/style.scss
+++ b/templates/style.scss
@@ -84,14 +84,7 @@ div.rustdoc {
         }
     }
 
-    // this is actual fix for docs.rs navigation and rustdoc sidebar
-    position: fixed;
-    top: $top-navbar-height;
-    bottom: 0;
-    left: 0;
-    right: 0;
-    display: block;
-    overflow-y: auto;
+    position: relative;
 }
 
 body {
--- a/templates/rustdoc/page.html
+++ b/templates/rustdoc/page.html
@@ -31,20 +31,34 @@
     </body>
 
     <script>
-        var doc_body = document.getElementById("rustdoc_body_wrapper");
+      // Reset the scroll offset on browsers that don't support
+      // scroll-padding-top (Desktop & Mobile Safari):
+      const maybeFixupViewPortPosition = function() {
         if (window.location.hash) {
-            var notFirefox = typeof InstallTrigger === 'undefined';
-            if (notFirefox) {
-                var hash = window.location.hash;
-                window.location.hash = "";
-                setTimeout(function () {
-                    window.location.hash = hash;
-                    doc_body.focus();
-                }, 1);
-            }
-        } else {
-            doc_body.focus();
+          const anchorElement = document.getElementById(window.location.hash.substr(1));
+          const navBarHeight = document.getElementsByClassName("nav-container-rustdoc")[0].offsetHeight;
+          if (anchorElement &&
+              anchorElement.getBoundingClientRect().top <= navBarHeight &&
+              Math.abs(anchorElement.getBoundingClientRect().top) >= 0) {
+            // It's just overlapped by the nav bar. This can't be a coincidence, scroll it into view:
+            requestAnimationFrame(function() {
+              scrollBy(0, -navBarHeight);
+            });
+          }
         }
+      };
+      window.addEventListener("hashchange", maybeFixupViewPortPosition, false);
+      // Fix up the scroll position if this was a direct visit to the page
+      // (not back/forward navigation):
+      if (window.performance) {
+        const navEntries = window.performance.getEntriesByType('navigation');
+        const usedBack = navEntries.length > 0 && navEntries[0].type === 'back_forward' ||
+              (window.performance.navigation &&
+               window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD);
+        if (!usedBack && window.location.hash) {
+          window.addEventListener("scroll", maybeFixupViewPortPosition, {"once": true});
+        }
+      }
     </script>
 
 </html>


jyn514:
Posting the status so I don't forget: This fixes #764, but breaks scrolling again (see https://github.com/rust-lang/docs.rs/pull/510, etc. for discussion). The proposed fix is to put the focusing back, but leave out the URL munging

antifuchs:
Updated to use `.scrollIntoView()` on all browsers (without the weird firefox detection code). The following work on firefox, chrome and safari for me:

* can scroll with the keyboard immediately upon loading the page
* element on the URL hash is in view whenever the page is visited.

antifuchs:
(sorry for the final commit --amend, I had done a variable rename that didn't carry over... and then tested with the wrong version of the code, ugh)

jyn514:
This doesn't fix scrolling if there is no hash in the URL. I think the fix should be as simple as moving that bit outside the `if`?

jyn514:
What is the setTimeout for, btw? Why not focus immediately?

antifuchs:
OK, so we had a good time on discord figuring out how to solve this for the `position: relative` main content container.

This weird javascript I'd touched earlier was necessary because of the fixed-position navbar; since this causes #595 and the solution in #510 looked much better, we decided to go with @Koxiaet's solution from #510 after all, and try fitting elements into the viewport despite the overlapping top navbar. The solution I pushed here uses javascript on page load (and on anchor jump links) to scroll just the right amount; unfortunately, no pure-CSS solution seemed to work:

* The `:target::before` solution from [this SO answer](https://stackoverflow.com/a/28824157/93405) doesn't function correctly in rustdoc pages because the jump targets are `display: flex` & that messes up the alignment of elements in the `[+] pub fn foo()     [src]` line.
* Using `scroll-padding-top` on the `body` doesn't have any effect, even though it's advertised as being useful for exactly this use case... I don't understand why; plus, it doesn't work in IE at all /:

So I went with the solution from https://github.com/rust-lang/docs.rs/pull/510#issuecomment-563415804, adjusting it so that back/forward buttons don't jump to where a user might have left off.

This solution isn't ideal either:
* Requires JS to be on, otherwise people see the crummy scroll positions
* Does not work in the "source code" view, as it's rustdoc's JS that scrolls to the first marked line... and that scrolls it out of the viewport.
* Generally seems displeasing, being a JS-based solution.

OTOH, My back/forward button work quite well now, and the highlighted elements go in the right places (:

antifuchs:
OK - pushed up yet another massive rework for different/maybe-better tradeoffs:

* Use a flexbox-based navigation system ([see here for an example and the technique](https://www.bitovi.com/blog/use-flexbox-to-create-a-sticky-header-and-sidebar-with-flexible-content))
* Since the scrolling container is now a sub-element and not `<body>`, browsers doesn't retain scroll positions as well; to fix that, save scroll positions in per-tab `sessionStorage` and restore upon visiting the page again (via reload or back/forward buttons). 
* The JS also focuses the "main content" body, so scrolling with the keyboard keeps working.

I tested the following things, with JS turned on, in chrome/safari/firefox & for good measure, on an iOS device:

* scrolling the body with a keyboard works.
* back/forward buttons jump directly to the previous page and scroll position
* reloading the page stays where you were previously
* jumping to a target (on a rustdoc page and on the source code view) has the whole target in the viewport.

templates/rustdoc.hbs
```diff
@@ -9,26 +9,35 @@
...
+ (function() {
+     var keyName = function() {
+         return 'rustdoc_body_scroll_posn_' + window.location.href;
+     }
+     var oldPosn = window.sessionStorage.getItem(keyName());
+     var wrapper = document.getElementById('rustdoc_body_wrapper');
+     var saveState = function() {
+         window.sessionStorage.setItem(keyName(), wrapper.scrollTop);
+     }
+     if(oldPosn) {
```
jyn514, line 32:
```suggestion
     if (oldPosn) {
```

templates/rustdoc.hbs
```diff
@@ -9,26 +9,35 @@
...
+         wrapper.scrollTop = oldPosn;
+     } else {
+         // we have no scroll state saved, set it now:
+         saveState();
+     }
+     wrapper.addEventListener("scroll", saveState);
+     window.addEventListener("hashchange", saveState);
+     wrapper.focus();
+ })()
```
jyn514, line 41: This is in an IIFE so that the variables aren't added to the global scope?

templates/style.scss
```diff
@@ -142,14 +160,11 @@ div.container-rustdoc {
 
 div.nav-container {
     // Nothing is supposed to be over or hovering the top navbar. Maybe add a few others '('? :)
```
jyn514, line 162: I think this comment was for the `z-index`.
```suggestion
```

jyn514:
Also I love that this is mostly CSS and the JavaScript is only for the scroll position :D

templates/rustdoc.hbs
antifuchs: Yup, I was getting worried that the previous change I'd had in mind used too many vars.

templates/style.scss
```diff
@@ -84,21 +84,33 @@ div.rustdoc {
...
 }
 
 body {
     padding: 0;
     margin: 0;
-    // Since top navbar is fixed, we need to add padding to the body content.
-    padding-top: $top-navbar-height;
+    height: 100%;
+    overflow: hidden;
+    margin: 0px;
```
GuillaumeGomez, line 99:
```suggestion
margin: 0;
```

templates/style.scss
```diff
@@ -84,21 +84,33 @@ div.rustdoc {
...
+html {
+    height: 100%;
 }
 
 body {
     padding: 0;
     margin: 0;
-    // Since top navbar is fixed, we need to add padding to the body content.
-    padding-top: $top-navbar-height;
+    height: 100%;
```
GuillaumeGomez, line 97: Normally, you don't need this, can you double-check please?

templates/style.scss
```diff
@@ -84,21 +84,33 @@ div.rustdoc {
...
+    overflow: hidden;
```
GuillaumeGomez, line 98: This one is making me really uncomfortable. What is it used for?

templates/style.scss
```diff
@@ -84,21 +84,33 @@ div.rustdoc {
...
+    margin: 0px;
+    display: flex;
+    box-sizing: border-box;
+}
+
+#page-container {
+    height: 100%;
+    width: 100%;
+    display: flex;
+    flex-direction: column;
```
GuillaumeGomez, line 108: It'll reduce a bit the current amount of devices able to use docs.rs, but not too much. So I don't know. What do you think @jyn514 ?

templates/rustdoc.hbs
```diff
@@ -9,26 +9,35 @@
...
-        }
-    } else {
-        doc_body.focus();
-    }
+ // restore scroll state across back/forward navigation & give focus
+ // to the scrolling container:
+ (function() {
+     var keyName = function() {
+         return 'rustdoc_body_scroll_posn_' + window.location.href;
+     }
```
GuillaumeGomez, line 26: You have the right to add ';'. :p

templates/rustdoc.hbs
```diff
@@ -9,26 +9,35 @@
...
-                doc_body.focus();
-            }, 1);
-        }
-    } else {
-        doc_body.focus();
-    }
+ // restore scroll state across back/forward navigation & give focus
+ // to the scrolling container:
+ (function() {
+     var keyName = function() {
```
GuillaumeGomez, line 24: Why not storing this information once and for all?

templates/rustdoc.hbs
```diff
@@ -9,26 +9,35 @@
...
+         return 'rustdoc_body_scroll_posn_' + window.location.href;
+     }
+     var oldPosn = window.sessionStorage.getItem(keyName());
+     var wrapper = document.getElementById('rustdoc_body_wrapper');
+     var saveState = function() {
+         window.sessionStorage.setItem(keyName(), wrapper.scrollTop);
+     }
```
GuillaumeGomez, line 31: Colon. :p

templates/style.scss
jyn514: https://caniuse.com/#search=flex-direction shows that it isn't supported on Firefox, but I didn't notice anything wrong when I visited the site. @antifuchs what does this do? Can you maybe show screenshots before and after?

antifuchs: The `flex-direction: column` is supposed to arrange child elements vertically below each other. I guess even if it's not supported, there is nowhere for the elements to flow to, since we're 100%ing them all.

GuillaumeGomez: Then is it useful to keep it?

templates/style.scss
antifuchs: This is the entire magic of the change: The "whole page" is divided up like this:

```
┌─ body ───────────────────────────────────────────────────────────────┐
│┌───────────────────────────────────────────────────────────────────┐ │
││                           nav bar                                 │ │
│├───────────────────────────────────────────────────────────────────┤ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                     actual page body with overflow-y: auto        │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
││                                                                   │ │
└┴───────────────────────────────────────────────────────────────────┴─┘
```

* The body does not scroll: Everything inside it remains in place
* The nav bar sits in the top portion of the body, taking up a fixed amount of space, and nothing is "under" (in the z direction) it.
* The main page body is the second, huge, portion of the viewport, and that has `overflow-y: auto` so that it scrolls.

This way, the "meat" of the page remains scrollable, and the nav bar remains where it should be.

templates/style.scss
antifuchs: maybe - I'll try removing it.

antifuchs: Yep, turns out it's necessary: Webkit at least reflows elements to sit next to each other.
screenshot showing the navbar on the left and the page body on the right

templates/rustdoc.hbs
antifuchs: Not sure what you mean by "once and for all" - the key thing that I want to achieve here is that people can navigate browser tabs back and forth and they get scrolled back to the place that they were when they viewed that page in that tab.

`sessionStorage` is tab-local; that means if you store a scroll position under one key for all pages, that means if you click on a link that takes you to, e.g. a source view in the same tab, this script would scroll you to the position that you were scrolled to on the previous page - not where you meant to go.

So, by keying this on the address that people are on, we restore their scroll position at exactly that address.

This actually reminds me of a thing that *could* occur: You could be on the same location.href in different places in your history & this script would restore your scroll location to the _last_ place you'd visited; this isn't ideal. Maybe I can add the history "depth" number to the storage key - that way, it's one position per page/tab/navigation history item position.

templates/style.scss
antifuchs: Without it, the viewport turns non-scrollable. /:

templates/style.scss
jyn514: Ok, if it's not a compatibility issue I don't see a problem. I'm still confused by the caniuse page but since it clearly works in my Firefox browser that page must have bad info.

antifuchs: Agreed - I don't find supporting evidence on the MDN page that any `about:config` settings would be necessary, as claimed on the caniuse page; maybe the problem was referring to `-webkit-flex-direction`?

GuillaumeGomez: My worry is mostly that most flex-based CSS properties were supported starting 2013-2014. So not that old (well, actually it's getting old now XD).

templates/rustdoc.hbs
GuillaumeGomez: No I mean: instead of making a function, why not just store what you do in the function in a variable?

antifuchs: Ah, sorry, I misunderstood: That's so that the `hashchange` event for intra-page navigation saves the scroll position under the right key; If I just close over the variable, it would save under the old location.href, but we want to save the new location.href. (:

GuillaumeGomez: Oh right, my bad! It's used in `savedState`. Sorry, didn't pay attention to that... Then it's solved! :)

templates/style.scss
antifuchs: Hah, I feel that worry - it definitely seems weirdly futuristic, even though it's 6 years old now & decently well-supported. From the [rust RFC on browser support](https://rust-lang.github.io/rfcs/1985-tiered-browser-support.html#supported-browsers), it does fit the support matrix (at least one firefox, at least one IE, two chromes, one of each for others); I guess this should make it fine?

GuillaumeGomez: This is what we need to agree on. I'll let that decision up to the other members of the @rust-lang/docs-rs team decide. :p

jyn514: If it works on Android, Safari, and iOS, I'm all in favor.

templates/rustdoc.hbs
jyn514: > You could be on the same location.href in different places in your history & this script would restore your scroll location to the last place you'd visited; this isn't ideal. Maybe I can add the history "depth" number to the storage key - that way, it's one position per page/tab/navigation history item position.

This is a nice improvement but it's certainly not necessary to merge this IMO.

templates/style.scss
antifuchs: I couldn't test on Android (don't have an Android device here), but once a staging server works, it should hopefully be easy to enlist help from the community (:

jyn514:
I tested on android
- Clicking on a hash link to a function keeps full function in view and the sidebar above the function
https://cdn.discordapp.com/attachments/541978667522195476/728376728669519972/Screenshot_20200702-182835_Ecosia.jpg
- Back button keeps the scroll position (if there's a forward button on mobile, I don't know about it)
- Reloading the page keeps the scroll position.

jyn514:
@antifuchs unfortunately this caused the about page to look very strange

image

jyn514:
I have a staging server with these changes running at http://50.116.47.176:3000/.  @Koxiaet , @cormacrelf, @imjasonmiller, @Nemo157, @rudedogg could you check if this fixes the issues you were having with slow scrolling and keyboard shortcuts?

This exact PR won't be merged (it needs changes once we finally finish up #786) but it will end up looking very similar to what I have on the staging server.

imjasonmiller:
@jyn514 it seems to be fixed for me on the main page, but not on the actual documentation pages themselves. The sidebar with `Methods`, `Trait implementations` and so on also doesn't appear to support smooth scrolling yet, but I think that might be fixed with [`-webkit-overflow-scrolling: touch`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling) on the scrolling container?

I could also send you screen recording, if that's okay?

jyn514:
A screen recording sounds great!

> it seems to be fixed for me on the main page, but not on the actual documentation pages themselves

That's a shame ... by main page you mean the /crate URLs? Those are using a different stylesheet than the rustdoc pages which might be why.

I'm really confused now why this doesn't work, the header is no longer `position: fixed` at all ...

antifuchs:
> I'm really confused now why this doesn't work, the header is no longer position: fixed at all ...

I'm guessing the way that the flex box top navbar fixation works causes behavior similar to the way a `position: fixed` container would? I'm not sure why only the left bar would be affected for you, though.

@imjasonmiller - can you tell me what exact configuration you're running (what browser, OS version, device type)? I'd love a  screen recording, too.

imjasonmiller:
> That's a shame ... by main page you mean the /crate URLs? Those are using a different stylesheet than the rustdoc pages which might be why.

@jyn514, I think this might be an error on my part. The actual documentation page (and sidebar) for a crate itself (e.g. http://50.116.47.176:3000/bat/0.15.4/bat/) seems to not support momentum scrolling, but the other pages seem to work fine. Are those pages using the separate style sheet you?

> ... can you tell me what exact configuration you're running (what browser, OS version, device type)? I'd love a screen recording, too.

@antifuchs, if the other tagged users aren't experiencing any issues, it might be considered legacy? It's an old iPhone 5S on iOS Safari and I've switched to Android quite some time ago.

It's currently past midnight, so I'll get to the screen recording this weekend, if you'd still like to?

antifuchs:
@imjasonmiller so the 5S is running iOS 12, I suppose? My 13.5 iPhone 11 does do momentum scrolling - I wonder if that's why we see different things. Either way, if you could get a screen recording, that'd be marvelous! (No rush, this PR won't be "ripe" until the template refactor is done) (:

Koxiaet:
It still doesn't work for me, but it's probably more the the fault of my browser which is unable to scroll anything other than the body itself.

GuillaumeGomez:
I don't think setting the body like that is a good idea. :-/

antifuchs:
> I don't think setting the body like that is a good idea. :-/

Well - I don't think solutions that please everyone exist. That's all due to the fact that a fixed-to-the-top navigation bar is a pain in general, and specifically with rustdoc pages:

* There's no reliable way to keep the bar from overlapping the elements on the scrolling page - workarounds exist, [see my investigation of the mess above](https://github.com/rust-lang/docs.rs/pull/873#issuecomment-652781292), but they do not work for rustdoc pages due to the flex layout that's going on in the headlines it generates.

* The page structure is always going to be somewhat wonky because rustdoc thinks it gets the whole page (and two separately-scrolling areas in that page) when docs.rs is only willing to give it a portion, claiming the root of the coordinate system for itself.

I'm myself quite unhappy with this navigation bar thing by now. How are people's thoughts about a ... *bottom* navigation bar?

antifuchs:
> Well - I don't think solutions that please everyone exist.

Let me eat my words, 30 minutes after posting this :D :D :D

Turns out there might be a solution that pleases everyone!! [`scroll-padding-top`](https://caniuse.com/#search=scroll-padding-top) is a CSS attribute you can attach to the `html` (and in chrome, must also attach to the `body`), which tells the browser where to move the viewport.

I'd been trying this before and discarded it because I hadn't attached the attribute to the right elements (been trying it with the headers that get jumped to). [This article](https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/) finally set me straight.

The solution works (as in keyboard scrolling, jump-target positioning, back/forward DTRT) in Chrome and Firefox, requires zero javascript and no page structure modification. I think we have a winner. 

Will commit what I have and push it up asap.

GuillaumeGomez:
Nice finding! If this is the miracle solution we've been looking for so long, I'll be really happy to finally close this chapter! :D

antifuchs:
Welp. Testing this in some more browsers, it looks like mobile and desktop Safari does not support the attribute for anchor elements: https://bugs.webkit.org/show_bug.cgi?id=179379

I guess we'll have to put a little bit of javascript back to fix safari positioning support, but it's better-enough of a solution that it'd probably be worth it.

antifuchs:
I rebased away the (confusing) flex layout changes) & pushed the scroll-padding-top based solution up now. The JS does look a little intimidating, but I hope it's not too bad. The only place where it really has any effect is in Safari.

templates/rustdoc.hbs
```diff
@@ -16,19 +16,37 @@
...
+        usedBack = true;
+      } else if(window.performance.navigation
+                && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD) {
+        usedBack = true;
+      }
+      if(!usedBack && window.location.hash) {
+        window.addEventListener("scroll", maybeFixupViewPortPosition, {"once": true});
+      }
     }
+  })();
```
Koxiaet, line 50: If we're using `const` is there any reason we aren't using `let`? Then we can drop the IIFE because we aren't polluting the global scope.

templates/rustdoc.hbs
```diff
@@ -16,19 +16,37 @@
...
+    window.addEventListener("hashchange", maybeFixupViewPortPosition, false);
+    var usedBack = false;
+    if(window.performance) {
+      var navEntries = window.performance.getEntriesByType('navigation');
+      if(navEntries.length > 0 && navEntries[0].type === 'back_forward') {
+        usedBack = true;
+      } else if(window.performance.navigation
+                && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD) {
+        usedBack = true;
+      }
```
Koxiaet, line 45: I refactored it to removed the unnecessary `if` and make its purpose slightly clearer. I'm not sure about the indentation, I don't know what JS conventions are.

```suggestion
    if(window.performance) {
      var navEntries = window.performance.getEntriesByType('navigation');
      var usedBack = navEntries.length > 0 && navEntries[0].type === 'back_forward'
        || (window.performance.navigation
            && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD);
```

templates/rustdoc.hbs
antifuchs: Ah, that would work. The other idea would be to check if we can just define this behavior in a function in a .js file that gets pulled in - menu.js maybe - and just call it from a top-level script tag.

antifuchs:
Thanks for the comments, @Koxiaet - that definitely makes the code easier to read; I've converted all the assignments to `const` (same scoping rules as `let`, signals they're convenience value holders) and shortened the expression as you suggested.

templates/rustdoc.hbs
```diff
@@ -16,19 +16,33 @@
...
+        requestAnimationFrame(function() {
+          scrollBy(0, -navBarHeight);
+        });
+      }
     }
+  };
+  window.addEventListener("hashchange", maybeFixupViewPortPosition, false);
+  // Fix up the scroll position if this was a direct visit to the page
+  // (not back/forward navigation):
+  if(window.performance) {
```
jyn514, line 38: What does looking at `window.performance` do?

nit:
```suggestion
  if (window.performance) {
```

templates/rustdoc.hbs
```diff
@@ -16,19 +16,33 @@
...
+    const navEntries = window.performance.getEntriesByType('navigation');
+    const usedBack = navEntries.length > 0 && navEntries[0].type === 'back_forward' ||
+        (window.performance.navigation &&
+         window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD);
+    if(!usedBack && window.location.hash) {
```
jyn514, line 43: nit:
```suggestion
    if (!usedBack && window.location.hash) {
```

templates/rustdoc.hbs
antifuchs: > What does looking at `window.performance` do?

This is for detecting whether the page was visited via the back/forward buttons: we should avoid scrolling to the anchor in that case, as people will expect to be returned to the place where they left the page in history.

Koxiaet:
> I redeployed the changes to http://50.116.47.176:3000/. This is really exciting!

I don't think you have. I checked it and it's still the old code, and my cache is off.

jyn514:
Oops, fixed. Good catch!

Koxiaet:
At least for me, it works perfectly!

antifuchs:
I asked my colleagues to try the staging server on Android devices, and they confirmed that it seems to work there in both Chrome and Firefox \o/

jyn514:
Ok great! Then I think this is only waiting on https://github.com/rust-lang/docs.rs/pull/879, then a rebase and we'll be good to go :) I'll see if I can review Tera this afternoon.

Kixiron:
I’ll have time later today/tonight to work on it, if you’re extra-free it’d help *tons* to investigate the "funky JS" issues on the templates I made but if not then no sweat

antifuchs:
OK, I rebased this onto master & ported the scroll-position fixing script to `rustdoc/page.html`. This should be the only place it's really needed for Safaris, but lmk and we can move it to another page template too.

I tried the change on Safari (mobile and desktop), and it works as it did on the old master branch.

PR #877, opened 2020-07-03
Use procfs only on Linux builds

antifuchs:
As we discussed on discord a few days ago, this fixes running the web service on macOS and freebsd machines as `procfs` is only available on Linux targets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ time = "0.1" # TODO: Remove once `iron` is removed
 version = "0.15"
 features = ["with-chrono", "with-serde_json"]
 
-[target.'cfg(linux)'.dependencies]
+[target.'cfg(target_os = "linux")'.dependencies]
 # Process information
 procfs = "0.7"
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -74,7 +74,7 @@ time = "0.1" # TODO: Remove once `iron` is removed
 version = "0.15"
 features = ["with-chrono", "with-serde_json"]
 
-[target.'cfg(not(windows))'.dependencies]
+[target.'cfg(linux)'.dependencies]
```
jyn514, line 77: Currently this is never true.
```suggestion
[target.'cfg(target_os = "linux")'.dependencies]
```

src/web/metrics.rs
```diff
@@ -128,7 +128,7 @@ pub fn metrics_handler(req: &mut Request) -> IronResult<Response> {
     PRIORITIZED_CRATES_COUNT.set(ctry!(queue.prioritized_count()) as i64);
     FAILED_CRATES_COUNT.set(ctry!(queue.failed_count()) as i64);
 
-    #[cfg(not(windows))]
+    #[cfg(linux)]
```
jyn514, line 131:
```suggestion
    #[cfg(target_os = "linux")]
```

Issue #740, opened 2020-05-01
Switch templating to Tera


Kixiron:
The current templates have lots of redundancy, breaking things into smaller units will help with consistency across pages and maintainability

Kixiron:
This has now changed slightly, I've been investigating the [tera](https://github.com/Keats/tera) templating engine as a suitable replacement for handlebars

jyn514:
Motivation for `tera`, as discussed on discord:
- hot reloading so that modifying templates isn't as painful
- inheritance in templates
- more processing done in templates instead of code so we can use more `derive(Serialize)`
- better docs

Kixiron:
All of that culminates into what at the end of the day is better, more modifiable and more maintainable code, which is a bonus on every front

pietroalbini:
I'd also be in favor of moving to tera.

jyn514:
Status: waiting on the bugs mentioned in https://github.com/rust-lang/docs.rs/pull/885 to be fixed. Then we can go another round at https://github.com/rust-lang/docs.rs/pull/879.

jyn514:
Implemented in https://github.com/rust-lang/docs.rs/pull/887

PR #879, opened 2020-07-04
The rest of Tera

### LINKED ISSUE #777: No link to builds tab on crate page
E.g. <https://docs.rs/crate/escargot/0.5.0> has no link to <https://docs.rs/crate/escargot/0.5.0/builds>, despite the latter having a "Builds" tab highlighted next to "Source".

Kixiron:
This PR is split into commits, that's by far the best way to review it

<details>
<summary>Commit 1: Builds</summary>

## Before

image

## After

image

</details>

<details>
<summary>Commit 2: Errors</summary>

This commit looks larger than it actually is since I had to update the call site of every `ctry!` and `cexpect!` invocation

## Before

image

## After

Error message is now the page description

image

</details>

<details>
<summary>Commit 3: Crate Details</summary>

## Before

image

## After

The builds page is now always displayed

image

</details>

<details>
<summary>Commit 4: Source</summary>

## Before

image

## After

Now there's a few more icons used

image

</details>

<details>
<summary>Commit 5: Rustdoc</summary>

## Before

image

## After

image

</details>

<details>
<summary>Commit 6: Release Activity</summary>

## Before

image

## After

image

</details>

<details>
<summary>Commit 7: Build Queue</summary>

## Before

image

## After

image

</details>

cc @antifuchs
Closes #777, #740 and #786


tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                                <a href="/releases/{{ author[1] }}" class="pure-menu-link">
+                                    {{ author[0] }}
+                                </a>
+                            </li>
+                        {%- endfor -%}
+
+                        <li class="pure-menu-heading">Links</li>
+
+                        {# If the crate has a homepage, show it #}
+                        {%- if details.homepage_url -%}
```
jyn514: This doesn't seem to handle the case when `details` is `None`. I think the best way to do that is by returning a 404 in `crate_details_handler` if the crate wasn't found.

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                                    <i class="fa fa-fw fa-file-text"></i> Documentation
+                                </a>
+                            </li>
+                        {%- endif -%}
+
+                        {# If the release has a repository, show it #}
+                        {%- if details.repository_url -%}
+                            <li class="pure-menu-item">
+                                <a href="{{ details.repository_url }}" class="pure-menu-link">
+                                    {# If the repo link is for github, show some github stats #}
```
jyn514:
```suggestion
                                    {# If the repo link is for github, show some github stats #}
                                    {# TODO: add support for hosts besides github (#35) #}
```

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                                        <i class="fa fa-code-fork fa-fw"></i> Repository
+                                    {%- endif -%}
+                                </a>
+                            </li>
+                        {%- endif -%}
+
+                        {# Show a link to the crate's Crates.io page #}
+                        <li class="pure-menu-item">
+                            <a href="https://crates.io/crates/{{ details.name }}" class="pure-menu-link"
+                                title="See {{ details.name }} in crates.io">
```
jyn514:
```suggestion
                                title="See {{ details.name }} on crates.io">
```

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                        <li class="pure-menu-heading">Dependencies</li>
+                        <li class="pure-menu-item">
+                            <div class="pure-menu pure-menu-scrollable sub-menu">
+                                <ul class="pure-menu-list">
+                                    {# List all dependencies that the current release has #}
+                                    {%- for dep in details.dependencies -%}
+                                        <li class="pure-menu-item">
+                                            <a href="/crate/{{ dep[0] }}/{{ dep[1] }}" class="pure-menu-link">
+                                                {{ dep[0] }} {{ dep[1] }}
+                                                <i class="dependencies {{ dep[2] }}">{{ dep[2] }}</i>
```
jyn514: I assume this was already present, but why do we add the dependency as a class? That seems odd.

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                    {# If there are no docs display a warning #}
+                    {%- if not details.rustdoc_status -%}
+                        <div class="warning">{{ details.name }}-{{ details.version }} doesn't have any documentation.</div>
+                    {%- endif -%}
+
+                    {# If there's a readme, display it #}
+                    {%- if details.readme -%}
+                        {{ details.readme | safe }}
+
+                    {# If there's not a readme then attempt to display docs #}
```
jyn514: We should probably rename the `rustdoc` variable, it's confusing. I noticed something was off because the docs are HTML, and so shouldn't have `| safe` applied.
```suggestion
                    {# If there's not a readme then display the long_description #}
```

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                    </div>
+
+                {# If the release has been yanked and is a library #}
+                {%- elif details.yanked -%}
+                    <div class="warning">
+                        {{ details.name }}-{{ details.version }} has been yanked.
+                    </div>
+
+                {# If the build succeeded, isn't yanked and is a library #}
+                {%- elif details.build_status -%}
```
jyn514: Why is this an else? If it failed to build and also was yanked, don't we want to show both warnings?

tera-templates/crate/details.html
```diff
@@ -0,0 +1,303 @@
...
+                return;
+            }
+
+            switch (getKey(ev)) {
+                case "s":
+                case "S":
+                    ev.preventDefault();
+                    document.getElementById("search").focus();
+                    break;
+            }
```
jyn514: Was all this here already? This is a lot of JavaScript, it looks like rustdoc code.

src/web/source.rs
```diff
@@ -1,69 +1,34 @@
...
-/// A source file
-#[derive(PartialEq, PartialOrd)]
+/// A source file's name and mime type
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Serialize)]
 struct File {
+    /// The name of the file
     name: String,
-    file_type: FileType,
+    /// The mime type of the file
+    mime: String,
```
jyn514, line 22: Why not store it as a FileType? I like having the type safety.

src/web/source.rs
```diff
@@ -126,19 +91,15 @@ impl FileList {
...
-                        } else if mime.starts_with("text") && path_splited[0].ends_with(".rs") {
-                            FileType::RustSource
-                        } else if mime.starts_with("text") {
-                            FileType::Text
+                        let mime = if path_splited.len() > 1 {
+                            "dir".to_owned()
                         } else {
-                            FileType::Binary
+                            mime.to_owned()
                         };
```
jyn514, line 98: Why did this change? IIRC we have some custom logic later that does different things for rust files than for anything else.

src/web/rustdoc.rs
```diff
@@ -215,6 +183,22 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
...
+    rustdoc_head: String,
+    rustdoc_body: String,
+    rustdoc_body_class: String,
+    krate: CrateDetails,
+}
+
+impl_webpage! {
+    RustdocPage = "rustdoc/page.html",
+}
+
```
jyn514, line 201: Can you move this above `RustLangRedirector` where it was before? It would make the diff smaller (and also there's no reason to change it).

tera-templates/rustdoc/navigation.html
```diff
@@ -0,0 +1,248 @@
...
+                                        {%- for author in krate.authors -%}
+                                            <li class="pure-menu-item">
+                                                <a href="/releases/{{ author[1] }}" class="pure-menu-link">
+                                                    <i class="fa fa-fw fa-user"></i> {{ author[0] }}
+                                                </a>
+                                            </li>
+                                        {%- endfor -%}
+                                    </ul>
+                                </div>
+                            </div>
```
jyn514: This has shown up twice now (I think the previous time was on rustdoc pages). Can we split it into a separate template?

tera-templates/rustdoc/navigation.html
```diff
@@ -0,0 +1,248 @@
...
+                                                        </li>
+                                                    {%- endfor -%}
+                                                </ul>
+                                            </div>
+                                        </li>
+                                    </ul>
+                                </div>
+                            </div>
+                        </div>
+                    </li>
```
jyn514: Ditto on the repeated template, it would be great to factor it out.

tera-templates/rustdoc/navigation.html
```diff
@@ -0,0 +1,248 @@
...
+
+                    {# Display the platforms that the release has been built for #}
+                    <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
+                        <a href="#" class="pure-menu-link" aria-label="Platform">
+                            <i class="fa fa-fw fa-gears"></i>
+                            <span class="title"> Platform</span>
+                        </a>
+
+                        {# Build the dropdown list showing avaliable targets #}
```
jyn514:
```suggestion
                        {# Build the dropdown list showing available targets #}
```

tera-templates/rustdoc/page.html
```diff
@@ -0,0 +1,56 @@
+{%- import "macros.html" as macros -%}
+
+<!DOCTYPE html>
+<html lang="en">
+
+    <head>
```
jyn514: Why does this use a new base instead of reusing `base.html`?

src/web/releases.rs
```diff
@@ -692,25 +688,30 @@ pub fn activity_handler(req: &mut Request) -> IronResult<Response> {
     .into_response(req)
 }
 
-pub fn build_queue_handler(req: &mut Request) -> IronResult<Response> {
-    let queue = extension!(req, BuildQueue);
+#[derive(Debug, Clone, PartialEq, Serialize)]
+struct BuildQueuePage {
+    description: Cow<'static, str>,
```
jyn514, line 693: Why not an `&'static str`? You only use this with string literals.

tera-templates/releases/build_queue.html
```diff
@@ -0,0 +1,126 @@
...
+{%- endblock header -%}
+
+{%- block body -%}
+    <div class="container">
+        <div class="recent-releases-container">
+
+            <div class="release">
+                {% set queue_length = queue | length -%}
+                {%- if queue_length == 0  -%}
+                    <strong>There is nothing in queue</strong>
```
jyn514:
```suggestion
                    <strong>There is nothing in the queue</strong>
```

tera-templates/releases/build_queue.html
```diff
@@ -0,0 +1,126 @@
...
+                return;
+            }
+            switch (getKey(ev)) {
+                case "s":
+                case "S":
+                    ev.preventDefault();
+                    document.getElementById("search").focus();
+                    break;
+            }
+        }
```
jyn514: This looks copy/pasted. Please refactor into a separate template (but see also the comment on the first time this showed up).

src/web/mod.rs
```diff
@@ -864,11 +847,6 @@ mod test {
         });
     }
 
-    #[test]
-    fn test_templates_are_valid() {
-        handlebars_engine().expect("Failed to load handlebar templates");
-    }
-
```
jyn514, line 828: Please add an equivalent of this test for tera.

src/web/mod.rs
```diff
@@ -606,6 +624,31 @@ impl MetaData {
...
+
+impl Default for ErrorPage {
+    fn default() -> Self {
+        Self {
+            title: Cow::Borrowed(""),
+            message: None,
+            status: Status::NotFound,
+        }
+    }
+}
```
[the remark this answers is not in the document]
Kixiron: It was at one point, but I forgot to remove it. Good catch!

src/web/rustdoc.rs
Kixiron: I put it here because it's right above `rustdoc_html_server_handler`, which is where it's used. I did this throughout the PR because I think it makes better sense and has better browse-locality to put the struct next to where it's used

src/web/source.rs
Kixiron: I left a more in-depth analysis of why on the previous comment, it's basically because the types weren't actually used any and it was only serving as an artificially limited UI piece

tera-templates/crate/details.html
Kixiron: I sanity-checked myself and it [was present](https://github.com/rust-lang/docs.rs/blob/master/templates/crate_details.hbs#L34), but I have no idea what it does or is or means

tera-templates/crate/details.html
Kixiron: `| safe` is required since Tera escapes HTML by default. You get a big ugly mess without the escaping

src/web/mod.rs
Kixiron: There already [is one](https://github.com/rust-lang/docs.rs/blob/master/src/web/page/templates.rs#L257-L265)

tera-templates/crate/details.html
Kixiron: Tera interprets `None`/`null` as false

tera-templates/crate/details.html
Kixiron: This is an exact recreation of the previous behavior, I don't know if multiple `warning`s will conflict in some way (CSS wise)

tera-templates/crate/details.html
Kixiron: This was part of where crate pages and rustdoc pages melded together, I honestly couldn't say

tera-templates/rustdoc/navigation.html
Kixiron: I'm not sure where a previous one is, this is the only instance I can find (the rustdoc navigation bar)

tera-templates/rustdoc/navigation.html
Kixiron: I think this is the same template as the last comment?

tera-templates/rustdoc/page.html
Kixiron: `base.html` injects a lot of extra style stuff for rustdoc pages, this was the simplest solution

tera-templates/crate/details.html
jyn514: Oh I see. Then this shouldn't have safe applied because long_description is not HTML.

Kixiron:
image

templates/crate/details.html
```diff
@@ -0,0 +1,304 @@
...
+                                <a href="/releases/{{ author[1] }}" class="pure-menu-link">
+                                    {{ author[0] }}
+                                </a>
+                            </li>
+                        {%- endfor -%}
+
+                        <li class="pure-menu-heading">Links</li>
+
+                        {# If the crate has a homepage, show it #}
+                        {%- if details.homepage_url -%}
```
jyn514, line 32: Can you give a screenshot of what the `details` page looks like when details is `None`?

templates/rustdoc/navigation.html
```diff
@@ -0,0 +1,248 @@
...
+                                        {%- for author in krate.authors -%}
+                                            <li class="pure-menu-item">
+                                                <a href="/releases/{{ author[1] }}" class="pure-menu-link">
+                                                    <i class="fa fa-fw fa-user"></i> {{ author[0] }}
+                                                </a>
+                                            </li>
+                                        {%- endfor -%}
+                                    </ul>
+                                </div>
+                            </div>
```
jyn514, line 110: > > This has shown up twice now (I think the previous time was on rustdoc pages). Can we split it into a separate template?

> I'm not sure where a previous one is, this is the only instance I can find (the rustdoc navigation bar)

The other place is in `crate/details.html`: https://github.com/rust-lang/docs.rs/pull/879/files#diff-2550c2ab5e8909adfed2ffb04e746370R19

templates/crate/details.html
```diff
@@ -0,0 +1,304 @@
...
+            }
+
+            switch (getKey(ev)) {
+                case "s":
+                case "S":
+                    ev.preventDefault();
+                    document.getElementById("search").focus();
+                    break;
+            }
+        }
```
jyn514, line 241: > > Was all this here already? This is a lot of JavaScript, it looks like rustdoc code.

> This was part of where crate pages and rustdoc pages melded together, I honestly couldn't say

Please find out.

templates/crate/details.html
```diff
@@ -0,0 +1,304 @@
...
+                        {{ details.name }}-{{ details.version }} is not a library.
+                    </div>
+
+                {# If the release has been yanked and is a library #}
+                {%- elif details.yanked -%}
+                    <div class="warning">
+                        {{ details.name }}-{{ details.version }} has been yanked.
+                    </div>
+
+                {# If the build succeeded, isn't yanked and is a library #}
```
jyn514, line 160: > > Why is this an else? If it failed to build and also was yanked, don't we want to show both warnings?

> This is an exact recreation of the previous behavior, I don't know if multiple warnings will conflict in some way (CSS wise)

This is not the same behavior, see https://github.com/rust-lang/docs.rs/blob/master/templates/crate_details.hbs#L47-L59. Please update this to match.

templates/crate/details.html
```diff
@@ -0,0 +1,304 @@
...
+                {%- elif details.build_status -%}
+                    {# If there are no docs display a warning #}
+                    {%- if not details.rustdoc_status -%}
+                        <div class="warning">{{ details.name }}-{{ details.version }} doesn't have any documentation.</div>
+                    {%- endif -%}
+
+                    {# If there's a readme, display it #}
+                    {%- if details.readme -%}
+                        {{ details.readme | safe }}
```
jyn514, line 169: Are we sure this should have `safe`? It's a markdown file, not HTML. Does this happen before or after markdown rendering?

templates/rustdoc/navigation.html
```diff
@@ -0,0 +1,248 @@
...
+                                                        </li>
+                                                    {%- endfor -%}
+                                                </ul>
+                                            </div>
+                                        </li>
+                                    </ul>
+                                </div>
+                            </div>
+                        </div>
+                    </li>
```
jyn514, line 184: > > Ditto on the repeated template, it would be great to factor it out.

> I think this is the same template as the last comment?

Right, it's still duplicated.

src/web/source.rs
jyn514: You didn't hit submit on the last comment (or at least it didn't show up).

Kixiron: Dammit, github’s lying to me again. It was artificially constrained and required code changes for something that’s entirely a UI issue, I made the templates use the actual mime types to choose file icons for the source view. Behavior hasn’t changed on the Rust side, if anything it’s less complex, it just enables the frontend/templates to do more

jyn514: Oh I see, previously this was only used to choose which icon to display? In that case this is fine, yeah.

templates/crate/details.html
Kixiron: The homepage just isn't there

templates/crate/details.html
Kixiron: After markdown rendering which happens during serialization

templates/crate/details.html
Kixiron: This is the same behavior. Handlebars follows this format

```py
if yanked:
    if build_status:
        {name-version} is yanked
    else:
        {name}-{this.version} is yanked and docs.rs failed to build it
    
else:
    if build_status:
        {version}
    else:
        docs.rs failed to build {name}-{version}
```

Which unrolled makes this

```py
if yanked and build_status:
    {name-version} is yanked
else if yanked:
    {name}-{this.version} is yanked and docs.rs failed to build it
else if build_status:
    {version}
else:
    docs.rs failed to build {name}-{version}
```

Which is the behavior and logic that the tera template has

templates/crate/details.html
Kixiron: It was a misunderstanding on my part, so that didn't need to be there at all

Kixiron:
I looked into trying to unify the details dropdown of [crate/details.html](https://github.com/Kixiron/docs.rs/blob/the-rest-of-tera/templates/crate/details.html#L18-L113) and [rustdoc/navigation.html](https://github.com/Kixiron/docs.rs/blob/the-rest-of-tera/templates/rustdoc/navigation.html#L33-L156) but I'm nervous to try since `navigation.html` has lots of extra CSS classes, and I think that's a bit out of scope for this PR. I'd be happy to open an issue after this PR lands and do it as a follow-up, but it's quite a lot to do for one PR

templates/crate/details.html
jyn514: Ah yep you're right. Sorry for the confusion.

PR #883, opened 2020-07-04
Reworked CI

Kixiron:
* Made caching better, it now cleans up uneeded garbage and recovers on cache misses
* Split testing, `clippy` and `rustfmt` into separate jobs so they can run independently and not interrupt each other
* Added `cargo-bloat-action` which will comment on PRs which affect dependencies and display the current number of dependencies, what dependencies have been added/removed and the final binary size each dependency inhabits
* Split the docker job into a separate file since it has nothing to do with the rest of that CI (it's self-contained)
* Added healthchecks to the compose file which should make running `docker-compose` more seamless

r? @Nemo157


.github/workflows/ci.yml
```diff
@@ -10,38 +10,43 @@ jobs:
...
+      - uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-sweep
 
-      # Caching
-      - name: Cache Cargo directories
-        uses: actions/cache@v2
+      - name: Cache directories
+        uses: actions/cache@v1
```
Nemo157, line 26: Why go back to v1?

.github/workflows/ci.yml
```diff
@@ -52,38 +57,87 @@ jobs:
           # Make sure the database is actually working
```
Nemo157, line 57: Given the health-check, can we replace the `sleep 5` above here with something that actually waits for the containers to be ready?

.github/workflows/ci.yml
```diff
@@ -52,38 +57,87 @@ jobs:
...
+          args: cargo-sweep
+
+      - name: Cache directories
+        uses: actions/cache@v1
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/bin
+            ~/.cargo/git
+          key: cargo-dirs-${{ hashFiles('**/Cargo.lock') }}
```
Nemo157, line 117: Clippy needs a disambiguator in its cache key, since it finished first it stored its cache and the main build failed to:

```
[warning]reserveCache failed: Cache already exists. Scope: refs/pull/883/merge, Key: cargo-build-c7087fe00-844e47a1d4d07d013bbf2fbfd5751995d510574cded08b4373348486baf5e48c, Version: (null)
```

https://github.com/rust-lang/docs.rs/pull/883/checks?check_run_id=837676712#step:22:2

.github/workflows/ci.yml
```diff
@@ -10,38 +10,43 @@ jobs:
...
-        run: rustup update stable && rustup default stable
-
-      - name: Set rustc version
-        run: echo "::set-env name=CURRENT_RUSTC_VERSION::$(rustc -V)"
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-sweep
```
Nemo157, line 23: `cargo-sweep` should be installed after restoring `~/.cargo/bin` so that we can use the cached build of it.

.github/workflows/ci.yml
```diff
@@ -10,38 +10,43 @@ jobs:
...
 
-      # Caching
-      - name: Cache Cargo directories
-        uses: actions/cache@v2
+      - name: Cache directories
+        uses: actions/cache@v1
         with:
           path: |
+            ~/.cargo/registry
```
Nemo157, line 29: This will slowly build up dependencies since it's not swept, though probably doesn't matter much given how small sources are.

.github/workflows/ci.yml
Kixiron: Whoops, my bad

.github/workflows/ci.yml
Kixiron: We could run [`cargo-cache`](https://github.com/matthiaskrgr/cargo-cache) to clean it up

.github/workflows/ci.yml
Kixiron: Apparently docker-compose only respects/runs health checks for a service's dependencies, I tried making a dummy service that just depended on the other two, but it didn't run. Any ideas?

Kixiron:
It looks like `cargo-bloat` [lacks the permissions to comment](https://github.com/rust-lang/docs.rs/pull/883/checks?check_run_id=838838273#step:8:2302) @pietroalbini I don't know how to fix that, could you take a look?

pietroalbini:
@Kixiron I'm not seeing any error related to cargo-bloat?

Kixiron:
Github linked improperly but [this should work](https://github.com/rust-lang/docs.rs/runs/838838273?check_suite_focus=true#step:8:2302)

pietroalbini:
@Kixiron that's not possible: the `GITHUB_TOKEN` available in PRs is read-only, and no other secret can be used in PRs.

.github/workflows/ci.yml
Nemo157: Ah, reading the docs sounds like the health-check is purely a debug status thing, it's not used by docker-compose in any way.

Might be good to use something like `wait-for`, but not necessary right now.

jyn514:
@Nemo157 you were the main reviewer for this, are you happy with it? If so I think it can be merged after a rebase.

PR #891, opened 2020-07-16
Reduce code duplication for releases page

jyn514:
r? @Nemo157

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -72,67 +72,37 @@ impl Default for Order {
 pub(crate) fn get_releases(conn: &Connection, page: i64, limit: i64, order: Order) -> Vec<Release> {
     let offset = (page - 1) * limit;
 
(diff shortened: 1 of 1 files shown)


src/web/releases.rs
```diff
@@ -72,67 +72,37 @@ impl Default for Order {
...
+        "SELECT crates.name,
+            releases.version,
+            releases.description,
+            releases.target_name,
+            releases.release_time,
+            releases.rustdoc_status,
+            crates.github_stars
+        FROM crates
+        INNER JOIN releases ON crates.id = releases.crate_id
+        {}
```
Nemo157, line 98: This could be something like

```sql
WHERE $3 OR (releases.build_status = FALSE AND releases.is_library = TRUE)
```

and pass `!failed` as a parameter.

jyn514: Good idea, updated.

PR #894, opened 2020-07-16
Add ayu.css to list of required files

### LINKED ISSUE #889: Ayu theme doesn't work
Go to a recently published crate like <https://docs.rs/streaming_algorithms/0.3.0/streaming_algorithms/>, and try to switch to ayu theme, the page loses all styling information because the ayu css is 404. I'm guessing we don't pull it in as part of the "per-nightly" files.

cc @Cldfire

Nemo157:
fixes #889

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -39,6 +39,7 @@ const ESSENTIAL_FILES_VERSIONED: &[&str] = &[
     "down-arrow.svg",
     "dark.css",
     "light.css",
+    "ayu.css",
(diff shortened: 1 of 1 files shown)


Nemo157:
Testing on my local instance this works fine:

image

It likely only works when it's the first build of a new nightly, otherwise it will use the cached essential files.

jyn514:
> It likely only works when it's the first build of a new nightly, otherwise it will use the cached essential files.

Ok, then after this is merged we should find some copies of ayu.css for the past few nightlies and upload them so the past docs work.

PR #901, opened 2020-07-18
Use a custom environment variable to configure logging

Nemo157:
Using `RUST_LOG` leads to issues with accidentally configuring logging for unrelated utilities when overriding it. The biggest one I notice is that setting `RUST_LOG=debug` temporarily to debug stuff results in `sccache` logging extra details during `cargo run` (which I've also opened an issue about moving to a custom variable, but in general I don't think using a single variable for all projects written in Rust makes sense).

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,6 @@
 CRATESFYI_PREFIX=ignored/cratesfyi-prefix
 CRATESFYI_DATABASE_URL=postgresql://cratesfyi:password@localhost:15432
-RUST_LOG=cratesfyi,rustwide=info
+DOCSRS_LOG=cratesfyi,rustwide=info
 AWS_ACCESS_KEY_ID=cratesfyi
 AWS_SECRET_ACCESS_KEY=secret_key
 S3_ENDPOINT=http://localhost:9000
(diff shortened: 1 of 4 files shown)


jyn514:
RUST_LOG=debug means to debug every crate and every dependency. What's the use case for that instead of `RUST_LOG=cratesfyi`? That would avoid the sccache logging.

Nemo157:
Sometimes you don't know which crate is causing a failure, so you want to just spam out all debug output and scan it for something that looks relevant. We also share dependencies with `sccache`, so using `RUST_LOG=tokio=debug` to debug something Tokio related would still have the same sort of issue.

jyn514:
Ah good point. Then it's indeed unfortunate we use the same env variable, it should be per-process, not global for every rust progress.

jyn514:
cc @rust-lang/docs-rs, everyone will have to update `.env` for it to work properly after this change.

jyn514:
@Nemo157 this looks good to merge after a rebase. I'll have to update a few things in prod after it's deployed.

Nemo157:
rebased

PR #902, opened 2020-07-19
Storage refactorings

pietroalbini:
This PR includes the first batch of refactorings to the storage subsystem to better support build tests. The most notable change is we're now using a single instance of `Storage` across the whole application, instead of creating a new one every time we need to use it.

This doesn't contain all the changes to the storage I need to make: namely I first need to move all the configuration to `Config`. This PR is starting to get big though and rebasing it is a pain, so I submitted it earlier.

This is best reviewed commit-by-commit.


src/storage/s3.rs
```diff
@@ -17,14 +17,17 @@ pub(crate) use test::TestS3;
 
 pub(crate) static S3_BUCKET_NAME: &str = "rust-docs-rs";
 
-pub(crate) struct S3Backend<'a> {
+pub(crate) struct S3Backend {
     client: S3Client,
-    bucket: &'a str,
+    bucket: String,
```
jyn514, line 22: I'd rather make this `&'static str` instead; it communicates the intent more clearly and avoids a copy.

src/storage/mod.rs
```diff
@@ -260,12 +260,6 @@ fn detect_mime(file_path: &Path) -> Result<&'static str, Error> {
     })
 }
 
-impl<'a> From<S3Backend<'a>> for Storage<'a> {
-    fn from(db: S3Backend<'a>) -> Self {
-        Self::S3(db)
-    }
-}
```
jyn514, line 267: Good call on removing this.

src/storage/s3.rs
```diff
@@ -63,19 +59,31 @@ impl<'a> S3Backend<'a> {
         })
     }
 
-    pub(super) fn store_batch(&mut self, batch: &[Blob]) -> Result<(), Error> {
-        use futures::stream::FuturesUnordered;
-        use futures::stream::Stream;
+    pub(super) fn start_storage_transaction(&self) -> Result<S3StorageTransaction, Error> {
+        Ok(S3StorageTransaction {
+            s3: self,
+            runtime: Runtime::new()?,
```
jyn514, line 68: Great idea making it so this is only constructed when uploading, not downloading :)

src/utils/daemon.rs
```diff
@@ -110,7 +111,9 @@ pub fn start_daemon(
     // at least start web server
     info!("Starting web server");
 
-    crate::Server::start(None, false, db, config, build_queue)?;
+    let storage = Arc::new(Storage::new(db.clone()));
+
+    crate::Server::start(None, false, db, config, build_queue, storage)?;
```
jyn514, line 116: The number of parameters keeps going up! 😆 It's fine for now, but we might want to think about factoring out some of these into a struct.

src/storage/mod.rs
```diff
@@ -117,23 +117,29 @@ pub fn get_file_list<P: AsRef<Path>>(path: P) -> Result<Vec<PathBuf>, Error> {
...
-    pub(crate) fn new(conn: &'a Connection) -> Self {
-        if let Some(c) = s3::s3_client() {
-            Storage::from(S3Backend::new(c, s3::S3_BUCKET_NAME))
+pub struct Storage {
+    backend: StorageBackend,
+}
+
+impl Storage {
+    pub fn new(pool: Pool) -> Self {
+        let backend = if let Some(c) = s3::s3_client() {
```
jyn514, line 131: Since this should now only be called by `Storage::new`, can we make `s3_client` `pub(super)`?

src/bin/cratesfyi.rs
```diff
@@ -473,8 +473,8 @@ impl DatabaseSubcommand {
             }
 
             Self::AddDirectory { directory, prefix } => {
-                add_path_into_database(ctx.pool()?, &prefix, directory)
-                    .context("Failed to add directory into database")?;
+                add_path_into_database(&*ctx.storage()?, &prefix, directory)
+                    .expect("Failed to add directory into database");
```
jyn514: This looks like a bad rebase, can you change it to use `context()` instead of `expect()`?

src/db/file.rs
```diff
@@ -13,20 +12,16 @@ use std::path::{Path, PathBuf};
 
 /// Store all files in a directory and return [[mimetype, filename]] as Json
 ///
-/// If there is an S3 Client configured, store files into an S3 bucket;
-/// otherwise, stores files into the 'files' table of the local database.
-///
```
jyn514, line 24: Why'd this comment disappear?

jyn514:
Wait hold on, it looks like a runtime is still created each time `start_storage_transaction` is called. Can you make the runtime a field on the backend instead and use `OnceCell::Lazy` so it's only created once?

src/storage/s3.rs
pietroalbini: It can't be `'static`, the bucket name is generated dynamically in `TestS3`.

src/utils/daemon.rs
pietroalbini: Yeah, we should eventually create a `trait Context`.

src/storage/mod.rs
pietroalbini: It's also called in `crate::db::delete`. I plan to clean that up in a future PR.

src/storage/database.rs
```diff
@@ -63,10 +63,39 @@ impl DatabaseBackend {
...
+pub(super) struct DatabaseStorageTransaction<'a> {
+    transaction: Option<Transaction<'a>>,
+}
+
+impl<'a> StorageTransaction for DatabaseStorageTransaction<'a> {
+    fn store_batch(&mut self, batch: &[Blob]) -> Result<(), Error> {
+        let transaction = self
+            .transaction
+            .as_ref()
+            .expect("called complete() before store_batch()");
```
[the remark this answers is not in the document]
pietroalbini: I'm wary of using the `Drop` impl here as you can't return errors from them.

src/db/file.rs
pietroalbini: So, that comment is going to be false in a future PR (I want to remove autodetection), dunno why I removed it now.

pietroalbini: Added the comment back.

src/storage/s3.rs
Kixiron: You could use a `Cow<'static, str>`

pietroalbini: This is initialized just a single time when the application starts, is that really worth?

pietroalbini:

PR #906, opened 2020-07-21
Fixed accidental URL escaping

Kixiron:
Tera escapes variables of any type by default, but we have urls that are stored within Tera inputs and urls that are built as temporary variables, so those were getting escaped by accident. This PR fixes that as well as a lot of formatting on the generated code

Before urls were taking this form in repo links, doc links, etc.
```html
<a href="https:&#x2F;&#x2F;github.com&#x2F;sigmaSd&#x2F;IRust" class="pure-menu-link">
```
and now they take this form
```html
<a href="https://github.com/sigmaSd/IRust" class="pure-menu-link">
```

As of now I don't, but I can run url encoding on *just* the crate's name & version and author names if we want to since they can contain more or less arbitrary characters (to varying degrees) (so from the link `/crate/:crate/:version` just `:crate` and `:version` have url encoding applied and in author links to `/releases/:author` only `:author` will be encoded)

Sorry about the large-ish diff, the vast majority is just self-nits for template formatting

### DIFF
--- a/templates/core/home.html
+++ b/templates/core/home.html
@@ -34,10 +34,13 @@
 
             <ul>
                 {%- for release in recent_releases -%}
+                    {# FIXME: https://github.com/Keats/tera/issues/539 #}
+                    {%- set name = release.name | escape -%}
+                    {%- set version = release.version | escape -%}
                     {%- if release.rustdoc_status -%}
-                        {%- set release_url = "/" ~ release.name ~ "/" ~ release.version ~ "/" ~ release.target_name -%}
+                        {%- set release_url = "/" ~ name ~ "/" ~ version ~ "/" ~ release.target_name -%}
                     {%- else -%}
-                        {%- set release_url = "/crate/" ~ release.name ~ "/" ~ release.version -%}
+                        {%- set release_url = "/crate/" ~ name ~ "/" ~ version -%}
                     {%- endif -%}
 
                     <li>
--- a/templates/core/home.html
+++ b/templates/core/home.html
@@ -41,7 +41,7 @@
                     {%- endif -%}
 
                     <li>
-                        <a href="{{ release_url }}"
+                        <a href="{{ release_url | safe }}"
                             class="release">
                             <div class="pure-g">
                                 <div class="pure-u-1 pure-u-sm-6-24 pure-u-md-5-24 name">
--- a/templates/core/home.html
+++ b/templates/core/home.html
@@ -34,13 +34,10 @@
 
             <ul>
                 {%- for release in recent_releases -%}
-                    {# FIXME: https://github.com/Keats/tera/issues/539 #}
-                    {%- set name = release.name | escape -%}
-                    {%- set version = release.version | escape -%}
                     {%- if release.rustdoc_status -%}
-                        {%- set release_url = "/" ~ name ~ "/" ~ version ~ "/" ~ release.target_name -%}
+                        {%- set release_url = "/" ~ release.name ~ "/" ~ release.version ~ "/" ~ release.target_name -%}
                     {%- else -%}
-                        {%- set release_url = "/crate/" ~ name ~ "/" ~ version -%}
+                        {%- set release_url = "/crate/" ~ release.name ~ "/" ~ release.version -%}
                     {%- endif -%}
 
                     <li>
--- a/templates/core/sitemap.xml
+++ b/templates/core/sitemap.xml
@@ -2,7 +2,7 @@
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
     {% for release in releases -%}
         <url>
-            <loc>https://docs.rs/{{ release[0] }}</loc>
+            <loc>https://docs.rs/{{ release[0] | urlencode }}</loc>
             <lastmod>{{ release[1] | escape_xml }}</lastmod>
         </url>
     {%- endfor %}
--- a/templates/core/sitemap.xml
+++ b/templates/core/sitemap.xml
@@ -2,7 +2,7 @@
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
     {% for release in releases -%}
         <url>
-            <loc>https://docs.rs/{{ release[0] | urlencode }}</loc>
+            <loc>https://docs.rs/{{ release[0] }}</loc>
             <lastmod>{{ release[1] | escape_xml }}</lastmod>
         </url>
     {%- endfor %}
--- a/templates/crate/details.html
+++ b/templates/crate/details.html
@@ -20,7 +20,7 @@
                         <li class="pure-menu-heading">Authors</li>
                         {%- for author in details.authors -%}
                             <li class="pure-menu-item">
-                                <a href="/releases/{{ author[1] | urlencode }}" class="pure-menu-link">
+                                <a href="/releases/{{ author[1] }}" class="pure-menu-link">
                                     {{ author[0] }}
                                 </a>
                             </li>
@@ -68,7 +68,7 @@
 
                         {# Show a link to the crate's Crates.io page #}
                         <li class="pure-menu-item">
-                            <a href="https://crates.io/crates/{{ details.name | urlencode }}" class="pure-menu-link"
+                            <a href="https://crates.io/crates/{{ details.name }}" class="pure-menu-link"
                                 title="See {{ details.name }} on crates.io">
                                 <i class="fa fa-cube fa-fw"></i> Crates.io
                             </a>
@@ -81,7 +81,7 @@
                                     {# List all dependencies that the current release has #}
                                     {%- for dep in details.dependencies -%}
                                         <li class="pure-menu-item">
-                                            <a href="/crate/{{ dep[0] | urlencode }}/{{ dep[1] | urlencode }}" class="pure-menu-link">
+                                            <a href="/crate/{{ dep[0] }}/{{ dep[1] }}" class="pure-menu-link">
                                                 {{ dep[0] }} {{ dep[1] }}
                                                 <i class="dependencies {{ dep[2] | default(value='') }}">{{ dep[2] | default(value="") }}</i>
                                             </a>
@@ -105,7 +105,7 @@
                         <li class="pure-menu-heading">Owners</li>
                         <li class="pure-menu-item">
                             {%- for owner in details.owners -%}
-                                <a href="/releases/@{{ owner[0] | urlencode }}">
+                                <a href="/releases/@{{ owner[0] }}">
                                     <img src="{{ owner[1] }}" alt="{{ owner[0] }}" class="owner">
                                 </a>
                             {%- endfor -%}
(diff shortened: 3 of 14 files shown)


jyn514:
> Sorry about the large-ish diff, the vast majority is just self-nits for template formatting

Can you please split that into a separate commit so it's easier to see the changes in functionality?

jyn514:
> As of now I don't, but I can run url encoding on just the crate's name & version and author names

That sounds like a good idea. The name and version have restrictions from crates.io but AFAIK the author name can be an arbitrary string, so we should escape at least that. Fortunately in the common case of a-zA-Z0-9 the url-escaped version is the same as the original.

Kixiron:
The first commit is the content changes, the second is as the name suggests

templates/core/home.html
```diff
@@ -41,7 +41,7 @@ <h1 class="brand"><i class="fa fa-cubes fa-fw"></i> Docs.rs</h1>
                     {%- endif -%}
 
                     <li>
-                        <a href="{{ release_url }}"
+                        <a href="{{ release_url | safe }}"
```
jyn514, line 44: This doesn't escape the name or version, is that intentional? I saw that you escaped them later.

templates/crate/details.html
```diff
@@ -31,7 +31,7 @@
                         {# If the crate has a homepage, show it #}
                         {%- if details.homepage_url -%}
                             <li class="pure-menu-item">
-                                <a href="{{ details.homepage_url }}" class="pure-menu-link">
+                                <a href="{{ details.homepage_url | safe }}" class="pure-menu-link">
```
jyn514, line 34: This is user-controlled input. I am not comfortable putting a blanket `| safe` here, no one should be trying to parse this anyway.

templates/crate/details.html
```diff
@@ -40,7 +40,7 @@
                         {# If the crate has a custom doc url, show it #}
                         {%- if details.documentation_url -%}
                             <li class="pure-menu-item">
-                                <a href="{{ details.documentation_url }}" title="Canonical documentation" class="pure-menu-link">
+                                <a href="{{ details.documentation_url | safe }}" title="Canonical documentation" class="pure-menu-link">
```
jyn514, line 43: This is user-controlled input.

templates/crate/details.html
```diff
@@ -49,7 +49,7 @@
                         {# If the release has a repository, show it #}
                         {%- if details.repository_url -%}
                             <li class="pure-menu-item">
-                                <a href="{{ details.repository_url }}" class="pure-menu-link">
+                                <a href="{{ details.repository_url | safe }}" class="pure-menu-link">
```
jyn514, line 52: This is user-controlled input.

templates/crate/details.html
```diff
@@ -68,7 +68,7 @@
 
                         {# Show a link to the crate's Crates.io page #}
                         <li class="pure-menu-item">
-                            <a href="https://crates.io/crates/{{ details.name }}" class="pure-menu-link"
+                            <a href="https://crates.io/crates/{{ details.name | urlencode }}" class="pure-menu-link"
```
jyn514, line 71: This is what I was referring to when I said 'I saw you escaped this later':

templates/rustdoc/navigation.html
```diff
@@ -55,7 +55,7 @@
                                         {# If the crate has a homepage, show a link to it #}
                                         {%- if krate.homepage_url -%}
                                             <li class="pure-menu-item">
-                                                <a href="{{ krate.homepage_url }}" class="pure-menu-link">
+                                                <a href="{{ krate.homepage_url | safe }}" class="pure-menu-link">
```
jyn514, line 58: This is user-controlled input.

templates/rustdoc/navigation.html
```diff
@@ -64,7 +64,7 @@
                                         {# If the crate has external docs, show a link #}
                                         {%- if krate.documentation_url -%}
                                             <li class="pure-menu-item">
-                                                <a href="{{ krate.documentation_url }}" title="Canonical documentation" class="pure-menu-link">
+                                                <a href="{{ krate.documentation_url | safe }}" title="Canonical documentation" class="pure-menu-link">
```
jyn514, line 67: This is user-controlled input.

templates/rustdoc/navigation.html
```diff
@@ -73,7 +73,7 @@
                                         {# If the crate is hosted on GitHub, show some stats #}
                                         {%- if krate.github -%}
                                             <li class="pure-menu-item">
-                                                <a href="{{ krate.repository_url }}" class="pure-menu-link">
+                                                <a href="{{ krate.repository_url | safe }}" class="pure-menu-link">
```
jyn514, line 76: This is user-controlled input.

templates/rustdoc/navigation.html
```diff
@@ -84,14 +84,14 @@
                                         {# If all the crate has is a repo url, show it #}
                                         {%- elif repository_url -%}
                                             <li class="pure-menu-item">
-                                                <a href="{{ krate.repository_url }}" class="pure-menu-link">
+                                                <a href="{{ krate.repository_url | safe }}" class="pure-menu-link">
```
jyn514, line 87: This is user-controlled input.

templates/rustdoc/page.html
```diff
@@ -31,34 +31,39 @@
...
+        if (window.performance) {
+            const navEntries = window.performance.getEntriesByType('navigation');
+            const usedBack = navEntries.length > 0
+                && navEntries[0].type === 'back_forward'
+                || (window.performance.navigation
+                    && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD);
+
+            if (!usedBack && window.location.hash) {
+                window.addEventListener("scroll", maybeFixupViewPortPosition, {"once": true});
+            }
```
jyn514, line 65: Do you know why this all shows as changed?

templates/core/sitemap.xml
```diff
@@ -2,7 +2,7 @@
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
     {% for release in releases -%}
         <url>
-            <loc>https://docs.rs/{{ release[0] }}</loc>
+            <loc>https://docs.rs/{{ release[0] | urlencode }}</loc>
```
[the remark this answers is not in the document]
Kixiron: The default behavior is to html encode, this just takes over from that with url encoding instead. It's a difference that probably doesn't matter a whole lot, so either works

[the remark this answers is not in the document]
jyn514: I think we actually do want HTML encoding here - the point of escaping is to avoid issues like a url of `"><script>...` which means we need to escape for the current context, not for a url.

templates/core/home.html
Kixiron: It should be escaped as a whole, I don't think escaping happens as part of variable building

jyn514: I'm not sure what you mean here. You pipe it to `safe` which means it won't be escaped, which isn't the same behavior as elsewhere.

jyn514:
Waiting on https://github.com/rust-lang/docs.rs/pull/906#discussion_r458533030 to be addressed.

templates/header/package_navigation.html
```diff
@@ -60,15 +60,15 @@ <h1 id="crate-title">
                     {%- if metadata.rustdoc_status -%}
                         <li class="pure-menu-item">
                             {# The docs tab redirects to the docs, so the tab will never be selected and seen #}
-                            <a href="/{{ crate_path }}/{{ metadata.target_name }}/" class="pure-menu-link">
+                            <a href="/{{ crate_path | safe }}/{{ metadata.target_name }}/" class="pure-menu-link">
```
jyn514, line 63: You need to escape the name and version above, before setting `crate_path`:
```
                    {%- set name = metadata.name | escape -%}
                    {%- set version = metadata.version | escape -%}
```

templates/macros.html
```diff
@@ -88,39 +88,42 @@
 {% macro releases_list(name, releases) %}
     {%- for release in releases -%}
         {# The url for the release, `/crate/:name/:version` #}
-        {%- set release_url = "/crate/" ~ name ~ "/" ~ release.version -%}
+        {# FIXME: https://github.com/Keats/tera/issues/539 #}
+        {%- set name = name | escape -%}
+        {%- set version = release.version | escape -%}
+        {%- set release_url = "/crate/" ~ name ~ "/" ~ version -%}
         {# The release's name and version, `:name-:version` #}
         {%- set release_name = name ~ "-" ~ release.version -%}
```
jyn514, line 96:
```suggestion
        {%- set release_name = name ~ "-" ~ version -%}
```

templates/releases/feed.xml
```diff
@@ -23,7 +23,7 @@
         <entry>
```
jyn514, line 23: A couple lines above you need `            {%- set link = "/crate/" ~ name ~ "/" ~ version -%}`, not `release.name`. Same for both branches of the `if`.

templates/releases/releases.html
```diff
@@ -28,7 +28,7 @@
                     {%- endif -%}
 
                     <li>
-                        <a href="{{ link }}" class="release">
+                        <a href="{{ link | safe }}" class="release">
```
jyn514, line 31: Need to escape `name` and `version` in link. Although maybe since crates.io disallows `/` in names we can just assume all of these don't need to be escaped? I just think we should be consistent.

templates/rustdoc/navigation.html
```diff
@@ -22,7 +22,7 @@
 
                 <ul class="pure-menu-list">
                     <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
-                        <a href="{{ crate_url }}" class="pure-menu-link" title="{{ krate.description }}">
+                        <a href="{{ crate_url | safe }}" class="pure-menu-link" title="{{ krate.description }}">
```
jyn514, line 25: Name and version are not escaped.

Nemo157:
Cargo requires the name to be "Unicode XID characters (numbers, `-`, `_`, or most letters)" which is probably safe to not escape. The version is even more restricted; ASCII numbers, ASCII letters, `-`, `+` and `.`; so also probably safe to not escape.

templates/releases/feed.xml
```diff
@@ -12,20 +12,17 @@
...
+            {%- set link = "/crate/" ~ release.name ~ "/" ~ release.version -%}
         {%- endif %}
 
         <entry>
-            <title>{{ name }}-{{ version }}</title>
+            <title>{{ release.name }}-{{ version }}</title>
 
             <link href="{{ link | safe }}" />
-            <id>urn:docs-rs:{{ name }}:{{ version }}</id>
+            <id>urn:docs-rs:{{ release.name }}:{{ version }}</id>
```
jyn514, line 25:
```suggestion
            <id>urn:docs-rs:{{ release.name }}:{{ release.version }}</id>
```

PR #908, opened 2020-07-23
Move postgres to a non-default port

jyn514:
@andy-thomason How does this look to you?

Addresses the issues in https://github.com/rust-lang/crates-build-env/pull/57.


jyn514:
Another alternative is to move the postgres server to a different port. I think @Kixiron has had trouble in the past with using his host server accidentally instead of the one in docker-compose.

andy-thomason:
Thanks! I hope that avoid further pitfalls.

jyn514:
cc @rust-lang/docs-rs , you'll have to update `.env` to point to port 15432 instead of the default 5432. See the change I made to `.env.sample`.

PR #909, opened 2020-07-25
Switch to one single runtime for S3 storage

Kixiron:
* Made `S3Backend::start_storage_transaction` stop creating a new `Runtime` on each call
* Added `parking_lot` as a direct dependency (was a transitive dependency)

This should allow updating `tokio` to v0.2 without the "fun little bugs" and hopefully allow async to actually be achievable

### DIFF
--- a/src/storage/s3.rs
+++ b/src/storage/s3.rs
@@ -24,7 +24,6 @@ static S3_RUNTIME: Lazy<Mutex<Runtime>> =
 pub(crate) struct S3Backend {
     client: S3Client,
     bucket: String,
-    runtime: &'static Mutex<Runtime>,
 }
 
 impl S3Backend {
@@ -32,7 +31,6 @@ impl S3Backend {
         Self {
             client,
             bucket: bucket.into(),
-            runtime: &*S3_RUNTIME,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/storage/s3.rs
```diff
@@ -16,17 +18,21 @@ mod test;
 pub(crate) use test::TestS3;
 
 pub(crate) static S3_BUCKET_NAME: &str = "rust-docs-rs";
+static S3_RUNTIME: Lazy<Mutex<Runtime>> =
+    Lazy::new(|| Mutex::new(Runtime::new().expect("Failed to create S3 runtime")));
 
 pub(crate) struct S3Backend {
     client: S3Client,
     bucket: String,
+    runtime: &'static Mutex<Runtime>,
```
jyn514, line 27: Is there a reason to make this part of `S3Backend` instead of using `S3_RUNTIME.lock().block_on()` directly?

Kixiron: It was based off of [this comment](https://github.com/rust-lang/docs.rs/pull/902#issuecomment-660662482)

jyn514: Gotcha. Looking at the code I don't think it's actually necessary, we can change it to be a field as part of the switch to tokio 0.2 but for now I think it's more clear to use the `Lazy` directly.

PR #910, opened 2020-07-25
Update tokio?

Kixiron:
* Updates tokio to v0.2
* Updates futures-util
* Updates rusoto crates

With one runtime being used for S3 uploads via #909, it may work now?
Hopefully more successful than #799 and #758

### DIFF
--- a/src/db/delete.rs
+++ b/src/db/delete.rs
@@ -1,7 +1,7 @@
-use crate::storage::s3::{s3_client, S3_BUCKET_NAME};
+use crate::storage::s3::{s3_client, S3Backend, S3_BUCKET_NAME};
 use failure::{Error, Fail};
 use postgres::Connection;
-use rusoto_s3::{DeleteObjectsRequest, ListObjectsV2Request, ObjectIdentifier, S3Client, S3};
+use rusoto_s3::{DeleteObjectsRequest, ListObjectsV2Request, ObjectIdentifier, S3};
 
 /// List of directories in docs.rs's underlying storage (either the database or S3) containing a
 /// subdirectory named after the crate. Those subdirectories will be deleted.
@@ -16,20 +16,22 @@ enum CrateDeletionError {
 pub fn delete_crate(conn: &Connection, name: &str) -> Result<(), Error> {
     let crate_id = get_id(conn, name)?;
     delete_crate_from_database(conn, name, crate_id)?;
-    if let Some(s3) = s3_client() {
+
+    if let Some(backend) = s3_client().map(|client| S3Backend::new(client, S3_BUCKET_NAME)) {
         for prefix in STORAGE_PATHS_TO_DELETE {
-            delete_prefix_from_s3(&s3, &format!("{}/{}/", prefix, name))?;
+            delete_prefix_from_s3(&backend, &format!("{}/{}/", prefix, name))?;
         }
     }
+
     Ok(())
 }
 
 pub fn delete_version(conn: &Connection, name: &str, version: &str) -> Result<(), Error> {
     delete_version_from_database(conn, name, version)?;
 
-    if let Some(s3) = s3_client() {
+    if let Some(backend) = s3_client().map(|client| S3Backend::new(client, S3_BUCKET_NAME)) {
         for prefix in STORAGE_PATHS_TO_DELETE {
-            delete_prefix_from_s3(&s3, &format!("{}/{}/{}/", prefix, name, version))?;
+            delete_prefix_from_s3(&backend, &format!("{}/{}/{}/", prefix, name, version))?;
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/storage/s3.rs
```diff
@@ -18,51 +19,60 @@ mod test;
...
+
+            let mut body = res
+                .body
+                .ok_or_else(|| failure::err_msg("Received a response from S3 with no body"))?;
 
-        let mut body = res.body.unwrap().into_blocking_read();
-        std::io::copy(&mut body, &mut content)?;
+            while let Some(data) = body.next().await.transpose()? {
+                content.write_all(data.as_ref())?;
+            }
```
jyn514, line 64: This calls `await` on each chunk in serial, which seems needlessly slow. Can we try to `await` all of them at once somehow? That's what `into_blocking_read` did before.

As an aside, where did you find `next()`? I don't see it on https://docs.rs/rusoto_core/0.45.0/rusoto_core/struct.ByteStream.html.

src/storage/s3.rs
```diff
@@ -76,44 +86,53 @@ pub(super) struct S3StorageTransaction<'a> {
...
+                                key: blob.path.clone(),
+                                body: Some(blob.content.clone().into()),
+                                content_type: Some(blob.mime.clone()),
+                                content_encoding: blob
+                                    .compression
+                                    .as_ref()
+                                    .map(|alg| alg.to_string()),
+                                ..Default::default()
+                            })
+                            .map_ok(|_ok| {
```
jyn514, line 109:
```suggestion
                            .map_ok(|_| {
```

src/storage/s3.rs
```diff
@@ -76,44 +86,53 @@ pub(super) struct S3StorageTransaction<'a> {
...
+                            })
+                            .map_err(|err| {
+                                log::error!("Failed to upload blob to S3: {:?}", err);
+                                // Reintroduce failed blobs for a retry
+                                blob
+                            }),
+                    );
+                }
+
+                while let Some(result) = futures.next().await {
```
jyn514, line 120: Same comment, this waits on each in serial. Can we wait in parallel instead? That was the whole point of switching to async in the first place.

src/storage/s3.rs
```diff
@@ -76,44 +86,53 @@ pub(super) struct S3StorageTransaction<'a> {
...
+                    // Push each failed blob back into the batch
+                    if let Err(blob) = result {
+                        batch.push(blob);
                     }
                 }
+
+                // If the batch is empty, we can exit
```
jyn514, line 127:
```suggestion
                // If we uploaded everything in the batch, we're done
```

src/storage/s3.rs
```diff
@@ -76,44 +86,53 @@ pub(super) struct S3StorageTransaction<'a> {
...
+                if batch.is_empty() {
+                    return Ok(());
+                }
             }
-        }
 
-        Ok(())
+            // If uploading failed 3 times, explode
```
jyn514, line 133: I don't think this needs a comment.
```suggestion
```

src/storage/s3.rs
```diff
@@ -76,44 +86,53 @@ pub(super) struct S3StorageTransaction<'a> {
...
+            // Attempt to upload the batch 3 times
+            for _ in 0..3 {
+                let mut futures = FuturesUnordered::new();
+                for blob in batch.drain(..) {
+                    futures.push(
+                        self.s3
+                            .client
+                            .put_object(PutObjectRequest {
+                                bucket: self.s3.bucket.to_string(),
+                                key: blob.path.clone(),
```
jyn514, line 100: Pre-existing, but can we avoid this clone here? It takes a [`ByteStream`](https://docs.rs/rusoto_core/0.45.0/rusoto_core/struct.ByteStream.html), which it looks like we can construct with only [`&mut blob.path`](https://docs.rs/futures-core/0.3.5/futures_core/stream/trait.Stream.html#impl-Stream-for-%26%27_%20mut%20S).

src/storage/s3.rs
Kixiron: `.next()` comes from the [`Stream` impl](https://docs.rs/rusoto_core/0.45.0/rusoto_core/struct.ByteStream.html#impl-Stream). This is also doing the [same thing `.into_blocking_read` does](https://docs.rs/rusoto_signature/0.45.0/src/rusoto_signature/stream.rs.html#139-141). With maintaining the order of the body, this is about as fast as it gets without major complication

src/storage/s3.rs
Kixiron: This also doesn't wait in serial since the containing collection is a [`FuturesUnordered`](https://docs.rs/futures-util/0.3.5/futures_util/stream/struct.FuturesUnordered.html). Calling `.poll_next()` (and transitively `.next()`) polls all contained futures until a ready one is found

src/db/delete.rs
```diff
@@ -16,20 +16,22 @@ enum CrateDeletionError {
 pub fn delete_crate(conn: &Connection, name: &str) -> Result<(), Error> {
     let crate_id = get_id(conn, name)?;
     delete_crate_from_database(conn, name, crate_id)?;
-    if let Some(s3) = s3_client() {
+
+    if let Some(backend) = s3_client().map(|client| S3Backend::new(client, S3_BUCKET_NAME)) {
```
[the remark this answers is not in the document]
Kixiron: That'd also require taking in a bucket name, so it'd be changing calls in a slightly different way

src/storage/s3.rs
Kixiron: Unfortunately that's an `impl Stream for &mut S where S: ?Sized + Stream + Unpin`, which an `&mut Vec<u8>` doesn't satisfy

src/storage/s3.rs
Nemo157: `FuturesUnordered` waits on all futures in itself concurrently.

src/storage/s3.rs
jyn514: Is this mostly the same as `into_async_read()`? If so, can we use that instead of a loop?

src/storage/s3.rs
jyn514: Ah gotcha, I misread it. This is fine then. I think I actually opened an issue for this a while ago: https://github.com/rusoto/rusoto/issues/234#issuecomment-632423950

src/storage/s3.rs
jyn514: I'm still not quite following ... but @Nemo157 is I think one of the authors of `futures` so I'll take his word for it 😆

src/db/delete.rs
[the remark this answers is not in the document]
jyn514: Do we need to take in the bucket name? Can we pass `S3_BUCKET_NAME` from `s3_client`? It seems silly to pass a constant that's always the same.

src/db/delete.rs
```diff
@@ -119,50 +121,57 @@ fn delete_crate_from_database(conn: &Connection, name: &str, crate_id: i32) -> R
...
-                    objects: to_delete,
-                    quiet: None,
-                },
-                ..DeleteObjectsRequest::default()
-            })
-            .sync()?;
-        if let Some(errs) = resp.errors {
-            for err in &errs {
-                log::error!("error deleting file from s3: {:?}", err);
+fn delete_prefix_from_s3(s3: &S3Backend, name: &str) -> Result<(), Error> {
```
jyn514, line 124: Now that we use a global `S3_RUNTIME`, we can change this back to just take the client. Same for various other places that changed to take a backend.

src/db/delete.rs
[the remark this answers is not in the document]
Kixiron: We need to take the bucket name at some point [due to tests](https://github.com/rust-lang/docs.rs/blob/master/src/storage/s3/test.rs#L14-L24) (Note that I've removed the `Box::leak` in my PR since that's not good)

src/storage/s3.rs
Kixiron: `.into_async_read()` doesn't allow for the same pre-allocation that we have, plus the loop's a pretty standard async pattern

jyn514: Got it. And we can't use `FuturesUnordered` because we do care about the order in this case. I'm still concerned that we're waiting in serial though - can we use [`FuturesOrdered`](https://docs.rs/futures/0.3.5/futures/stream/struct.FuturesOrdered.html) instead? That says

> While futures in the set will race to completion in parallel, results will only be returned in the order their originating futures were added to the queue.

> Futures are pushed into this queue and their realized values are yielded in order. This structure is optimized to manage a large number of futures. Futures managed by FuturesOrdered will only be polled when they generate notifications. This reduces the required amount of work needed to coordinate large numbers of futures.

src/storage/s3.rs
jyn514: For the record, Nemo explained it in more depth here: https://discordapp.com/channels/442252698964721669/541978667522195476/736708951332225226

src/storage/s3.rs
Kixiron: `FuturesUnordered` isn't used here, only for uploading batched files while this is downloading the contents of a singular future

src/db/delete.rs
```diff
@@ -16,20 +16,22 @@ enum CrateDeletionError {
 pub fn delete_crate(conn: &Connection, name: &str) -> Result<(), Error> {
     let crate_id = get_id(conn, name)?;
     delete_crate_from_database(conn, name, crate_id)?;
-    if let Some(s3) = s3_client() {
+
+    if let Some(backend) = s3_client() {
```
jyn514, line 20: nit: you changed the types back, but not the variable names. All this code is going to be refactored soon though, so no need to fix it unless you want to.
```suggestion
    if let Some(client) = s3_client() {
```

PR #911, opened 2020-07-26
Unify storage tests

pietroalbini:
This PR improves the test coverage of the `Storage` by moving most of the tests into "backend tests". Storage backend tests are shared between backend implementations, and executed on both backends (database and S3) when running `cargo test`. This reduces the amount of tests that needs to be written for the storage (as we don't have to duplicate tests between the database and S3 anymore) and increases the confidence in switching backends.

Along with those changes:

* The compression code was split in its own finme
* The S3 bucket name is configurable through `Config`
* `TestS3` is removed, and it's behavior is moved to the normal S3 backend when used during tests
* Deleting prefixes was moved from `db/delete.rs` to `Storage`
* Use a runtime in the `S3Backend` struct instead of a global one

r? @jyn514 
cc @Kixiron


jyn514:
@pietroalbini do you need this change for build tests? Or is it a general refactor?

pietroalbini:
Kinda, it started for buildtests and then I started on a tangent :sweat_smile:

pietroalbini:
Rebased and did this other change:

* Use a runtime in the S3Backend struct instead of a global one

src/bin/cratesfyi.rs
```diff
@@ -584,7 +585,9 @@ impl Context {
     fn storage(&self) -> Result<Arc<Storage>, Error> {
         Ok(self
             .storage
-            .get_or_try_init::<_, Error>(|| Ok(Arc::new(Storage::new(self.pool()?))))?
+            .get_or_try_init::<_, Error>(|| {
+                Ok(Arc::new(Storage::new(self.pool()?, &*self.config()?)?))
+            })?
             .clone())
     }
```
Kixiron, line 592: It's unrelated to your PR, but we should be using `once_cell::sync::Lazy`s inside of `Context` instead of `OnceCell`s

src/storage/mod.rs
```diff
@@ -279,215 +260,322 @@ fn detect_mime(file_path: &Path) -> Result<&'static str, Error> {
...
+        check_mime("hello.css", "text/css");
+        check_mime("hello.js", "application/javascript");
+        check_mime("hello.html", "text/html");
+        check_mime("hello.hello.md", "text/markdown");
+        check_mime("hello.markdown", "text/markdown");
+        check_mime("hello.json", "application/json");
+        check_mime("hello.txt", "text/plain");
+        check_mime("file.rs", "text/rust");
+        check_mime("important.svg", "image/svg+xml");
+    }
```
Kixiron, line 289: These were really needed, thanks!

src/storage/s3.rs
```diff
@@ -77,6 +104,29 @@ impl S3Backend {
     pub(super) fn start_storage_transaction(&self) -> Result<S3StorageTransaction, Error> {
         Ok(S3StorageTransaction { s3: self })
     }
+
+    #[cfg(test)]
+    pub(super) fn cleanup_after_test(&self) -> Result<(), Error> {
+        if !self.temporary {
+            return Ok(());
+        }
```
Kixiron, line 112: Is this check needed if there's the panic below?

src/test/mod.rs
```diff
@@ -99,7 +98,8 @@ pub(crate) struct TestEnvironment {
     db: OnceCell<TestDatabase>,
     storage: OnceCell<Arc<Storage>>,
     frontend: OnceCell<TestFrontend>,
-    s3: OnceCell<TestS3>,
+    s3: OnceCell<Arc<Storage>>,
+    storage_db: OnceCell<Arc<Storage>>,
```
Kixiron, line 102: `Lazy` is always better than `OnceCell`, it removes the user overhead of filling them manually

Kixiron, line 102: Is there a reason to have everything inside of the test environment `Arc`'d instead of simply having the environment itself be `Arc`'d?

src/bin/cratesfyi.rs
```diff
@@ -584,7 +585,9 @@ impl Context {
     fn storage(&self) -> Result<Arc<Storage>, Error> {
         Ok(self
             .storage
-            .get_or_try_init::<_, Error>(|| Ok(Arc::new(Storage::new(self.pool()?))))?
```
Kixiron, line 587: Is there a reason to have `Context`'s members all be in `Arc`s instead of it?

src/db/delete.rs
```diff
@@ -13,26 +12,27 @@ enum CrateDeletionError {
...
     Ok(())
 }
 
-pub fn delete_version(conn: &Connection, name: &str, version: &str) -> Result<(), Error> {
+pub fn delete_version(
+    conn: &Connection,
+    storage: &Storage,
+    name: &str,
+    version: &str,
+) -> Result<(), Error> {
```
Kixiron, line 31: Maybe switch away from the standalone function since you've already moved the meat of it to the associated function on `Storage`

src/storage/compression.rs
```diff
@@ -0,0 +1,136 @@
...
+        $vis enum $name {
+            $($variant = $discriminant,)*
+        }
+
+        impl $name {
+            #[cfg(test)]
+            const AVAILABLE: &'static [Self] = &[$(Self::$variant,)*];
+        }
+
+        impl fmt::Display for CompressionAlgorithm {
```
Kixiron, line 18: Bit of a nit, might as well use the fully qualified paths here and remove the external import

src/bin/cratesfyi.rs
pietroalbini: We can't use `Lazy` (or it's impractical anyway). Those `OnceCell`s depend on each other, but `Lazy` requires the initializer closure to be defined when it's instantiated.

src/storage/s3.rs
pietroalbini: Yes, this way if we create a non-temporary S3 bucket during tests we avoid shredding data.

src/test/mod.rs
pietroalbini: See https://github.com/rust-lang/docs.rs/pull/911#discussion_r460587158

pietroalbini: Yes: having each component in its own `Arc` allows cloning it and passing it around without carrying the whole context, especially when only one of them is present. Also, currently the contextes of `cratesfyi` and the test suite are different, so we couldn't do that anyway.

src/bin/cratesfyi.rs
pietroalbini: See https://github.com/rust-lang/docs.rs/pull/911#discussion_r460587484

src/storage/s3.rs
```diff
@@ -1,50 +1,77 @@
...
-            client,
-            bucket: bucket.into(),
+    pub(super) fn new(client: S3Client, config: &Config) -> Result<Self, Error> {
+        let runtime = Runtime::new()?;
+
+        // Create the temporary S3 bucket during tests.
+        if config.s3_bucket_is_temporary {
+            if cfg!(not(test)) {
+                panic!("safeguard to prevent creating temporary buckets outside of tests");
+            }
```
jyn514, line 35: Can we instead enforce that this won't compile unless it's in test mode? Something like this:

```suggestion
        extern crate criterion;
```

That should give a compile error without dev-dependencies even if the `cfg(test)` somehow disappears.

Kixiron: We could also use a compile error within the function
```suggestion
            #[cfg(not(test))]
            compile_error!("You fool, you'll kill us all!");
```

src/storage/s3.rs
```diff
@@ -1,50 +1,77 @@
...
+                .await;
+
+            let res = match response {
+                Ok(res) => res,
+                Err(RusotoError::Service(GetObjectError::NoSuchKey(_))) => {
+                    return Err(super::PathNotFoundError.into());
+                }
+                Err(RusotoError::Unknown(http)) if http.status == 404 => {
+                    return Err(super::PathNotFoundError.into());
+                }
```
Kixiron, line 72: Maybe these errors should be logged with more details than just the `PathNotFoundError`
```suggestion
                Err(RusotoError::Service(GetObjectError::NoSuchKey(key))) => {
                    log::error!("S3 didn't recognize a key: {:?}", key);
                    return Err(super::PathNotFoundError.into());
                }
                Err(RusotoError::Unknown(http)) if http.status == 404 => {
                    log::error!("Attempted to get an unknown file from S3: {:?}", http);
                    return Err(super::PathNotFoundError.into());
                }
```

src/storage/s3.rs
```diff
@@ -1,50 +1,77 @@
...
+                Err(err) => return Err(err.into()),
+            };
```
Kixiron, line 74: These can be simplified to remove the returns
```suggestion
            let res = match response.map_err(|err| match err {
                RusotoError::Service(GetObjectError::NoSuchKey(_)) => super::PathNotFoundError.into(),
                RusotoError::Unknown(http)) if http.status == 404 => super::PathNotFoundError.into(),
                err => err.into(),
            })?;
```
With the above suggestion that makes this
```suggestion
            let res = match response.map_err(|err| match err {
                RusotoError::Service(GetObjectError::NoSuchKey(_)) => {
                    log::error!("S3 didn't recognize a key: {:?}", key);
                    super::PathNotFoundError.into()
                },
                RusotoError::Unknown(http)) if http.status == 404 => {
                    log::error!("Attempted to get an unknown file from S3: {:?}", http);
                    super::PathNotFoundError.into()
                },
                err => err.into(),
            })?;
```

src/storage/s3.rs
```diff
@@ -132,6 +182,61 @@ impl<'a> StorageTransaction for S3StorageTransaction<'a> {
...
+                        ..DeleteObjectsRequest::default()
+                    })
+                    .await?;
+
+                if let Some(errs) = resp.errors {
+                    for err in &errs {
+                        log::error!("error deleting file from s3: {:?}", err);
+                    }
+
+                    failure::bail!("uploading to s3 failed");
```
Kixiron, line 229:
```suggestion
                    failure::bail!("deleting from s3 failed");
```

src/storage/s3.rs
```diff
@@ -132,6 +182,61 @@ impl<'a> StorageTransaction for S3StorageTransaction<'a> {
...
+                }
+
+                continuation_token = list.continuation_token;
+                if continuation_token.is_none() {
```
Kixiron, line 233: Is the `continuation_token` temp var even needed here? It looks like you could just use
```rust
if list.continuation_token.is_none() {
    return Ok(());
}
```

src/storage/s3.rs
```diff
@@ -77,6 +104,29 @@ impl S3Backend {
...
+        if !self.temporary {
+            return Ok(());
+        }
+
+        if cfg!(not(test)) {
+            panic!("safeguard to prevent deleting the production bucket");
+        }
+
+        let mut transaction = Box::new(self.start_storage_transaction()?);
+        transaction.delete_prefix("")?;
```
Kixiron, line 119: What does deleting an empty prefix do?

pietroalbini: It removes all the contents.

src/storage/s3.rs
pietroalbini: The variable is used in the next iteration to continue with the listing, so yes, it's needed.

src/storage/mod.rs
```diff
@@ -119,21 +68,38 @@ pub fn get_file_list<P: AsRef<Path>>(path: P) -> Result<Vec<PathBuf>, Error> {
 
 enum StorageBackend {
     Database(DatabaseBackend),
-    S3(S3Backend),
+    S3(Box<S3Backend>),
```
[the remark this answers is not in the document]
pietroalbini: Yes.

src/storage/s3.rs
pietroalbini: I mean, they're just 404. Not sure why the extra logging would be useful.

src/storage/s3.rs
pietroalbini: It's already `#[cfg(test)]`, I don't see a reason to add even that.

pietroalbini:
Addressed most of the feedback. The few items left are things I think can be done in a followup PR.

src/storage/s3.rs
Kixiron: Why was this closed?

pietroalbini: It **was** resolved, then I force-pushed by accident overriding it :sweat_smile:

Good catch!

jyn514:
> Use a runtime in the S3Backend struct instead of a global one

Can you expand on the rationale for this? Are we sure the backend is only ever constructed once? Tokio is still [leaking file descriptors](https://github.com/tokio-rs/tokio/issues/2635) so we need to be careful not to create too many runtimes.

pietroalbini:
Yeah we're only building a single instance of the backend across the whole application.

Issue #907, opened 2020-07-22
New release for badge crate


jyn514:
Is it possible to create a new release for the `badge` crate? I use the crate myself and got the same deprecation warning for `stb_truetype`.

_Originally posted by @vbrandl in https://github.com/rust-lang/docs.rs/issues/869#issuecomment-662569458_

jyn514:
@onur are you still around github? I don't think anyone else has permissions to publish `badge`.

jyn514:
Just published a new release!

PR #912, opened 2020-07-26
Release badge 0.3

jyn514:
- Add myself and @Kixiron as authors
- Add a changelog
- Update `base64` dependency. This changes no dependencies in the docs.rs tree (because `iron` depended on 0.9 and we already depended directly on 0.12), but gets all the improvements from 0.9 to 0.12: https://github.com/marshallpierce/rust-base64/blob/master/RELEASE-NOTES.md

Closes https://github.com/rust-lang/docs.rs/issues/907

@vbrandl how does this look to you? Anything missing that you'd like to see in a new release?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,7 +93,7 @@ dependencies = [
 
 [[package]]
 name = "badge"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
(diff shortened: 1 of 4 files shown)


vbrandl:
Nothing that I can think of. I had no problems with the API, just with `stb_truetype` being outdated. Looks good.

Issue #357, opened 2019-05-20
Please install `rustfmt` to the docs.rs doc build environment


Lokathor:
My crate builds bindings, and the bindings output comes out as [all one line](https://docs.rs/fermium/0.0.11/src/fermium/home/cratesfyi/cratesfyi/debug/build/fermium-2e11252b8a4909e1/out/bindings.rs.html#3) without rustfmt available.

jyn514:
@Lokathor  It looks like this is fixed in new versions of fermium, probably because of #407. Can you take a look at https://docs.rs/fermium/0.1.4/src/fermium/SDL2-v2.0.8-x86_64-pc-windows-msvc.rs.html#4692-4695 and make sure it looks about right?

Lokathor:
Well, that's correct, but probably because I have since stopped using bindgen on `docs.rs`. The file you linked is just permanently in the repo: https://github.com/Lokathor/fermium/blob/master/src/SDL2-v2.0.8-x86_64-pc-windows-msvc.rs

but i guess we can close this and someone else will report if it's still a problem later.

jyn514:
I'd actually rather keep it open until I know it's fixed, sorry we took so long you found a workaround instead.

jyn514:
Mentoring instructions:
- In [`update_toolchain`](https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L143), call [`toolchain.add_component("rustfmt")`](https://docs.rs/rustwide/0.5.1/rustwide/toolchain/struct.Toolchain.html#method.add_component).

If you have any questions, feel free to reach out here or [on Discord](https://discord.gg/f7mTXPW).

elmurci:
`add_component` requires two parameters:

`self.toolchain.add_component(&self.workspace, "rustfmt");`

elmurci:
the solution above didn't work...

jyn514:
What do you mean by didn't work? Can you put up a draft branch?

elmurci:
adding the command:

`self.toolchain.add_component(&self.workspace, "rustfmt")?;`

to `update_toolchain`

jyn514:
@elmurci can you join the discord so I can help you debug without spamming notifications for people watching the issue? https://discord.gg/f7mTXPW

elmurci:
I managed to fix it. Putting it here as it may help others.
[...]

PR #913, opened 2020-07-27
Add rustfmt when installing a new toolchain

jyn514:
Example build:

```
2020/07/26 20:02:52 [INFO] rustwide::toolchain: adding component rustfmt for toolchain nightly
2020/07/26 20:02:52 [INFO] rustwide::cmd: running `Command { std: "/home/joshua/src/rust/docs.rs/.rustwide/cargo-home/bin/rustup" "component" "add" "--toolchain" "nightly" "rustfmt", kill_on_drop: false }`
2020/07/26 20:02:53 [INFO] rustwide::cmd: [stderr] info: downloading component 'rustfmt'
2020/07/26 20:02:54 [INFO] rustwide::cmd: [stderr] info: installing component 'rustfmt'
```

This was tested by building `fermium 0.0.11` with the new changes. I made sure rustfmt was actually visible to the crate by looking at the generated source. Here is `/fermium/0.0.11/src/fermium/opt/rustwide/target/debug/build/fermium-d794f31f687205c6/out/bindings.rs.html#714-726`:

image

I have not tested the upgrade behavior, so possibly this shouldn't be
merged until I do that.

Closes https://github.com/rust-lang/docs.rs/issues/357.

r? @pietroalbini

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -166,6 +166,9 @@ impl RustwideBuilder {
         for target in &targets_to_install {
             self.toolchain.add_target(&self.workspace, target)?;
         }
+        // NOTE: rustup will automatically refuse to update the toolchain
+        // if `rustfmt` is not available in the newer version
+        self.toolchain.add_component(&self.workspace, "rustfmt")?;
 
         self.rustc_version = self.detect_rustc_version()?;
         if old_version.as_deref() != Some(&self.rustc_version) {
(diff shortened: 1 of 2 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -166,6 +166,9 @@ impl RustwideBuilder {
         for target in &targets_to_install {
             self.toolchain.add_target(&self.workspace, target)?;
         }
+        // NOTE: rustup will automatically refuse to update the toolchain
+        // if `rustfmt` is not available in the newer version
```
Nemo157, line 170: Do we _want_ this behavior, or should we not block the update on it and only have optional `rustfmt` support? This will currently result in occasionally having like a week without updating the nightly.

jyn514: I think going a week without updating is better than having rustfmt support you can only sometimes depend on. See https://github.com/rust-lang/docs.rs/issues/357#issuecomment-606252725.

Issue #803, opened 2020-06-02
404s from module item links when module url doesn't with a /


dnsmt:
I regularly navigate back to a module page from a function description by removing the function name from the url. If I also remove the `/` the module overview page loads just fine, but the urls to its children all 404.

Example:
1. Initial url: https://docs.rs/tokio/0.2.21/tokio/time/fn.advance.html
2. Remove the `/fn.advance.html` to navigate to https://docs.rs/tokio/0.2.21/tokio/time
3. Observe that all items at the bottom of the page listed under modules, structs, and functions now 404. This does not happen on https://docs.rs/tokio/0.2.21/tokio/time/ (note the final slash).

The 404s are correct, since the url for the children lacks module name (`time`). The submodule `delay_queue` links to https://docs.rs/tokio/0.2.21/tokio/delay_queue/index.html instead of https://docs.rs/tokio/0.2.21/tokio/time/delay_queue/index.html and `tokio::delay_queue` of course does not exist.

I'd prefer to either be redirected from https://docs.rs/tokio/0.2.21/tokio/time (without slash) to https://docs.rs/tokio/0.2.21/tokio/time/ (with slash) or that the urls on https://docs.rs/tokio/0.2.21/tokio/time (without slash) are derived properly.

Nemo157:
The redirect will much easier (and more correct) to do. Around here we just need to check if the original url ended with a `/` when trying to load the `index.html`, if it didn't and we successfully found an `index.html` then we should redirect to the current url + `/`.

https://github.com/rust-lang/docs.rs/blob/9bd032f0ef07b217450cee3fda4906086de9cc93/src/web/rustdoc.rs#L300-L305

Kixiron:
I think this may be a related issue, but when in a source view navigating to sub-directories requires a trailing `/`, for the case of [tokio](https://docs.rs/crate/tokio/0.2.21/source/) [this](https://docs.rs/crate/tokio/0.2.21/source/src/) brings you to the `src` folder but [this](https://docs.rs/crate/tokio/0.2.21/source/src) doesn't

PR #914, opened 2020-07-27
Redirect module pages to /index.html if they're missing a trailing slash

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/803

r? @Kixiron

### DIFF
--- a/src/storage/s3.rs
+++ b/src/storage/s3.rs
@@ -10,8 +10,8 @@ use log::warn;
 use rusoto_core::{region::Region, RusotoError};
 use rusoto_credential::DefaultCredentialsProvider;
 use rusoto_s3::{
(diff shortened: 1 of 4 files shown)


jyn514:
This should probably wait until after https://github.com/rust-lang/docs.rs/pull/911 so I don't have to write two different tests (and to avoid having to rebase that PR again).

src/storage/s3.rs
```diff
@@ -35,6 +35,20 @@ impl S3Backend {
         }
     }
 
+    pub(super) fn exists(&self, path: &str) -> Result<bool, Error> {
+        S3_RUNTIME.handle().block_on(async {
+            let resp = self
+                .client
+                .list_objects_v2(ListObjectsV2Request {
```
pietroalbini, line 42: A list request is ~10x more expensive money wise than a get request.

jyn514: I changed it to use HEAD instead.

PR #916, opened 2020-07-27
Add link to the Rust Cookbook to the topbar

### LINKED ISSUE #137: Adding the Rust Cookbook to the "Rust" dropdown
I think that the [`rust-cookbook`](https://rust-lang-nursery.github.io/rust-cookbook/) should also be added in the `Rust` dropdown(where you have links to the reference manual, std...).
It probably falls in the same ballpark as `Rust by Example`

belak:
Fixes #137

Screen Shot 2020-07-26 at 8 03 14 PM

### DIFF
--- a/templates/header/topbar.html
+++ b/templates/header/topbar.html
@@ -53,10 +53,17 @@
                                 </a>
                             </li>
 
-                            <li class="pure-menu-item menu-item-divided">
+                            <li class="pure-menu-item">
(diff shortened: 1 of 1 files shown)


jyn514:
The drop-down is getting a little big, maybe eventually we could offload some of these links to rust-lang.org/learn and only link to pages that are harder to find, like rustc-dev-guide. Or maybe have the windows approach with submenus.

This PR looks great though, thanks so much!

Issue #158, opened 2017-10-05
Remove crate version from docs.rs


sunjay:
Is there a way to remove certain versions of a crate from docs.rs?

onur:
Currently there isn't a way to remove crates from docs.rs. I am planning to add a method to check yanked crates and show a warning or remove the crate completely from docs.rs.

For now, I can manually remove the crate if there is an urgency.

sunjay:
Could you remove the yanked versions of the turtle crate? That's currently all the versions. 

Thanks

pravic:
+ futures 0.2.x have been yanked as well, see #221

jyn514:
We've since redirected the docs for yanked crates for the latest version. It would be nice to add a badge saying that the crate has been yanked e.g. on https://docs.rs/crate/futures/0.2.3-docs-yank.4.

jyn514:
Oops, I think that might have been done by `futures` in their docs actually. We have a long way to go for handling yanks :sweat_smile: 

See also #112, #322 , #561

Nemo157:
@sunjay with #614 fixed so that the yanked status of 0.1.* is updated, do you still want the docs fully removed? The old versions show warnings that they are yanked and suggest redirecting to the prereleases.

sunjay:
> @sunjay with #614 fixed so that the yanked status of 0.1.* is updated, do you still want the docs fully removed? 

Yes, that would be great. Those versions were for a different crate that used to use that name. Removing them will avoid some confusion so please do that. Thanks!

jyn514:
We have a way to delete entire crates (`delete-crate`), but it sounds like you're asking for specific versions to be removed which is a little harder.

sunjay:
Yes. I only want the 0.x versions removed. Back when this issue was created those were the only versions. :sweat_smile:

pietroalbini:
I'm wondering if we should just hide the documentation of all yanked crates.

jyn514:
Currently what we do is we don't show them unless you explicitly go to the link, and all the links to it have a big warning sign in addition to the 'this crate has been yanked' banner.

![image](https://user-images.githubusercontent.com/23638587/81220725-ca0b2c00-8faf-11ea-9e56-da430fc6593e.png)

![image](https://user-images.githubusercontent.com/23638587/81220798-e909be00-8faf-11ea-8035-ec25ab73f856.png)


[...]

PR #917, opened 2020-07-28
Update owner details more fully during build and add CLI command to trigger it

### LINKED ISSUE #900: `delete version` doesn't remove the owner if there are no more releases with that owner
See for example https://docs.rs/crate/pioneer/0.0.1 which still says it was authored by `@polygraphy`. This may need https://github.com/rust-lang/docs.rs/issues/766 to fix.

_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/issues/837#issuecomment-660153609_

Nemo157:
First two commits improve handling of owner details during a build, if an owners mutable details have been changed since the last time we got them they will be updated in the database, and if an owner has been removed from a crate the relationship will be removed.

Next three commits are cherry-picked from #898, they standardise the prefix handling into the config so that it's easier to correctly load the registry details outside of the `DocBuilder`.

Final commit adds a CLI command to update the details for a specific crate from the registry API, so that we can solve requests like #158/#837 directly. New (renewed) and existing crates will be fixed automatically when they publish a new version, and doing a mass update of existing bad data should be part of #766, so this should be a temporary command just to solve the problem for users that request it.

fixes #900

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use cratesfyi::db::{self, add_path_into_database, Pool};
+use cratesfyi::index::Index;
 use cratesfyi::utils::{remove_crate_priority, set_crate_priority};
 use cratesfyi::{
     BuildQueue, Config, DocBuilder, DocBuilderOptions, RustwideBuilder, Server, Storage,
@@ -417,6 +418,12 @@ enum DatabaseSubcommand {
     /// Updates github stats for crates.
     UpdateGithubFields,
 
+    /// Updates info for a crate from the registry's API
+    UpdateCrateRegistryFields {
+        #[structopt(name = "CRATE")]
+        name: String,
+    },
+
     AddDirectory {
         /// Path of file or directory
         #[structopt(name = "DIRECTORY")]
@@ -452,6 +459,16 @@ impl DatabaseSubcommand {
             Self::UpdateGithubFields => {
                 cratesfyi::utils::GithubUpdater::new(&*ctx.config()?, ctx.pool()?)?
                     .update_all_crates()?;
+            }
+
+            Self::UpdateCrateRegistryFields { name } => {
+                let index = Index::new(&ctx.config()?.registry_index_path)?;
+
+                db::update_crate_data_in_database(
+                    &*ctx.conn()?,
+                    &name,
+                    &index.api().get_crate_data(&name),
+                )?;
             }
 
             Self::AddDirectory { directory, prefix } => {
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -467,7 +467,7 @@ impl DatabaseSubcommand {
                 db::update_crate_data_in_database(
                     &*ctx.conn()?,
                     &name,
-                    &index.api().get_crate_data(&name),
+                    &index.api().get_crate_data(&name)?,
                 )?;
             }
 
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -281,7 +281,9 @@ struct Build {
 impl Build {
     pub fn handle_args(self, ctx: Context) -> Result<(), Error> {
         let docbuilder = {
-            let mut doc_options = DocBuilderOptions::new(&*ctx.config()?);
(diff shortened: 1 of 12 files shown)


src/db/add_package.rs
```diff
@@ -328,31 +327,103 @@ fn add_authors_into_database(
...
+                &[&owner.login, &owner.avatar, &owner.name, &owner.email],
+            )?
+            .get(0)
+            .get(0)
         };
 
         // add relationship
-        let _ = conn.query(
+        let updated = conn.query(
             "INSERT INTO owner_rels (cid, oid) VALUES ($1, $2)",
```
jyn514, line 392: Maybe this should use `ON CONFLICT DO NOTHING`? Not sure if that would let you still debug that it was pre-existing though.

src/db/add_package.rs
```diff
@@ -328,31 +327,103 @@ fn add_authors_into_database(
...
+        FROM owners
+        INNER JOIN owner_rels
+            ON owner_rels.oid = owners.id
+        WHERE owner_rels.cid = $1
+    ",
+            &[&crate_id],
+        )?
+        .into_iter()
+        .map(|row| row.get(0))
+        .collect();
```
jyn514, line 365: You only use this later as an iterator, so the `collect()` isn't necessary.
```suggestion
        .map(|row| row.get(0));
```

src/db/add_package.rs
```diff
@@ -328,31 +327,103 @@ fn add_authors_into_database(
...
+            Err(e) => return Err(e.into()),
+        }
     }
+
+    let to_remove = existing_owners
+        .iter()
+        .filter(|login| !owners.iter().any(|owner| &&owner.login == login));
+
+    for login in to_remove {
+        debug!("Removing owner relationship {}", login);
```
jyn514, line 413: In https://github.com/rust-lang/docs.rs/pull/917/commits/de38887296bd4e6799ce98cf1093aa195aee67c9#diff-9297ac45637dfecac9d9fb58922d4609R66-R69 we default to no owners if an error happens. But here we remove any owners that aren't in the new list. So we could end up deleting all owners by accident. Instead, can we return a `Result` from `get_crate_owners` and don't do the update if there's an error?

src/test/fakes.rs
```diff
@@ -167,6 +167,11 @@ impl<'a> FakeRelease<'a> {
         self.source_file("README.md", content.as_bytes())
     }
 
+    pub(crate) fn add_owner(mut self, owner: CrateOwner) -> Self {
+        self.registry_crate_data.owners.push(owner);
```
jyn514, line 171: Hmm ... now I wonder if `owners` should be a set instead. I guess it doesn't really matter either way since we're using UPSERT.

src/web/crate_details.rs
```diff
@@ -610,4 +611,95 @@ mod tests {
...
+                })
+                .create()?;
+
+            let details = CrateDetails::new(&db.conn(), "foo", "0.0.1").unwrap();
+            assert_eq!(
+                details.owners,
+                vec![("barfoo".into(), "https://example.org/barfoo".into())]
+            );
+
+            // Changing owner details on another of there crates applies too
```
jyn514, line 684:
```suggestion
            // Changing owner details on another of their crates applies the change to both
```
and thank you for the test!

src/bin/cratesfyi.rs
```diff
@@ -454,6 +461,16 @@ impl DatabaseSubcommand {
...
             }
 
+            Self::UpdateCrateRegistryFields { name } => {
+                let index = Index::new(&ctx.config()?.registry_index_path)?;
+
+                db::update_crate_data_in_database(
+                    &*ctx.conn()?,
+                    &name,
+                    &index.api().get_crate_data(&name),
+                )?;
```
jyn514, line 471: So, this is ok for now because it's just one or two crates. But when we automate this for the full consistency check, it should use the database dump to avoid spamming the crates.io API. @pietroalbini will not be happy if we make 300k requests in a day 😆

src/db/add_package.rs
Nemo157: Hmm, maybe with `RETURNING cid` it will either return one or zero rows, I'll try it.

Nemo157: Didn't see a way to get the debug message, but it doesn't matter too much.

PR #926, opened 2020-08-01
Fix `owner_rels.cid` foreign key reference

Nemo157:
Also improves testing to catch attempts to use incorrect foreign keys.

### DIFF
--- a/src/db/delete.rs
+++ b/src/db/delete.rs
@@ -49,9 +49,8 @@ fn get_id(conn: &Connection, name: &str) -> Result<i32, Error> {
 
 // metaprogramming!
 // WARNING: these must be hard-coded and NEVER user input.
-const METADATA: [(&str, &str); 5] = [
+const METADATA: &[(&str, &str)] = &[
     ("author_rels", "rid"),
-    ("owner_rels", "cid"),
     ("keyword_rels", "rid"),
     ("builds", "rid"),
     ("compression_rels", "release"),
@@ -60,7 +59,7 @@ const METADATA: [(&str, &str); 5] = [
 fn delete_version_from_database(conn: &Connection, name: &str, version: &str) -> Result<(), Error> {
     let crate_id = get_id(conn, name)?;
     let transaction = conn.transaction()?;
-    for &(table, column) in &METADATA {
+    for &(table, column) in METADATA {
         transaction.execute(
             &format!("DELETE FROM {} WHERE {} IN (SELECT id FROM releases WHERE crate_id = $1 AND version = $2)", table, column),
             &[&crate_id, &version],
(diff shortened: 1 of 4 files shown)


Nemo157:
I also checked the code, everywhere uses `owner_rels.cid` to refer to `crates`, not `releases`.

src/test/mod.rs
```diff
@@ -239,6 +239,29 @@ impl TestDatabase {
...
+        // Move all sequence start positions 10000 apart to avoid overlapping primary keys
+        let query: String = conn
+            .query(
+                "
+                    SELECT relname
+                    FROM pg_class
+                    INNER JOIN pg_namespace
+                        ON pg_class.relnamespace = pg_namespace.oid
+                    WHERE pg_class.relkind = 'S'
+                        AND pg_namespace.nspname = $1
```
jyn514, line 251: Dark magics 😨

src/db/migrate.rs
```diff
@@ -364,6 +364,22 @@ pub fn migrate(version: Option<Version>, conn: &Connection) -> CratesfyiResult<(
...
+        migration!(
+            context,
+            // version
+            15,
+            // description
+            "Fix owner_rels.cid foreign key reference",
+            // upgrade query
+            "
+            ALTER TABLE owner_rels DROP CONSTRAINT owner_rels_cid_fkey;
+            ALTER TABLE owner_rels ADD FOREIGN KEY (cid) REFERENCES crates(id);
```
jyn514, line 376: So if I understand this correctly:

- The code uses `cid` to mean `crate_id` everywhere.
- The database foreign key referenced the `releases.id` column, which didn't match the usage.
- The reason we didn't catch it until now is because both are strictly increasing, and there are always more releases than crates. So the foreign constraint held even though the semantic meaning was incorrect.
- This changes the foreign key constraint but nothing else, because all the joins are already done on `owners INNER JOIN crates ON owners.cid = crates.id`, which is the correct behavior.

However I don't think the last point is entirely true: When I wrote `delete_crate` I thought this was actually the release id. So I think that `delete_crate` is actually very broken right now and needs to be fixed as part of this change: https://github.com/rust-lang/docs.rs/blob/7ab83067e3519e30d3ca3214c49b8bbddceca34c/src/db/delete.rs#L65 and https://github.com/rust-lang/docs.rs/blob/7ab83067e3519e30d3ca3214c49b8bbddceca34c/src/db/delete.rs#L102

cc @pietroalbini: **do not delete any more crates** until this is fixed.

src/db/delete.rs
```diff
@@ -105,6 +104,7 @@ fn delete_crate_from_database(conn: &Connection, name: &str, crate_id: i32) -> R
             &[&crate_id],
         )?;
     }
+    transaction.execute("DELETE FROM owner_rels WHERE cid = $1;", &[&crate_id])?;
```
jyn514, line 107: I was going to say you only do this for deleting a crate, not a version, then I realized that's the right behavior and it was just buggy before 😅

src/db/migrate.rs
Nemo157: Yep, that all sounds correct, I've fixed it in the latest commit. I don't see any easy way to really test the behaviour.

jyn514: Fortunately this part of the code is (since #917) self-correcting so the worst that can happen is that the owner is missing until the next publish. Thanks for working on this :)

Issue #607, opened 2020-02-13
The source view is not available when crate failed to build.


pftbest:
It should be possible to view the source code of a crate even if it failed to compile for some reason. I think some time ago this was working correctly, but I'm not sure.

It seems docs.rs is the only website on the internet which allows you to read the source code of a crate, because github link is often missing and even if it's not it contains a different source code than an actual crate version. It is very unfortunate that you are not able to see the source just because it failed to compile on docs.rs platform.

Example links:
https://docs.rs/crate/libui-sys/0.1.9/source/
https://docs.rs/crate/stm32ral/0.3.1/source/

jyn514:
I think this should be as simple as moving this block outside of the `if res.result.successful` block: https://github.com/rust-lang/docs.rs/blob/master/src/docbuilder/rustwide_builder.rs#L313

jyn514:
Testing instructions: Once you've gotten set up (either with [docker-compose](https://github.com/rust-lang/docs.rs#development) or [natively](https://github.com/rust-lang/docs.rs/wiki/Developing-without-docker-compose)), run a build for a crate that you know will fail to build. You can choose any crate off of https://docs.rs/releases/recent-failures. 

```
# for docker-compose
docker-compose run web build crate <name> <version>
# running locally
cargo run build crate <name> <version>
```

You can then go to http://localhost:3000, click on the crate name, and click 'Source' in the top-left. It should show a list of all the files in the crate, including a `Cargo.toml`.

I expect this to also add sources for binaries when fixed, if you could test that as well that would be great, for example with `bat 0.13.0`.

pftbest:
The same problem with binary crates. They all lost the source code too. Examples: cargo-bloat, ripgrep.

pftbest:
Now even [log](https://docs.rs/crate/log/0.4.8/source/) crate lost its source code. It seems like soon there will be no source left on docs.rs :(

jyn514:
That seems like a different bug, log 0.4.8 built successfully.

jyn514:
This would be fixed by https://github.com/rust-lang/docs.rs/issues/826.

jyn514:
> Now even [log](https://docs.rs/crate/log/0.4.8/source/) crate lost its source code. It seems like soon there will be no source left on docs.rs :(

This was fixed by https://github.com/rust-lang/docs.rs/pull/830.

pftbest:
Thank you!

Issue #520, opened 2019-12-13
"Source" tab is shown even when no source is present


pietroalbini:
https://github.com/rust-lang/docs.rs/pull/519 fixed a panic happening when visiting a crate's source page when no source was available (due to a build failure), by showing a 404 page instead.

Still, the link to the source tab is unconditionally present in the navbar, even when we know it will lead to a 404 page. We should fix this by hiding the link in those cases.

jyn514:
This is now much less of an issue due to https://github.com/rust-lang/docs.rs/pull/928 (since there will always be source present from now on). However there will still be 404s for crates built before that PR.

PR #928, opened 2020-08-02
Upload sources even for binaries and failed builds

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/607. Makes https://github.com/rust-lang/docs.rs/issues/520 much less of an issue (since it will only happen for crates built before this PR).

Tested with:
- notcurses 1.6.10, a library that failed to build
- bat 0.12.1, a binary
- hexponent 0.3.0, a library that built successfully

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -351,12 +351,6 @@ impl RustwideBuilder {
                     other_targets,
                 } = metadata.targets();
 
-                // Store the sources even if the build fails
-                debug!("adding sources into database");
-                let prefix = format!("sources/{}/{}", name, version);
-                let (files_list, mut algs) =
-                    add_path_into_database(&self.storage, &prefix, build.host_source_dir())?;
-
                 // Perform an initial build
                 let res = self.execute_build(default_target, true, &build, &limits, &metadata)?;
                 if res.result.successful {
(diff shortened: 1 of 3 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -343,26 +343,23 @@ impl RustwideBuilder {
...
                     default_target,
                     other_targets,
                 } = metadata.targets();
 
-                // Do an initial build and then copy the sources in the database
+                // Store the sources even if the build fails
+                debug!("adding sources into database");
+                let prefix = format!("sources/{}/{}", name, version);
+                let (files_list, mut algs) =
+                    add_path_into_database(&self.storage, &prefix, build.host_source_dir())?;
```
Nemo157, line 358: It feels like this should still be done down around `add_package_into_database` until we proactively insert the release details into the database before building, otherwise in cases where we completely fail and don't insert the release details there will be uploaded files unassociated with any release.

jyn514: Hmm, that seems like a different refactor? The error case you're talking about makes sense, but if I move uploading the sources later we'll still upload all the documentation if an error occurs.

jyn514: And actually I don't see a way to have `add_package` come first because we record whether or not the build succeeded. So many we should the steps of building the docs and uploading the docs? Have one come before update the DB and the other come after? That definitely seems like a larger PR though.

Nemo157: It seems like the `execute_build` call is the most likely to fail, so specifically doing the upload before there has the highest chance of getting uploaded files unassociated with a release.

jyn514: Updated!

PR #930, opened 2020-08-02
Switch from kuchiki to LOL_HTML

### LINKED ISSUE #876: Switch from html5ever to lol_html
See https://github.com/kuchiki-rs/kuchiki/pull/74#issuecomment-653184255 for discussion. In a nutshell, kuchiki is not intended for low-memory usage: it uses lots of Rc and RefCell (one for each node in the tree!) LOL HTML is intended for exactly our use case:

> Low Output Latency streaming HTML rewriter/parser with CSS-selector based API.

This would allow us to step the size of files rendered _way_ up, possibly removing the limit altogether (https://github.com/rust-lang/docs.rs/pull/834).

jyn514:
Closes #876

My styles haven't been working since ~forever, if someone else could check this looks right that would be great. However this looks approximately the same before and after to me.

Ideas for testing this are also welcome - the previous tests were only testing the parsing, not the rendered code, so they're not very useful here.

Before:

image

After:

image

r? @Kixiron 
cc @nataliescottdavidson


jyn514:
Source before: https://gist.github.com/jyn514/421101e65a28bb67720f0fd31628b4c6
Source after: https://gist.github.com/jyn514/08ee9831ad5cb853cb528adb6aa2ea37

jyn514:
Going to try this on the 300 MB file and see what happens.

jyn514:
Once https://github.com/cloudflare/lol-html/issues/56 is fixed, we could also remove `SizedBuffer` and use [`MemorySettings`](https://docs.rs/lol_html/0.2.0/lol_html/struct.MemorySettings.html) instead (technically we could do it earlier but I'm lazy). That can wait for a follow-up PR though.

jyn514:
> Going to try this on the 300 MB file and see what happens.

It worked 🎉 🎉 🎉 Took about a minute to serve then another 5 to load but that's to be expected :)

jyn514:
It used ~350 MB of memory to serve which seems about right.

src/web/rustdoc.rs
```diff
@@ -377,18 +353,33 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+        tera::Context::from_serialize(RustdocPage {
+            latest_path,
+            latest_version,
+            inner_path,
+            is_latest_version,
+            krate,
+        })
+    );
+    // Extract the head and body of the rustdoc file so that we can insert it into our own html
+    let html = ctry!(req, utils::rewrite_lol(file_content, ctx, templates));
```
Kixiron, line 375: I think using a [`HtmlRewriter`](https://docs.rs/lol_html/0.2.0/lol_html/struct.HtmlRewriter.html) would be better here for two reasons, first it means we can use `MemorySettings` since `rewrite_str` currently doesn't and second it means that we can avoid parsing the file's content as utf8 since `HtmlRewriter` [takes in bytes](https://docs.rs/lol_html/0.2.0/lol_html/struct.HtmlRewriter.html#method.write)

src/web/rustdoc.rs
```diff
@@ -377,18 +353,33 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+
+    rendering_time.step("serve html");
+    use iron::{headers::ContentType, status::Status};
+    let mut response = Response::with((Status::Ok, html));
+    response.headers.set(ContentType::html());
+
+    Ok(response)
```
Kixiron, line 382: Could most of this be abstracted out into a struct, I think it'd make the code much cleaner

jyn514: If you mean `impl_webpage!`, I don't see how. `WebPage` runs templating _after_ everything else, but this runs it _before_. So the only way to make it work is to have every `impl WebPage` take a custom callback, which seems like overkill if it's only going to be used here.

jyn514:
> makes me wonder if we shouldn't add a time limit of some sort later on down the line to prevent us getting swamped (alternatively having a higher size cap would probably work, letting us not even try to parse files that are over x MB)

For now I left the cap at 5 MB. I planned to make a follow-up PR removing `SizedBuffer` in favor of memory limits, but since you mentioned it I'll make it part of this one.

>  You may need to update the benchmarks, and I'd be really interested to know if there's a speedup of any sort after the switch

Good point, I forgot to update them. The benchmarks are not going to be one-to-one though - before we only measured the parse time, while now it measures both parsing and templating (and that's sort of a fundamental part of this change, there's not a way to do one without the other).

src/web/rustdoc.rs
Kixiron: Nah, not an `impl_webpage!`, that'd require reworking how the rendering works. I just mean a straight-up normal struct to make things a little cleaner. (As a side note, the imports from `iron` can probably just go at the top of the page)

jyn514: I can make it a separate function if you think that would make it easier to follow. I don't see a point in having a struct that we make exactly once just to immediately call a method on.

jyn514: err actually I say that but this would look really nice on `RustdocPage` :laughing:

src/web/rustdoc.rs
jyn514: @Kixiron what size do you think should we use here? Previously we were guesstimating based on the file size, but this lets us configure the actual memory usage of the parser.

I'm going to go with 350 MB for now and we can change it later.

src/web/rustdoc.rs
jyn514: Done.

src/web/rustdoc.rs
Kixiron: Yah, that's something you or Pietro would know better based off of the available hardware. iirc the limits are set somewhere within our `Config` struct, if not then adding env vars would be a good path I'd think

src/utils/html.rs
```diff
@@ -1,62 +1,79 @@
...
 
-/// Extracts the contents of the `<head>` and `<body>` tags from an HTML document, as well as the
-/// classes on the `<body>` tag, if any.
-pub fn extract_head_and_body(html: &str) -> Result<(String, String, String)> {
-    let dom = kuchiki::parse_html().one(html);
+pub(crate) fn rewrite_lol(
+    html: &[u8],
+    ctx: Context,
+    templates: &TemplateData,
+) -> Result<Vec<u8>, RewritingError> {
```
Kixiron, line 9: Maybe document what the return value *means*, that the `Vec<u8>` is the fully rendered HTML ready to be served

src/utils/html.rs
```diff
@@ -1,62 +1,79 @@
...
-        .attributes
-        .borrow()
-        .get("class")
-        .map(|v| v.to_owned())
-        .unwrap_or_default();
+    let head_handler = |head: &mut Element| {
+        head.append(&tera_head, ContentType::Html);
+        Ok(())
+    };
+    // Before: <body> ... rustdoc content ... </body>
```
Kixiron, line 21:
```suggestion
    };

    // Before: <body> ... rustdoc content ... </body>
```

src/web/rustdoc.rs
```diff
@@ -598,7 +592,7 @@ mod test {
     ) -> Result<Option<String>, failure::Error> {
         assert_success(path, web)?;
         let data = web.get(path).send()?.text()?;
-        println!("{}", data);
+        println!("fetched path {} and got content {}", path, data);
```
Kixiron, line 595: Wouldn't a log be better here?
```suggestion
        log::info!("fetched path {} and got content {}", path, data);
```

jyn514: This is in a test, so it doesn't make a big deal either way. But I can change it to `log` so it has the line numbers.

jyn514:
@Kixiron how set are you on having benchmarks? So far I have
```rust
    let html = std::fs::read_to_string("benches/struct.CaptureMatches.html").unwrap();
    let page = RustdocPage {

    };
    let config = unimplemented!();
    let conn = Connection::connect(config.database_url.as_str(), postgres::TlsMode::None)?;
    let ctx = Context::from_serialize(&page).unwrap();
    let templates = TemplateData::new(conn).unwrap();
```
and it shows no signs of slowing down, I just need more and more things to be public.

src/utils/html.rs
jyn514: Done, what do you think?

Kixiron:
Benches don't matter a ton, I can follow-up with them if need be

src/web/rustdoc.rs
jyn514: Those are limits for the file size, not for the memory used to parse it. I'll add an environment variable.

src/utils/html.rs
Kixiron: Looks good, maybe add a note that this should/will be valid UTF-8, but we just haven't checked ourselves

jyn514:
I also raised the default HTML size limit from 5 MB to 50 MB. Since I tested on a 300 MB file this shouldn't give the prod server any problems.

inikulin:
@jyn514 If understand the context correctly then 5Mb should be enough. Lol-html requires to allocate memory only for really long tags (without content, just start tag tokens). So if you don't have/don't plan to have start tags longer than 5Mb than it should be more than enough for the observable future. (just in case: increasing the buffer size doesn't boost performance)

jyn514:
> @jyn514 If understand the context correctly then 5Mb should be enough. Lol-html requires to allocate memory only for really long tags (without content, just start tag tokens). So if you don't have/don't plan to have start tags longer than 5Mb than it should be more than enough for the observable future. (just in case: increasing the buffer size doesn't boost performance)

Wait, did I read that right? Lol-html will only allocate as much memory as the size of the start tag, even for multi-megabyte files? That's amazing!

inikulin:
@jyn514 yep, it's a streaming parser, so we don't allocate as less as possible

jyn514:
Switched the limit to 5 MB and it still loaded the 300 MB file fine 🎉

jyn514:
Hmm, not as much of an improvement in response times as I hoped (deploy is at 15:07):

image

I expect the memory usage will go down a lot though, I'll post back in a few hours once there's more data.

jyn514:
Aha, I was looking at the wrong screen! Those are some _nice_ numbers :D

image
image
image

inikulin:
@jyn514 btw, I believe you can push the memory limit even further to 3-5Kb, I believe in your case that should be enough

PR #936, opened 2020-08-04
Log HTML rewriting errors

Kixiron:
Logs any HTML rewriting errors to the console using `log::error!()` and increments a metric for failed html rewrites. Could be useful to see if we ever reach a point where we need to raise our memory limit

### DIFF
--- a/src/utils/html.rs
+++ b/src/utils/html.rs
@@ -23,6 +23,7 @@ pub(crate) fn rewrite_lol(
 
     let head_handler = |head: &mut Element| {
         head.append(&tera_head, ContentType::Html);
+
         Ok(())
     };
 
@@ -77,11 +78,14 @@ pub(crate) fn rewrite_lol(
 
     // The input and output are always strings, we just use `&[u8]` so we only have to validate once.
     let mut buffer = Vec::new();
+    // TODO: Make the rewriter persistent?
     let mut writer = HtmlRewriter::try_new(settings, |bytes: &[u8]| {
         buffer.extend_from_slice(bytes);
     })
     .expect("utf8 is a valid encoding");
+
     writer.write(html)?;
     writer.end()?;
+
     Ok(buffer)
 }
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -14,6 +14,7 @@ use iron::{
     modifiers::Redirect,
     status, Handler, IronError, IronResult, Request, Response, Url,
 };
+use lol_html::errors::RewritingError;
 use router::Router;
 use serde::Serialize;
 
@@ -197,6 +198,7 @@ impl RustdocPage {
         rustdoc_html: &[u8],
         max_parse_memory: usize,
         req: &mut Request,
+        file_path: &str,
     ) -> IronResult<Response> {
         use iron::{headers::ContentType, status::Status};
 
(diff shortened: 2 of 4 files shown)


src/utils/html.rs
```diff
@@ -77,11 +80,33 @@ pub(crate) fn rewrite_lol(
...
+
+/// Logs OOM errors encountered while rewriting a html file
+fn log_memory_errors<T>(
+    result: Result<T, RewritingError>,
+    file_path: &Path,
+) -> Result<T, RewritingError> {
+    match result {
+        err @ Err(RewritingError::MemoryLimitExceeded(..)) => {
+            crate::web::metrics::HTML_REWRITE_OOMS.inc();
+            log::error!(
```
Nemo157, line 103: Rather than adding an extra log message here, does chaining the message on via `.context` result in a decent error message from the standard logging?

Nemo157: (though I guess that doesn't help with the metrics)

Kixiron: It'd probably have roughly the same effect, but this makes sure that an error is always printed, regardless of if the error from the function is caught later on or not

jyn514:
A few points:
1. I don't expect to ever hit the memory limit in practice. From what @inikulin said in https://github.com/rust-lang/docs.rs/pull/930#issuecomment-668707731, it seems like LOL uses basically no additional memory usage compared to storing the file in memory (❤️). However I'm ok with having this log just so we know that for a fact.
2. The log should be in the web module, not in the rewrite itself, i.e. it should `match` on the value instead of moving globals into the rewriter. Otherwise we're making it even harder to eventually add benchmarks.
3. This reasoning seems really strange to me:
> It'd probably have roughly the same effect, but this makes sure that an error is always printed, regardless of if the error from the function is caught later on or not
Why should rewriting be special? If we want logging for requests that go wrong, we should add that directly instead of special-casing it here.

Kixiron:
Addressed comments

src/web/rustdoc.rs
```diff
@@ -206,12 +208,23 @@ impl RustdocPage {
...
+            Err(RewritingError::MemoryLimitExceeded(..)) => {
+                crate::web::metrics::HTML_REWRITE_OOMS.inc();
+
+                let config = extension!(req, Config);
+                let err = failure::err_msg(format!(
+                    "Failed to serve the rustdoc file '{}' because rewriting it surpassed the memory limit of {} bytes",
+                    file_path, config.max_parse_memory,
+                ));
+
+                ctry!(req, Err(err))
```
jyn514, line 224: This is basically the same as the `MemoryLimitExceededErrror` [Display impl](https://docs.rs/lol_html/0.2.0/src/lol_html/memory/limiter.rs.html#11), can you use `.context()` instead? That also keeps the `failure` backtrace.

```suggestion
                let err = err.context(format!(
                    "failed to serve the rustdoc file '{}' with memory limit of {} bytes",
                    file_path, max_parse_memory,
                ));

                ctry!(req, Err(err))
```

Kixiron: I originally tried this but got the error of
```rust
error[E0599]: no method named `context` found for enum `std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>` in the current scope
   --> src\web\rustdoc.rs:219:31
    |
219 |                 let err = err.context(format!(
    |                               ^^^^^^^ method not found in `std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>`
    |
   ::: C:\Users\Chase\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\src\libcore\result.rs:247:1
    |
247 | pub enum Result<T, E> {
    | --------------------- doesn't satisfy `_: failure::result_ext::ResultExt<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>`
    |
   ::: C:\Users\Chase\.cargo\registry\src\github.com-1ecc6299db9ec823\lol_html-0.2.0\src\rewriter\mod.rs:60:1
    |
60  | pub enum RewritingError {
    | ----------------------- doesn't satisfy `lol_html::rewriter::RewritingError: failure::Fail`
    |
   ::: C:\Users\Chase\.cargo\registry\src\github.com-1ecc6299db9ec823\quick-error-1.2.3\src\lib.rs:983:8
    |
983 |     fn context<X>(self, x: X) -> Result<T, Context<X, E>>;
    |        -------
    |        |
    |        the method is available for `std::boxed::Box<std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>>` here
    |        the method is available for `std::sync::Arc<std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>>` here
    |        the method is available for `std::rc::Rc<std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>>` here
    |
    = note: the method `context` exists but the following trait bounds were not satisfied:
            `lol_html::rewriter::RewritingError: failure::Fail`
            which is required by `std::result::Result<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>: failure::result_ext::ResultExt<std::vec::Vec<u8>, lol_html::rewriter::RewritingError>`
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use quick_error::ResultExt;`

warning: unused import: `failure::ResultExt`
   --> src\web\rustdoc.rs:218:21
    |
218 |                 use failure::ResultExt;
    |                     ^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default
```

jyn514: Ah right, this is https://github.com/cloudflare/lol-html/issues/54. Don't worry about it then, we can add it later.

PR #937, opened 2020-08-04
Cleanup from LOL HTML rewrite

jyn514:
- Add missing index.js
- Add missing CSS files
- Add Cargo.lock

r? @Kixiron 

I don't know how Cargo.lock slipped though, that seems like a regression from #883.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,6 +71,11 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
+version = "0.1.7"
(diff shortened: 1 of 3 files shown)


Nemo157:
Lockfile issue is from rebasing #930 over master, there was no git-detected conflict, but there is some sort of semantic conflict between the lockfiles that required updating it (would be avoided by using `bors` for testing the merged changes instead of the pre-merge changes :wink:).

jyn514:
Ah, that makes sense. CI is still working fine then.

Issue #934, opened 2020-08-04
Provide a way to copy dependency string from docs.rs


inikulin:
Usually, before adding a dependency, I first open docs.rs and check crate documentation. Then, to add the latest version I need to open crates.io from dropdown menu and only then copy `Cargo.toml` entry. These doesn't sound like a big deal, but then you repeat this process over and over again it may become bothersome. 

Probably there are hidden shortcuts that I'm missing, but it would be great if docs.rs allowed to copy a `Cargo.toml` entry for the current version like crates.io does.

Nemo157:
(I'm agnostic on this proposal itself, but I highly recommend the [`cargo-edit`](https://crates.io/crates/cargo-edit) set of utilities so that you can just `cargo add foo bar baz` and avoid having to manually edit `Cargo.toml` for dependencies at all).

inikulin:
@Nemo157 true, but it's still easier copy-paste a string into Cargo.toml. Also, since it's not a part of the official cargo distribution, very small fraction of people is aware of `cargo-edit`.

jyn514:
This already exists: 
![image](https://user-images.githubusercontent.com/23638587/89295144-8b2c5700-d62e-11ea-848e-d567da5c076d.png)

Maybe it's hard to find because it's on /crate, not on the documentation pages? I'm not sure where we'd put it on the rustdoc page, there's not a lot of space.

jyn514:
Another alternative people have suggested is making that whole top tag show up even for the documentation, but then it's taking up a lot of screen real estate when you're trying to read the docs themselves.

inikulin:
@jyn514 Agh, didn't realise it's there! That's handy. Though, it's more a UI design issue then - it's not obvious that's it's a "copy to clipboard" button (though maybe it's just me).

inikulin:
Oh, it's only on the "Source" tab, it would be cool to have this button in this dropdown too:
![image](https://user-images.githubusercontent.com/453071/89315033-47f3d780-d672-11ea-88cd-f2f03fbe2db3.png)

PR #938, opened 2020-08-04
Allow copying title and version from rustdoc pages

jyn514:
Same as https://github.com/rust-lang/docs.rs/pull/619 but for rustdoc pages.
Closes https://github.com/rust-lang/docs.rs/issues/934.

@inikulin does this look like about what you want? I know you mentioned the old icon was not very intuitive, I changed it to 'fa-copy' instead.

/crate:
 image

rustdoc pages:
new-clipboard

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,6 +71,11 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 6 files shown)


inikulin:
Yes, this is great, thank you!

PR #939, opened 2020-08-05
Add doc coverage information

GuillaumeGomez:
This PR adds the usage of the `--show-coverage` option of rustdoc. It currently looks like this:

Screenshot from 2020-08-07 00-08-51

Screenshot from 2020-08-07 00-08-42

A few things I think remains to be done (to be debated for some of them I guess?):
* Add tests
* Add this score in crates search results?
* Add color code (like if it's 100% it's green and the closer to 0, the whiter?)


GuillaumeGomez:
Not sure to understand the issue with clippy here:

```
error: in an `if` condition, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
##[error]   --> src/docbuilder/rustwide_builder.rs:550:43
    |
550 |               .process_lines(&mut |line, _| {
    |  ___________________________________________^
551 | |                 if line.starts_with('{') && line.ends_with('}') {
552 | |                     json = line.to_owned();
553 | |                 }
554 | |             })
    | |_____________^
    |
    = note: `-D clippy::blocks-in-if-conditions` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#blocks_in_if_conditions
```

Nemo157:
Looks like a false-positive from the lint.

GuillaumeGomez:
And I think I forgot to add something in the migration tests, but not sure what exactly...

src/db/migrate.rs
```diff
@@ -381,6 +381,25 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
+            "
+            CREATE TABLE doc_coverage (
+                release_id INT NOT NULL REFERENCES releases(id),
+                total_items INT,
+                documented_items INT
+            );
+            ",
+            // downgrade query
+            "
+            -- None.
```
Nemo157, line 400: This needs to drop the table, the migration downgrade fails because migration 1 removes the `releases` table, leaving the foreign key here dangling.

src/docbuilder/rustwide_builder.rs
```diff
@@ -469,6 +476,106 @@ impl RustwideBuilder {
...
+            .env("DOCS_RS", "1")
+            .args(&cargo_args)
+            .log_output(false)
+            .process_lines(&mut |line, _| {
+                if line.starts_with('{') && line.ends_with('}') {
+                    json = line.to_owned();
+                }
+            })
+            .run()
+            .is_ok()
```
Nemo157, line 556: Oh :grin: the clippy lint is sort of correct, just off on what it should be spanning, this whole expression should be pulled out to a separate variable and just `if build_result.is_ok()` or something here.

src/db/add_package.rs
```diff
@@ -130,6 +130,22 @@ pub(crate) fn add_package_into_database(
...
+pub(crate) fn add_doc_coverage(
+    conn: &mut Client,
+    release_id: i32,
+    total_items: i32,
+    documented_items: i32,
+) -> Result<i32> {
+    debug!("Adding doc coverage into database");
+    let rows = conn.query(
+        "INSERT INTO doc_coverage (release_id, total_items, documented_items)
+            VALUES ($1, $2, $3)
```
Nemo157, line 142: This should have an `ON CONFLICT UPDATE` for cases where we re-build the docs for an existing release.

src/docbuilder/rustwide_builder.rs
Nemo157: (though, at a glance it seems like there's a lot of duplication between here and the actual build, is it possible to pull that out to a function that prepares most of the command the same?)

src/db/add_package.rs
GuillaumeGomez: Ok!

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: I'll do that in a second commit then.

GuillaumeGomez:
Updated!

src/docbuilder/rustwide_builder.rs
```diff
@@ -535,50 +516,181 @@ impl RustwideBuilder {
...
+                            Ok(Some((total, documented)))
+                        }
+                        _ => Ok(None),
+                    }
+                } else {
+                    Ok(None)
+                }
+            },
+        )
+    }
```
Kixiron, line 597: This should fix the clippy lint
```suggestion
    fn get_coverage(
        &self,
        target: &str,
        build: &Build,
        metadata: &Metadata,
        limits: &Limits,
    ) -> Result<Option<(i32, i32)>> {
        self.make_build_object(
            target,
            metadata,
            limits,
            |cargo_args, mut rustdoc_flags, _| {
                rustdoc_flags.extend(vec![
                    "--output-format".to_string(),
                    "json".to_string(),
                    "--show-coverage".to_string(),
                ]);

                let mut doc_coverage_json = None;
                build
                    .cargo()
                    .timeout(Some(limits.timeout()))
                    .no_output_timeout(None)
                    .env(
                        "RUSTFLAGS",
                        metadata
                            .rustc_args
                            .as_ref()
                            .map(|args| args.join(" "))
                            .unwrap_or_default(),
                    )
                    .env("RUSTDOCFLAGS", rustdoc_flags.join(" "))
                    // For docs.rs detection from build script:
                    // https://github.com/rust-lang/docs.rs/issues/147
                    .env("DOCS_RS", "1")
                    .args(&cargo_args)
                    .log_output(false)
                    .process_lines(&mut |line, _| {
                        if line.starts_with('{') && line.ends_with('}') {
                            json = Some(line.to_owned());
                        }
                    })
                    .run()
                    .ok();

                if let Some(json) = doc_coverage_json {
                    if let Ok(Value::Object(m)) = serde_json::from_str(&json) {
                        let (mut total, mut documented) = (0, 0);

                        for entry in m.values() {
                            if let Some(Value::Number(n)) = entry.get("total") {
                                total += n.as_i64().unwrap_or(0) as i32;
                            }

                            if let Some(Value::Number(n)) = entry.get("with_docs") {
                                documented += n.as_i64().unwrap_or(0) as i32;
                            }
                        }

                        return Ok(Some((total, documented)));
                    }
                }

                Ok(None)
            },
        )
    }
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -535,50 +516,181 @@ impl RustwideBuilder {
...
+                                .map(|args| args.join(" "))
+                                .unwrap_or_default(),
+                        )
+                        .env("RUSTDOCFLAGS", rustdoc_flags.join(" "))
+                        // For docs.rs detection from build script:
+                        // https://github.com/rust-lang/docs.rs/issues/147
+                        .env("DOCS_RS", "1")
+                        .args(&cargo_args)
+                        .run()
+                        .is_ok()
```
Kixiron, line 655: This same sequence comes up a lot, could you pull it into a method on `DocBuilder`?

src/docbuilder/rustwide_builder.rs
```diff
@@ -631,4 +743,7 @@ pub(crate) struct BuildResult {
     pub(crate) docsrs_version: String,
     pub(crate) build_log: String,
     pub(crate) successful: bool,
+    // The two next fields are used for the doc coverage.
+    pub(crate) total_items: Option<i32>,
+    pub(crate) documented_items: Option<i32>,
```
Kixiron, line 748:
```suggestion
    /// The total items that could be documented in the current crate, used to calculate documentation coverage
    pub(crate) total_items: Option<i32>,
    /// The items of the crate that are documented, used to calculate documentation coverage
    pub(crate) documented_items: Option<i32>,
```

src/web/crate_details.rs
```diff
@@ -180,6 +184,16 @@ impl CrateDetails {
...
+            doc_coverage: {
+                let total: Option<i32> = krate.get("total_items");
+                let documented: Option<i32> = krate.get("documented_items");
+                match (total, documented) {
+                    (Some(total), Some(documented)) => {
+                        Some(documented as f32 * 100. / total as f32)
+                    }
+                    _ => None,
+                }
+            },
```
Kixiron, line 196: Maybe pull this into a separate variable

src/web/crate_details.rs
```diff
@@ -44,6 +44,7 @@ pub struct CrateDetails {
     pub(crate) doc_targets: Vec<String>,
     license: Option<String>,
     documentation_url: Option<String>,
+    doc_coverage: Option<f32>,
```
Kixiron, line 47: This could actually be calculated on the template side of things from `total_items` and `documented_items` so we could not only have the coverage percentage but also a `100 out of 200 items documented` line in the coverage display

GuillaumeGomez: Sure!

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: The sequence is quite small and can only be found in two places. Is it really worth it considering that even though they look similar, they have differences?

GuillaumeGomez:
Updated! I updated the screenshot as well.

src/docbuilder/rustwide_builder.rs
Kixiron: I guess it doesn't matter a ton, but some of the common stuff could be (like setting the env vars) so that there's only one place to change

jyn514: I would rather have this in one place so that people don't accidentally update it in only one place. This is the part of docs.rs that has the fewest tests so it should be as simple to modify as possible.

jyn514:
This says '50 out of 52 items documented' in the screenshot - does it have a subpage that shows which items are documented and which aren't? And if not, could you add that? Maybe `/crate/:crate/:version/coverage` or something, you could have it as one of the tabs after 'Builds'.

templates/crate/details.html
```diff
@@ -16,6 +16,13 @@
             <div class="pure-u-1 pure-u-sm-7-24 pure-u-md-5-24">
                 <div class="pure-menu package-menu">
                     <ul class="pure-menu-list">
+                        {%- if details.documented_items and details.total_items -%}
+                            {% set pourcent = details.documented_items * 100 / details.total_items %}
+                            <li class="pure-menu-heading">Documentation<br>coverage</li>
+                            <li class="pure-menu-item" style="text-align:center;"><b>{{ pourcent | round(precision=2) }} %</b><br>
```
jyn514, line 22:
```suggestion
                            {% set percent = details.documented_items * 100 / details.total_items %}
                            <li class="pure-menu-heading">Documentation<br>coverage</li>
                            <li class="pure-menu-item" style="text-align:center;"><b>{{ percent | round(precision=2) }} %</b><br>
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -428,6 +429,12 @@ impl RustwideBuilder {
                     algs,
                 )?;
 
+                if let (Some(total), Some(documented)) =
+                    (res.result.total_items, res.result.documented_items)
```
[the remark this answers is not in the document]
Kixiron: +1 on this, we need to sub-struct more often, some of these are incredibly unwieldy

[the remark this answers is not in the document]
Kixiron: +1 on this, we need to sub-struct more often, some of these are incredibly unwieldy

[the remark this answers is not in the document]
GuillaumeGomez: Absolutely, it's just that I find tuples harder to understand. But I can create a new struct if you prefer?

src/db/delete.rs
```diff
@@ -65,6 +66,12 @@ fn delete_version_from_database(conn: &mut Client, name: &str, version: &str) ->
             &[&crate_id, &version],
         )?;
     }
+    transaction.execute(
+        "DELETE FROM doc_coverage WHERE release_id = (
+            SELECT id FROM releases WHERE crate_id = $1 AND version = $2
+        )",
+        &[&crate_id, &version],
+    )?;
```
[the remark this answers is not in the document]
GuillaumeGomez: Oh, I already added it into `METADATA`. ^^'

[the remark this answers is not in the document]
GuillaumeGomez: Are you sure we're not supposed to put it into both?

src/docbuilder/rustwide_builder.rs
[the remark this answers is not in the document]
GuillaumeGomez: Arf, didn't see your comment @Kixiron. >< Then let's go for a struct!

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: Fair enough. I'll try to move code into a function.

templates/crate/details.html
```diff
@@ -16,6 +16,13 @@
             <div class="pure-u-1 pure-u-sm-7-24 pure-u-md-5-24">
                 <div class="pure-menu package-menu">
                     <ul class="pure-menu-list">
+                        {%- if details.documented_items and details.total_items -%}
+                            {% set percent = details.documented_items * 100 / details.total_items %}
+                            <li class="pure-menu-heading">Documentation<br>coverage</li>
```
jyn514, line 21: This makes it span 1 line instead of two, which makes it easier to read IMO. Plus the whole site is already about documentation so I don't think we have to repeat it everywhere.

```suggestion
                            <li class="pure-menu-heading">Coverage</li>
```

There should also be a corresponding section in `templates/rustdoc/body.html` - there's a lot of duplicated content between those two files I haven't gotten around to refactoring.

GuillaumeGomez:
Updated! I also added it into the "popup" as @jyn514 said. :) Screenshots updated.

src/docbuilder/rustwide_builder.rs
```diff
@@ -469,6 +537,50 @@ impl RustwideBuilder {
...
+        metadata: &Metadata,
+        limits: &Limits,
+    ) -> Result<Option<DocCoverage>> {
+        let mut doc_coverage_json = None;
+
+        let rustdoc_flags = vec![
+            "--output-format".to_string(),
+            "json".to_string(),
+            "--show-coverage".to_string(),
+        ];
```
jyn514, line 553: Shouldn't this also include the rustdoc flags in `metadata`? Otherwise it might be missing a `--cfg` which would have different output.

src/docbuilder/rustwide_builder.rs
```diff
@@ -68,6 +69,69 @@ const ESSENTIAL_FILES_UNVERSIONED: &[&str] = &[
 const DUMMY_CRATE_NAME: &str = "empty-library";
 const DUMMY_CRATE_VERSION: &str = "1.0.0";
 
+macro_rules! config_command {
```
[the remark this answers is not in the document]
GuillaumeGomez: Good idea!

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: Done!

src/docbuilder/rustwide_builder.rs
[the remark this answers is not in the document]
GuillaumeGomez: Done!

GuillaumeGomez:
Updated! :)

jyn514:
> This says '50 out of 52 items documented' in the screenshot - does it have a subpage that shows which items are documented and which aren't? And if not, could you add that? Maybe `/crate/:crate/:version/coverage` or something, you could have it as one of the tabs after 'Builds'.

I ran this locally and it turns out rustdoc doesn't output this info! Maybe at some point rustdoc could add this and you could make a follow-up PR? No need to block on it though.

```
{"src/fpformat.rs":{"total":2,"with_docs":2},"src/lib.rs":{"total":17,"with_docs":17}}
```

It might be nice to break down the coverage by file, though.

GuillaumeGomez:
I had something different in mind: just using this coverage as a "grade" and maybe link people to how they can improve it. I don't think that `docs.rs` should provide such information. After all, it sounds like it's something only the maintainer would be interested in, no?

jyn514:
Well, it's hard to improve it if you don't know what items are missing documentation 😆

>  it sounds like it's something only the maintainer would be interested in, no?

Sure, but docs.rs is just as much for crate maintainers as it is for end-users. This is the same reason we have the build logs instead of just 'this crate failed to build'.

GuillaumeGomez:
They can run the doc coverage themselves or even use warnings to see which items aren't documented. I actually don't know how to display such information. Also: it only shows modules, not items themselves. So not sure how useful it would be...

jyn514:
Ok, we can leave it for later if you're not sure how to implement it. I do think it would be nice to have so you don't have to mess around with unstable flags locally.

>  it only shows modules, not items themselves. So not sure how useful it would be...

Right, this is also a feature request for rustdoc :P I'll open a separate issue so we don't lose the discussion.

GuillaumeGomez:
> Ok, we can leave it for later if you're not sure how to implement it. I do think it would be nice to have so you don't have to mess around with unstable flags locally.

I think it'll take a little while before we can actually know for sure how it should look like and where to put it. I'm not convinced that docs.rs should provide such information directly though but if we want to do it, we'll have to think where to put this information. Maybe a new tab?

Anyway, what remains to be done for this PR?

src/docbuilder/rustwide_builder.rs
```diff
@@ -69,6 +69,14 @@ const ESSENTIAL_FILES_UNVERSIONED: &[&str] = &[
 const DUMMY_CRATE_NAME: &str = "empty-library";
 const DUMMY_CRATE_VERSION: &str = "1.0.0";
 
+// This macro exists because of lifetimes issues surrounding the `Command::process_lines` method:
+// it expects a mutable reference to a closure, in which we capture a variable in order to store
+// the JSON output. Unfortunately, we *need* to create the command in the same scope level otherwise
+// rustc becomes very grumpy about the fact that the variable needs to be static because it is
+// captured in a mutably referenced closure.
```
jyn514, line 76: Would this be fixed by taking a closure by value instead? If so, let's open an issue on https://github.com/rust-lang/rustwide/ and put a TODO here to remove the macro once it's fixed.

src/docbuilder/rustwide_builder.rs
```diff
@@ -109,6 +117,11 @@ macro_rules! config_command {
             "--cap-lints".to_string(),
             "warn".to_string(),
         ];
+
+        if let Some(package_rustdoc_args) = &$metadata.rustdoc_args {
+            rustdoc_flags.append(&mut package_rustdoc_args.iter().map(|s| s.to_owned()).collect());
```
jyn514, line 122:
```suggestion
            rustdoc_flags.append(&mut package_rustdoc_args.clone());
```

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: I honestly don't know but I would expect it to solve our issue.

GuillaumeGomez:
I fixed the nit and added a TODO. I can take a look at this issue after this PR is merged if you want?

jyn514:
That would be great!

GuillaumeGomez:
Ok, will do then!

jyn514:
> Add tests

Whoops, I forgot about this one. The builds will be hard to test (#822) but you can test the frontend using [kuchiki](https://docs.rs/kuchiki/0.8.0/kuchiki/). See https://github.com/rust-lang/docs.rs/blob/master/src/web/mod.rs#L665 for an example - you'll also have to update [`FakeRelease`](https://github.com/rust-lang/docs.rs/blob/master/src/test/fakes.rs#L33) to include the coverage.

GuillaumeGomez:
Ok! Thanks for the directions. :)

pietroalbini:
https://github.com/rust-lang/rustwide/pull/36 will fix the rustwide issue.

jyn514:
Blocked on rustwide publishing a new version.

pietroalbini:
Pushed two commits to this branch, bumping rustwide to 0.10.0 and removing the macro.

GuillaumeGomez:
:confetti_ball:

src/db/add_package.rs
```diff
@@ -130,6 +130,29 @@ pub(crate) fn add_package_into_database(
...
 }
 
+pub(crate) fn add_doc_coverage(
+    conn: &mut Client,
+    release_id: i32,
+    doc_coverage: DocCoverage,
+) -> Result<i32> {
+    debug!("Adding doc coverage into database");
+    let rows = conn.query(
+        "INSERT INTO doc_coverage (release_id, total_items, documented_items)
```
pietroalbini, line 140: Why is this a separate table?

GuillaumeGomez: We plan to add extra data. `doc-coverage` will be extended into the future and @jyn514 also wanted to include the current data (the by-module results). Since we didn't agree on how to display the information, we decided to do it in a later PR.

pietroalbini: Hmm ok.

sunjay:
Hey @GuillaumeGomez, thanks for taking the time to add this feature. :) 

I was wondering if we could move the Coverage section to a metrics page or something? I'm not sure if it makes sense to display this so prominently in the crate information dropdown. Documentation coverage is a great thing to know, but it doesn't indicate the quality of the crate or even the quality of its documentation. (It has the same downsides as using code coverage as the only metric for how good your tests are.)

I think having access to this information is good for library authors, but we should really find a proper place for it where we can make sure it's as useful as possible. If we have a separate page or tab or something, we can keep adding more things to it to help library authors while avoiding overpopulating the crate information dropdown.

I think it's great that you added it to the Crate tab as well as the dropdown. Maybe we could remove it from the dropdown for the timebeing while we figure out how to best go about displaying this information? :)

GuillaumeGomez:
Thanks for the feedback @sunjay!

For now, the feature itself is pretty limited, and we agree: in some cases, the doc coverage isn't an important information (pure FFI bindings for examples). However, for everything else, we've been trying in the documentation team to enforce people to document all items by adding lints (`missing_docs` and `missing_doc_code_examples`) to help crates' maintainer in this task. Putting this information forward is just the next step and I really think that it's an important information for anyone that is comparing multiple crates. Documentation quantity isn't an insurance for *good* documentation, but no documentation is also a good factor.

For now it's only about how many items are documented, but I intend to modify this value by also adding more checks like "how much doc comments are actually providing examples". So this is just a first step, and I am curious to see how it'll be used by users. The "hidden" goal being to force people to document more of course. :3

yaahc:
> The "hidden" goal being to force people to document more of course. :3

I agree with this goal but not the approach. This feels like negative re-enforcement to me and I think this will have negative impacts on maintainers emotionally when they're being guilt-tripped into adding docs. I'm not sure it should be removed but I think you should tread carefully here and be sure you've considered the emotional impact of this change.

GuillaumeGomez:
I wanted to add this information in the crate results directly, but for now I think it's too early for that and that we should let maintainers get to know how it works before putting everything in place. For now, I think the display is quite discrete: when browsing through docs.rs, it's "rare" to take a look at those places, which is why we think it's fine for the moment.

The goal isn't to shame anyone for their work. Everyone should be thankful for the incredible work of all rustaceans. We're just trying to help the crate maintainers by giving them tools to improve their crate even more, and to users to help them to have a crate that matches what they need better. Some crates are not meant to be used by "external" users, which is perfectly fine, but in this case, it's "normal" (still think it's a good thing to have documentation in any case, if not others at least for yourself :p) to have less documentation and it's a good indicator that this crate might not be the one you want.

However, once again: we didn't promote this display more specifically because we want to help people understand and adjust to the system. It'll also allow us to provide even more information and tools for both crate's owners and users.

pietroalbini:
Maybe we could make it less prominent?

GuillaumeGomez:
@pietroalbini Might be a bit hard to get what it's talking about, no? Not really sure... However, maybe we could move the documentation coverage block under "dependencies" and "versions" blocks. What do you think?

yaahc:
> For now, I think the display is quite discrete: when browsing through docs.rs, it's "rare" to take a look at those places, which is why we think it's fine for the moment.

I looked more carefully and you're right, I had mistakenly assumed it was in the sidebar on the landing page for the crate docs when you navigate to something like `docs.rs/foo`, so things are already not as bad as I'd assumed.

GuillaumeGomez:
We can add it later on, but for now (and the next months in fact...) it's not scheduled. Next step will be to provide a page to help crates' maintainers to understand how this score is computed.

pietroalbini:
> Might be a bit hard to get what it's talking about, no?

@GuillaumeGomez tried moving it at the bottom and making the text more clear:

GuillaumeGomez:
But still not keeping what the number is about? :p

pietroalbini:
I don't think the exact number of items documented is *that* important. We can make that field a link to the about page though, which already shows the detailed data.

GuillaumeGomez:
Saying what it is seems quite important from my point of view. But as you prefer...

pietroalbini:
Opened https://github.com/rust-lang/docs.rs/pull/961 with my proposed change

PR #943, opened 2020-08-06
Added tooltips to the About and Releases links

Kixiron:
Before/Unhovered:
image

After/Hovered:
image
image

The tooltip text isn't final, I just couldn't think of better ones.

Engaged in a *minor* CSS splitting, no actual refactoring happened since it was just moved to separate files and then `@include`'d

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,7 @@ fn main() {
 
     write_git_version();
     if let Err(sass_err) = compile_sass() {
-        panic!("Error compiling SASS: {}", sass_err);
+        panic!("Error compiling sass: {}", sass_err);
     }
     copy_js();
 }
--- a/build.rs
+++ b/build.rs
@@ -1,18 +1,25 @@
 use git2::Repository;
-use std::env;
-use std::fs::{self, File};
-use std::io::Write;
-use std::path::Path;
+use std::{
+    env,
+    error::Error,
+    fs::{self, File},
+    io::Write,
+    path::Path,
+};
 
 fn main() {
     // Set the host target
     println!(
         "cargo:rustc-env=CRATESFYI_HOST_TARGET={}",
-        env::var("TARGET").unwrap()
+        env::var("TARGET").unwrap(),
     );
 
     // Don't rerun anytime a single change is made
-    println!("cargo:rerun-if-changed=templates/style.scss");
+    println!("cargo:rerun-if-changed=templates/style/base.scss");
+    println!("cargo:rerun-if-changed=templates/style/_rustdoc.scss");
+    println!("cargo:rerun-if-changed=templates/style/_vars.scss");
+    println!("cargo:rerun-if-changed=templates/style/_utils.scss");
+    println!("cargo:rerun-if-changed=templates/style/_navbar.scss");
     println!("cargo:rerun-if-changed=templates/menu.js");
     println!("cargo:rerun-if-changed=templates/index.js");
     // TODO: are these right?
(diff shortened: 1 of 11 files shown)


jyn514:
So, this is an improvement over the current situation, but since we're putting in the time anyway I think it would be nice to have a dropdown the way we do for platforms and 'Rust'. Then you can go directly to the page you're looking for without having to click twice. How hard would that be to do?

jyn514:
Why is this a 350 line diff?

Kixiron:
Updated screenshots, the diff is because of the css splitting, sass allows you to split & include files with no loss of semantic meaning, it's part of a longer term plan to make our css more understandable and hackable. No actual *changing* of it was done

jyn514:
Ok, can you separate the split into a separate commit so it's easier to see what the changes in behavior are?

src/web/page/templates.rs
```diff
@@ -65,7 +64,9 @@ impl TemplateData {
 
             while rx.recv().is_ok() {
                 if let Err(err) = reload(&template_data, &pool) {
-                    log::error!("failed to reload templates: {:?}", err);
+                    log::error!("Failed to reload templates:\n{}", err);
+                } else {
+                    log::info!("Reloaded templates");
```
jyn514, line 69: Logging should be lower-case so it doesn't look odd after another prefixed message: https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err
```suggestion
                    log::error!("failed to reload templates: {}", err);
                } else {
                    log::info!("reloaded templates");
```

templates/header/topbar.html
```diff
@@ -25,26 +25,73 @@
...
 
                         <ul class="pure-menu-children">
-                            <a href="/about" class="pure-menu-link">
-                                Documentation about builds, badges and more
-                            </a>
+                            <li class="pure-menu-item">
+                                <a href="/about/badges" class="pure-menu-link">
+                                    Badges
+                                </a>
+                            </li>
```
jyn514, line 37: Can you use the `active_link` macro here?
```suggestion
                            {{ macros::active_link(expected="never", href="/about/badges", text="Badges") }}
```
Same for all the following links.

build.rs
```diff
@@ -1,26 +1,35 @@
...
     println!("cargo:rerun-if-changed=templates/index.js");
     // TODO: are these right?
     println!("cargo:rerun-if-changed=.git/HEAD");
     println!("cargo:rerun-if-changed=.git/index");
 
     write_git_version();
-    compile_sass();
+    if let Err(sass_err) = compile_sass() {
+        panic!("Error compiling sass: {}", sass_err);
+    }
```
jyn514, line 32:
```suggestion
    compile_sass().expect("error compiling sass");
```

Kixiron: I did this because it makes a better error. Using the `Display` impl allows the error produced to actually be shown in a readable form instead of `being\nlike\nthis\n`

jyn514: Ah right, this uses Display instead of Debug. I wish `expect` would use that by default :/ This change looks fine though.

templates/header/topbar.html
Kixiron: I *might* have gone overboard

templates/header/topbar.html
```diff
@@ -24,66 +24,80 @@
...
+                    <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
                         <a href="/releases" class="pure-menu-link">
                             Releases
                         </a>
+
+                        <ul class="pure-menu-children">
+                            {{ macros::menu_link(href="/releases/stars", text="Releases by Stars") }}
+                            {{ macros::menu_link(href="/releases/recent-failures", text="Recent Build Failures") }}
+                            {{ macros::menu_link(href="/releases/failures", text="Build Failures by Stars") }}
+                            {{ macros::menu_link(href="/releases/failures", text="Release Activity") }}
```
jyn514, line 49:
```suggestion
                            {{ macros::menu_link(href="/releases/activity", text="Release Activity") }}
```

Issue #945, opened 2020-08-07
Font rendering got weird recently


benesch:
docs.rs suddenly started rendering monospace fonts quite differently.

Here's the before:

<img width="1026" alt="image" src="https://user-images.githubusercontent.com/882976/89691710-9fdf4800-d8d7-11ea-9ed5-f0dd8e210c65.png">

And here's the after:

<img width="1009" alt="image" src="https://user-images.githubusercontent.com/882976/89691629-67d80500-d8d7-11ea-8a5b-435d1722a564.png">

Was this intentional? The only obviously glitch is the `fi` ligature (which is _weird_ for a monospace font; wonder why that's happening), but as a matter of personal preference I also preferred the smaller and lighter weight fonts in the before screenshot.

jyn514:
https://docs.rs/protobuf/2.16.2/protobuf/descriptor/struct.FileDescriptorSet.html

This is probably fallout from #930, it's not an intentional change. According to devtools this is using `font-face: monospace,monospace` so your default monospace font is mostly to blame, but it's still a bug on our end - we should be using rustdoc fonts.

jyn514:
I expect this is the same root cause as https://github.com/rust-lang/docs.rs/issues/935

jyn514:
Actually I think this is #937, the pure-css files are being loaded where they weren't before. Let me try reverting those locally and see if it helps.

jyn514:
Fixed by #946

benesch:
Ah yep, that fixed it right up. Thanks very much for the quick fix!

PR #946, opened 2020-08-07
Remove CSS files that shouldn't be on rustdoc pages

jyn514:
These should only have been on the homepage. Not sure if this actually fixes the problem, my styles continue to be messed up locally. I'm not sure what else it could be though.

Addresses https://github.com/rust-lang/docs.rs/issues/945

r? @Kixiron


jyn514:
Ok I checked locally and this does now have `font-family: "Source Serif Pro", Georgia, Times, "Times New Roman", serif`. No idea how it looks though, this is what I have locally:

image

Kixiron:
Seems reasonable

PR #947, opened 2020-08-08
Refactor queries to use the more semantic postgres::Client methods

Nemo157:


Nemo157:
(I'm sure touching almost every database query is unlikely to conflict with other merge requests 🤭)

jyn514:
I'm really uncomfortable making changes like this without https://github.com/rust-lang/docs.rs/issues/874. The new code is definitely better - but are we sure all of it is tested? Changes like this have been a major source of outages in the past.

This is the same reason I don't want to refactor the builder too much until we have https://github.com/rust-lang/docs.rs/issues/822.

Nemo157:
(I'm planning to go through this and cherry-pick out all the changes that are covered in existing tests, just got distracted from finishing it on Sunday).

Nemo157:
(superseded by moving to sqlx soon)

PR #951, opened 2020-08-08
Update `cargo run` commands

### LINKED ISSUE #950: Errors following _Getting Started_ instructions
I  tried following [Getting Started instructions](https://github.com/rust-lang/docs.rs#getting-started).

```bash
git clone https://github.com/rust-lang/docs.rs.git docs.rs
cd docs.rs
# Configure the default settings for external services
cp .env.sample .env
# Builds the docs.rs binary
cargo build
# Start the extenal services
docker-compose up -d db s3
# Build a sample crate to make sure it works
# This sets up the docs.rs build environment, including installing the nightly
# Rust toolchain. This will take a while the first time but will be cached afterwards.
cargo run web build crate regex 1.3.1
# This starts the web server but does not build any crates.
cargo run start-web-server
```

But I'm getting the following failure:

```bash
     Running `target/debug/cratesfyi web build crate regex 1.3.1`
error: Found argument 'web' which wasn't expected, or isn't valid in this context
 
USAGE:
    cratesfyi <SUBCOMMAND>
 
For more information try --help
```

Full log is attached.
[run.log](https://github.com/rust-lang/docs.rs/files/5046153/run.log)

Nemo157:
Consistently use `--` to separate arguments to `cratesfyi`, remove `web` subcommand that is only relevant for `docker-compose` commands.

fixes #950

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ See `cargo run -- --help` for a full list of commands.
 
 ```sh
 # This command will start web interface of docs.rs on http://localhost:3000
-cargo run start-web-server
+cargo run -- start-web-server
 ```
 
 #### `build` subcommand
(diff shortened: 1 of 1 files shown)


jyn514:
Why'd you add the `--`? They look a lot uglier IMO and they're not required.

Nemo157:
I find it much less likely to cause issues when you do want to pass flags to `cargo` if you just consistently use them when `cargo run`ning something.

Nemo157:
Also, it only works when you pass a subcommand rather than an argument, `cargo run --help` will give the help for `cargo run`.

PR #953, opened 2020-08-09
Add the Context trait and move Metrics out of the global state 

pietroalbini:
This PR does three things:

* Creates a `Context` trait, allowing to pass multiple "components" of docs.rs around without adding countless function params all around the codebase.
* Adds a new `Metrics` struct which replaces the lazy statics. This has the side effect of making `cargo test` work when running it in parallel :tada:, and allows testing that metrics are properly recorded.
* Adds a couple metrics tests around the codebase.

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -7,7 +7,8 @@ use cratesfyi::db::{self, add_path_into_database, Pool};
 use cratesfyi::index::Index;
 use cratesfyi::utils::{remove_crate_priority, set_crate_priority};
 use cratesfyi::{
-    BuildQueue, Config, Context, DocBuilder, DocBuilderOptions, RustwideBuilder, Server, Storage,
+    BuildQueue, Config, Context, DocBuilder, DocBuilderOptions, Metrics, RustwideBuilder, Server,
+    Storage,
 };
 use failure::{err_msg, Error, ResultExt};
 use once_cell::sync::OnceCell;
@@ -316,7 +317,8 @@ impl BuildSubcommand {
             Self::World => {
                 docbuilder.load_cache().context("Failed to load cache")?;
 
-                let mut builder = RustwideBuilder::init(ctx.pool()?, ctx.storage()?)?;
+                let mut builder =
+                    RustwideBuilder::init(ctx.pool()?, ctx.metrics()?, ctx.storage()?)?;
                 builder
                     .build_world(&mut docbuilder)
                     .context("Failed to build world")?;
@@ -330,8 +332,9 @@ impl BuildSubcommand {
                 local,
             } => {
                 docbuilder.load_cache().context("Failed to load cache")?;
-                let mut builder = RustwideBuilder::init(ctx.pool()?, ctx.storage()?)
-                    .context("failed to initialize rustwide")?;
+                let mut builder =
+                    RustwideBuilder::init(ctx.pool()?, ctx.metrics()?, ctx.storage()?)
+                        .context("failed to initialize rustwide")?;
 
                 if let Some(path) = local {
                     builder
@@ -367,14 +370,16 @@ impl BuildSubcommand {
                     }
                 }
 
-                let mut builder = RustwideBuilder::init(ctx.pool()?, ctx.storage()?)?;
+                let mut builder =
+                    RustwideBuilder::init(ctx.pool()?, ctx.metrics()?, ctx.storage()?)?;
                 builder
                     .update_toolchain()
                     .context("failed to update toolchain")?;
             }
 
             Self::AddEssentialFiles => {
-                let mut builder = RustwideBuilder::init(ctx.pool()?, ctx.storage()?)?;
+                let mut builder =
+                    RustwideBuilder::init(ctx.pool()?, ctx.metrics()?, ctx.storage()?)?;
                 builder
                     .add_essential_files()
                     .context("failed to add essential files")?;
@@ -546,6 +551,7 @@ struct BinContext {
     storage: OnceCell<Arc<Storage>>,
     config: OnceCell<Arc<Config>>,
     pool: OnceCell<Pool>,
+    metrics: OnceCell<Arc<Metrics>>,
 }
 
 impl BinContext {
@@ -555,6 +561,7 @@ impl BinContext {
             storage: OnceCell::new(),
             config: OnceCell::new(),
             pool: OnceCell::new(),
+            metrics: OnceCell::new(),
         }
     }
 
@@ -573,7 +580,11 @@ impl Context for BinContext {
         Ok(self
             .build_queue
             .get_or_try_init::<_, Error>(|| {
-                Ok(Arc::new(BuildQueue::new(self.pool()?, &*self.config()?)))
+                Ok(Arc::new(BuildQueue::new(
+                    self.pool()?,
+                    self.metrics()?,
+                    &*self.config()?,
+                )))
             })?
             .clone())
     }
@@ -582,7 +593,11 @@ impl Context for BinContext {
         Ok(self
             .storage
             .get_or_try_init::<_, Error>(|| {
-                Ok(Arc::new(Storage::new(self.pool()?, &*self.config()?)?))
+                Ok(Arc::new(Storage::new(
+                    self.pool()?,
+                    self.metrics()?,
+                    &*self.config()?,
+                )?))
             })?
             .clone())
     }
@@ -597,7 +612,14 @@ impl Context for BinContext {
     fn pool(&self) -> Result<Pool, Error> {
         Ok(self
             .pool
-            .get_or_try_init::<_, Error>(|| Ok(Pool::new(&*self.config()?)?))?
+            .get_or_try_init::<_, Error>(|| Ok(Pool::new(&*self.config()?, self.metrics()?)?))?
+            .clone())
+    }
+
+    fn metrics(&self) -> Result<Arc<Metrics>, Error> {
+        Ok(self
+            .metrics
+            .get_or_try_init::<_, Error>(|| Ok(Arc::new(Metrics::new()?)))?
             .clone())
     }
 }
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -116,14 +116,7 @@ impl CommandLine {
                 socket_addr,
                 reload_templates,
             } => {
-                Server::start(
-                    Some(&socket_addr),
-                    reload_templates,
-                    ctx.pool()?,
-                    ctx.config()?,
-                    ctx.build_queue()?,
-                    ctx.storage()?,
-                )?;
+                Server::start(Some(&socket_addr), reload_templates, &ctx)?;
             }
             Self::Daemon {
                 foreground,
@@ -134,6 +127,7 @@ impl CommandLine {
                 }
 
                 cratesfyi::utils::start_daemon(
+                    &ctx,
                     ctx.config()?,
                     ctx.pool()?,
                     ctx.build_queue()?,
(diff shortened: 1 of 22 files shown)


src/web/mod.rs
```diff
@@ -403,14 +403,6 @@ impl Server {
         template_data: Arc<TemplateData>,
         context: &dyn Context,
     ) -> Result<Self, Error> {
-        // poke all the metrics counters to instantiate and register them
-        metrics::TOTAL_BUILDS.inc_by(0);
-        metrics::SUCCESSFUL_BUILDS.inc_by(0);
-        metrics::FAILED_BUILDS.inc_by(0);
-        metrics::NON_LIBRARY_BUILDS.inc_by(0);
-        metrics::UPLOADED_FILES_TOTAL.inc_by(0);
-        metrics::FAILED_DB_CONNECTIONS.inc_by(0);
```
jyn514, line 434: Why don't we need these any more? (and why did we need them in the first place?)

src/metrics/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// Number of files uploaded to the storage backend
+        uploaded_files_total: IntCounter,
+
+        /// The number of attempted files that failed due to a memory limit
+        html_rewrite_ooms: IntCounter,
+    }
+
+    metrics visibility: pub(crate),
+    namespace: "docsrs",
```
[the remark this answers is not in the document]
Kixiron: Namespace is for the prometheus namespace I assume, but I think the visibility could be better done with a visibility capture on each field, a la
```rust
    (
        $vis:vis struct $name:ident {
            $(#[doc = $help:expr] $metric_vis:vis $metric:ident: $ty:ty $([$($label:expr),*])?,)*
        }
        namespace: $namespace:expr,
    ) => {
        $vis struct $name {
            registry: Registry,
            $($metric_vis $metric: $ty,)*
        }
        ...
    }
```

src/metrics/macros.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[macro_export]
+macro_rules! metrics {
+    (
+        $vis:vis struct $name:ident {
+            $(#[doc = $help:expr] $metric:ident: $ty:ty $([$($label:expr),*])?,)*
+        }
+        metrics visibility: $metric_vis:vis,
+        namespace: $namespace:expr,
+    ) => {
```
Kixiron, line 13:
```suggestion
    (
        $vis:vis struct $name:ident {
            $(
                #[doc = $help:expr]
                $metric:ident: $ty:ty $([$($label:expr),* $(,)?])?
            ),* $(,)?
        }
        metrics visibility: $metric_vis:vis,
        namespace: $namespace:expr,
    ) => {
```

src/metrics/macros.rs
```diff
@@ -0,0 +1,65 @@
...
+        $vis struct $name {
+            registry: Registry,
+            $($metric_vis $metric: $ty,)*
+        }
```
Kixiron, line 17:
```suggestion
    (
        $vis:vis struct $name:ident {
            $(#[doc = $help:expr]
            $metric_vis:vis $metric:ident: $ty:ty $([$($label:expr),*])?,)*
        }
        namespace: $namespace:expr,
    ) => {
        $vis struct $name {
            registry: Registry,
            $($metric_vis $metric: $ty,)*
        }
```

src/web/mod.rs
Kixiron: I think what they're for is to poke the metrics so their `Lazy`s are initialized

src/test/mod.rs
```diff
@@ -314,22 +312,10 @@ pub(crate) struct TestFrontend {
 }
 
 impl TestFrontend {
-    fn new(
-        db: &TestDatabase,
-        config: Arc<Config>,
-        build_queue: Arc<BuildQueue>,
-        storage: Arc<Storage>,
-    ) -> Self {
+    fn new(context: &dyn Context) -> Self {
```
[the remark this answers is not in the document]
pietroalbini: I think generics are overkill for this: the context is never stored anywhere, it's just passed around for the initialization, so we don't care about the performance of accessing it. Passing `&dyn` makes the code more simple and readable.

src/utils/daemon.rs
```diff
@@ -106,8 +90,6 @@ pub fn start_daemon(
         },
     )?;
 
-    // TODO: update ssl certificate every 3 months
-
```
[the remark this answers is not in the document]
pietroalbini: Yes.

src/metrics/mod.rs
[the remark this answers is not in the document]
pietroalbini: The namespace is a prefix for all the metrics: a metric named `foo` with a namespace of `docsrs` produces the `docsrs_foo` metric. I might just rename that to `metric names prefix:`.

src/web/mod.rs
pietroalbini: Yeah, we don't need them anymore.

src/bin/cratesfyi.rs
```diff
@@ -571,6 +571,17 @@ impl Context {
         }
     }
 
+    fn conn(
+        &self,
+    ) -> Result<
+        r2d2::PooledConnection<r2d2_postgres::PostgresConnectionManager<postgres::NoTls>>,
+        Error,
+    > {
+        Ok(self.pool()?.get()?)
```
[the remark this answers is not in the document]
pietroalbini: Nope, `get()` returns a different error type.

src/metrics/mod.rs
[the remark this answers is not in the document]
pietroalbini: Added a comment explaining what the namespace is.

pietroalbini:

src/bin/cratesfyi.rs
```diff
@@ -554,28 +546,40 @@ enum DeleteSubcommand {
     },
 }
 
-struct Context {
+struct BinContext {
     build_queue: OnceCell<Arc<BuildQueue>>,
     storage: OnceCell<Arc<Storage>>,
     config: OnceCell<Arc<Config>>,
     pool: OnceCell<Pool>,
+    metrics: OnceCell<Arc<Metrics>>,
```
Kixiron, line 554: `Metrics` has no cross-dependencies so could be a `Lazy` instead of a `OnceCell`

pietroalbini: I'd prefer to keep `OnceCell` for consistency sake with the rest of the others. Using `OnceCell` is also not as painful here as it is everywhere else, since we only access the underlying `OnceCell` in only one place.

jyn514:
LGTM after a rebase

pietroalbini:
Rebased.

src/metrics/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        used_db_connections: IntGauge,
+        /// The maximum number of database connections
+        max_db_connections: IntGauge,
+        /// Number of attempted and failed connections to the database
+        pub(crate) failed_db_connections: IntCounter,
+
+        /// The number of currently opened file descriptors
+        open_file_descriptors: IntGauge,
+        /// The number of threads being used by docs.rs
+        running_threads: IntGauge,
```
Kixiron, line 36: Need to add `$(#[$meta:meta])*` captures to the macro so that these two metrics can be cfg'd out for `#[cfg(target_os = "linux")]`
```suggestion
        /// The number of currently opened file descriptors
        #[cfg(target_os = "linux")]
        open_file_descriptors: IntGauge,
        /// The number of threads being used by docs.rs
        #[cfg(target_os = "linux")]
        running_threads: IntGauge,
```

src/metrics/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        self.failed_crates_count.set(queue.failed_count()? as i64);
+
+        self.gather_system_performance();
+        Ok(self.registry.gather())
+    }
+
+    #[cfg(windows)]
+    fn gather_system_performance(&self) {}
+
+    #[cfg(not(windows))]
```
Kixiron, line 90:
```suggestion
    #[cfg(not(target_os = "linux"))]
    fn gather_system_performance(&self) {}

    #[cfg(target_os = "linux")]
```

src/metrics/macros.rs
```diff
@@ -0,0 +1,67 @@
...
+                $(
+                    let $metric = <$ty>::from_opts(
+                        prometheus::Opts::new(stringify!($metric), $help)
+                            .namespace($namespace)
+                            $(.variable_labels(vec![$($label.into()),*]))?
+                    )?;
+                    registry.register(Box::new($metric.clone()))?;
+                )*
+                Ok(Self { registry, $($metric,)* })
+            }
```
Kixiron, line 32:
```suggestion
                #[doc = $help:expr]
                $(#[$meta:meta])*
                $metric_vis:vis $metric:ident: $ty:ty $([$($label:expr),* $(,)?])?
            ),* $(,)?
        }
        namespace: $namespace:expr,
    ) => {
        $vis struct $name {
            registry: prometheus::Registry,
            $(
                $(#[$meta])*
                $metric_vis $metric: $ty,
            )*
        }
        impl $name {
            $vis fn new() -> Result<Self, prometheus::Error> {
                let registry = prometheus::Registry::new();
                $(
                    $(#[$meta])*
                    {
                        let $metric = <$ty>::from_opts(
                            prometheus::Opts::new(stringify!($metric), $help)
                                .namespace($namespace)
                                $(.variable_labels(vec![$($label.into()),*]))?
                        )?;
                        registry.register(Box::new($metric.clone()))?;
                    }
                )*
                Ok(Self { registry, $($metric,)* })
            }
```

src/metrics/mod.rs
pietroalbini: Why do you want to cfg out the metrics? We can just avoid setting them.

Kixiron: If they aren't cfg'd out then Windows/MacOS users will receive dead code warnings (and if isn't changed to `#[cfg(not(target_os = "linux"))]` then MacOS users won't be able to compile)

pietroalbini: Makes sense! `#[cfg(linux)]` also works btw.

src/metrics/macros.rs
pietroalbini: Note that this is not correct: you can't use a block here because then the `let $metric` is scoped only inside the block, and is not available in the constructor below. Applying the attribute on each statement works though.

pietroalbini:
Addressed the comment.

Kixiron:
I had the macro capture as
```rust
#[doc = $help:expr]
$(#[$meta:meta])*
```
So we wouldn't hit the "doc comment on non-top level item" lint, I guess you could just annotate the generated functions with allow's

src/metrics/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        /// The maximum number of database connections
+        max_db_connections: IntGauge,
+        /// Number of attempted and failed connections to the database
+        pub(crate) failed_db_connections: IntCounter,
+
+        /// The number of currently opened file descriptors
+        #[cfg(linux)]
+        open_file_descriptors: IntGauge,
+        /// The number of threads being used by docs.rs
+        #[cfg(linux)]
```
Kixiron, line 37:
```suggestion
        #[cfg(target_os = "linux")]
        open_file_descriptors: IntGauge,
        /// The number of threads being used by docs.rs
        #[cfg(target_os = "linux")]
```

src/metrics/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.failed_crates_count.set(queue.failed_count()? as i64);
+
+        self.gather_system_performance();
+        Ok(self.registry.gather())
+    }
+
+    #[cfg(not(linux))]
+    fn gather_system_performance(&self) {}
+
+    #[cfg(linux)]
```
Kixiron, line 92:
```suggestion
    #[cfg(not(target_os = "linux"))]
    fn gather_system_performance(&self) {}

    #[cfg(target_os = "linux")]
```

Issue #952, opened 2020-08-09
Document read-only and writable file system


schulzch:
[About/Builds](https://docs.rs/about/builds) does not mention that the sandbox file system is mostly read-only, except `OUT_DIR` (?). I think this should be documented, along with maximum writable disk space and best practices.

Several crates seem to have run into this issue already. Just to mention a few:
- https://github.com/rtbo/rust-xcb/issues/74
- https://github.com/avr-rust/avrd/issues/18
- https://github.com/rust-lang/docs.rs/issues/727
- https://github.com/rust-lang/docs.rs/issues/811

jyn514:
Mentoring instructions: near https://github.com/rust-lang/docs.rs/blob/master/templates/core/about/builds.html#L5, add a line about the source directory being read-only. It should probably also mention that this affects dependencies as well as the crate you published.

jyn514:
Documented in #954

PR #954, opened 2020-08-09
Rework About/Build

schulzch:
Rework About/Build in response to #952

- Splitted h4 headings into notes and diagnosis
- Use the buzzword "sandbox" more often
- Add section about read-only and writable locations

### DIFF
--- a/templates/core/about/builds.html
+++ b/templates/core/about/builds.html
@@ -3,77 +3,106 @@
 {%- block title -%} Builds {%- endblock title -%}
 
 {%- block body -%}
-	{%- set docsrs_repo = "https://github.com/rust-lang/docs.rs" -%}
-	<h1>Builds</h1>
-	<div class="container about">
-	<p>
-		Docs.rs automatically builds crates' documentation released on
-		<a href="https://crates.io/">crates.io</a>
-		using the nightly release of the Rust compiler.
-		Builds can take a while depending how many crates are in <a href="/releases/queue">the queue</a>.
-	</p>
+    {%- set docsrs_repo = "https://github.com/rust-lang/docs.rs" -%}
+    <h1>Builds</h1>
+    <div class="container about">
+    <p>
+        Docs.rs automatically builds documentation for crates released on <a href="https://crates.io/">crates.io</a>.
+         
+        It may take a while to build your crate, depending on how many crates are in <a href="/releases/queue">the queue</a>.
+    </p>
 
-	<p>
-		{%- if rustc_version %}
-		The current version of the Rust compiler in use is <code>{{ rustc_version }}</code>.
-		{%- endif -%}
-	</p>
+    <p>
+        All crates are built in a sandbox using the nightly release of the Rust compiler.
+        {%- if rustc_version %}
+        The current version in use is <code>{{ rustc_version }}</code>.
+        {%- endif -%}
+    </p>
 
-	<p>
-		The README of a crate is taken from the <code>readme</code> field defined in
-		<code>Cargo.toml</code>. If a crate doesn't have this field,
-		no README will be displayed.
-	</p>
+    <h3>Notes on using Docs.rs</h3>
 
-	<h3>Diagnosing a failed build</h3>
+    <h4>Setting a README</h4>
+    <p>
+        The README of a crate is taken from the <code>readme</code> field defined in
+        <code>Cargo.toml</code>. If this field is not set, no README will be displayed.
+    </p>
 
-	<h4>Missing dependencies</h4>
-	<p>
-		Missing dependencies are a common reason for a failed build.
-		Docs.rs dependencies are managed through
-		<a href="https://github.com/rust-lang/crates-build-env">crates-build-env</a>;
-		see <a href="https://forge.rust-lang.org/docs-rs/add-dependencies.html">Forge</a> for instructions
-		on how to make a PR. You can always <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>
-		if you're having trouble.
-	</p>
+    <h4>Detecting Docs.rs</h4>
+    <p>
+        To recognize Docs.rs from <code>build.rs</code> files, you can test for the environment variable <code>DOCS_RS</code>, e.g.:
+        {% filter dedent -%}
+        <pre>
+            <code>
+                if let Ok(_) = std::env::var("DOCS_RS") {
+                    // ... your code here ...
+                }        
+            </code>
+        </pre>
+        {%- endfilter %}
+        This approach can be helpful if you need dependencies for building the library, but not for building the documentation.
+    </p>
 
-	<h4>Detecting Docs.rs from <code>build.rs</code></h4>
-	<p>
-		Docs.rs builds crates with the environment variable <code>DOCS_RS</code> set to
-		<code>1</code>, which enables the crate to detect docs.rs and build the
-		crate differently. This can be helpful if you need
-		dependencies for building the library, but not for building the documentation.
-	</p>
+    <p>
+        You can configure how your crate is built by adding <a href="metadata">package metadata</a> to your <code>Cargo.toml</code>, e.g.:
+        {% filter dedent -%}
+        <pre>
+            <code>
+                [package.metadata.docs.rs]
+                rustc-args = ["--cfg", "docsrs"]
+            </code>
+        </pre>
+        {%- endfilter %}
+        Here, the compiler arguments are set so that <code>#[cfg(docsrs)]</code> (not to be confused with <code>#[cfg(doc)]</code>) can be used for conditional compilation.
+        This approach is also useful for setting <a href="https://doc.rust-lang.org/cargo/reference/features.html">cargo features</a>.
+    </p>
 
-	<h4>Detecting Docs.rs from <code>#[cfg]</code> attributes</h4>
-	<p>
-		You can detect Docs.rs by having a <a href="https://doc.rust-lang.org/cargo/reference/features.html">feature</a>
-		which is only set by Docs.rs. See <a href="metadata">Metadata</a> for more information.
-	</p>
+    <h4>Testing documentation builds locally</h4>
+    {%- set build_subcommand = docsrs_repo ~ "/blob/master/README.md#build-subcommand" -%}
+    <p>
+        The Docs.rs <a href="{{ build_subcommand | safe }}">README</a> describes how to build
+        unpublished crate documentation locally using the same build environment as the Docs.rs build agent.
+    </p>
 
-	<h4>Global sandbox limits</h4>
+    <h3>Diagnosing failed builds</h3>
 
-	<p>
-		All the builds on docs.rs are executed inside a sandbox with limited
-		resources. The current limits are the following:
-	</p>
+    <h4>Missing dependencies</h4>
+    <p>
+        Missing dependencies are a common reason for failed builds.
+        Docs.rs dependencies are managed through
+        <a href="https://github.com/rust-lang/crates-build-env">crates-build-env</a>.
+        See <a href="https://forge.rust-lang.org/docs-rs/add-dependencies.html">Forge</a> for how to add a dependency.
+    </p>
 
-	{{ macros::crate_limits(limits=limits) }}
+    <h4>Write attempt on read-only directories</h4>
+    <p>
+        Most of the sandbox is a read-only file system, including the source directory of your crate and its dependencies.
+        If your <code>build.rs</code> generates files that are relevant for documentation, consider writing to the
+        <a href="https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates">cargo output directory</a>,
+        passed in the environment variable <code>OUT_DIR</code>.
+    </p>
 
-	<p>
-		If a build fails because it hit one of those limits please
-		<a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
-		to get them increased for your crate.
-		Note that network access will not be enabled for any crate.
-	</p>
+    <h4>Hitting resource limits</h4>
+    <p>
+        All the builds are executed inside a sandbox with limited resources. The current limits are:
+    </p>
 
-	<h4>Test crate documentation build locally</h4>
-	{%- set build_subcommand = docsrs_repo ~ "/blob/master/README.md#build-subcommand" -%}
-	<p>
-		The <a href="{{ build_subcommand | safe }}">docs.rs README</a> describes how to build an
-		unpublished crate's documentation locally using the same build environment as the build agent.
-	</p>
-	</div>
+    {{ macros::crate_limits(limits=limits) }}
+
+    <p>
+        If your build fails because it hit one of these limits, please
+        <a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
+        to get them increased for your crate.
+        Since our build agent has finite resources, we have to consider each case individually.
+        However, network access will not be enabled for any crate.
+    </p>
+
+    <h4>Other failures and requesting rebuilds</h4>
+    <p>
+        If your crate fails to build for a reason not listed here, please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
+        Some build failures can be fixed by Docs.rs, e.g., by issuing a rebuild after a bug in Docs.rs has been fixed.
+    </p>
+
+    </div>
 {%- endblock body %}
 
 {% block css -%}


templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
...
 	<p>
-		You can detect Docs.rs by having a <a href="https://doc.rust-lang.org/cargo/reference/features.html">feature</a>
-		which is only set by Docs.rs. See <a href="metadata">Metadata</a> for more information.
+		Docs.rs sets a <a href="https://doc.rust-lang.org/cargo/reference/features.html">feature</a>. See <a href="metadata">Metadata</a> for details.
+	</p>
+
+	<h4>Testing documentation builds locally</h4>
+	{%- set build_subcommand = docsrs_repo ~ "/blob/master/README.md#build-subcommand" -%}
+	<p>
+		The Doc.rs <a href="{{ build_subcommand | safe }}">README</a> describes how to build
```
Kixiron, line 46:
```suggestion
		The Docs.rs <a href="{{ build_subcommand | safe }}">README</a> describes how to build
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
...
+		unpublished crate documentation locally using the same build environment as the Doc.rs build agent.
```
Kixiron, line 47:
```suggestion
		unpublished crate documentation locally using the same build environment as the Docs.rs build agent.
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
...
 	<p>
-		If a build fails because it hit one of those limits please
+		All the builds on Docs.rs are executed inside a sandbox with limited
+		resources. If a build fails because it hit limits please
 		<a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
 		to get them increased for your crate.
-		Note that network access will not be enabled for any crate.
+		Network access will not be enabled for any crate.
+		The current limits are:
 	</p>
```
Kixiron, line 77: Could you add a "Requesting Rebuilds" section that outlines how to request a rebuild and why you can/can't?
You can request a rebuild by opening an issue on our repo, but we can rebuild crates that fail due to something that is *our* fault, like a missing dependency or docs.rs bug. Crates that hit resource limits *might* be rebuilt if we have the resources to do so (since our builder itself has finite resources), but failures due to networking will not be rebuilt

templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
 	<h1>Builds</h1>
 	<div class="container about">
 	<p>
-		Docs.rs automatically builds crates' documentation released on
-		<a href="https://crates.io/">crates.io</a>
-		using the nightly release of the Rust compiler.
-		Builds can take a while depending how many crates are in <a href="/releases/queue">the queue</a>.
+		Docs.rs automatically builds documentation for crates released on
+		<a href="https://crates.io/">crates.io</a>.
```
Kixiron, line 11:
```suggestion
		Docs.rs automatically builds documentation for crates released on <a href="https://crates.io/">crates.io</a>.
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
...
+		 
+		It may take a while to build your crate depending on how many crates are in <a href="/releases/queue">the queue</a>.
 	</p>
 
 	<p>
+		All crates are build in a sandbox using the nightly release of the Rust compiler.
 		{%- if rustc_version %}
-		The current version of the Rust compiler in use is <code>{{ rustc_version }}</code>.
+		The current version in use is <code>{{ rustc_version }}</code>.
```
Kixiron, line 19:
```suggestion
		The current version of <code>rustc</code> in use is <code>{{ rustc_version }}</code>.
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,76 @@
...
 		{%- endif -%}
 	</p>
 
-	<p>
-		The README of a crate is taken from the <code>readme</code> field defined in
-		<code>Cargo.toml</code>. If a crate doesn't have this field,
-		no README will be displayed.
-	</p>
+	<h3>Notes on using Doc.rs</h3>
```
Kixiron, line 23:
```suggestion
	<h3>Notes on using Docs.rs</h3>
```

templates/core/about/builds.html
Kixiron: This may not be needed, @jyn514 I don't remember the format that `rustc_version` takes, either just a version & hash or the full thing

jyn514: You can see it on the website: 
image

templates/core/about/builds.html
jyn514: And if it needs a change to the source code a rebuild won't help (e.g. a dependency was fixed, but in a semver breaking bump).

templates/core/about/builds.html
schulzch: Currently the string emitted by `rustc_version` is "rustc 1.47.0-nightly (6c8927b0c 2020-07-26)". Thus adding another "rustc" would be redundant.

templates/core/about/builds.html
schulzch: Good idea. Will do so.

templates/core/about/builds.html
```diff
@@ -7,72 +7,75 @@
...
 
+	<h4>Hitting sandbox limits</h4>
 	<p>
-		If a build fails because it hit one of those limits please
+		All the builds on Docs.rs are executed inside a sandbox with limited
+		resources. If a build fails because it hit limits please
 		<a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
 		to get them increased for your crate.
-		Note that network access will not be enabled for any crate.
+		Network access will not be enabled for any crate.
```
jyn514, line 73:
```suggestion
		However, network access will not be enabled for any crate.
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,75 @@
...
+		See <a href="https://forge.rust-lang.org/docs-rs/add-dependencies.html">Forge</a> for instructions
+		on how to make a PR. You can always <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>
+		when having trouble.
 	</p>
 
-	{{ macros::crate_limits(limits=limits) }}
+	<h4>Writing files to read-only directories</h4>
+	<p>
+		Most of the sandbox is a read-only file system, including the source directory of your crate and its dependencies.
+		If you <code>build.rs</code> generates files that are relevant for documentation, consider writing to the output directory which is passed as the environment variable <code>OUT_DIR</code>.
```
jyn514, line 64:
```suggestion
		If your <code>build.rs</code> generates files that are relevant for documentation, consider writing to the <a href="https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates">cargo output directory</a>, passed in the environment variable <code>OUT_DIR</code>.
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,75 @@
...
+		Missing dependencies are a common reason for failed builds.
+		Docs.rs dependencies are managed through
+		<a href="https://github.com/rust-lang/crates-build-env">crates-build-env</a>.
+		See <a href="https://forge.rust-lang.org/docs-rs/add-dependencies.html">Forge</a> for instructions
+		on how to make a PR. You can always <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>
+		when having trouble.
 	</p>
 
-	{{ macros::crate_limits(limits=limits) }}
+	<h4>Writing files to read-only directories</h4>
```
jyn514, line 61: Crate authors are _not_ writing to the read-only directories ;)
```suggestion
	<h4>Read-only directories</h4>
```

templates/core/about/builds.html
```diff
@@ -7,72 +7,75 @@
...
+		Docs.rs sets the environment variable <code>DOCS_RS</code> to <code>1</code>.
+		Detecting Doc.rs and building your crate differently can be helpful if you need
 		dependencies for building the library, but not for building the documentation.
 	</p>
 
 	<h4>Detecting Docs.rs from <code>#[cfg]</code> attributes</h4>
 	<p>
-		You can detect Docs.rs by having a <a href="https://doc.rust-lang.org/cargo/reference/features.html">feature</a>
-		which is only set by Docs.rs. See <a href="metadata">Metadata</a> for more information.
+		Docs.rs sets a <a href="https://doc.rust-lang.org/cargo/reference/features.html">feature</a>. See <a href="metadata">Metadata</a> for details.
```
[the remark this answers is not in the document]
Kixiron: I think we should add a better snippet giving an example though, like
```toml
[package.metadata.docs.rs]
rustc-args = ["--cfg", "docsrs"]
```
for `#[cfg(docsrs)]`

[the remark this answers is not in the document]
jyn514: cfg isn't a great example though because you already have `cfg(doc)`. I would rather give the example `all-features = true`.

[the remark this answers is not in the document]
Kixiron: Now that I think about it, should probably have a "common metadata" section in the Metadata tab that gives common/useful configurations (separate PR though)

templates/core/about/builds.html
```diff
@@ -3,80 +3,92 @@
...
+        Network access will not be enabled for any crate.
+        The current limits are:
+    </p>
+
+    {{ macros::crate_limits(limits=limits) }}
+
+    <h4>Something else went wrong</h4>
+    <p>
+        Happens. Feel free to <a href="{{ docsrs_repo | safe }}/issues">file an issue</a> when having trouble.
+    </p>
```
jyn514, line 81: This section seems kind of odd. Don't we already say to open an issue above?

If we do keep it the tone should be more polite:
```suggestion
    <h4>Something else went wrong</h4>
    <p>
        Please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a> if you have trouble building a crate on docs.rs.
    </p>
```

templates/core/about/builds.html
```diff
@@ -3,80 +3,92 @@
...
+
+    <h3>Requesting rebuilds</h3>
+    <p>
+        Sometimes builds fail as a result of something that is our fault.
```
jyn514, line 85:
```suggestion
        Some build failures can be fixed by docs.rs.
```

templates/core/about/builds.html
```diff
@@ -3,80 +3,92 @@
...
+        For example, your crate may fail to build because of a missing dependency or a bug in Docs.rs.
+        Crates that hit resource limits might be rebuilt if we have the resources to do so. Failures caused by networking will not be rebuilt.
```
jyn514, line 87: I don't think this line belongs here, you already said everything relevant in the section on sandbox limits.

templates/core/about/builds.html
```diff
@@ -3,80 +3,92 @@
...
+    </p>
+
+    <h3>Diagnosing failed builds</h3>
+
+    <h4>Missing dependencies</h4>
+    <p>
+        Missing dependencies are a common reason for failed builds.
+        Docs.rs dependencies are managed through
+        <a href="https://github.com/rust-lang/crates-build-env">crates-build-env</a>.
+        See <a href="https://forge.rust-lang.org/docs-rs/add-dependencies.html">Forge</a> for further instructions.
```
jyn514, line 56: Why did this section change?

templates/core/about/builds.html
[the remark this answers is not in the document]
Nemo157: `cfg(doc)` is very different than `cfg(docsrs)` because docs.rs runs nightly, so I think it's a very useful example to allow docs to start testing nightly features before they're stabilized.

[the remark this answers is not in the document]
schulzch: I've rewritten the section in hope for more clarity and added an example (`all-features = true`).

templates/core/about/builds.html
schulzch: I find it less confusing to offer a single entry point into dependency management. Maybe the sentence is too short now?

templates/core/about/builds.html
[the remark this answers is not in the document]
Kixiron: Could you use the example I have above specifically?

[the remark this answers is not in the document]
schulzch: Done

schulzch:
Should be good now and sorry for the noise :)

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+        If your <code>build.rs</code> generates files that are relevant for documentation, consider writing to the <a href="https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates">cargo output directory</a>, passed in the environment variable <code>OUT_DIR</code>.
+    </p>
+
+    <h4>Hitting sandbox limits</h4>
+    <p>
+        All the builds on Docs.rs are executed inside a sandbox with limited
+        resources. If a build fails because it hit limits please
+        <a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
+        to get them increased for your crate.
+        Since our build agent has finite resources, we have to consider each case indidually.
```
Kixiron, line 77:
```suggestion
        Since our build agent has finite resources, we have to consider each case individually.
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+
+    <h3>Notes on using Docs.rs</h3>
+
+    <h4>Setting a README</h4>
+    <p>
+        The README of a crate is taken from the <code>readme</code> field defined in
+        <code>Cargo.toml</code>. Omitting this field means that no README will be displayed.
+    </p>
+
+    <h4>Detecting Docs.rs from <code>build.rs</code></h4>
```
Kixiron, line 30:
```suggestion
    <h4>Detecting Docs.rs from the <code>build.rs</code></h4>
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+    <p>
+        Docs.rs sets the environment variable <code>DOCS_RS</code> to <code>1</code>.
+        Detecting Docs.rs and building your crate differently can be helpful if you need
+        dependencies for building the library, but not for building the documentation.
+    </p>
+
+    <h4>Detecting Docs.rs from <code>Cargo.toml</code></h4>
+    <p>
+        Doc.rs recognizes <a href="metadata">package metadata</a> in your <code>Cargo.toml</code>.
```
Kixiron, line 39:
```suggestion
        Docs.rs recognizes <a href="metadata">package metadata</a> in your <code>Cargo.toml</code>.
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+        This can be useful to configure the compilation environment, e.g.:
+        <code>
+            [package.metadata.docs.rs]
+            rustc-args = ["--cfg", "docsrs"]
+        </code>
```
Kixiron, line 44: As-is this codeblock is improperly formatted
```suggestion
        {% filter dedent -%}
        <pre>
            <code>
                [package.metadata.docs.rs]
                rustc-args = ["--cfg", "docsrs"]
            </code>
        </pre>
        {%- endfilter %}
```

Before:
image

After:
image

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+        Passing this flag sets <code>#[cfg(docsrs)]</code> (not to be confused with <code>#cfg(doc)</code>) for conditional compilation.
+    </p>
```
Kixiron, line 46: These two sections on their own don't currently contain enough content to warrant being separate, plus they're both the same concept of "Detecting Docs.rs builds". I think they'd be better combined, with a small section at the end that gives examples of both (using the examples that are there right now)

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+    <p>
+        Please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a> if you have trouble building a crate on Docs.rs.
+    </p>
+
+    <h3>Requesting rebuilds</h3>
+    <p>
+        Some build failures can be fixed by Docs.rs.
+        For example, your crate may fail to build because of a missing dependency or a bug in Docs.rs.
+        You can request a rebuild by <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
+    </p>
```
Kixiron, line 94: These three sections are also incredibly similar and separating them is fairly redundant, it could be one paragraph that follows the scaffold of:

- We build in a sandbox, these are the limits
- If you hit those limits, they can be fixed (but no guarantee since our machine can only handle so much)
    - Networking will never be given to a crate
- If you encounter an issue that falls under these criteria, please open one

templates/core/about/builds.html
```diff
@@ -3,77 +3,97 @@
...
+
+    <h4>Detecting Docs.rs from <code>build.rs</code></h4>
+    <p>
+        Docs.rs sets the environment variable <code>DOCS_RS</code> to <code>1</code>.
+        Detecting Docs.rs and building your crate differently can be helpful if you need
+        dependencies for building the library, but not for building the documentation.
+    </p>
+
+    <h4>Detecting Docs.rs from <code>Cargo.toml</code></h4>
```
Kixiron, line 37: This isn't really what's happening, you aren't detecting docs.rs from your Cargo.toml and are instead setting conditional compilation flags from within it.
```suggestion
    <h4>Detecting Docs.rs with conditional compilation</h4>
```

templates/core/about/builds.html
jyn514: That wording looks more odd to me, `build.rs` is a filename - there's many of them, one for each project. We could use 'your build.rs file' but I think the existing wording is fine.

schulzch: If "build.rs" is more like a name, then I'd omit "the", but if its a countable noun, then I'd add a your/the. However, I am not a native speaker.

templates/core/about/builds.html
schulzch: I kept the "other failures" section, since it always feels nice if other problems are allowed ;)

templates/core/about/builds.html
```diff
@@ -74,29 +81,25 @@ <h4>Write attempt on read-only directories</h4>
...
     </p>
 
-    <h3>Requesting rebuilds</h3>
+    <h4>Other failures and requesting rebuilds</h4>
     <p>
-        Some build failures can be fixed by Docs.rs.
-        For example, your crate may fail to build because of a missing dependency or a bug in Docs.rs.
-        You can request a rebuild by <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
+        If you encounter build failures that do not fall under these criteria, please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
+        Some build failures can be fixed by Docs.rs, e.g., by issuing a rebuild after a bug has been fixed.
```
jyn514, line 102: This change no longer says 'a bug in Docs.rs' and just says 'a bug'. This is a change in meaning - rebuilds can't fix a bug in the crate itself.

templates/core/about/builds.html
```diff
@@ -3,77 +3,106 @@
...
+        The current version in use is <code>{{ rustc_version }}</code>.
+        {%- endif -%}
+    </p>
+
+    <h3>Notes on using Docs.rs</h3>
+
+    <h4>Setting a README</h4>
+    <p>
+        The README of a crate is taken from the <code>readme</code> field defined in
+        <code>Cargo.toml</code>. Omitting this field means that no README will be displayed.
```
jyn514, line 27: nit: this uses a little more complicated language which is more concise but might be harder for some readers. Is there a reason you changed it?

templates/core/about/builds.html
```diff
@@ -3,77 +3,106 @@
...
+                    // ... your code here ...
+                }        
+            </code>
+        </pre>
+        {%- endfilter %}
+        This approach can be helpful if you need dependencies for building the library, but not for building the documentation.
+    </p>
+
+    <p>
+        To configure the compilation environment, you can add <a href="metadata">package metadata</a> to your <code>Cargo.toml</code>, e.g.:
```
jyn514, line 46: This seems a little more wordy than it needs to be (same as above where it might be harder to read).
```suggestion
        You can configure how your crate is built by adding <a href="metadata">package metadata</a> to your <code>Cargo.toml</code>, e.g.:
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,106 @@
...
+        {% filter dedent -%}
+        <pre>
+            <code>
+                [package.metadata.docs.rs]
+                rustc-args = ["--cfg", "docsrs"]
+            </code>
+        </pre>
+        {%- endfilter %}
+        Here, the compiler arguments are set so that <code>#[cfg(docsrs)]</code> (not to be confused with <code>#[cfg(doc)]</code>) can be used for conditional compilation.
+        This approach is also useful for setting <a href="https://doc.rust-lang.org/cargo/reference/features.html">features</a>.
```
jyn514, line 56: nit: 'features' on its own seems a little ambiguous
```suggestion
        This approach is also useful for setting cargo <a href="https://doc.rust-lang.org/cargo/reference/features.html">features</a>.
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,106 @@
...
+
+    <h4>Hitting resource limits</h4>
+    <p>
+        All the builds are executed inside a sandbox with limited resources. The current limits are:
+    </p>
+
+    {{ macros::crate_limits(limits=limits) }}
+
+    <p>
+        If your build fails because it these limits, please
```
jyn514, line 92:
```suggestion
        If your build fails because it hit one of these limits, please
```

templates/core/about/builds.html
```diff
@@ -3,77 +3,106 @@
...
+        <a href="{{ docsrs_repo | safe }}/issues/new/choose">open an issue</a>
+        to get them increased for your crate.
+        Since our build agent has finite resources, we have to consider each case individually.
+        However, network access will not be enabled for any crate.
+    </p>
+
+    <h4>Other failures and requesting rebuilds</h4>
+    <p>
+        If you encounter build failures that do not fall under these criteria, please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
```
jyn514, line 101: Same comment about using simple language.
```suggestion
        If your crate fails to build for a reason not listed here, please <a href="{{ docsrs_repo | safe }}/issues">file an issue</a>.
```

jyn514:
LGTM once the last few nits are addressed. Could you also squash the commit history a little?

schulzch:
Done. I've replaced the "Omitting this field means that no README will be displayed." with "If this field is not set, no README will be displayed.". Otherwise it is just a squash including the requested changes.

jyn514:
Thanks so much for the PR! The page is definitely a lot more helpful now :)

PR #956, opened 2020-08-11
Run git gc periodically on the crates.io index (#778)

### LINKED ISSUE #778: Run git gc periodically on the crates.io index
We discovered that the clone on the index managed by docs.rs reached 2k+ packfiles, which caused spikes of thousands of open FDs and 20% CPU usage every time the repository was queried. We should add code that performs `git gc` periodically.

ohaddahan:
Forking `git gc` periodically from `start_registry_watcher`.

Closes #778.


src/utils/daemon.rs
```diff
@@ -11,20 +11,32 @@ use crate::{
...
 use log::{debug, error, info};
+use std::process::Command;
 use std::sync::Arc;
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+fn run_git_gc() {
+    Command::new("git")
+        .args(&["gc"])
```
Kixiron, line 21: Can tweak the parameters here a little bit, adding [`--auto`](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---auto) will only actually gc when it's needed and we could maybe add a [`--prune=<date>`](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt) since the repo has fairly regular updates
```suggestion
        .args(&["gc", "--auto"])
```
we *could* run [`--aggressive`](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---aggressive), but I'm not entirely sure if the cost to benefit ratio is worth it

Kixiron: cc @jyn514 for the options of `--prune` and `--aggressive`

jyn514: Auto is a great idea. We definitely don't need aggressive since that will basically rewrite the whole index every time there's an update. Prune sounds interesting but since we only have 1 branch and no force pushes I don't think it's necessary.

Really this is just trying to avoid the worst case scenario of '3 years without git gc', it doesn't need to be super efficient.

src/utils/daemon.rs
```diff
@@ -11,20 +11,33 @@ use crate::{
...
 use failure::Error;
 use log::{debug, error, info};
+use std::process::Command;
 use std::sync::Arc;
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+fn run_git_gc() {
+    let gc = Command::new("git").args(&["gc", "--auto"]).output();
```
Nemo157, line 20: This needs the path to the repo to clean. I think it'd make most sense to move this to a method on `index::Index` since that's what needs cleaning and has the path available. You can then access it via `docbuilder.index` to call.

src/config.rs
```diff
@@ -61,6 +63,7 @@ impl Config {
             // LOL HTML only uses as much memory as the size of the start tag!
             // https://github.com/rust-lang/docs.rs/pull/930#issuecomment-667729380
             max_parse_memory: env("DOCSRS_MAX_PARSE_MEMORY", 5 * 1024 * 1024)?,
+            registry_gc_interval: env("REGISTRY_GC_INTERVAL", 60 * 60)?,
```
pietroalbini, line 66: Can we prefix the environment variable with `DOCSRS_`? All the variables not starting with the prefix are legacy ones we didn't migrate to the new naming scheme yet.

src/utils/daemon.rs
ohaddahan: @Nemo157 inside what method? 

How about `RustwideBuilder::build_package` ? Doesn't it run in the same location as the `git` repo of the crate?

src/config.rs
ohaddahan: Sure, will do.

src/utils/daemon.rs
Nemo157: A new method, probably just keep this name for it `run_git_gc`.

I have vague ideas of eventually removing the index to queue handling from `docbuilder` completely, it isn't really related to building the docs.

src/web/metrics.rs
```diff
@@ -130,6 +130,7 @@ pub static MAX_DB_CONNECTIONS: Lazy<IntGauge> = Lazy::new(|| {
 });
 
 #[cfg(not(windows))]
+#[allow(dead_code)]
```
ohaddahan, line 133: clippy failed without these on local runs, not sure why it's not failing in GitHub actions running it.

git_hooks/pre-commit
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env sh
```
ohaddahan, line 1: I skipped these steps a couple of times, I think it's useful to provide this an an optional utility.

src/web/metrics.rs
Kixiron: Instead of this I'd change the [`#[cfg(target_os = "linux")]`](https://github.com/rust-lang/docs.rs/blob/525c6f855571962bd594b659f0da552dfbb3146a/src/web/metrics.rs#L171) to `#[cfg(not(windows))]`

src/web/metrics.rs
```diff
@@ -139,6 +140,7 @@ pub static OPEN_FILE_DESCRIPTORS: Lazy<IntGauge> = Lazy::new(|| {
...
```
Kixiron, line 143: Same as the other one

src/utils/daemon.rs
ohaddahan: @Nemo157 what I meant was, where in the `Index` flow to add it.

 I checked now and the `Index` is called from `RustwideBuilder` which is called from the `Queue`.

So maybe add `last_gc_run` property on the `Queue` and call `run_git_gc` conditionally from inside `build_package`? Something along these lines?

git_hooks/pre-commit
Kixiron: I don't know if this is something for us to provide, I know that I already have hooks running (plus you need to run clippy before check, it's a known issue)

src/web/metrics.rs
ohaddahan: @Kixiron sure I'll change that.

What about the `dead_code` it's OK to add it?

ohaddahan: Changing to `#[cfg(target_os = "linux")]` make my IDE (InteliJ) grey it out since I'm on a Mac.

Which also has file descriptors, so maybe keeping the `not windows` is OK?

Kixiron: No, that should be able to be removed, we don't want it lying around if we don't actually need it

git_hooks/pre-commit
ohaddahan: @Kixiron regarding the order, if I'll first run `clippy` and then run `fmt` I won't get the warning? So we can remove the `dead_code` that I added?

Also, it's an **optional** utility to make the contributor local  environment match the CI/CD, I actually find these sort of helpers missing from various projects and as you mentioned yourself, we all end up recreating them locally which is literally the opposite intention an open source project like this one.

src/utils/daemon.rs
Nemo157: I was thinking just keep calling it from here, this is the place that causes us to create new packfiles so should be responsible for knowing when to clean them up.

src/web/metrics.rs
Nemo157: That can't be `cfg(not(windows))` since `procfs` is a linux only thing (it'd fail on macOS). This should probably be `#[cfg(target_os = "linux")]` as well.

git_hooks/pre-commit
Kixiron: If you run `cargo check` before `cargo clippy` it doesn't output anything, even when there's lints (usually) and emoving the dead code attribute comes from fixing the problem it's talking about

src/web/metrics.rs
ohaddahan: I actually think it's a false alarm , I found a couple of threads about similar issues.

> src/web/metrics.rs:134:pub static OPEN_FILE_DESCRIPTORS: Lazy<IntGauge> = Lazy::new(|| {
> src/web/metrics.rs:176:        OPEN_FILE_DESCRIPTORS.set(process.fd().unwrap().len() as i64);

> src/web/metrics.rs:144:pub static CURRENTLY_RUNNING_THREADS: Lazy<IntGauge> = Lazy::new(|| {
> src/web/metrics.rs:177:        CURRENTLY_RUNNING_THREADS.set(process.stat().unwrap().num_threads as i64);

src/utils/daemon.rs
ohaddahan: @Nemo157 what is the location you refer to by `calling it from here`?

src/web/metrics.rs
Kixiron: On non-linux and non-windows systems the static exists but the invocation doesn't, please change the `cfg`s on the statics to `#[cfg(target_os = "linux")]` and remove the `#[allow(dead_code)]`s, that should fix things

ohaddahan: Yeah just noticed and pushed it, the declaration and usage had different conditions.

git_hooks/pre-commit
jyn514: +1 to having pre-commit hooks, otherwise you end up with people having to force-push three times because they forgot to run `cargo fmt`. That said I think this should be either a hard or soft link so that it doesn't get outdated when the commit script is updated: `ln -s ../../git_hooks/pre-commit .git/hooks` (not sure the windows equivalent - maybe `mklink` @Kixiron ?). Also, does it really need to be in a subdirectory if it's only one file?

git_hooks/pre-commit
```diff
@@ -0,0 +1,19 @@
...
+cargo fmt -- --check
+if [ "$?" != "0" ] ; then
```
jyn514, line 3:
```suggestion
if ! cargo fmt -- --check; then
```

git_hooks/pre-commit
```diff
@@ -0,0 +1,19 @@
...
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tcargo fmt -- --check\033[0m\n"
+  exit 1
+fi
+cargo clippy --locked -- -D warnings
+if [ "$?" != "0" ] ; then
```
jyn514, line 10:
```suggestion
if ! cargo clippy --locked -- -D warnings; then
```

git_hooks/pre-commit
```diff
@@ -0,0 +1,19 @@
...
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tclippy --locked -- -D warning\033[0m\n"
+  exit 1
+fi
+
+printf "\n"
+printf "\033[0;32mpre-commit hook succeeded\033[0m\n"
+exit 0
```
jyn514, line 19: Missing newline at EOF; this will break `cat`.
```suggestion
exit 0
```

src/utils/daemon.rs
jyn514: I assume he meant `daemon.rs`. So `fn run_git_gc` gets moved to `Index`, but still gets called from `start_registry_watcher`.

git_hooks/pre-commit
Kixiron: It doesn't really matter what command you run, git runs hooks in a bash environment

jyn514: Right, I'm asking about the command to set up the hook in the first place. Does windows have soft or hard links?

Kixiron: Windows does both, `mklink Link Target` for a soft link at a file, `mklink /D Link Target` for a soft link at a directory, `mklink /H Link Target` for a hard link at a file and `mklink /J Link Target` for a hard link at a directory

Nemo157: > Also, does it really need to be in a subdirectory if it's only one file?

If anything I would rename this to `.git_hooks` to remove it from view since it's not something most people will need to touch.

src/utils/daemon.rs
ohaddahan: @jyn514 @Nemo157 updated, let me know if that what you meant 😊

Kixiron:
@ohaddahan Could you push your commits so we can see the changes?

ohaddahan:
@Kixiron ops sorry, forgot to merge my fork branch 😊

Kixiron:
Also as a heads-up, we no merge-commit policy like [rustc does](https://rustc-dev-guide.rust-lang.org/contributing.html#pull-requests)

README.md
```diff
@@ -22,6 +22,12 @@ The recommended way to develop docs.rs is a combination of `cargo run` for
 the main binary and [docker-compose](https://docs.docker.com/compose/) for the external services.
 This gives you reasonable incremental build times without having to add new users and packages to your host machine.
 
+### Git Hooks
+For ease of use, `git_hooks` directory contains useful `git hooks` to make your development easier.
+```bash
```
Kixiron, line 27:
```suggestion

### Git Hooks

For ease of use, `git_hooks` directory contains useful `git hooks` to make your development easier.
```bash
```

README.md
```diff
@@ -22,6 +22,12 @@ The recommended way to develop docs.rs is a combination of `cargo run` for
...
+cd .git/hooks && ln -s ../../.git_hooks/* . && cd ../..
+```
```
Kixiron, line 29:
```suggestion
```bash
# Unix
cd .git/hooks && ln -s ../../.git_hooks/* . && cd ../..

# Powershell
cd .git/hooks && New-Item -Path ../../.git_hooks/* -ItemType SymbolicLink -Value . && cd ../..
```
```

File: src/index/mod.rs [hunk 12]
```diff
@@ -50,6 +51,14 @@ impl Index {
         Ok(Self { path, api })
     }
 
+    pub fn run_git_gc(&self) {
+        let cmd = format!("cd {} && git gc --auto", self.path.to_str().unwrap());
+        let gc = Command::new("sh").args(&["-c", cmd.as_str()]).output();
+        if let Err(err) = gc {
```
Kixiron, line 57:
```suggestion
        let gc = Command::new("sh").args(&["-c", cmd.as_str()]).output();

        if let Err(err) = gc {
```

src/index/mod.rs
```diff
@@ -50,6 +51,14 @@ impl Index {
         Ok(Self { path, api })
     }
 
+    pub fn run_git_gc(&self) {
+        let cmd = format!("cd {} && git gc --auto", self.path.to_str().unwrap());
+        let gc = Command::new("sh").args(&["-c", cmd.as_str()]).output();
```
Kixiron, line 56: `sh` isn't available on all platforms
```suggestion
        #[cfg(not(windows))]
        const SHELL: &str = "sh";
        #[cfg(windows)]
        const SHELL: &str = "cmd";

        let gc = Command::new(SHELL).args(&["-c", cmd.as_str()]).output();
```

src/index/mod.rs
```diff
@@ -50,6 +51,14 @@ impl Index {
         Ok(Self { path, api })
     }
 
+    pub fn run_git_gc(&self) {
+        let cmd = format!("cd {} && gc --auto", self.path.to_str().unwrap());
+        let gc = Command::new("sh").args(&["-c", cmd.as_str()]).output();
```
Nemo157:
```suggestion
        let gc = Command::new("git").arg("-C").arg(&self.path).args(&["gc", "--auto"]).output();
```

Avoids the use of `sh` and potential panic if the path is non-UTF-8.

src/index/mod.rs
```diff
@@ -1,4 +1,5 @@
 use std::path::{Path, PathBuf};
+use std::process::Command;
```
Kixiron, line 2:
```suggestion
use std::{path::{Path, PathBuf}, process::Command};
```

src/index/mod.rs
Kixiron: I didn't see Nemo's, do what [he said](https://github.com/rust-lang/docs.rs/pull/956#discussion_r470110557)

jyn514:
@ohaddahan Hi, your commits messages say 'rebase' but rebases do not have commits, they apply the original commits on top of the original branch. As it stands this has 143 commits and a 900 line diff. Can you please rebase using something like the following:

```
git checkout master
git fetch origin
git rebase -i origin/master
```

Since you have several merge commits there will be many conflicts, so you should drop all commits that are not related to this PR. Alternatively, it might be easier to start over on a new branch:

```
git checkout origin/master
git checkout --branch git-gc
git cherry-pick all the relevent commit hashes
```

However, as-is, this is very difficult to review.

jyn514:
If you want help feel free to join the discord channel: https://discord.gg/f7mTXPW

ohaddahan:
@jyn514 yeah my fork went down the drain, I'll clean it up.

jyn514:
Closing in favor of https://github.com/rust-lang/docs.rs/pull/975

PR #957, opened 2020-08-11
Update Font Awesome  & Tera and refactor templates

Kixiron:
* Update to Font Awesome 5
* Update to Tera 1.5
* Use custom filters for making icons
* Made some of the icons fancier

### Before:

image
image

### After:

image
image

r? @pietroalbini


Nemo157:
Can we make the icons in the file list [fixed width](https://fontawesome.com/how-to-use/on-the-web/styling/fixed-width-icons)?

templates/macros.html
```diff
@@ -126,26 +126,26 @@
             {# If the release isn't a library, then display that warning #}
             {% if not release.is_library -%}
                 <a href="{{ release_url | safe }}" class="pure-menu-link warn" title="{{ release_name }} is not a library">
-                    <i class="fa fa-fw fa-warning"></i> {{ release.version }}
+                    {{ "exclamation-triangle" | fas }} {{ release.version }}
```
Nemo157, line 129: Actually, a bunch of places seem to have been fixed-width previously and are now not.

src/web/page/templates.rs
```diff
@@ -147,6 +148,9 @@ pub(super) fn load_templates(conn: &mut Client) -> Result<Tera> {
     tera.register_filter("timeformat", timeformat);
     tera.register_filter("dbg", dbg);
     tera.register_filter("dedent", dedent);
+    tera.register_filter("fas", IconType::Strong);
+    tera.register_filter("far", IconType::Regular);
+    tera.register_filter("fab", IconType::Brand);
```
pietroalbini, line 153: I'd add `fa[srb]_fw` filters for the fixed width icons.

src/web/page/templates.rs
```diff
@@ -250,6 +254,43 @@ fn dedent(value: &Value, _args: &HashMap<String, Value>) -> TeraResult<Value> {
...
+
+impl tera::Filter for IconType {
+    fn filter(&self, value: &Value, args: &HashMap<String, Value>) -> TeraResult<Value> {
+        let icon = format!(
+            r#"<span class="{} fa-{}" aria-hidden="true" {}></span>"#,
+            self,
+            value.as_str().expect("Icons only take strings"),
+            args.get("extra")
+                .and_then(Value::as_str)
+                .unwrap_or_default(),
```
pietroalbini, line 283: We need to be careful here, as we might get XSS issues if we ever allow users to choose a dynamic icon. A possible approach would be:

```rust
args.entry("class".to_string())
    .or_insert_with(String::new())
    .push_str(&format!(
		"{} fa-{}",
		self,
		value.as_str().expect("Icons only take strings"),
	));
args.entry("aria-hidden".to_string()).or_insert_with(|| "true".to_string());

let tag = "<span".to_string();
for (key, value) in args.iter() {
    tag.push(' ');
    tag.push_str(&tera::escape_html(key));
    tag.push_str("=\"");
    tag.push_str(&tera::escape_html(value));
    tag.push('"');
}
tag.push_str("></span>");
Ok(Value::String(tag));
```

This way, it's possible to turn:

```
{{ "copy" | far(extra='id="clipboard" aria-label="Copy crate name and version information"') }}
```

...into:

```
{{ "copy" | far(id="clipboard" aria-label="Copy crate name and version information") }}
```

...and we also get HTML escaping too!

templates/header/package_navigation.html
```diff
@@ -21,7 +22,7 @@ <h1 id="crate-title">
                     {{ title }}
                 {%- else -%}
                     {{ metadata.name }} {{ metadata.version }}
-                    <i class="fa fa-copy fa-1" id="clipboard" aria-label="Copy crate name and version information"></i>
+                    {{ "copy" | far(extra='id="clipboard" aria-label="Copy crate name and version information"') | safe }}
```
pietroalbini, line 25: You forgot to remove the `| safe` here. Also, this needs to have `aria-hidden="false"`, since we want screen readers to see this icon (since we added a `aria-label` attribute).

templates/rustdoc/body.html
```diff
@@ -33,19 +33,18 @@
                             <ul class="pure-menu-list menu-item-divided">
                                 <li class="pure-menu-heading" id="crate-title">
                                     {{ krate.name }} {{ krate.version }}
-                                    <i class="fa fa-copy fa-1" id="clipboard" aria-label="Copy crate name and version information"></i>
+                                    {{ "copy" | far(extra='id="clipboard" aria-label="Copy crate name and version information"') | safe }}
```
pietroalbini, line 36: You forgot to remove the `| safe` here. Also, this needs to have `aria-hidden="false"`, since we want screen readers to see this icon (since we added a `aria-label` attribute).

src/lib.rs
```diff
@@ -31,7 +31,11 @@ pub(crate) static GLOBAL_ALERT: Option<GlobalAlert> = Some(GlobalAlert {
     url: "https://blog.rust-lang.org/2019/09/18/upcoming-docsrs-changes.html",
     text: "Upcoming docs.rs breaking changes!",
     css_class: "error",
+<<<<<<< HEAD
     fa_icon: "warning",
+=======
+    fa_icon: "exclamation-triangle",
+>>>>>>> 6e25e2c... Pause for upstream
```
jyn514, line 38: Please fix this.

templates/rustdoc/head.html
```diff
@@ -3,8 +3,12 @@
             media="all" />
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/pure/0.6.0/grids-min.css" type="text/css"
             media="all" />
+<<<<<<< HEAD
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css"
             type="text/css" media="all" />
+=======
+        <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.14.0/css/all.css" />
+>>>>>>> 6e25e2c... Pause for upstream
```
jyn514, line 11: Please fix this.

Kixiron:
The upgrade is just bumping versions, does that matter?

jyn514:
Yes, not all projects correctly follow semver (or make accidental breaking changes) and even if there's no breaking change there could be a large performance difference, since most projects don't track that. For example this happened with tokio 0.2 and html5ever 0.24.

templates/rustdoc/head.html
```diff
@@ -3,8 +3,12 @@
             media="all" />
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/pure/0.6.0/grids-min.css" type="text/css"
             media="all" />
+<<<<<<< HEAD
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css"
             type="text/css" media="all" />
+=======
+        <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.14.0/css/all.css" />
```
Nemo157, line 10: Where does this URL come from? I don't see anywhere on the main website listing it as a CDN?

Kixiron: [This page](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4) in the codeblock of the `Upgrading Manually` section

src/lib.rs
Kixiron: No idea how I missed these, but they're fixed now

jyn514:
> Can we separate the upgrade from the refactor? I think I heard the refactor depends on the upgrade, so we could do the upgrade first without any large changes to docs.rs itself.

We discussed this on discord and decided that since there have been [very few changes](https://github.com/Keats/tera/blob/master/CHANGELOG.md#150-2020-08-10) since 1.3.1 it's ok to do both at once.

PR #960, opened 2020-08-12
Switch to grass for sass compilation

Kixiron:
Removes our dependency on `sass_rs`/`libsass`, which reduces our total # of native dependencies
* Helps contributors since it reduces the number of native dependencies as well as fixing the interments issues with `libsass` randomly breaking
* `grass` supports more features, namely sass modules which are currently unsupported by `sass_rs`

This PR was made out of anger when `libsass` suddenly broke

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,11 @@ version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "ahash"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jyn514:
> libsass suddenly broke

I didn't hear about this, what went wrong?

jyn514:
Code changes look fine, but have you run a diff on the generated CSS file before and after? I don't want to make changes blind.

Kixiron:
I have pretty much no idea, it just suddenly [died when linking](https://hastebin.com/zafuyapema.js)

Kixiron:
The two outputs are very nearly identical, only difference is that `grass` [doesn't minify yet](https://github.com/connorskees/grass/pull/33), but I don't think that's a huge deal considering how small our css is anyways

jyn514:
This bit about a complete rewrite really worries me: https://github.com/connorskees/grass/pull/33#issuecomment-673110393

I understand `grass` is a new project and has a lot of advantages over libsass, but I don't want to switch to a project that's about to be completely redone. Can we come back to this after the proposed rewrite?

connorskees:
If you're still interested in this, `grass` is at a point where I feel quite confident in its output. For docs.rs, the output only seems to differ from `dart-sass` in whitespace.

jyn514:
@connorskees I probably won't have time to work on switching myself, but if it reduces our dependencies I'd be willing to take a PR for it :)

PR #961, opened 2020-08-12
Improve the styling of the coverage in the dropdown

pietroalbini:
New styling (the text is clickable and leads to the crate page):

r? @GuillaumeGomez 
cc @yaahc @sunjay

### DIFF
--- a/templates/rustdoc/body.html
+++ b/templates/rustdoc/body.html
@@ -117,21 +117,7 @@
                                 </div>
                             </div>
 
-                            {%- if krate.documented_items and krate.total_items -%}
-                                {% set percent = krate.documented_items * 100 / krate.total_items %}
-                                <div class="pure-g menu-item-divided">
-                                    <div class="pure-u-1">
-                                        <ul class="pure-menu-list">
-                                            <li class="pure-menu-heading">Coverage</li>
-                                            <li class="pure-menu-link"><b>{{ percent | round(precision=2) }} %</b><br>
-                                                <span><b>{{ krate.documented_items }}</b> out of <b>{{ krate.total_items }}</b> items documented</span>
-                                            </li>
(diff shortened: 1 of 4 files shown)


sunjay:
Thanks @pietroalbini this looks great! Appreciate you being open to this @GuillaumeGomez. I definitely agree with the overall goal of getting people to write more docs and it will be awesome to see more tools/metrics come out to help with that. :)

GuillaumeGomez:
Like I said, I don't completely agree with this change but if it makes people more at ease with this feature, this is acceptable.

templates/rustdoc/body.html
```diff
@@ -117,21 +117,7 @@
...
-                                            <li class="pure-menu-link"><b>{{ percent | round(precision=2) }} %</b><br>
-                                                <span><b>{{ krate.documented_items }}</b> out of <b>{{ krate.total_items }}</b> items documented</span>
-                                            </li>
-                                        </ul>
-                                    </div>
-                                </div>
-                            {%- endif -%}
-
-                            <div class="pure-g">
+                            <div class="pure-g menu-item-divided">
```
GuillaumeGomez, line 120: Isn't this part supposed to just have `pure-g` as class?

pietroalbini: No, the `menu-item-divided` adds the line at the bottom.

GuillaumeGomez: Oh I see!

jyn514:
Looks like you need to update the tests too:

```
thread 'main' panicked at 'assertion failed: check_doc_coverage_is_present_for_path("/fake_crate/0.0.1/fake_crate", web)', src/web/mod.rs:702:13
```

Looks fine to me other than that.

PR #964, opened 2020-08-13
Rust dropdown now directly links to rust-lang.org (#959)

### LINKED ISSUE #959: 'Rust' dropdown should take you to rust-lang.org
Right now, we have three dropdown tabs: 'About', 'Releases', and 'Rust'. The first two take you to a separate page when you click on them. The last does nothing - you have to click on one of the drop-down items. To make it consistent, 'Rust' should take you to rust-lang.org and the 'Rust Programming Language' item should be removed.

shorsher:
* Rust topbar dropdown now directly links to https://www.rust-lang.org/
* Removed "The Rust Programming Language" link 

dropdown_rs

Closes #959

### DIFF
--- a/templates/header/topbar.html
+++ b/templates/header/topbar.html
@@ -53,15 +53,9 @@
 
                     {# The Rust dropdown menu #}
                     <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
-                        <a href="#" class="pure-menu-link">Rust</a>
+                        <a href="https://www.rust-lang.org/" target="_blank" class="pure-menu-link">Rust</a>
 
                         <ul class="pure-menu-children">
-                            {{ macros::menu_link(
-                                href="https://www.rust-lang.org/",
-                                text="The Rust Programming Language",
-                                target="_blank"
-                            ) }}
-
                             {{ macros::menu_link(
                                 href="https://doc.rust-lang.org/book/",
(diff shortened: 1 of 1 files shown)


templates/header/topbar.html
```diff
@@ -53,15 +53,9 @@
...
-                        <a href="#" class="pure-menu-link">Rust</a>
+                        <a href="https://www.rust-lang.org/" target="_blank" class="pure-menu-link">Rust</a>
 
                         <ul class="pure-menu-children">
-                            {{ macros::menu_link(
-                                href="https://www.rust-lang.org/",
-                                text="The Rust Programming Language",
-                                target="_blank"
-                            ) }}
-
```
Kixiron, line 64: I think having both links is fine, this was just about being consistent with other dropdowns

shorsher: Sounds good, I can add that back in after lunch

shorsher: One quick question, @jyn514 mentioned to remove the "Rust Programming Language" link in the original issue. Just want to have some consensus to add it back, even though it's a super small change :)

jyn514: I think we should be consistent. If we have the additional link here, we should have it for the other dropdowns. But I think it looks fine without.

jyn514:
Thanks for the PR!

PR #972, opened 2020-08-15
Don't create a duplicate title on rustdoc pages

### LINKED ISSUE #971: Avoid creating two <title>s
docs.rs creates two titles for most pages, the Rustdoc title and docs.rs title:

two-titles

This is [invalid HTML](https://www.w3.org/TR/html53/document-metadata.html#the-title-element) and causes my browser to sometimes just display the URL on the titlebar, since it doesn't know which title to use.

I would actually prefer to use just the Rustdoc title because it provides more information about the site.

jyn514:
Since the LOL HTML rewrite, we show both the rustdoc title and the
docs.rs title. We should only show the rustdoc title.

Closes #971 
r? @Kixiron


jyn514:
Before:  before-title

After: 
title

PR #973, opened 2020-08-15
pre-commit hooks

ohaddahan:
@Kixiron @jyn514 @pietroalbini 

Deleted my fork and restart my changes, also separated them into smaller chunks.

(https://github.com/rust-lang/docs.rs/pull/956)

### DIFF
--- a/.git_hooks/pre-commit
+++ b/.git_hooks/pre-commit
@@ -1,0 +1,18 @@
+#!/usr/bin/env sh
+if ! cargo fmt -- --check ; then
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tcargo fmt -- --check\033[0m\n"
+  exit 1
+fi
+
+if ! cargo clippy --locked -- -D warnings ; then
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tclippy --locked -- -D warning\033[0m\n"
+  exit 1
+fi
+
+printf "\n"
+printf "\033[0;32mpre-commit hook succeeded\033[0m\n"
+exit 0
--- a/.git_hooks/pre-commit
+++ b/.git_hooks/pre-commit
@@ -15,4 +15,3 @@ fi
 
 printf "\n"
 printf "\033[0;32mpre-commit hook succeeded\033[0m\n"
-exit 0
--- a/README.md
+++ b/README.md
@@ -22,6 +22,17 @@ The recommended way to develop docs.rs is a combination of `cargo run` for
 the main binary and [docker-compose](https://docs.docker.com/compose/) for the external services.
 This gives you reasonable incremental build times without having to add new users and packages to your host machine.
 
+### Git Hooks
+
+For ease of use, `git_hooks` directory contains useful `git hooks` to make your development easier.
+
+```bash
+# Unix
+cd .git/hooks && ln -s ../../.git_hooks/* . && cd ../..
+# Powershell
+cd .git/hooks && New-Item -Path ../../.git_hooks/* -ItemType SymbolicLink -Value . && cd ../..
+```
+
 ### Dependencies
 
 Docs.rs requires at least the following native C dependencies.


.git_hooks/pre-commit
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
+if ! cargo fmt -- --check ; then
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tcargo fmt -- --check\033[0m\n"
```
jyn514, line 5: Do these escape characters work on Windows? cc @Kixiron

.git_hooks/pre-commit
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
```
jyn514, line 1: This seems weird, is there a reason you didn't use `/bin/sh`?
```suggestion
#!/bin/sh
```

.git_hooks/pre-commit
```diff
@@ -0,0 +1,18 @@
...
+if ! cargo clippy --locked -- -D warnings ; then
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tclippy --locked -- -D warning\033[0m\n"
+  exit 1
+fi
+
+printf "\n"
+printf "\033[0;32mpre-commit hook succeeded\033[0m\n"
+exit 0
```
jyn514, line 18:
```suggestion
```

.git_hooks/pre-commit
```diff
@@ -0,0 +1,18 @@
...
+  exit 1
+fi
+
+if ! cargo clippy --locked -- -D warnings ; then
+  printf "\n"
+  printf "\033[0;31mpre-commit hook failed during:\033[0m\n"
+  printf "\033[0;31m\tclippy --locked -- -D warning\033[0m\n"
+  exit 1
+fi
+
```
jyn514, line 15: No `cargo test`?

README.md
```diff
@@ -22,6 +22,17 @@ The recommended way to develop docs.rs is a combination of `cargo run` for
 the main binary and [docker-compose](https://docs.docker.com/compose/) for the external services.
 This gives you reasonable incremental build times without having to add new users and packages to your host machine.
 
+### Git Hooks
+
+For ease of use, `git_hooks` directory contains useful `git hooks` to make your development easier.
+
+```bash
+# Unix
+cd .git/hooks && ln -s ../../.git_hooks/* . && cd ../..
```
jyn514, line 31: Does this need to be a whole directory? It's only one file.

Also the command can be simplified:
```suggestion
ln -s ../../.git_hooks/* .git/hooks
```

.git_hooks/pre-commit
Kixiron: Yah, running them in git bash works. Generally ANSI escapes work everywhere

.git_hooks/pre-commit
ohaddahan: Not a major difference for most cases, but using `/usr/bin/env` will take `sh` from the environment used while `/bin/sh` assumes this is the path to the shell installation which is common but not mandatory.

Especially if the environment has multiple shell versions, for example if `PATH` contains "/path_to_sh/:/bin".

`/user/bin/env sh` will take `sh` from `path_to_sh` which is first in `PATH` while `/bin/sh` will use `/bin/sh`.

I fine either way, let me know.

.git_hooks/pre-commit
ohaddahan: `cargo test` is somewhat slow, maybe we should put it in `pre-push` and not `pre-commit`?

README.md
ohaddahan: `zsh: no matches found: ../../.git_hooks/*` you need the `cd` without it it won't find the source file.

Since the `README.md` is assumes you're in the root (also `.git/hooks` path assumes that too).

Also if you try `ln -s .git_hooks/* .git/hooks` the link created isn't correct, the combo with `cd` seems to be the most robust way I found.

ohaddahan: Regarding the one file, I assume more will be added, `pre-push` for `cargo test` for example as discussed in the previous comment.

.git_hooks/pre-commit
jyn514: Ooh, I like that idea :smile: ignore my later comment about removing the `.git_hooks` dir then.

README.md
jyn514: Oh I see, because you're using shell expansion here. Yeah this is fine then.

.git_hooks/pre-commit
ohaddahan: Already working on its PR 😄

.git_hooks/pre-commit
jyn514: > common but not mandatory

I think posix requires `/bin/sh`. But I don't feel strongly about it, it's fine to keep it as-is.

.git_hooks/pre-commit
jyn514: Feel free to add it to this one, it's still part of the pre-commit script :)

.git_hooks/pre-commit
ohaddahan: Mentioning POSIX is a good way to get me to double check myself 😊 .

https://pubs.opengroup.org/onlinepubs/009695399/utilities/sh.html

> Applications should note that the standard PATH to the shell cannot be assumed to be either /bin/sh or /usr/bin/sh, 
> and should be determined by interrogation of the PATH returned by getconf PATH , ensuring that the returned 
> pathname is an absolute pathname and not a shell built-in.

.git_hooks/pre-commit
ohaddahan: Preferred to create a separate PR https://github.com/rust-lang/docs.rs/pull/977.
I think we can merge this one, I don't see any pending comment.

PR #975, opened 2020-08-15
Run git gc periodically on the crates.io index

ohaddahan:
@Kixiron @jyn514 @pietroalbini

Deleted my fork and restart my changes, also separated them into smaller chunks.

Hopefully I got understood and implemented your comments properly.

(#956)

### DIFF
--- a/src/docbuilder/queue.rs
+++ b/src/docbuilder/queue.rs
@@ -84,4 +84,8 @@ impl DocBuilder {
 
         Ok(processed)
     }
+
+    pub fn run_git_gc(&self) {
+        self.index.run_git_gc();
+    }
 }
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -84,7 +84,7 @@ impl Index {
 
         if let Err(err) = gc {
             log::error!(
-                "Failed to run `git gc --auto`\nPath: {:#?}\nError:{:#?}",
+                "failed to run `git gc --auto`\npath: {:#?}\nerror: {:#?}",
                 &self.path,
                 err
             );
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -81,6 +81,7 @@ impl Index {
             .arg(&self.path)
             .args(&["gc", "--auto"])
             .output();
+
         if let Err(err) = gc {
             log::error!(
                 "Failed to run `git gc --auto`\nPath: {:#?}\nError:{:#?}",
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -1,4 +1,7 @@
-use std::path::{Path, PathBuf};
+use std::{
+    path::{Path, PathBuf},
+    process::Command,
+};
 
 use url::Url;
 
@@ -70,6 +73,21 @@ impl Index {
 
     pub fn api(&self) -> &Api {
         &self.api
+    }
+
+    pub fn run_git_gc(&self) {
+        let gc = Command::new("git")
+            .arg("-C")
+            .arg(&self.path)
+            .args(&["gc", "--auto"])
+            .output();
+        if let Err(err) = gc {
+            log::error!(
+                "Failed to run `git gc --auto`\nPath: {:#?}\nError:{:#?}",
+                &self.path,
+                err
+            );
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/index/mod.rs
```diff
@@ -71,6 +74,21 @@ impl Index {
...
         &self.api
     }
+
+    pub fn run_git_gc(&self) {
+        let gc = Command::new("git")
+            .arg("-C")
+            .arg(&self.path)
+            .args(&["gc", "--auto"])
+            .output();
+        if let Err(err) = gc {
```
Kixiron, line 84:
```suggestion
            .output();

        if let Err(err) = gc {
```

src/utils/daemon.rs
```diff
@@ -10,16 +10,18 @@ use chrono::{Timelike, Utc};
...
 use log::{debug, error, info};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 fn start_registry_watcher(opts: DocBuilderOptions, context: &dyn Context) -> Result<(), Error> {
     let pool = context.pool()?;
     let build_queue = context.build_queue()?;
+    let config = context.config()?;
     thread::Builder::new()
```
Kixiron, line 19:
```suggestion
    let config = context.config()?;

    thread::Builder::new()
```

src/utils/daemon.rs
```diff
@@ -10,16 +10,18 @@ use chrono::{Timelike, Utc};
...
         .name("registry index reader".to_string())
         .spawn(move || {
             // space this out to prevent it from clashing against the queue-builder thread on launch
             thread::sleep(Duration::from_secs(30));
+            let mut last_gc = Instant::now();
```
Kixiron, line 24:
```suggestion
            thread::sleep(Duration::from_secs(30));

            let mut last_gc = Instant::now();
```

src/docbuilder/queue.rs
```diff
@@ -84,4 +84,8 @@ impl DocBuilder {
 
         Ok(processed)
     }
+
+    pub fn run_git_gc(&self) {
+        self.index.run_git_gc();
+    }
```
jyn514, line 90: I don't think this is necessary, you can call `self.index.run_git_gc()` directly in the function.

src/index/mod.rs
```diff
@@ -71,6 +74,21 @@ impl Index {
...
+            log::error!(
+                "Failed to run `git gc --auto`\nPath: {:#?}\nError:{:#?}",
```
jyn514, line 86: errors should be lowercase: https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err
```suggestion
                "failed to run `git gc --auto`\npath: {:#?}\nerror: {:#?}",
```

src/docbuilder/queue.rs
ohaddahan: In which function?

If I'm replacing:
`doc_builder.run_git_gc()` with `doc_builder.index.run_git_gc()` it won't compile since `index` is private.

jyn514: Ah, I see. This seems fine then.

ohaddahan:
BTW, maybe we should add some logging on success too, just to be sure it's actually working as expected.
What do you think?

https://github.com/rust-lang/docs.rs/blob/508a9d07d4be90f401f8a7795bc5c0bffb647bc7/src/index/mod.rs#L78

jyn514:
Sounds good to me. Time for another PR? :laughing:

Byron:
It's my stated goal to as-soon-as-reasonable switch `crates-index-diff` over to use [gitoxide](https://github.com/Byron/gitoxide), the main motivation being slightly faster pulls and smart-gc. I have been bitten by this myself more than once, and believe this pitfall shouldn't be one anymore.

I was thinking that we really want to run GC if there are too many packs, and too many really means more than we are allowed to open as per the file handle limit. What if `gitoxide` would track the amount of open handles and GC automatically when they run out. The GC could also be optimized to do only minimal work in favor of producing a combined pack fast, which from what I know now is totally possible.

[Here is the tracking issue](https://github.com/Byron/gitoxide/issues/15) for feedback, comments and general brainstorming.

PR #977, opened 2020-08-15
Pre push

ohaddahan:
Adding `pre-push` git hook that runs `docker-compose up -d db s3` and `cargo test`.


.git_hooks/pre-push
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env sh
+if ! docker-compose up -d db s3 ; then
```
jyn514, line 2: Hmm, my only concern is that this breaks workflows for people not using docker-compose ... but I'm not sure anyone is doing that? cc @rust-lang/docs-rs, are you happy with this?

Kixiron: I know I use containers for the side services, but this may re-spawn them if they're already running

GuillaumeGomez:
I don't have docker-compose so that would prevent me to send any PRs?

.git_hooks/pre-push
ohaddahan: IIRC when I ran `docker-compose up -d db s3` twice it just said it's already running.

We can put this conditional on some ENVAR you can set to enable it and keep `cargo run` as default.

Kixiron: I think just leaving service management up to the user is the best thing to do

ohaddahan: Done.

.git_hooks/pre-push
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+if ! cargo test ; then
```
jyn514, line 2: One more thing I thought of: this is a good opportunity to fix the intermittent 'too many open files' errors:

```
thread 'r2d2-worker-2' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Other, message: "Too many open files" }', /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/postgres-0.17.5/src/config.rs:318:27
```
```suggestion
ulimit -n 4096
if ! cargo test ; then
```

Nemo157:
@GuillaumeGomez git hooks are opt-in, so you'd just not set it up. I'm not planning to use them either so I have no opinion on what it does.

ohaddahan:
Screenshot as requested.

image

.git_hooks/pre-push
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env sh
+fd_hard_limit="$(ulimit -n -H)"
+if [ "$fd_hard_limit" = "unlimited" ] ; then
+  ulimit -n 4096
+else
+  ulimit -n "$fd_hard_limit"
```
jyn514, line 6: Are you sure this is the right thing to do? On my machine `ulimit -n -H` gives `1048576` which seems _way_ too high.

.git_hooks/pre-push
ohaddahan: To prevent trying to set 4096 and it being over the hard limit I inserted a check before it.
Also keep in mind that these settings won't persist in your env but only during each pre-push script running.
Furthermore, `ulimit -n` isn't POSIX compatible but I think it's safe enough, I literally couldn't find access to an old enough Linux that won't have `ulimit -n` working on it 😄 

https://github.com/koalaman/shellcheck/wiki/SC2039

> In .git_hooks/pre-push line 6:
>  ulimit -n "$fd_hard_limit"
>        ^-- SC2039: In POSIX sh, ulimit -n is undefined.

.git_hooks/pre-push
ohaddahan: Given that the soft limits are just a recommendation and basically anyone can ignore them and use whatever the hard limits allow, I don't see it as a big issue.

But I don't mind capping it at 4096 if you prefer it this way.

jyn514: I prefer that I think. We have a limit of 8096 in production so if this fails with 4096 that's a canary that something's going wrong and lets us fix it before we have an outage.

.git_hooks/pre-push
jyn514: I'm not worried about compatibility with very old Linux machines; if you're using something that old you'll know what the equivalent of `ulimit -n` is. The larger issue is that `cargo run build crate` only works on Linux currently (I've been meaning to open an issue).

Issue #979, opened 2020-08-16
https://docs.rs does not load in China anymore


Byron:
This is what I see when debugging this in Firefox - it appears only a few resources are not loading:

<img width="684" alt="Screenshot 2020-08-16 at 17 49 40" src="https://user-images.githubusercontent.com/63622/90331835-4d461000-dfea-11ea-8125-58b99f47dc46.png">

I thought it might be related to the recent move of blocking TLS 1.3 and ESNI, but I couldn't verify it for sure.

I am happy to provide further data to help resolving this issue, if at all possible.
Thank you

jyn514:
I guess this is pretty strong motivation for #921. We did switch to fontawesome 5 recently which uses a different CDN, but from your screenshot it looks like even the old files aren't loading.

jyn514:
Is the site completely broken or is it just the css missing? We can move it from `<head>` to `<body>` as a short term workaround which would let you at least use the site.

Byron:
Thanks for taking a look! It's completely broken as one only sees a white screen, due to the CSS not loading. The HTML is present though, it just won't ever be displayed. Firefox would display 'ghosts' of the page while in dev mode for some reason, and Curl can also get anything hosted on docs.rs.

That workaround of moving the CSS sounds great, I would love to give it a try.

pietroalbini:
I think vendoring all those files is as easy as shuffling things around, I'd avoid messing with the ordering.

Kixiron:
If we need to we can vendor our CSS, that's the recommended way to use FA 5

jyn514:
> I have no idea if a Chinese Rust community even exist

@Byron you might be interested in https://www.youtube.com/watch?v=j_ha1-1w6sw

PR #980, opened 2020-08-16
Allow the site to load even if the external CSS does not

jyn514:
Our external CSS is only used for the header bar and a few other things - the site will look ugly without, but the rustdoc pages should still be usable.

This also changes several `<script>` tags to be inside the body, since
it's invalid for them to come after: 
script

Partially addresses #979, cc @Byron. I can set up a staging server in a little bit so you can verify this actually helps.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -9,20 +9,11 @@
         <meta name="generator" content="docs.rs {{ docsrs_version() }}">
         {%- block meta -%}{%- endblock meta -%}
 
-        {# External styles #}
-        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/pure/0.6.0/pure-min.css" type="text/css"
-            media="all" />
-        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/pure/0.6.0/grids-responsive-min.css"
-            type="text/css" media="all" />
-        <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.14.0/css/all.css" />
-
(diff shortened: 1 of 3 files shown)


Byron:
I am happy to help :)! Thanks so much!

jyn514:
Since #981 is waiting on some design decisions, I'm thinking about this more to fix the immediate issues.

> putting the css in the body might cause some flicker in the screen, which imo is worse

@pietroalbini I actually think it's better to allow loading to fail, if you have a slow connection and the CSS times out you should still be able to see the rest of the site.

Here's loading with 'Good 2G' (whatever firefox thinks that is): https://youtu.be/tfXVmckH_j4
With 'Wifi', it still looks pretty good - there's a noticable load time, but no flicker: https://youtu.be/MOAJ53C8EMY
With the CSS in the head as currently deployed, it looks about the same, I didn't see a noticeable difference.

Without any CSS at all (I just commented it out) it looks pretty bad but you can at least see the documentation:

image
image

@Byron what do you think about those last two images? Is that usable or not?

Byron:
@jyn514 Thanks a lot for pushing for an intermediate solution - I can't wait to have docs.rs back in any condition.

Byron:
It looks like this one is ready to go, and I along with the entirety of China would greatly benefit from this PR being merged.

Is that correct, or is there anything else impeding it?

notriddle:
Now that docs.rs doesn't have any external resources, does this PR still make sense?

jyn514:
Closing this; it's no longer necessary after #981

PR #981, opened 2020-08-16
Vendor FontAwesome and Pure-CSS

### LINKED ISSUE #979: https://docs.rs does not load in China anymore

Kixiron:
Vendor all of our CSS, allows more users to access our site while simultaneously making less requests to load our page
To load the home page it's currently 14 requests, with this PR

| Page | Current Requests | PR Requests |
| - | - | - |
| Home page | 14 | 11  |
| Rustdoc page | 29 | 19 |

Since we're using SASS for everything, this also allows for us to serve one single CSS file instead of the previous 4+

Closes #979


src/web/routes.rs
```diff
@@ -18,6 +18,7 @@ pub(super) fn build_routes() -> Routes {
     routes.static_resource("/robots.txt", super::sitemap::robots_txt_handler);
     routes.static_resource("/sitemap.xml", super::sitemap::sitemap_handler);
     routes.static_resource("/opensearch.xml", super::opensearch_xml_handler);
+    routes.static_resource("/-/static/:file", super::statics::static_handler);
```
jyn514, line 21: While you're at it, can you change `index.js`, `style.css`, and `menu.js` to be under `/-/static`?

src/web/statics.rs
```diff
@@ -0,0 +1,97 @@
+use super::error::Nope;
+use iron::{
+    headers::CacheDirective,
+    headers::{CacheControl, ContentType},
+    status::Status,
+    IronError, IronResult, Request, Response,
+};
+use router::Router;
+
+const EOT: &str = "application/vnd.ms-fontobject";
```
jyn514, line 10: What is EOT?

src/web/statics.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[test]
+fn static_file_that_doesnt_exist() {
+    crate::test::wrapper(|env| {
+        let web = env.frontend();
+        assert!(web
+            .get("/-/static/whoop-de-do.png")
+            .send()?
+            .status()
+            .is_client_error());
```
jyn514, line 93:
```suggestion
        assert_eq!(web
            .get("/-/static/whoop-de-do.png")
            .send()?
            .status()
            .as_u16(), 404);
```

src/web/statics.rs
Kixiron: An embedded opentype font file

jyn514:
You also need to update the dockerfile.

```
--- stderr
thread 'main' panicked at 'Error compiling sass: Error: File to import not found or unreadable: fontawesome.
        on line 2:1 of templates/style/base.scss
>> @import "vars", "rustdoc", "utils", "navbar", "fontawesome", "regular", "sol
   ^
```

vendor/fontawesome/scss/_variables.scss
```diff
@@ -0,0 +1,1468 @@
+// Variables
+// --------------------------
+
+$fa-font-path: "../-/static" !default;
```
pietroalbini, line 4: Instead of changing the path in the vendored file (which would override it when we update FA), could you set the variable in our own `base.scss`?

src/web/statics.rs
```diff
@@ -0,0 +1,97 @@
...
+    "fa-solid-900.ttf"     => TTF,
+    "fa-solid-900.woff2"   => WOFF2,
+    "fa-brands-400.svg"    => SVG,
+    "fa-brands-400.woff"   => WOFF,
+    "fa-regular-400.eot"   => EOT,
+    "fa-regular-400.ttf"   => TTF,
+    "fa-regular-400.woff2" => WOFF2,
+    "fa-solid-900.svg"     => SVG,
+    "fa-solid-900.woff"    => WOFF,
+}
```
pietroalbini, line 83: Why did you take this approach?

I think it'd be less maintenance to have a map of paths and their source directories:

```
/-/static/            => {style.css,*.js}
/-/static/fontawesome => vendor/fontawesome/webfonts
```

And then map the extensions with the correct MIME type. That way, nothing is hardcoded and we can add or remove files without having to touch the source code.

src/web/statics.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[test]
+fn static_file_that_doesnt_exist() {
```
pietroalbini, line 86: Tests should be in a `mod tests`.

vendor/fontawesome/scss/_variables.scss
Kixiron: This was the recommendation in [fontawesome's docs](https://fontawesome.com/how-to-use/on-the-web/using-with/sass#compile), but I can do whatever

src/web/statics.rs
Kixiron: I agree that that's a better approach, I'm just forever afraid of path traversal and don't trust myself

vendor/fontawesome/scss/_variables.scss
pietroalbini: Why are they recommending that :/

pietroalbini:
If the goal is getting docs.rs up and running in China as soon as possible, I'm happy to not block on my comments and merge, if a followup PR is made addressing them.

Kixiron:
I'm so sorry, I've been taking a breather the last few days but I'll get right on this

vendor/fontawesome/scss/_variables.scss
Kixiron: I have no idea to be perfectly honest

src/web/statics.rs
Kixiron: What approach to this should I take?

vendor/fontawesome/scss/_variables.scss
Kixiron: Note that this will be fixed whenever we can use [scss modules](https://sass-lang.com/documentation/at-rules/use), at that time we can use something along these lines
```scss
@use "fontawesome" with (
    $fa-font-path: "../-/static"
);
// For more info see https://sass-lang.com/documentation/at-rules/use#configuration
```
Modules are nice for other reasons as well, they allow namespacing and explicit access to external variables, mixins and rules

src/web/statics.rs
```diff
@@ -49,9 +49,22 @@ fn serve_file(req: &Request, file: &str) -> IronResult<Response> {
     let contents = ctry!(req, fs::read(&path));
 
     // If we can detect the file's mime type, set it
-    let content_type = MimeGuess::from_path(&path)
-        .first()
-        .map(|mime| ContentType(mime.as_ref().parse().unwrap()));
+    // MimeGuess misses a lot of the file types we need, so there's a small wrapper
+    // around it
```
jyn514, line 53: Can we upstream this patch? I don't want to have to special case things if I can at all help it.

Kixiron: I'm going to guess not since there's lots of open issues & prs for exactly this with no response. It shouldn't really be a huge deal since there's not very many exceptions

PR #983, opened 2020-08-19
update latest version calculation

### LINKED ISSUE #749: Go to latest should not consider prereleases (unless there are only prereleases)
Example: <https://docs.rs/nemo157/0.3.0/nemo157/> shows "Go to latest" linking to <https://docs.rs/nemo157/0.3.1-pre.1/nemo157/index.html>.

Need to also take yanked into account, I think it should go through these four subsets, showing the first that gives any result:

 * latest non-prerelease non-yanked version
 * latest prerelease non-yanked version
 * latest non-prerelease yanked version
 * latest prerelease yanked version

(Probably in <https://github.com/rust-lang/docs.rs/blob/ece1b2115957ad606e252a88acec2c0ca0e061d4/src/web/crate_details.rs#L281>, if it doesn't affect anything else).

shorsher:
Updated the latest_version algo to consider the latest **non-prerelease**, non-yanked release. 

Closes #749

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -534,7 +534,10 @@ mod tests {
 
             for version in &["0.0.1", "0.0.2", "0.0.3"] {
                 let details = CrateDetails::new(&mut db.conn(), "foo", version).unwrap();
-                assert_eq!(details.latest_release().version, semver::Version::parse("0.0.3")?);
+                assert_eq!(
+                    details.latest_release().version,
+                    semver::Version::parse("0.0.3")?
+                );
             }
 
             Ok(())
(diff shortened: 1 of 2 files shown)


shorsher:
Can someone assign @Nemo157, I'd like his feedback. It was suggested checking each version for the following criteria:

```
    latest non-prerelease non-yanked version
    latest prerelease non-yanked version
    latest non-prerelease yanked version
    latest prerelease yanked version
```

However, since we sort releases in descending order it would still pick up prereleased/yanked releases if they are the latest. I think this should create an acceptable outcome, which is only consider the latest yanked/prereleased release if that is all that exist.

I

Nemo157 (approved):
Looks good to me, I think my 4-step lookup process was probably over-engineered :grin:, if you only have some set of yanked/prerelease versions available with no stable-unyanked version, you probably don't care about exactly which version is shown as latest.

PR #1000, opened 2020-08-23
Separate metadata parsing into a library

jyn514:
This is intended to eventually be published and used by crater for https://github.com/rust-lang/crater/issues/532.

r? @pietroalbini


jyn514:
This is ready for review.

src/docbuilder/metadata/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
pietroalbini, line 1: We should make the metadata crate part of a workspace instead of ignoring the `Cargo.lock`.

src/docbuilder/metadata/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[lib]
+path = "lib.rs"
+
+[dependencies]
+toml = { version = "0.5", default-features = false }
+thiserror = "1"
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--document-private-items"]
```
pietroalbini, line 20: Is there a reason why? People would mistake that for the crate's public API.

src/docbuilder/metadata/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "metadata"
```
pietroalbini, line 2: To publish the crate on crates.io I'd call it `docsrs-metadata`.

Nemo157, line 2: If this is going to be published, it needs a less generic name.

src/docbuilder/metadata/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    /// Return the environment variables that should be set when building this crate.
+    pub fn environment_variables(&self) -> HashMap<&'static str, String> {
+        let joined = |v: &Option<Vec<_>>| v.as_ref()
+            .map(|args| args.join(" "))
+            .unwrap_or_default();
+
+        let mut map = HashMap::new();
+        map.insert("RUSTFLAGS", joined(&self.rustc_args));
+        map.insert("RUSTDOCFLAGS", joined(&self.rustdoc_args));
```
pietroalbini, line 243: We're not adding `-Z unstable-options`, so the user's `RUSTDOCFLAGS` could not work. Also, we're missing a lot of other flags set by docs.rs.

src/docbuilder/metadata/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    /// See [`BuildTargets`].
+    default_target: Option<String>,
+    targets: Option<Vec<String>>,
+
+    /// List of command line arguments for `rustc`.
+    pub rustc_args: Option<Vec<String>>,
+
+    /// List of command line arguments for `rustdoc`.
+    pub rustdoc_args: Option<Vec<String>>,
```
pietroalbini, line 118: Hmm, why are the fields public?

pietroalbini (changes requested):
Other than the comments I left, I'd put the crate on the root directory:

```
/docsrs-metadata
/src
...
```

It's quite confusing to have other crates nested inside `src/` (it's the same for `badge`).

src/docbuilder/metadata/.gitignore
pietroalbini: This is also preventing running the metadata tests with `cargo test --workspace`. We're not executing them at all on CI at the moment.

pietroalbini:
We'll also need to change `cargo test --locked` to `cargo test --workspace --locked` on CI.

src/docbuilder/metadata/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+    "i686-pc-windows-msvc",
+    "i686-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-linux-gnu",
+];
+
+/// The possible errors for [`Metadata::from_crate_root`].
+#[derive(Debug, Error)]
+pub enum MetadataError {
```
Nemo157, line 65: `#[non_exhaustive]`?

src/docbuilder/metadata/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+                .get("metadata")?
+                .as_table()?
+                .get("docs")?
+                .as_table()?
+                .get("rs")?
+                .as_table()
+        }
+
+        if let Some(table) = fetch_manifest_tables(&manifest) {
+            // TODO: all this `to_owned` is inefficient, this should use explicit matches instead.
```
Nemo157, line 274: Could this block just be [`table.try_into()?`](https://docs.rs/toml/0.5.6/toml/value/enum.Value.html#method.try_into) with `derive(Deserialize)`?

src/docbuilder/metadata/Cargo.toml
jyn514: It's left over from when the docs on `BuildTargets` were still on the private fields. I'll remove it.

src/docbuilder/metadata/lib.rs
jyn514: The other flags are set in the rustwide builder: https://github.com/rust-lang/docs.rs/pull/1000/files#diff-8fac5c4dd8b0132a470c525431c6c4bcR609-R611. My reasoning is that crater doesn't need these flags, so they don't need to be part of the library (and moving over anything besides `-Z unstable-options` will be annoying because it needs access to the full TOML, not just `[package.metadata.docs.rs]`). Do you want -Z unstable-options in the library?

src/docbuilder/metadata/lib.rs
jyn514: This lets the API user add more flags if they want to. It also makes it possible to construct a `Metadata` without having a TOML file already.

jyn514:
> It's quite confusing to have other crates nested inside src/ (it's the same for badge).

Good point, I'll change it here and make a follow-up PR for badge.

src/docbuilder/metadata/lib.rs
jyn514: Is it non-exhaustive? I don't see this ever expanding in the future. If for some reason it does we can always make a major version bump.

src/docbuilder/metadata/.gitignore
jyn514: Can do, but I found a bug in clippy while changing this - it will ignore `--all` and only run clippy on the top-level directory. I'll change CI to run it once for each crate.

src/docbuilder/metadata/lib.rs
jyn514: I'm not sure what you mean here. This is a `Value` we already have, not a string, there's no deserializers in `toml` I see that take a `Value`.

src/docbuilder/metadata/Cargo.toml
jyn514: Done, I changed it to `docsrs-metadata`.

jyn514:
Updated!

src/docbuilder/metadata/lib.rs
jyn514: Oh I missed the link - good idea, I'll try that

jyn514: Done!

.github/workflows/ci.yml
```diff
@@ -58,10 +58,10 @@ jobs:
           psql "${CRATESFYI_DATABASE_URL}"
 
       - name: Build
-        run: cargo build --locked
+        run: cargo build --all --locked
```
Nemo157, line 61: `--all` is deprecated, `--workspace` is the replacement

src/docbuilder/metadata/lib.rs
pietroalbini: You can't construct the metadata since `default_target` and `targets` are not public.

metadata/lib.rs
```diff
@@ -0,0 +1,629 @@
+#![deny(missing_docs)]
```
Nemo157, line 1: Can this just be `warn` and rely on the `-Dwarnings` in CI to avoid blocking developers locally.

src/docbuilder/metadata/lib.rs
pietroalbini: I'd prefer to avoid making major version bumps if we can avoid id.

metadata/lib.rs
jyn514: Good idea, done.

src/docbuilder/metadata/.gitignore
jyn514: Err this was just me misunderstanding how clippy works, it does listen to `--all`.

src/docbuilder/metadata/lib.rs
jyn514: `let metadata = Metadata { your_fields, ..Metadata::default() };`

But if you don't think it's useful, I can remove it.

pietroalbini: Yeah I'd avoid making the fields public.

jyn514: The only downside here is that the documentation for the fields no longer shows up on docs.rs, you have to read the source.

src/docbuilder/metadata/lib.rs
jyn514: Added

src/docbuilder/metadata/lib.rs
pietroalbini: Well, the documentation for those fields is in https://docs.rs/about/metadata

src/docbuilder/metadata/lib.rs
pietroalbini: So, this is tricky: ideally I'd prefer to have all the flags being returned by `Metadata`, so Crater and other tools can fully test the exact setup docs.rs uses. On the other hand, docs.rs relies on unstable flags, and we often run Crater on the beta branch, so adding `-Z unstable-options` would prevent this from being used in beta runs :/

jyn514: If it helps, the only difference this should make is for users passing unstable `RUSTDOCFLAGS` in CI, which seems like a pretty niche thing to do. Most of the unstable options affect only what documentation is _generated_, not the code rustdoc sees.

src/docbuilder/metadata/lib.rs
jyn514: Ooh good idea, let me link to that.

src/docbuilder/metadata/lib.rs
Nemo157: Most of the rustdocflags I've seen in the wild enable unstable compiler features via `cfg_attr` (mostly `#![feature(doc_cfg)]`), so do affect the code rustdoc sees.

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: fmt
-          args: --workspace -- --check
+          args: --all -- --check
```
Nemo157, line 91: `--workspace` behaviour is actually the default, `--all` does something different (finds `path` dependencies and formats those, even outside the workspace), https://github.com/rust-lang/rustfmt/issues/3911

jyn514: Wow, that's confusing 😆

src/docbuilder/metadata/lib.rs
pietroalbini: Or I guess we could add everything to `.environment_variables()` and trick Crater to think it's on nightly even on beta when the "docs.rs mode" is used.

jyn514: `stable()` and `unstable()` doesn't seem great because we don't know which of the user flags are which, we'd have to put them all in `unstable()`. I think the best way is to use `RUSTC_BOOTSTRAP` in crater, yeah.

pietroalbini: Ok, let's move all environment variables handling in the library then.

Cargo.toml
```diff
@@ -8,6 +8,8 @@ repository = "https://github.com/rust-lang/docs.rs"
 build = "build.rs"
 edition = "2018"
 
+[workspace]
```
Kixiron, line 11: What's the purpose of the empty workspace?

metadata/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    // Set the host target
+    println!(
+        "cargo:rustc-env=DOCS_RS_METADATA_HOST_TARGET={}",
```
Kixiron, line 4: For using in a library context I'd prefix the env var with `__` on the astronomical chance there's a conflict and just to communicate that this is a "don't touch" variable

src/docbuilder/mod.rs
```diff
@@ -1,12 +1,10 @@
 mod crates;
 mod limits;
-mod metadata;
 pub(crate) mod options;
 mod queue;
 mod rustwide_builder;
 
 pub(crate) use self::limits::Limits;
-pub(self) use self::metadata::Metadata;
```
Kixiron, line 9: I'd remove the re-exporting of a library if at all possible

src/docbuilder/metadata/lib.rs
jyn514: Sorry - I was suggesting to move `-Z unstable-options` into the library, not `--extern-html-root`. Aside from the mess that makes on the command line, it's not relevant at all to crater since it will still see the same code either way, only the generated links will be different. And it will be very hard to make that part of the library.

metadata/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+pub const DEFAULT_TARGETS: &[&str] = &[
+    "i686-pc-windows-msvc",
+    "i686-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-linux-gnu",
+];
+
+/// The possible errors for [`Metadata::from_crate_root`].
+#[derive(Debug, Error)]
```
Kixiron, line 66: We shouldn't need to use thiserror here, implementing `std::error::Error` should be fine & pretty small and would allow a lib with less dependencies

metadata/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+/// ```
+///
+/// You can define one or more fields in your `Cargo.toml`.
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct Metadata {
+    /// List of features to pass on to `cargo`.
+    ///
+    /// By default, docs.rs will only build default features.
+    features: Option<Vec<String>>,
```
Kixiron, line 104: I'd rather this be named `docsrs_features` or `enabled_features` since I'm going to be having this parse all features soon

Cargo.toml
jyn514: https://doc.rust-lang.org/cargo/reference/workspaces.html#workspaces

> A workspace can be created by adding a [workspace] section to Cargo.toml. This can be added to a Cargo.toml that already defines a [package], in which case the package is the root package of the workspace. The workspace root is the directory where the workspace's Cargo.toml is located.

metadata/lib.rs
jyn514: I tried doing this at first and it exploded into another 100 lines of code. Since we already depend on thiserror through docs.rs I don't think it's a big deal here.

src/docbuilder/mod.rs
jyn514: It is removed? I don't understand the comment.

metadata/build.rs
jyn514: I'm pretty sure this environment variable is only set for docs.rs, not for anyone else: https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-envvarvalue. The `DOCS_RS_METADATA` prefix is only so it doesn't conflict with environment variables you set in your shell, but this will override those anyway.

metadata/lib.rs
jyn514: It's named after the format in `Cargo.toml`, which can't be changed without breaking all crates. I can use `serde(rename = "enabled_features")` if you prefer?

Cargo.toml
Kixiron: Well yah, but why a workspace at all since this is essentially just a dependency

src/docbuilder/mod.rs
Kixiron: The GH app was lying to me, never mind

metadata/lib.rs
Kixiron: Does this get the features table of the Cargo.toml or the `package.metadata.docs.rs.features` section?

metadata/lib.rs
Kixiron: Strange, it should just be this
```rust
use std::{io, fmt};

#[derive(Debug)]
pub enum MetadataError {
    /// The error returned when the manifest could not be read.
    IO(io::Error),
    /// The error returned when the manifest could not be parsed.
    Parse(toml::de::Error),
}
impl fmt::Display for MetadataError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::IO(err) => write!(f, "failed to read manifest from disk: {}", err),
            Self::Parse(err) => write!(f, "failed to parse manifest: {}", err),
        }
    }
}
impl std::error::Error for MetadataError {}
```

src/docbuilder/metadata/lib.rs
jyn514: Pushed a commit adding `-Z unstable-options`, let me know if you still have concerns. IMO the remaining options don't make sense to use with crater:
- `-j2`
- `--target`
- `--static-root-path /`
- `--cap-lints warn`
- `--resource-suffix`
- `--extern-html-root-url`

Cargo.toml
jyn514: See @pietroalbini's comments above - without a workspace the tests won't be run in CI. https://github.com/rust-lang/docs.rs/pull/1000#discussion_r475421475

metadata/lib.rs
jyn514: No, it also requires `From` impls and a `cause()` function for `MetadataError`. I really don't think adding boilerplate helps any.

metadata/lib.rs
jyn514: The second, `package.metadata.docs.rs.features`.

jyn514:
Is this waiting on anything? @Kixiron did I address your suggestions or is there something else I should do?

pietroalbini:
I'll take a look at this tomorrow.

pietroalbini (approved):

Issue #1001, opened 2020-08-25
wavecar_rs needs more build time


Ionizing:
**Crate name:** wavecar_rs
**Build failure link:** https://docs.rs/crate/wavecar_rs/0.1.0/builds/285405

**Requested RAM limit:** 3GB
**Requested timeout:** 30min
**Requested number of targets:** 10

**Why your crate needs the resource increases:** That crate depends on `fftw` and `openblas`, which consumes a lot of time during the build.

jyn514:
30 minutes is longer than we normally extend for. Would 20 or 25 minutes work instead?

Note that this is for each target, so it needs to be multiplied by 5 - please first set `targets = []` in `package.metadata.docs.rs` so that your crate only builds for one target.

Ionizing:
> Would 20 or 25 minutes work instead?

Okay, 25 minutes may solve it. If it still fails I would try other solutions.

> please first set targets = [] in package.metadata.docs.rs

Okay, I'll fix it soon.

jyn514:
In the meantime I can limit it in production so you don't have to make a new release.

Ionizing:
Thank you very much~

Ionizing:
It seems the doc build is completed successfully, thanks for your help!

PR #1002, opened 2020-08-25
Document that time extensions will be limited to one target

jyn514:
Thought of this in https://github.com/rust-lang/docs.rs/issues/1001, it seems like a good policy.
r? @pietroalbini

### DIFF
--- a/src/docbuilder/limits.rs
+++ b/src/docbuilder/limits.rs
@@ -37,11 +37,14 @@ impl Limits {
             if let Some(memory) = row.get::<_, Option<i64>>("max_memory_bytes") {
                 limits.memory = memory as usize;
             }
-            if let Some(timeout) = row.get::<_, Option<i32>>("timeout_seconds") {
+            let timeout = row.get::<_, Option<i32>>("timeout_seconds");
+            if let Some(timeout) = timeout {
                 limits.timeout = Duration::from_secs(timeout as u64);
             }
             if let Some(targets) = row.get::<_, Option<i32>>("max_targets") {
                 limits.targets = targets as usize;
+            } else if timeout.is_some() {
+                limits.targets = 1;
             }
         }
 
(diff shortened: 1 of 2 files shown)


jyn514:
image

Kixiron:
If we're going to have a policy, then a method of "enforcement" should probably be implemented because there's nothing stopping someone from getting an increase and then switching to a higher number of targets later on (either forgetting that they have an increase or maliciously)

jyn514:
Actually this can be enforced in a database constraint too, let me do that.

jyn514:
@Kixiron there is already a way to enforce it: `update sandbox_overrides set max_targets = 1 where timeout_seconds is not null;`. But that requires manual action - if I add a database constraint I can't forget to run that command.

pietroalbini:
I wouldn't put a hard constraint on the database schema. What we *could* do is change the limits code so that when the number of targets is not overridden, it defaults to `n` when the timeout is not overridden and to `1` when it is.

Kixiron:
I think displaying to the user on the builds page that their builds are limited is important too if we aren't

jyn514:
> I think displaying to the user on the builds page that their builds are limited is important too if we aren't

We already do: https://docs.rs/crate/haxcel/0.1.0/builds

Kixiron:
I mean specially limited, like in the situation where we extend build time in exchange for a single target (only if we don't already)

jyn514:
Do you think that's necessary? We already document the effective limits build page and the policy on /about/builds. Putting in on a successful build seems like overkill.

Kixiron:
I mean on a crate-by-crate basis, if I own the crate `foo-bar-baz` and I get time extended, when I look at my builds page I should see
| Limit | Setting |
| :- | :-: |
| Maximum rustdoc execution time | 10 hours |
| Maximum number of build targets | 1 |

jyn514:
That's what it says already, that's the link I gave.

jyn514:
The limits for that page are calculated the same way they're calculated for builds, they don't show exactly what's in the database.

jyn514:
@pietroalbini is this waiting on anything?

pietroalbini:
Looks good to me.

PR #1016, opened 2020-08-29
Remove DocBuilder, part 1

pietroalbini:
This PR is the first of the two PRs I plan to send to remove `DocBuilder` and `DocBuilderOptions` from the docs.rs codebase: both those structs serve as "intermediaries" between `RustwideBuilder` and the rest of the application, but now that we have `Config` and `Context` that's not really needed anymore. Not having the middle layer will help with building buildtests a lot.

Notable changes:

* `cratesfyi build --skip-if-log-exists <command>` was removed: the `--skip-if-log-exists` flag was really fragile, relying on a text log of all the builds updated every 10 builds. There is also no need I can see for it, as the practically equivalent (and better) `--skip-if-exists` already exists.
* `cratesfyi build --skip-if-exists <command>` was improved: now it checks if a *successful build* is present, instead of checking whether the release metadata is stored in the database.
* All the `RustwideBuilder` configuration was moved over to `Config`.
* All of `DocBuilderOptions` was moved over to `Config`.
* The `Index` is now stored in `Context` and a single instance is reused across the whole application. This won't affect the file descriptors problem, as the underlying git instance is still re-created every time it is accessed.

In the next PR I'll move away all the rest of `DocBuilder`. This PR is best reviewed commit-by-commit.


src/utils/queue_builder.rs
```diff
@@ -48,21 +48,13 @@ pub fn queue_builder(
 
         if status.count() >= 10 {
             // periodically, we need to flush our caches and ping the hubs
-            debug!("10 builds in a row; flushing caches");
+            debug!("10 builds in a row; pinging pubsubhubhub");
             status = BuilderState::QueueInProgress(0);
 
             match pubsubhubbub::ping_hubs() {
```
jyn514, line 47: I didn't know we had this 😆 do we know anyone using this? I'd rather have #809 I think.

src/utils/queue_builder.rs
```diff
@@ -48,21 +48,13 @@ pub fn queue_builder(
 
         if status.count() >= 10 {
             // periodically, we need to flush our caches and ping the hubs
```
jyn514, line 43:
```suggestion
            // periodically, ping the hubs
```

src/docbuilder/rustwide_builder.rs
```diff
@@ -648,6 +649,29 @@ impl RustwideBuilder {
...
+        docbuilder: &DocBuilder,
+        name: &str,
+        version: &str,
+    ) -> Result<bool> {
+        if docbuilder.options().skip_if_exists {
+            // Check whether no successful builds are present in the database.
+            Ok(conn
+                .query(
+                    "SELECT 1 FROM crates, releases, builds
+                     WHERE crates.id = releases.crate_id AND releases.id = builds.rid
```
jyn514, line 642: This needs an `INNER JOIN` or you'll get duplicates.

src/config.rs
```diff
@@ -64,6 +70,12 @@ impl Config {
             // https://github.com/rust-lang/docs.rs/pull/930#issuecomment-667729380
             max_parse_memory: env("DOCSRS_MAX_PARSE_MEMORY", 5 * 1024 * 1024)?,
             registry_gc_interval: env("DOCSRS_REGISTRY_GC_INTERVAL", 60 * 60)?,
+
+            rustwide_workspace: env("CRATESFYI_RUSTWIDE_WORKSPACE", PathBuf::from(".workspace"))?,
```
jyn514, line 74: Eventually we should change this to `DOCSRS_RUSTWIDE_WORKSPACE`. But no need to do that here.

src/docbuilder/rustwide_builder.rs
```diff
@@ -61,59 +59,45 @@ const DUMMY_CRATE_VERSION: &str = "1.0.0";
...
-        let workspace_path = env_workspace_path
-            .as_ref()
-            .map(|v| v.as_str())
-            .unwrap_or(DEFAULT_RUSTWIDE_WORKSPACE);
-        let is_docker = std::env::var("DOCS_RS_DOCKER")
-            .map(|s| s == "true")
-            .unwrap_or(false);
-        let mut builder = WorkspaceBuilder::new(Path::new(workspace_path), USER_AGENT)
-            .running_inside_docker(is_docker);
-        if let Ok(custom_image) = std::env::var("DOCS_RS_LOCAL_DOCKER_IMAGE") {
```
jyn514, line 83: This is great :D

src/bin/cratesfyi.rs
```diff
@@ -297,19 +278,17 @@ enum BuildSubcommand {
...
+
         let rustwide_builder = || -> Result<RustwideBuilder, Error> {
-            Ok(RustwideBuilder::init(
-                ctx.config()?,
-                ctx.pool()?,
-                ctx.metrics()?,
-                ctx.storage()?,
-            )?)
+            let mut builder =
+                RustwideBuilder::init(ctx.config()?, ctx.pool()?, ctx.metrics()?, ctx.storage()?)?;
```
jyn514: This has the same duplication you removed in the previous commit because it's a closure in a separate function. Could you make `rustwide_builder` a method on `Context` instead so it can be reused here?

src/utils/daemon.rs
```diff
@@ -51,14 +51,10 @@ fn start_registry_watcher(opts: DocBuilderOptions, context: &dyn Context) -> Res
 
 pub fn start_daemon(context: &dyn Context, enable_registry_watcher: bool) -> Result<(), Error> {
     let config = context.config()?;
-    let dbopts = DocBuilderOptions::new(config.prefix.clone(), config.registry_index_path.clone());
-
-    // check paths once
-    dbopts.check_paths().unwrap();
```
jyn514, line 57: Hmm, shouldn't we keep this check in? What happens now if the path doesn't exist?

src/bin/cratesfyi.rs
jyn514: Oops, you did that in a later commit 😆 `RustwideBuilder::init` taking a context is just as good.

src/utils/daemon.rs
pietroalbini: The index fails to initialize.

src/utils/queue_builder.rs
pietroalbini: Dunno why it exists, but let's not remove it for now until we get more insights on who's using it.

src/docbuilder/rustwide_builder.rs
pietroalbini: Hmm? This is working as expected when testing manually. Why would you expect it to fail?

jyn514: Huh, I expected it to fail when you have three different crates, each of which satisfies _one_ of these conditions.
```
CREATE TEMPORARY TABLE crates (
  id SERIAL PRIMARY KEY,
  name VARCHAR(255) UNIQUE NOT NULL
);

CREATE TEMPORARY TABLE releases (
   id SERIAL PRIMARY KEY,
   crate_id INT NOT NULL REFERENCES crates(id),
   version VARCHAR(100)
);

CREATE TEMPORARY TABLE builds (
   id SERIAL,
   rid INT NOT NULL REFERENCES releases(id),
   build_status BOOL NOT NULL
);

insert into crates (name) values ('a'), ('b');
insert into releases (crate_id, version) values (1, '0.1.0'), (2, '0.1.0');
insert into builds (rid, build_status) values (1, true), (2, true);

SELECT 1 FROM crates, releases, builds
WHERE crates.id = releases.crate_id AND releases.id = builds.rid
  AND crates.name = 'a' AND releases.version = '0.1.0'  AND builds.build_status = TRUE;
```
But that doesn't happen - the `WHERE id = crate_id AND id = rid` filters out the duplicates. It's more clear when you select everything instead of `1`:
```
select * from crates, releases, builds WHERE crates.id = releases.crate_id AND releases.id = builds.rid;
 id | name | id | crate_id | version | id | rid | build_status 
----+------+----+----------+---------+----+-----+--------------
  1 | a    |  1 |        1 | 0.1.0   |  1 |   1 | f
  2 | b    |  2 |        2 | 0.1.0   |  2 |   2 | t
joshua=> select * from crates, releases, builds WHERE crates.name = 'a' AND releases.version = '0.1.0'  AND builds.build_status = TRUE;
 id | name | id | crate_id | version | id | rid | build_status 
----+------+----+----------+---------+----+-----+--------------
  1 | a    |  1 |        1 | 0.1.0   |  2 |   2 | t
  1 | a    |  2 |        2 | 0.1.0   |  2 |   2 | t
```
So this _works_, it's just confusing.

PR #1017, opened 2020-08-30
Update to lockfile version 2

Nemo157:
New nightly cargos are finally updating to v2 when they have to change the lockfile, so retaining compatibility with v1 requires bringing in `cargo lock translate` to convert it back to v1 (previously they would rewrite a v1 lockfile if the existing one was v1). This has been supported since around 1.39 so it shouldn't be a breaking change for anyone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,4171 +4,4169 @@
 name = "ab_glyph_rasterizer"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2b7e4e8cf778db814365e46839949ca74df4efb10e87ba4913e6ec5967ef0285"
 
 [[package]]
 name = "addr2line"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


jyn514:
How can I verify that this has the same dependencies as the original lockfile and cargo didn't bump the versions during the switch?

jyn514:
Or alternatively, you could first make another PR running `cargo update` so that the lockfile cargo generates by default would match the one here.

Nemo157:
you can `git show 6d459cb | rg '\+version'` to see that it doesn't change any of the `version=` lines

jyn514:
```
git show 6d459cb3ae4bb3378cdc7a816f463cf7a98aeb37 | grep '[+-]version'
- "unic-ucd-version 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unic-ucd-version",
 name = "unic-ucd-version"
-"checksum unic-ucd-version 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)" = "96bd2f2237fe450fcd0a1d2f5f4e91711124f7857ba2e964247776ebeeb7b0c4"
```
Looks good to me :)

PR #1019, opened 2020-08-30
Add metrics tracking number of recently loaded crates/versions/platforms

Nemo157:
These metrics are intended to be useful for evaluating changes for #1004. Knowing how many different crates get accessed over short timespans could help us know things like the potential memory cost of keeping cached indexes around short term.

I'm not sure what the performance impact of these might be in production. The most likely issue would be a large memory usage if the number of different platforms requested per-hour is very high, leading to a lot of small strings kept around to track these.

(based on #1017, so to review the lockfile changes look at the last commit in isolation)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,13 @@ mod test;
 pub mod utils;
 mod web;
 
+#[allow(dead_code)]
+mod target {
+    //! [`crate::target::TargetAtom`] is an interned string type for rustc targets, such as
+    //! `x86_64-unknown-linux-gnu`. See the [`string_cache`] docs for usage examples.
+    include!(concat!(env!("OUT_DIR"), "/target_atom.rs"));
+}
+
 use web::page::GlobalAlert;
 
 // Warning message shown in the navigation bar of every page. Set to `None` to hide it.
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -48,7 +48,7 @@ metrics! {
         pub(crate) rustdoc_rendering_times: HistogramVec["step"],
 
         /// Count of recently accessed crates
-        pub(crate) recent_krates: IntGaugeVec["duration"],
+        pub(crate) recent_crates: IntGaugeVec["duration"],
         /// Count of recently accessed versions of crates
         pub(crate) recent_versions: IntGaugeVec["duration"],
         /// Count of recently accessed platforms of versions of crates
@@ -130,7 +130,7 @@ impl RecentReleases {
                 .set(five_minute_count);
         }
 
-        inner(&self.krates, &metrics.recent_krates);
+        inner(&self.krates, &metrics.recent_crates);
         inner(&self.versions, &metrics.recent_versions);
         inner(&self.platforms, &metrics.recent_platforms);
     }
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -3,6 +3,7 @@ mod macros;
 
 use self::macros::MetricFromOpts;
 use crate::db::Pool;
+use crate::target::TargetAtom;
 use crate::BuildQueue;
 use dashmap::DashMap;
 use failure::Error;
@@ -81,7 +82,7 @@ metrics! {
 pub(crate) struct RecentlyAccessedReleases {
     krates: DashMap<i32, Instant>,
     versions: DashMap<i32, Instant>,
-    platforms: DashMap<(i32, String), Instant>,
+    platforms: DashMap<(i32, TargetAtom), Instant>,
 }
 
 impl RecentlyAccessedReleases {
@@ -93,7 +94,7 @@ impl RecentlyAccessedReleases {
         self.krates.insert(krate, Instant::now());
         self.versions.insert(version, Instant::now());
         self.platforms
-            .insert((version, target.to_owned()), Instant::now());
+            .insert((version, TargetAtom::from(target)), Instant::now());
     }
 
     pub(crate) fn gather(&self, metrics: &Metrics) {
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -91,10 +91,11 @@ impl RecentlyAccessedReleases {
     }
 
     pub(crate) fn record(&self, krate: i32, version: i32, target: &str) {
-        self.crates.insert(krate, Instant::now());
-        self.versions.insert(version, Instant::now());
+        let now = Instant::now();
+        self.crates.insert(krate, now);
+        self.versions.insert(version, now);
         self.platforms
-            .insert((version, TargetAtom::from(target)), Instant::now());
+            .insert((version, TargetAtom::from(target)), now);
     }
 
     pub(crate) fn gather(&self, metrics: &Metrics) {
@@ -116,7 +117,7 @@ impl RecentlyAccessedReleases {
                 }
 
                 // Only retain items accessed within the last hour
-                return elapsed < Duration::from_secs(60 * 60);
+                elapsed < Duration::from_secs(60 * 60)
             });
 
             metric.with_label_values(&["one hour"]).set(hour_count);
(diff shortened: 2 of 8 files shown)


jyn514:
> The most likely issue would be a large memory usage if the number of different platforms requested per-hour is very high, leading to a lot of small strings kept around to track these.

We already run high at about 5 GB in prod. I find it hard to think the number of crate/version/platform pairs would take up anything close to that, do you have a guesstimate for 'lot of small strings'?

src/metrics/mod.rs
```diff
@@ -44,6 +47,13 @@ metrics! {
         /// The time it takes to render a rustdoc page
         pub(crate) rustdoc_rendering_times: HistogramVec["step"],
 
+        /// Count of recently accessed crates
+        pub(crate) recent_krates: IntGaugeVec["duration"],
```
pietroalbini, line 51: Nit: let's call this recent_crates.

pietroalbini:
I think we can avoid the problem with strings by not storing them at all. My understanding is we don't care about the crate names/versions, so we can just store `i32`s:

* For crates we could map the crate ID with the instant.
* For versions we could map the release ID with the instant.
* For targets we could map (release ID, interned target name) with the instant.

jyn514:
When you get a chance, please rebase over master so the diff no longer shows up as 1500 lines changed ;)

src/metrics/mod.rs
```diff
@@ -67,6 +78,64 @@ metrics! {
     namespace: "docsrs",
 }
 
+#[derive(Debug, Default)]
+pub(crate) struct RecentlyAccessedReleases {
+    krates: DashMap<i32, Instant>,
```
jyn514, line 83: This could be `crates` too.
```suggestion
    crates: DashMap<i32, Instant>,
```

src/metrics/mod.rs
```diff
@@ -67,6 +78,64 @@ metrics! {
...
+                if elapsed > Duration::from_secs(30 * 60) {
+                    return true;
+                }
+                half_hour_count += 1;
+                if elapsed > Duration::from_secs(5 * 60) {
+                    return true;
+                }
+                five_minute_count += 1;
+                true
+            });
```
jyn514, line 120: This is really confusing. What is it doing? Why does it return `false` after an hour? Why does it return early if _more_ than the time elapsed has passed - couldn't it skip recording an interval?

Nemo157: It goes through from longest to shortest interval, checking if the value is within the interval, short-circuiting and exiting if it's not, otherwise incrementing that counter and moving on to check the next interval. The first check returns `false` to remove the entry from the map, the others return `true` to keep it for future checks.

pietroalbini:
Other than Joshua's comments, this looks good to me!

src/metrics/mod.rs
Nemo157: I can probably rewrite it to be more obvious, and just rely on the optimizer to reorder the duplicate checks that that would need.

src/metrics/mod.rs
```diff
@@ -104,19 +104,19 @@ impl RecentlyAccessedReleases {
...
-                if elapsed > Duration::from_secs(5 * 60) {
-                    return true;
+                if elapsed < Duration::from_secs(5 * 60) {
+                    five_minute_count += 1;
                 }
-                five_minute_count += 1;
-                true
+
+                // Only retain items accessed within the last hour
+                return elapsed < Duration::from_secs(60 * 60);
```
jyn514, line 119:
```suggestion
                elapsed < Duration::from_secs(60 * 60)
```

src/lib.rs
```diff
@@ -26,6 +26,13 @@ mod test;
 pub mod utils;
 mod web;
 
+#[allow(dead_code)]
```
jyn514, line 29: Why is this here?

src/metrics/mod.rs
```diff
@@ -67,6 +78,64 @@ metrics! {
...
+impl RecentlyAccessedReleases {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    pub(crate) fn record(&self, krate: i32, version: i32, target: &str) {
+        self.crates.insert(krate, Instant::now());
+        self.versions.insert(version, Instant::now());
+        self.platforms
+            .insert((version, TargetAtom::from(target)), Instant::now());
```
jyn514, line 97: Let's make these all the same instant.
```suggestion
        let now = Instant::now();
        self.crates.insert(krate, now);
        self.versions.insert(version, now);
        self.platforms.insert((version, TargetAtom::from(target)), now);
```

src/metrics/mod.rs
```diff
@@ -67,6 +78,64 @@ metrics! {
...
+    platforms: DashMap<(i32, TargetAtom), Instant>,
+}
+
+impl RecentlyAccessedReleases {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    pub(crate) fn record(&self, krate: i32, version: i32, target: &str) {
+        self.crates.insert(krate, Instant::now());
```
jyn514, line 94: Should this have a cap on the number of crates recorded? Right now it will grow without bound and only evict entries each hour, which seems like it would be really easy to DOS.

On a related point, is it possible to do all this calculation on the grafana side? That way we'd only have to hold on to the entries long enough to send them to `metrics`, and could delete them immediately after.

src/lib.rs
Nemo157: There's a lot of unused `const`s since we're not using the `match` based way of looking up interned strings.

src/metrics/mod.rs
Nemo157: It evicts entries every time metrics are requested, if they're an hour old. But I can probably throw some massive limit on there just in case (say 64 bytes per entry (4 byte key, 8 byte hash, 16 byte value, padding, overhead), 100,000 entries = 6.4M, times two for hashmap overhead, times two for margin = 25.6M per map; I'm less sure about what that would do to the metrics response times going through all those entries every time, so maybe 10,000 as a more conservative limit; some very quick benchmarking on my machine shows perfect linear scaling and ~11ms for 100,000 entries in all 3 maps).

I have no idea how you'd do it in grafana, it doesn't fit into any of the prometheus metric types directly. (If we had logs going into kibana, then I'd have some idea about how to pull this data from there instead of tracking it in service).

jyn514: 75 MB seems like more than enough and won't affect our memory usage too much, let's go with that. Ideally it would be a LRU cache but if that's hard to do I think it's fine to just refuse to record more entries.

src/metrics/mod.rs
```diff
@@ -91,10 +91,16 @@ impl RecentlyAccessedReleases {
     }
 
     pub(crate) fn record(&self, krate: i32, version: i32, target: &str) {
-        self.crates.insert(krate, Instant::now());
-        self.versions.insert(version, Instant::now());
+        if self.platforms.len() > 100_000 {
+            // Avoid filling the maps _too_ much, we should never get anywhere near this limit
```
jyn514, line 95:
```suggestion
            // Avoid filling the maps _too_ much, we should never get anywhere near this limit
            debug!("more than 100_000 hits in the last hour, disabling further metrics");
```

Nemo157: We will see it capping out on the metrics, I think it would be a waste to spam a message per-request to the logs too.

jyn514:
Looks good to me after a rebase :)

src/web/rustdoc.rs
```diff
@@ -415,16 +415,24 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
 
+    metrics
+        .recently_accessed_releases
+        .record(krate.crate_id, krate.release_id, target);
+
+    let target = if target == "" {
+        String::new()
+    } else {
+        format!("{}/", target)
+    };
```
Nemo157, line 434: this `target` handling was the only major conflict.

Nemo157 (commented):
Rebased, some minor conflicts but I've identified the only major conflict below, both this and `master` had started tracking the `target` when extracting that from the URL but this wants it without trailing `/`.

PR #1022, opened 2020-09-01
ignore rustwide from Cargo's workspace

pietroalbini:


Cargo.toml
```diff
@@ -9,6 +9,10 @@ build = "build.rs"
 edition = "2018"
 
 [workspace]
+exclude = [
+    "ignored",
+    ".workspace",
+]
```
jyn514, line 15: Can we instead use `members =` so we don't have to worry about any other folders the user has?
```suggestion
members = ["metadata"]
```

jyn514:
For posterity: this is fallout from #1000.

Cargo.toml
pietroalbini: Dunno why, but members doesn't work.

jyn514: This is https://github.com/rust-lang/cargo/issues/5418, it seems like there's no way to override this 🤦

jyn514: Ok, in that case looks good once the docker workspace is ignored too.
```suggestion
exclude = [
    "ignored",
    ".workspace",
    ".rustwide-docker",
]
```

PR #1023, opened 2020-09-01
Add missing build command

GuillaumeGomez:


README.md
```diff
@@ -68,6 +68,8 @@ docker-compose up -d db s3
 # including running the migrations and installing the nightly Rust toolchain.
 # This will take a while the first time but will be cached afterwards.
 cargo run -- build crate regex 1.3.1
+# Generate important files for the web navigation
+cargo run -- build add-essential-files
```
jyn514, line 72: `build crate` should do this automatically ... did `build crate` succeed for you locally?

Nemo157: Ah, it is meant to, but it fails the check and doesn't
```
[src/docbuilder/rustwide_builder.rs:162] old_version.as_deref() = Some(
    "rustc 1.48.0-nightly (d006f5734 2020-08-28)",
)
[src/docbuilder/rustwide_builder.rs:162] Some(self.rustc_version.as_str()) = Some(
    "rustc 1.48.0-nightly (d006f5734 2020-08-28)",
)
[src/docbuilder/rustwide_builder.rs:162] dbg!(old_version . as_deref()) != dbg!(Some(self . rustc_version . as_str())) = false
```
I think this is if you clear your docs.rs data, but not your rustwide data.

Nemo157: Yes, if I `rm -rf .workspace` then do a build on a fresh db it will add essential files automatically.

Nemo157: (probably means the check for whether to add essential files should be comparing the "rustc resource suffix" and the current toolchain or something instead of old and new toolchain versions).

jyn514: Ugh, ok. Let's just recommend always running `add-essential-files` then, this sounds painful to fix automatically.

jyn514: > (probably means the check for whether to add essential files should be comparing the "rustc resource suffix" and the current toolchain or something instead of old and new toolchain versions).

feel free to make a PR with this but I don't intend to change it myself

GuillaumeGomez: Well, for information, I removed the folder and rebuilt and it worked. :)

PR #1026, opened 2020-09-01
Unify crate pages rendering

GuillaumeGomez:
The result looks like this:

Screenshot from 2020-09-01 21-55-21
Screenshot from 2020-09-01 17-26-18
Screenshot from 2020-09-01 17-23-50
Screenshot from 2020-09-01 17-23-52

As you can see, we now keep the sidebar in all cases and there are no more weird tabs.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -26,8 +26,6 @@
     <body>
         {%- include "header/topbar.html" -%}
 
-        {%- block header %}{% endblock header -%}
-
         {%- block body -%}{%- endblock body -%}
     </body>
 
--- a/templates/crate/builds.html
+++ b/templates/crate/builds.html
@@ -5,78 +5,82 @@
     {{ macros::doc_title(name=metadata.name, version=metadata.version) }}
 {%- endblock title -%}
 
-{%- block header -%}
-    {{ navigation::package_navigation(metadata=metadata, active_tab="builds") }}
-{%- endblock header -%}
-
 {%- block body -%}
     <div class="container">
-        <div class="recent-releases-container">
-            {# If there is a build log then show it #}
-            {# TODO: When viewing a build log, show a back button or a hide button #}
-            {%- if build_details -%}
-                <div class="release">
-                    <strong>Build #{{ build_details.id }} {{ build_details.build_time | date(format="%+") }}</strong>
-                </div>
-
-                {%- filter dedent -%}
-                    <pre>
-                        # rustc version
-                        {{ build_details.rustc_version }}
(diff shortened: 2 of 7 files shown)


jyn514:
I like this idea, but the top bar looks a little repetitive:

image

We should only show the title and clipboard once. Maybe we should remove it on the sidebar, so there's space to put the description below on the main page?

jyn514:
> Maybe we should remove it on the sidebar

Hmm, then it wouldn't show up on `/builds` or `/source` though.

Nemo157:
One big difference I notice with this is that the rustdoc pages are left-aligned, while this is centered, if your screen exceeds the max width.

Some more minor things (first two are already existing issues, but I found them more noticeable with this layout):
 - no indication of the current path when viewing a subfolder in source
 - no indication of the current file in "source"
 - big white page when looking at root/subfolder in "source"
 - white highlight of selected "builds" overlaps the bottom border of the dark panel
 - dark background of build log or file source merges with the dark panel of the sidebar, the "crate" page avoids this with a white gutter between the sidebar and the page contents

(It looks like existing ayu.rs styling does integrate well with it without any changes 😀)

GuillaumeGomez:
Actually I removed the clipboard thing from the document and only kept it inside the sidebar (otherwise it duplicated an idea and made a test fail).

As for the points @Nemo157 reported, well, it's more or less already the case. But I'm open to ideas for improvement. :)

Nemo157:
One other issue I just noticed is in the mobile rendering, before you get a row of icons, now you get a column of icons:

image

(mobile rendering is really not great as it stands, but it might make sense to fall back to the old layout if possible for it)

GuillaumeGomez:
Forgot to check mobile, good point!

GuillaumeGomez:
Now the mobile looks like this:

Screenshot from 2020-09-02 13-29-23

Not great, however I can suggest something similar to what we have in rustdoc: adding a burger button which will then display the menu.

jyn514:
Actually I like the way you currently have it :) I don't think we should show a hamburger menu because most people are looking for the documentation, not /crate - we shouldn't make that harder to find.

GuillaumeGomez:
Okay, I started something to check out how it looked but fine by me! However, I can keep the links and put the other information in the "hamburger menu"? That'd be a good middle thing.

GuillaumeGomez:
Well, it can be done as a second step. Let's just focus on the sidebar for the moment. :)

jyn514 (approved):

Issue #996, opened 2020-08-22
Add UI tests


jyn514:
Would have been very helpful to catch #992, #935, the issues from #981.

syphar:
solved since https://github.com/rust-lang/docs.rs/pull/1698

PR #1032, opened 2020-09-03
Transform "documentation" tab into button

GuillaumeGomez:
Same as before but with a fix for the 404 pages and a test to avoid regressions.

EDIT: Back to tab style but the "documentation" tab is now a button:

mobile:

Screenshot from 2020-09-08 17-50-40

"normal":

Screenshot from 2020-09-08 17-50-37

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -93,3 +93,33 @@ impl From<PoolError> for IronError {
         IronError::new(err.compat(), Status::InternalServerError)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test::wrapper;
+    use kuchiki::traits::TendrilSink;
+
+    #[test]
+    fn check_404_page_content() {
+        wrapper(|env| {
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/page-which-doesnt-exist")
+                    .send()?
+                    .text()?,
+            );
+            assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+            assert_eq!(
+                page.select("#crate-title")
+                    .unwrap()
+                    .next()
+                    .unwrap()
+                    .text_contents(),
+                "The requested resource does not exist",
+            );
+
+            Ok(())
+        });
+    }
+}
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -96,7 +96,6 @@ impl From<PoolError> for IronError {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use crate::test::wrapper;
     use kuchiki::traits::TendrilSink;
 
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -1213,4 +1213,58 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    fn check_releases_page_content() {
+        // NOTE: this is a little fragile and may have to be updated if the HTML layout changes
+        let sel = ".pure-menu-horizontal>.pure-menu-list>.pure-menu-item>.pure-menu-link>.title";
+        wrapper(|env| {
+            let tester = |url| {
+                let page = kuchiki::parse_html()
+                    .one(env.frontend().get(url).send().unwrap().text().unwrap());
+                assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+                let not_matching = page
+                    .select(sel)
+                    .unwrap()
+                    .map(|node| node.text_contents())
+                    .zip(
+                        [
+                            "Recent",
+                            "Stars",
+                            "Recent Failures",
+                            "Failures By Stars",
+                            "Activity",
+                            "Queue",
+                        ]
+                        .iter(),
+                    )
+                    .filter(|(a, b)| a.as_str() != **b)
+                    .collect::<Vec<_>>();
+                if not_matching.len() > 0 {
+                    let not_found = not_matching.iter().map(|(_, b)| b).collect::<Vec<_>>();
+                    let found = not_matching.iter().map(|(a, _)| a).collect::<Vec<_>>();
+                    assert!(
+                        not_matching.is_empty(),
+                        "Titles did not match for URL `{}`: not found: {:?}, found: {:?}",
+                        url,
+                        not_found,
+                        found,
+                    );
+                }
+            };
+
+            for url in &[
+                "/releases",
+                "/releases/stars",
+                "/releases/recent-failures",
+                "/releases/failures",
+                "/releases/activity",
+                "/releases/queue",
+            ] {
+                tester(url);
+            }
+
+            Ok(())
+        });
+    }
 }
--- a/templates/about-base.html
+++ b/templates/about-base.html
@@ -1,32 +1,34 @@
 {% extends "base.html" %}
 
 {% block header %}
-	<div class="cratesfyi-package-container">
+    <div class="cratesfyi-package-container">
         <div class="container">
-            <h1 id="crate-title" class="no-description">Docs.rs documentation</h1>
-            <div class="pure-menu pure-menu-horizontal">
-                <ul class="pure-menu-list">
-                    {% set text = "info-circle" | fas(fw=true) %}
-                    {% set text = text ~ ' <span class="title">About</span>' %}
-                    {{ macros::active_link(expected="index", href="/about", text=text) }}
+            <div class="description-container">
+                <h1 id="crate-title" class="no-description">Docs.rs documentation</h1>
+                <div class="pure-menu pure-menu-horizontal">
+                    <ul class="pure-menu-list">
+                        {% set text = "info-circle" | fas(fw=true) %}
+                        {% set text = text ~ ' <span class="title">About</span>' %}
+                        {{ macros::active_link(expected="index", href="/about", text=text) }}
 
-                    {% set text = "fonticons" | fab(fw=true) %}
-                    {% set text = text ~ ' <span class="title">Badges</span>' %}
-                    {{ macros::active_link(expected="badges", href="/about/badges", text=text) }}
+                        {% set text = "fonticons" | fab(fw=true) %}
+                        {% set text = text ~ ' <span class="title">Badges</span>' %}
+                        {{ macros::active_link(expected="badges", href="/about/badges", text=text) }}
 
-                    {% set text = "cogs" | fas(fw=true) %}
-                    {% set text = text ~ ' <span class="title">Builds</span>' %}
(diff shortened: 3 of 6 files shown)


jyn514:
Now that I think about this more, I'm not sure about this change ... it makes the /crate page very inconsistent with respect to the release pages, e.g. https://docs.rs/releases/stars.

GuillaumeGomez:
It's voluntary and I proposed to discussed about it. Discussion is still open. :)

pietroalbini:
Yeah, I think chasing the rustdoc styling for docs.rs is not the best choice, especially if we want to eventually make docs.rs in line with the website styling.

GuillaumeGomez:
Agreed, that's why when we're still in "crate focus", we have a sidebar simply because the tab system is just not adapted. However, for the release pages, it totally makes sense (in my opinion).

GuillaumeGomez:
Except that the current tab style is not appropriate for the crate pages because the behavior of the tabs isn't consistent.

pietroalbini:
I think having the button be more visible should help:

pietroalbini:
Looks good! Two notes:

* Let's add some margin between the description and the button, to avoid having the ellipsis so close to the button itself.
* I'd add a bit more padding to the button (dunno, `10px 15px`?).

pietroalbini:
Also, it seems a bit off-centered.

GuillaumeGomez:
> Also, it seems a bit off-centered.

A bit more of precision on this one please. :laughing: (I'll implement the other remarks in the meantime)

pietroalbini:
I think the best solution to fix this is to put everything in a flexbox (not just the description) and have all the contents of the header on the left and the button on the right-

pietroalbini:
Mostly looks good now! The only two changes I'd make:

* There are merge conflicts :)
* There should be some `padding-bottom` on mobile below the crate name, ideally the same as the `padding-top`. Currently it's too close to the tabs.

GuillaumeGomez:
Done!

pietroalbini (approved):

jyn514:
I'm unsure about this change too :P I like making 'Documentation' a separate button, because it takes you to a whole new page. I don't like though that it's over on the right side of the screen - it's harder to find if you're looking in the old place and it's also away from the main focus, which is the crate title.

Maybe we could put it in the center instead (literally front and center ;)? Then it would be easier to find.

GuillaumeGomez:
It's quite difficult to miss it though. But just by curiosity, where would you put it? (screenshot done with gimp please!)

src/web/error.rs
```diff
@@ -93,3 +93,33 @@ impl From<PoolError> for IronError {
         IronError::new(err.compat(), Status::InternalServerError)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
jyn514, line 99:
```suggestion
```

src/web/releases.rs
```diff
@@ -1213,4 +1213,57 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    fn check_releases_page_content() {
```
jyn514, line 1218: Isn't there already a test for this?

jyn514 (commented):
This changed the 'about' page to look rather strange, I think I prefered it before:
image

src/web/releases.rs
GuillaumeGomez: I thought to but couldn't find it.

GuillaumeGomez:
> This changed the 'about' page to look rather strange, I think I prefered it before:

Then it's a bug! I'll fix it.

templates/style/base.scss
```diff
@@ -444,87 +444,118 @@ div.cratesfyi-package-container {
...
+    .container {
+        display: flex;
+        align-items: center;
 
-        &.no-description {
-            padding-bottom: 15px;
-        }
-    }
+        .description-container {
+            flex-grow: 3;
```
jyn514, line 452: Without this it looks a lot better IMO, the 'Documentation' tab is centered:
image

src/web/releases.rs
jyn514: Here it is: https://github.com/rust-lang/docs.rs/blob/d688dc78f5d6e63f07b7c40b275e77817ef466c5/src/web/releases.rs#L1024
So I think I would prefer to remove this duplicate test.

templates/style/base.scss
pietroalbini: I'm pretty sure this way the button will jump around the page if the description is longer.

Nemo157: An example of a longer description: <https://docs.rs/crate/tokio/0.2.22>

jyn514: This is what it looks like on mobile: 
image

jyn514: > I'm pretty sure this way the button will jump around the page if the description is longer.

Hmm, maybe we can limit the length of the description and show this button after?

GuillaumeGomez: It's kind of the case. I just don't like the fact that the button position might change, which is why I added `flex-grow: 3` so like that, whatever the content length, it remains on the right.

jyn514: Ok yeah, with `width: 50%;` it looks consistent:
image
image

jyn514: I actually prefer it with the text limited as well, it makes the lines easier to read when they're shorter.

GuillaumeGomez: Fine by me. We'll have both then! :)

pietroalbini: It feels... weird. Dunno why, but I'm not sure it's the right approach :/

GuillaumeGomez: I don't like it either. Didn't add the line yet so feel free to debate about it. :)

src/web/releases.rs
GuillaumeGomez: It doesn't test the same things. My test added checks specifically to ensure the links were not removed by mistake (which I did while changing the templates when working on this PR).

templates/style/base.scss
jyn514: We decided to instead keep the right position, but make the button more noticable by adding a black background.

jyn514:
About is fixed: 
image
But releases are still broken: 
image

We should really think about #996 :/

I'll review the rest in the meantime so you can fix bugs in parallel ;)

src/web/releases.rs
```diff
@@ -1213,4 +1213,57 @@ mod tests {
...
+        let sel = ".pure-menu-horizontal>.pure-menu-list>.pure-menu-item>.pure-menu-link>.title";
+        wrapper(|env| {
+            let tester = |url| {
+                let page = kuchiki::parse_html()
+                    .one(env.frontend().get(url).send().unwrap().text().unwrap());
+                assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+                let not_maching = page
```
jyn514, line 1225: nit: `not_matching`

src/web/releases.rs
```diff
@@ -1213,4 +1213,57 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    fn check_releases_page_content() {
+        let sel = ".pure-menu-horizontal>.pure-menu-list>.pure-menu-item>.pure-menu-link>.title";
```
jyn514, line 1219: Hmm ... this seems a little fragile but I'm not sure how else to do it. I guess we're the only ones updating the releases page so whoever modifies the HTML can be in charge of fixing the test, but if so there should be a comment:
```suggestion
        // NOTE: this is a little fragile and may have to be updated if the HTML layout changes
        let sel = ".pure-menu-horizontal>.pure-menu-list>.pure-menu-item>.pure-menu-link>.title";
```

src/web/releases.rs
```diff
@@ -1213,4 +1213,57 @@ mod tests {
...
+                            "Recent Failures",
+                            "Failures By Stars",
+                            "Activity",
+                            "Queue",
+                        ]
+                        .iter(),
+                    )
+                    .filter(|(a, b)| a.as_str() != **b)
+                    .collect::<Vec<_>>();
+                if !not_maching.is_empty() {
```
jyn514, line 1242: nit: double negatives are hard to read
```suggestion
                if not_maching.len() > 0 {
```

src/web/releases.rs
```diff
@@ -1213,4 +1213,57 @@ mod tests {
...
+                    let not_found = not_maching.iter().map(|(_, b)| b).collect::<Vec<_>>();
+                    let found = not_maching.iter().map(|(a, _)| a).collect::<Vec<_>>();
+                    assert!(
+                        not_maching.is_empty(),
+                        "Titles not found for URL `{}`: {:?}, found: {:?}",
```
jyn514, line 1247:
```suggestion
                        "Titles did not match for URL `{}`: not found: {:?}, found: {:?}",
```

templates/about-base.html
```diff
@@ -1,32 +1,34 @@
...
-                </ul>
+                        {% set text = "road" | fas(fw=true) %}
+                        {% set text = text ~ ' <span class="title">Shorthand URLs</span>' %}
+                        {{ macros::active_link(expected="redirections", href="/about/redirections", text=text) }}
+                    </ul>
+                </div>
             </div>
         </div>
-	</div>
+    </div>
```
jyn514, line 33: Why is this changing `about-base` at all? Shouldn't this only need to modify /crate pages?

templates/style/base.scss
```diff
@@ -444,87 +444,119 @@ div.cratesfyi-package-container {
     border-bottom: 1px solid $color-border;
     margin-bottom: 20px;
 
-    h1 {
-        margin: 0;
-        padding: 15px 0 0 14px;
+    .container {
+        display: flex;
+        align-items: center;
```
jyn514, line 449: I don't feel comfortable reviewing all this CSS, but since you and pietro worked it out I trust your judgement. The end result looks good :)

templates/about-base.html
GuillaumeGomez: Oh it was just a tab so I guess it was done by my IDE...

GuillaumeGomez:
Fixed release page as well and fixed the nits.

> We should really think about #996 :/

Yep but it's a tricky thing to do... We can use the same system we use in rustdoc but we still have the issue with getting the screenshot differences. But in here, I think this problem would be very limited considering that the build is quite small in comparison.

templates/about-base.html
jyn514: No, it added `             <div class="description-container">` above, which doesn't seem right.

jyn514:
Thanks for working on this!

PR #1033, opened 2020-09-03
Switch fontawesome from font files to embedded SVG

### LINKED ISSUE #790: Update Font Awesome
We currently use Font Awesome 4 which was replaced in 2017ish by Font Awesome 5 which has 1500+ icons (opposed to the ~300 of our version). v5 also uses SVGs, which would help with some of the aliasing problems that the current icons can have on screens with larger resolution. There's also some icon-related stuff I'd like to do which is only possible with some of the extended icons that v5 offers

notriddle:
Fixes #790


jyn514:
Can you post some screenshots of what this looks like after the change? I'm a little worried about changing so much of the CSS without any frontend tests ...

notriddle:
## If anybody notices a change, then I messed up

image

image

image

pietroalbini:
cc @Kixiron 

This is great! I'm wondering, could we inline the SVG directly without having to download the `.svg` file? That would cut a network request completly, and since we don't use icons *that much* (except for maybe the source listing page?) it should also reduce the amount of data sent to viewers.

pietroalbini:
To be clear, landing this is also fine, if you don't have the time to make the other change or you prefer to do it in a followup :)

notriddle:
Yeah, we probably should switch to inlined SVG. I didn't originally do it, because there's no FontAwesome HOWTO for it, but I've done it before and know how. Also, IE11 shows blank spots where the icons ought to be, and inlining them would fix that.

notriddle:
Okay, it embeds the SVG directly now. It looks identical.

jyn514:
It looks like you moved all the vendor files into a new repo, which you just published as a crate? https://github.com/notriddle/font-awesome-as-a-crate

What's the rationale for doing that? What's the long-term maintenance plan if we need to update? I want to avoid adding and deleting 4k lines of code every 2 weeks.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ once_cell = { version = "1.4.0", features = ["parking_lot"] }
 base64 = "0.12.1"
 strum = { version = "0.18.0", features = ["derive"] }
 lol_html = "0.2"
+font-awesome-as-a-crate = "0.1.0"
```
Kixiron, line 53: I'd prefer if we didn't take this approach and vendored as normal, it's needless complexity to maintain as well as being fragile in the first place

The more I think about it the less I like it, I'd really prefer going back to in-house vendoring

templates/header/global_alert.html
```diff
@@ -5,7 +5,7 @@
 {%- if global_alert -%}
     <li class="pure-menu-item">
         <a href="{{ global_alert.url | safe }}" class="pure-menu-link {{ global_alert.css_class }}">
-            <i class="{{ global_alert.fa_icon | fas(fw=true) }}"></i>
+            <svg class=fa-svg viewBox="0 0 512 512"><use xlink:href="/-/static/fa-solid.svg#{{ global_alert.fa_icon }}"></svg>
```
Kixiron, line 8: Why is this the only invocation changed? Why doesn't the `fas` function work?

Kixiron, line 8: Additionally I don't think this link will work since you removed the routing to the files that were served

Kixiron (changes requested):

templates/header/global_alert.html
notriddle: Oh, right. I missed that one.

Cargo.toml
notriddle: It's not fragile. Embedding the SVG files directly in the binary should be quite robust.

notriddle:
> It looks like you moved all the vendor files into a new repo, which you just published as a crate? https://github.com/notriddle/font-awesome-as-a-crate

Yes.

> What's the rationale for doing that?

font-awesome-as-a-crate doesn't just vendor the SVG files. It [concatenates them into a massive rust source file](https://github.com/notriddle/font-awesome-as-a-crate/blob/46108a44096227fd7af9d3ab771b53d09d65f134/build.rs#L14).

The rationale for embedding the SVG files directly in your binary:

* avoid file I/O occurring in your Tera Filter, which would likely have complex knock-on effects when combined with `async`
* avoid complex error handling in your Tera Filter, which is the inevitable outcome of file I/O
* avoid having to carefully construct filesystem paths, with all the terrible possible outcomes and failure modes that come with that (using the infrastructure that the existing static file server uses is not an option, because it's not an HTTP request)
* I'm guessing it's faster to pull a string out of the binary's text section than it is to open a file for every request, though I have done no benchmarking to justify this wild speculation

The rationale for doing it in its own crate:

* we're not using any of docs.rs's file handling infrastructure anyway, so all the other stuff in its build.rs was just getting in the way while I coded up its codegen script
* it should be useful in other Rust web projects, regardless of the web framework they're using
* let's not force rustc to parse a 1.3MiB source file every time you change docs.rs; this seems like prime "separate translation unit" material

jyn514:
All your points make sense, except that I'm a little unsure about having font_awesome_as_a_crate in a separate repository, since we're the only ones currently using it. Do you mind making it part of the docs.rs workspace and adding I and @pietroalbini as authors on crates.io? See [#1000](https://github.com/rust-lang/docs.rs/pull/1000/files#diff-80398c5faae3c069e4e6aa2ed11b28c0) for an example of what that would look like.

notriddle:
Sure, let's do that.

vendor/font-awesome-as-a-crate/build.rs
```diff
@@ -0,0 +1,38 @@
...
+
+fn main() {
+    println!("cargo:rustc-cfg=font_awesome_out_dir");
+    write_fontawesome_sprite();
+}
+
+fn write_fontawesome_sprite() {
+    let dest_path = Path::new(&env::var("OUT_DIR").unwrap()).join("fontawesome.rs");
+    let mut dest_file = File::create(&dest_path).unwrap();
+    dest_file.write("fn fontawesome_svg(dir:&str,file:&str)->&'static str{match(dir,file){".as_bytes()).expect("fontawesome fn write");
```
jyn514, line 16:
```suggestion
    dest_file.write_all(b"fn fontawesome_svg(dir:&str,file:&str)->&'static str{match(dir,file){").expect("fontawesome fn write");
```

vendor/font-awesome-as-a-crate/build.rs
```diff
@@ -0,0 +1,38 @@
...
+        let mut data = String::new();
+        for file in dir {
+          let file = file.expect("fontawesome directory access");
+          let filename = file.file_name().into_string().expect("fontawesome filenames are unicode");
+          let mut file = File::open(file.path()).expect("fontawesome file access");
+          data.clear();
+          file.read_to_string(&mut data).expect("fontawesome file read");
+          // if this assert goes off, add more hashes here and in the format! below
+          assert!(data.find("###").is_none(), "file {} breaks raw string", filename);
+          dest_file.write(format!(
```
jyn514, line 28:
```suggestion
          dest_file.write_all(format!(
```

vendor/font-awesome-as-a-crate/build.rs
```diff
@@ -0,0 +1,38 @@
...
+              r####"("{dirname}","{filename}")=>r###"{data}"###,"####,
+              data = data,
+              dirname = dirname,
+              filename = filename.replace(".svg", ""),
+          ).as_bytes()).expect("write fontawesome file");
+        }
+    }
+    dest_file.write("_=>panic!(\"missing image file\")}}".as_bytes()).expect("fontawesome fn write");
```
jyn514, line 36:
```suggestion
    dest_file.write_all(b"_=>panic!(\"missing image file\")}}").expect("fontawesome fn write");
```

notriddle:
Did you get my invite on crates.io?

jyn514:
Got it, thanks.

Don't have time to do a full review of this right now but the general idea looks right. Thanks for working on this!

vendor/font-awesome-as-a-crate/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "font-awesome-as-a-crate"
+version = "0.1.2"
+# https://github.com/FortAwesome/Font-Awesome/blob/master/composer.json
+authors = ["Michael Howell <michael@notriddle.com>", "Travis Chase", "Dave Gandy", "Rob Madole", "Jory Raphael", "Geremia Taglialatela", "Brian Talbot", "Mike Wilkerson", "Fonticons Inc <hello@fontawesome.com>"]
+edition = "2018"
+license = "CC-BY-4.0 AND MIT"
+description = "Font Awesome Free, packaged as a crate"
+repository = "https://github.com/rust-lang/docs.rs"
+
```
jyn514, line 10: nit: extra newline
```suggestion
```

vendor/font-awesome-as-a-crate/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "font-awesome-as-a-crate"
+version = "0.1.2"
+# https://github.com/FortAwesome/Font-Awesome/blob/master/composer.json
+authors = ["Michael Howell <michael@notriddle.com>", "Travis Chase", "Dave Gandy", "Rob Madole", "Jory Raphael", "Geremia Taglialatela", "Brian Talbot", "Mike Wilkerson", "Fonticons Inc <hello@fontawesome.com>"]
```
jyn514, line 5: See #1039 
```suggestion
authors = ["Michael Howell <michael@notriddle.com>", "The Rust Project Developers", "Travis Chase", "Dave Gandy", "Rob Madole", "Jory Raphael", "Geremia Taglialatela", "Brian Talbot", "Mike Wilkerson", "Fonticons Inc <hello@fontawesome.com>"]
```

vendor/font-awesome-as-a-crate/build.rs
```diff
@@ -0,0 +1,54 @@
...
+                .write_all(
+                    format!(
+                        r####"("{dirname}","{filename}")=>r###"{data}"###,"####,
+                        data = data,
+                        dirname = dirname,
+                        filename = filename.replace(".svg", ""),
+                    )
+                    .as_bytes(),
+                )
+                .expect("write fontawesome file");
```
jyn514, line 48: Wow, this is _clever_ 😆 I don't know whether to be impressed or horrified, I'll go with impressed.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ once_cell = { version = "1.4.0", features = ["parking_lot"] }
 base64 = "0.12.1"
 strum = { version = "0.18.0", features = ["derive"] }
 lol_html = "0.2"
+font-awesome-as-a-crate = { path = "vendor/font-awesome-as-a-crate" }
```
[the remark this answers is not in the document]
jyn514: Opened #1041 to unblock this.

jyn514 (approved):
LGTM once https://github.com/rust-lang/docs.rs/pull/1041 unblocks switching `vendor/fontawesome` to `crates/fontawesome`

jyn514:
Can you rebase this over master and change the new crate to be in `crates/font-awesome-as-a-crate`? Looks good to me after that.

notriddle:
shucks

jyn514:
I'd take a PR adding triagebot ;)

jyn514:
Thanks for the PR!

PR #1034, opened 2020-09-04
Fix styling of the about page header

pietroalbini:

### DIFF
--- a/templates/style/base.scss
+++ b/templates/style/base.scss
@@ -447,6 +447,10 @@ div.cratesfyi-package-container {
     h1 {
         margin: 0;
         padding: 15px 0 0 14px;
+
+        &.no-description {
+            padding-bottom: 15px;
+        }
     }
 
     div.description {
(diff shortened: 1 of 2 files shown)


templates/style/base.scss
```diff
@@ -447,6 +447,10 @@ div.cratesfyi-package-container {
     h1 {
         margin: 0;
         padding: 15px 0 0 14px;
+
+        &.no-description {
```
GuillaumeGomez, line 451: You don't need the `&` here.

jyn514:
I kind of liked the underscore before, is there a reason you removed it?
image

pietroalbini:
It's not consistent with the other pages that have tabs.

templates/style/base.scss
pietroalbini: Why not? I need to apply it to `<h1 class="no-description">`.

GuillaumeGomez: Oooooh. I see what you did there! Indeed, my bad!

jyn514:
Seems fine to me.

PR #1041, opened 2020-09-10
Rename crates to be in the same top-level crates/ directory

jyn514:
Unblocks https://github.com/rust-lang/docs.rs/pull/1033.
r? @pietroalbini


Nemo157:
Whats wrong with having them at the root? Outside `rust-lang/rust` (because of its ridiculously high number of crates) adding in an extra level of nesting is uncommon.

pietroalbini:
I feel like we're reaching a point where there is too much going on in the root directory, especially as we add more standalone crates (docsrs-metadata, badge, font-awesome-as-a-crate, a possible future archive crate...).

jyn514:
> adding in an extra level of nesting is uncommon.

If it helps, it's the same number of nesting as normal because there's no `src/` directory (I don't like those at _all_ in shared crates). This also gives them a common `crates/` prefix - before you couldn't tell `metadata` was a crate from the name. Changing it to `docsrs-metadata` in the top-level is also possible but not as appealing to me.

Nemo157:
> If it helps, it's the same number of nesting as normal because there's no src/ directory (I don't like those at all in shared crates).

Another thing I find really weird about this setup, and breaks my muscle memory for editing files from sibling crates :grin:; but I don't have any major issues with this setup, most workspaced projects seem to grow some weird unique setup eventually.

Issue #55, opened 2016-09-23
"The requested crate does not exist" message if a version does not exist


bluss:
"The requested crate does not exist" is the message if you try to load up docs for a **crate** that exists, in a **version** that does **not**. Could just use a small tweak in how it reports the error.

onur:
Error messages definitely needs improvements.

docs.rs is also giving same error message when an author not exists, i.e: https://docs.rs/releases/non-existent-author

Kixiron:
Triage: This should be an easy fix, error messages in general do need improvement still

Kixiron:
Mentor: 
1. The current error is `Nope::CrateNotFound` which is accurate but still not a good error. As of now, a better error doesn't exist, so we'll make it
2. Go to [`src/web/error.rs`](https://github.com/rust-lang/docs.rs/blob/master/src/web/error.rs#L9) and find the `Nope` enum. Add a new variant named `VersionNotFound`, this will be our new error
3. Update the `Display` implementation for `Nope` and add the new variant with a helpful message
4. Docs.rs uses `iron` for its web framework and `iron` operates off of a hander system, where individual objects (structs, enums, etc.) define their behavior as the responder to a web request by implementing the `Handler` trait. The default page that the user sees is created by using the `Page` struct. Add the `VersionNotFound` variant to `Nope`'s handler.
5. Within the `VersionNotFound` branch of the handler, construct a new page
    -  `Page::new("the requested crate version does not exist".to_owned())` Create a new page with `the requested crate version does not exist` as the content
    - `.set_status(status::NotFound)` Set the status of the response to 404
    - `.title("The requested version does not exist")` Set the title of the page
    - `.to_resp("error")` Render the page using the `error` template
6. Go to [`src/web/rustdoc.rs`](https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L179) and change the error to the new `Nope::VersionNotFound`

robinhundt:
I'll start working on this. Thank you @Kixiron for the detailed description of what to do :blush:

robinhundt:
Mhh, doesn't seem to be as easy as it first looked. As a start, I'm not even able to reproduce the error.  
Opening any of the following
- https://docs.rs/non-existent
- https://docs.rs/sqlx/42
- https://docs.rs/releases/non-existent-author (example from above)
all produce a 404 page with the content
```
The requested resource does not exist
no such resource
```
[...]

PR #1043, opened 2020-09-10
Introduced Nope::VersionNotFound error

robinhundt:
Trying to reproduce the issue #55 revealed another bug: the order in which the different handlers were called in CratesfyiHandler::handle resulted in the more specific errors from the router_handler being dropped, due to them being 404 and the last handler being the static_handler which would just return a ResourceNotFound error.
It should be noted that the different execution order might affect the performance in production.

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -10,6 +10,7 @@ use std::{error::Error, fmt};
 pub enum Nope {
     ResourceNotFound,
     CrateNotFound,
+    VersionNotFound,
     NoResults,
     InternalServerError,
 }
@@ -19,6 +20,7 @@ impl fmt::Display for Nope {
         f.write_str(match *self {
             Nope::ResourceNotFound => "Requested resource not found",
             Nope::CrateNotFound => "Requested crate not found",
+            Nope::VersionNotFound => "Requested crate does not have specified version",
             Nope::NoResults => "Search yielded no results",
             Nope::InternalServerError => "Internal server error",
         })
@@ -47,6 +49,17 @@ impl Handler for Nope {
                 ErrorPage {
                     title: "The requested crate does not exist",
                     message: Some("no such crate".into()),
+                    status: Status::NotFound,
+                }
+                .into_response(req)
+            }
+
+            Nope::VersionNotFound => {
+                // user tried to navigate to a crate with a version that does not exist
+                // TODO: Display the attempted crate and version
+                ErrorPage {
+                    title: "The requested version does not exist",
+                    message: Some("no such version for this crate".into()),
                     status: Status::NotFound,
                 }
                 .into_response(req)
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -113,11 +113,37 @@ mod tests {
     use kuchiki::traits::TendrilSink;
 
     #[test]
-    fn check_404_page_content() {
+    fn check_404_page_content_crate() {
         wrapper(|env| {
             let page = kuchiki::parse_html().one(
                 env.frontend()
-                    .get("/page-which-doesnt-exist")
+                    .get("/crate-which-doesnt-exist")
+                    .send()?
+                    .text()?,
+            );
+            assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+            assert_eq!(
+                page.select("#crate-title")
+                    .unwrap()
+                    .next()
+                    .unwrap()
+                    .text_contents(),
+                "The requested crate does not exist",
+            );
+
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn check_404_page_content_resource() {
+        // Resources with a `.js` and `.ico` extension are special cased in the
+        // routes_handler which is currently run last. This means that `get("resource.exe")` will
+        // fail with a `no so such crate` instead of 'no such resource'
+        wrapper(|env| {
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/resource-which-doesnt-exist.js")
                     .send()?
                     .text()?,
             );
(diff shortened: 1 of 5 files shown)


src/web/mod.rs
```diff
@@ -358,13 +371,18 @@ fn match_version(conn: &mut Client, name: &str, version: Option<&str>) -> Option
...
-            corrected_name,
-            version: MatchSemver::Semver((v.0.to_string(), v.1)),
-        });
+        return versions_sem
+            .first()
+            .map(|v| MatchVersion {
+                corrected_name,
+                version: MatchSemver::Semver((v.0.to_string(), v.1)),
+            })
+            .ok_or(Nope::CrateNotFound);
```
jyn514, line 380:
```suggestion
            .ok_or(Nope::VersionNotFound);
```

jyn514: Please also add a test for this case by requesting `/:crate/*` for a crate that only has pre-releases, then make sure that it fails before and succeeds after this change.

src/web/mod.rs
```diff
@@ -320,24 +326,31 @@ fn match_version(conn: &mut Client, name: &str, version: Option<&str>) -> Option
...
-        return Some(MatchVersion {
+        return Ok(MatchVersion {
             corrected_name,
             version: MatchSemver::Exact((version.to_owned(), *id)),
         });
     }
 
     // Now try to match with semver
-    let req_sem_ver = VersionReq::parse(&req_version).ok()?;
+    let req_sem_ver = VersionReq::parse(&req_version).map_err(|_| Nope::CrateNotFound)?;
```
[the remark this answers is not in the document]
jyn514: Please also add a test for this case by requesting `/:crate/some_version/`, then make sure that it fails before and succeeds after this change.

src/web/mod.rs
```diff
@@ -581,7 +599,13 @@ mod test {
...
     fn version(v: Option<&str>, db: &TestDatabase) -> Option<String> {
         match_version(&mut db.conn(), "foo", v)
-            .and_then(|version| version.assume_exact().map(|semver| semver.into_parts().0))
+            .ok()
+            .and_then(|version| {
+                version
+                    .assume_exact()
+                    .ok()
+                    .map(|semver| semver.into_parts().0)
+            })
```
jyn514, line 608: nit: this would be easier to read with `?`:
```suggestion
         match_version(&mut db.conn(), "foo", v).ok()?.assume_exact().ok()?.into_parts().0
```

src/web/rustdoc.rs
```diff
@@ -286,28 +286,29 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
-        match match_vers.version {
-            MatchSemver::Exact((version, _)) => {
-                // Redirect when the requested crate name isn't correct
-                if let Some(name) = match_vers.corrected_name {
-                    return redirect(&name, &version, &req_path);
-                }
+    let version = match match_version(&mut conn, &name, url_version) {
+        Ok(match_vers) => {
+            match match_vers.version {
+                MatchSemver::Exact((version, _)) => {
```
jyn514, line 292: Lots of 'match' going on 😆 maybe call the local variable `release_found`?

src/web/rustdoc.rs
```diff
@@ -286,28 +286,29 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
     // * If there is a semver (but not exact) match, redirect to the exact version.
     // * Otherwise, return a 404.
-    let version = if let Some(match_vers) = match_version(&mut conn, &name, url_version) {
-        match match_vers.version {
-            MatchSemver::Exact((version, _)) => {
-                // Redirect when the requested crate name isn't correct
-                if let Some(name) = match_vers.corrected_name {
-                    return redirect(&name, &version, &req_path);
-                }
+    let version = match match_version(&mut conn, &name, url_version) {
```
jyn514, line 289: To simplify this, could you use `map_err` instead? Then you wouldn't need to match on the `Err` case.
```suggestion
    let version = match match_version(&mut conn, &name, url_version).map_err(|err|  IronError::new(err, status::NotFound))?.version {
```

src/web/rustdoc.rs
```diff
@@ -562,7 +563,7 @@ pub fn badge_handler(req: &mut Request) -> IronResult<Response> {
                 return Ok(super::redirect(iron_url));
             }
 
-            None => BadgeOptions {
+            Err(_) => BadgeOptions {
```
jyn514, line 566: Can you add a separate badge for `VersionNotFound` that say "version not found" or something?
```suggestion
            Err(Nope::CrateNotFound) => BadgeOptions {
```

jyn514 (commented):
Lots of little nit-picky comments, but the approach looks right. Thanks for working on this!

jyn514:
Can you also add some tests for the new behavior? There's examples around https://github.com/rust-lang/docs.rs/blob/master/src/web/rustdoc.rs#L821 - I don't think anything tests the error text currently, but it shouldn't be too hard to get it from the response: https://docs.rs/reqwest/0.10.8/reqwest/struct.Response.html#method.text

jyn514:
Looks like you need to update the text for one of the tests:
```
[2020-09-11T18:37:05Z INFO  docs_rs::web] Running docs.rs web server on http://127.0.0.1:39637
thread 'web::error::tests::check_404_page_content' panicked at 'assertion failed: `(left == right)`
  left: `"The requested crate does not exist"`,
 right: `"The requested resource does not exist"`', src/web/error.rs:125:13
```

robinhundt:
Yea, I've already updated it. I was just confused for a moment because the failing test wasn't present in my branch so I had to fetch the origin and rebase.
I'll probably write the tests sometime this weekend :blush:

jyn514:
Sounds great! Thanks so much for working on this :)

src/web/mod.rs
robinhundt: Such a request would return the correct crate before and after the PR. The difference in behavior is for a `/:crate/*` request for a crate that only has yanked versions in which case `match_version` would have returned a None and the resulting page would have displayed `no such resource`.  
I've added a test case asserting that in this case the correct `VersionNotFound` error is returned.

src/web/error.rs
```diff
@@ -113,11 +113,37 @@ mod tests {
...
+
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn check_404_page_content_resource() {
+        // Resources with a `.js` and `.ico` extension are special cased in the
+        // routes_handler which is currently run last. This means that `get("resource.exe")` will
+        // fail with a `no so such crate` instead of 'no such resource'
```
jyn514, line 142: Ooh, good catch. This seems like an ok bug to have though - if you're trying to download .exes docs.rs doesn't want to help you even as much to have the right error.

src/web/error.rs
```diff
@@ -134,4 +160,66 @@ mod tests {
...
+                .create()?;
+            let page = kuchiki::parse_html().one(env.frontend().get("/dummy/*").send()?.text()?);
+            assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+            assert_eq!(
+                page.select("#crate-title")
+                    .unwrap()
+                    .next()
+                    .unwrap()
+                    .text_contents(),
+                "The requested version does not exist",
```
jyn514, line 219: Eventually it would be nice to instead say 'all versions of this crate are yanked', which makes more sense since this didn't request a specific version. But that can be a follow-up PR.

jyn514 (approved):
Looks great!

robinhundt:
Thanks for the help and quick, detailed reviews :blush:

PR #1045, opened 2020-09-13
Display "Go to latest stable" for pre-release

### LINKED ISSUE #1044: Show different message than 'Go to latest version' on pre-releases
We automatically redirect to the first non-prerelease version when available, but on pre-release versions we still display the "Go to latest version" box in the tab, see [sqlx v0.4.0-beta.1](https://docs.rs/sqlx/0.4.0-beta.1/sqlx/index.html)

robinhundt:
closes #1044

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -189,6 +189,7 @@ struct RustdocPage {
     latest_version: String,
     inner_path: String,
     is_latest_version: bool,
+    is_prerelease: bool,
     krate: CrateDetails,
 }
 
@@ -367,6 +368,19 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     // Get the latest version of the crate
     let latest_version = latest_release.version.to_string();
     let is_latest_version = latest_version == version;
+    let is_prerelease = semver::Version::parse(&version)
+        // should be impossible unless there is a semver incompatible version in the db
+        // Note that there is a redirect earlier for semver matches to the exact version
+        .map_err(|err| {
+            log::error!(
+                "invalid semver in database for crate {}: {}. Err: {}",
+                name,
+                &version,
+                err
+            );
+            IronError::new(Nope::InternalServerError, status::InternalServerError)
+        })?
+        .is_prerelease();
 
     // If the requested crate version is the most recent, use it to build the url
     let latest_path = if is_latest_version {
@@ -408,6 +422,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         latest_version,
         inner_path,
         is_latest_version,
+        is_prerelease,
         krate,
     }
     .into_response(&file.0.content, config.max_parse_memory, req, &path)
--- a/templates/rustdoc/header.html
+++ b/templates/rustdoc/header.html
@@ -181,23 +181,24 @@
                             </span>
                         </li>
 
-                    {# If the crate isn't the latest & is yanked, display a warning and offer a link to the latest release #}
-                    {%- elif not is_latest_version and krate.yanked -%}
-                        <li class="pure-menu-item">
-                            <a href="{{ latest_path | safe }}" class="pure-menu-link warn"
-                                title="You are seeing a yanked version of {{ krate.name }} crate. Click here to go to latest version.">
-                                {{ "exclamation-triangle" | fas(fw=true) }}
-                                <span class="title"> This release has been yanked, go to latest version</span>
-                            </a>
-                        </li>
-
-                    {# If this isn't the most recent release, offer a link to the latest #}
+                    {# If this isn't the most recent stable release, offer a link to the latest #}
                     {%- elif not is_latest_version -%}
+                        {%- if krate.yanked -%}
+                            {%- set tooltip = "You are seeing a yanked version of the " ~ krate.name ~ "crate. Click here to go to the latest version." -%}
+                            {%- set title = "This release has been yanked, go to latest version" -%}
+                        {%- elif is_prerelease -%}
+                            {%- set tooltip = "You are seeing a pre-release version of the " ~ krate.name ~ "crate. Click here to go to the latest stable version." -%}
+                            {%- set title = "Go to latest stable release" -%}
+                        {%- else -%}
+                            {%- set tooltip = "You are seeing an outdated version of the " ~ krate.name ~ "crate. Click here to go to the latest version." -%}
+                            {%- set title = "Go to latest version" -%}
+                        {%- endif -%}
+
                         <li class="pure-menu-item">
                             <a href="{{ latest_path | safe }}" class="pure-menu-link warn"
-                                title="You are seeing an outdated version of {{ krate.name }} crate. Click here to go to latest version.">
+                                title="{{ tooltip }}">
                                 {{ "exclamation-triangle" | fas(fw=true) }}
-                                <span class="title"> Go to latest version</span>
+                                <span class="title">{{ title }}</span>
                             </a>
                         </li>
                     {%- endif -%}
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -348,11 +348,12 @@ fn match_version(
         for version in versions.iter().filter(|(_, _, yanked)| !yanked) {
             // in theory a crate must always have a semver compatible version,
             // but check result just in case
-            let version_sem = Version::parse(&version.0).map_err(|_| {
+            let version_sem = Version::parse(&version.0).map_err(|err| {
                 log::error!(
-                    "invalid semver in database for crate {}: {}",
+                    "invalid semver in database for crate {}: {}. Err: {}",
                     name,
-                    version.0
+                    version.0,
+                    err
                 );
                 Nope::InternalServerError
             })?;


src/web/rustdoc.rs
```diff
@@ -367,6 +368,10 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     // Get the latest version of the crate
     let latest_version = latest_release.version.to_string();
     let is_latest_version = latest_version == version;
+    let is_prerelease = semver::Version::parse(&version)
+        // should be impossible unless there is a bug in match_version
+        .map_err(|_| Nope::InternalServerError)?
+        .is_prerelease();
```
jyn514, line 374: I don't think this is true, we allow exact matches before checking semver: https://github.com/rust-lang/docs.rs/blob/fb06f95e4a8537fa321c3b192a73ffa6dbd68e97/src/web/mod.rs#L333
```suggestion
    // If this is an exact match (not a valid semver), assume it's not a stable version.
    let is_prerelease = semver::Version::parse(&version).unwrap_or(true);
```

templates/rustdoc/header.html
```diff
@@ -191,6 +191,16 @@
                             </a>
                         </li>
 
+                    {# If the crate is a pre-release, display a warning and offer a link to the latest stable release#}
+                    {%- elif is_prerelease -%}
+                    <li class="pure-menu-item">
+                        <a href="{{ latest_path | safe }}" class="pure-menu-link warn"
+                           title="You are seeing a pre-release version of {{ krate.name }} crate. Click here to go to latest stable version.">
```
jyn514, line 198: I don't think this is quite right ... You could be on a pre-release version with no stable version available. I think this block should be inside `{%- elif not is_latest_version -%}` and only change the text, not the behavior.

src/web/rustdoc.rs
```diff
@@ -367,6 +368,10 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     // Get the latest version of the crate
     let latest_version = latest_release.version.to_string();
     let is_latest_version = latest_version == version;
+    let is_prerelease = semver::Version::parse(&version)
+        // should be impossible unless there is a bug in match_version
+        .map_err(|_| IronError::new(Nope::InternalServerError, status::InternalServerError))?
```
jyn514, line 373: Hmm, I think we could `impl From<Nope> for IronError` to make this easier ... probably should go in a separate PR though.

src/web/rustdoc.rs
robinhundt: Oh, you're right. But the suggested change won't work, because then `GET /sqlx/0.3` would be treated as a pre-release although the semver compatible version 0.3.5 would be returned.  
I thought just parsing the Version here would be the least invasive change, but it seems like I'll have to get the actual version returned from the DB for the release serverd.

robinhundt: Mhh, i looked at it again. The `version` is actually guaranteed to be semver compatible, because in the case of an exact match the version is semver compatible (if not, the db contains an invalid semver version) and in the case of an inexact match `rustdoc_html_server_handler` will redirect to the exact version.
https://github.com/rust-lang/docs.rs/blob/c444b639dd489a8985426217feea5b1ec8d97c2c/src/web/rustdoc.rs#L303-L310

jyn514: I don't follow, `"0.3"` is a valid semver and would be parsed successfully.

jyn514: > if not, the db contains an invalid semver version

Hmm ... I don't think we enforce semantic versioning in the database, we depend on cargo to do it for us. In principle a 500 is probably the right thing to do, but it seems a shame to return that when it's so easy to return the page with a different text instead.

templates/rustdoc/header.html
robinhundt: `{%- elif not is_latest_version -%}` is already the condition for the normal `Go to latest` message.  
I think 
```
{# If the crate is a pre-release, display a warning and offer a link to the latest stable
release. If a crate only has pre-release the warning won't be displayed for the latest #}
{%- elif is_prerelease and not is_latest_version -%}
```
is what we want. This won't display a warning if the user views the latest pre-release for a crate that has no stable releases but displays it for all other pre-releases.

src/web/rustdoc.rs
robinhundt: Yep, I had the same thought :smile:

templates/rustdoc/header.html
jyn514: Yes, that sounds right. If you put it inside `{%- elif not is_latest_version -%}` you only have to change the text displayed without having to copy the HTML around it.

src/web/rustdoc.rs
jyn514: Opened #1046.

templates/rustdoc/header.html
robinhundt: I haven't used Tera a lot. I know how to use the condition for the text inside the span but I don't know how to do that elegantly for the title attribute. I'm also not sure if it would make the template cleaner if I somehow used interpolation for the title.

jyn514: Like this:
```html
                    {# If this isn't the most recent release, offer a link to the latest #}
                    {%- elif not is_latest_version -%}
                      {%- if not is_latest_version -%}
                        {% set title = "You are seeing an outdated version of the " ~ {{ krate.name }} ~ "crate. Click here to go to latest version."
                      {%- else %}
                        {% set title = "You are seeing a pre-release version of the " ~ {{ krate.name }} ~ "crate. Click here to go to latest stable version."
                      {%- endif %}
                        <li class="pure-menu-item">
                            <a href="{{ latest_path | safe }}" class="pure-menu-link warn"
                                title="{% title %}">
                                {{ "exclamation-triangle" | fas(fw=true) }}
                                <span class="title"> Go to latest version</span>
                            </a>
                        </li>
                    {%- endif -%}
```

jyn514: but if you think it's cleaner the other way, that's fine too, I have a tendency to DRY too much sometimes

robinhundt: I've changed it and included the `not is_latest_version and krate.yanked` variant as well.

templates/rustdoc/header.html
```diff
@@ -181,23 +181,24 @@
...
+                    {# If this isn't the most recent stable release, offer a link to the latest #}
                     {%- elif not is_latest_version -%}
+                        {%- if krate.yanked -%}
+                            {%- set tooltip = "You are seeing a yanked version of the " ~ krate.name ~ "crate. Click here to go to latest version." -%}
+                            {%- set title = "This release has been yanked, go to latest version" -%}
+                        {%- elif is_prerelease -%}
+                            {%- set tooltip = "You are seeing a pre-release version of the " ~ krate.name ~ "crate. Click here to go to latest stable version." -%}
+                            {%- set title = "Go to latest stable release" -%}
+                        {%- else -%}
+                            {%- set tooltip = "You are seeing an outdated version of the " ~ krate.name ~ "crate. Click here to go to latest version." -%}
```
jyn514, line 193: nit: grammar
```suggestion
                            {%- set tooltip = "You are seeing a yanked version of the " ~ krate.name ~ "crate. Click here to go to the latest version." -%}
                            {%- set title = "This release has been yanked, go to latest version" -%}
                        {%- elif is_prerelease -%}
                            {%- set tooltip = "You are seeing a pre-release version of the " ~ krate.name ~ "crate. Click here to go to the latest stable version." -%}
                            {%- set title = "Go to latest stable release" -%}
                        {%- else -%}
                            {%- set tooltip = "You are seeing an outdated version of the " ~ krate.name ~ "crate. Click here to go to the latest version." -%}
```

src/web/rustdoc.rs
robinhundt: Would you rather add a new `Nope` variant for this case? I feel like in this case it should be fine to just return a 500. We do the same in the `match_version` func https://github.com/rust-lang/docs.rs/blob/c444b639dd489a8985426217feea5b1ec8d97c2c/src/web/mod.rs#L348-L360

src/web/rustdoc.rs
```diff
@@ -367,6 +368,11 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     // Get the latest version of the crate
     let latest_version = latest_release.version.to_string();
     let is_latest_version = latest_version == version;
+    let is_prerelease = semver::Version::parse(&version)
+        // should be impossible unless there is a semver incompatible version in the db
+        // Note that there is a redirect earlier for semver matches to the exact version
+        .map_err(|_| IronError::new(Nope::InternalServerError, status::InternalServerError))?
```
jyn514, line 374: Can you add a log message here that says the error was invalid semver? That would help us notice these bugs.

jyn514 (approved):
LGTM with nits fixed

src/web/rustdoc.rs
jyn514: Hmm, the `Nope` error would just turn into 'InternalServerError', right? I think that would be useful after #1046 to print the log message automatically (in the `from` impl) but other than that I don't think it buys us much.

jyn514:
Thanks for working on this (and putting up with all my nits)!

robinhundt:
No problem :) The code is better that way and you're reviewing really fast :smile: :+1:

Issue #1005, opened 2020-08-26
Styles are broken on old rustdoc pages


jyn514:
Probably because of #981.
![Screenshot_20200826-075814_Ecosia](https://user-images.githubusercontent.com/23638587/91300919-0e465600-e772-11ea-8b0f-07c9c39d4451.jpg)

jyn514:
Hmm it's not actually related to mobile but to the version of the page. Maybe the rustdoc page structure changed and #930 didn't take the old structure into account?
![Screenshot_20200826-080015_Ecosia](https://user-images.githubusercontent.com/23638587/91301040-42ba1200-e772-11ea-921d-70e84f41ef18.jpg)

jyn514:
I tried removing the duplicate `rustdoc-navigation` (https://github.com/rust-lang/docs.rs/issues/1027#issuecomment-685980959) without success. I think the issue is that `vendored.css` is missing on old rustdoc pages.

Issue #992, opened 2020-08-22
Monospace fonts are broken again


jyn514:
Mostly likely by #981.

https://docs.rs/pulldown-cmark/0.7.2/pulldown_cmark/struct.Parser.html#method.new_with_broken_link_callback
Before: ![image](https://user-images.githubusercontent.com/23638587/90943948-27ee4300-e3ea-11ea-93da-adcf9ff32fab.png)
Currently (but system-dependent): ![image](https://user-images.githubusercontent.com/23638587/90943962-36d4f580-e3ea-11ea-849c-fee313c7ceb0.png)

The culprit is `style.css`, which, now that all CSS is in one file, is not particularly helpful.

```
code, kbd, pre, samp {
    font-family: monospace,monospace;
    font-size: 1em;
}
```

cc @Kixiron

PR #1059, opened 2020-09-19
Fix rustdoc styles for very old docs

jyn514:
Previously this used the selector `link[type='text/css'][href*='normalize']` to inject the vendored HTML.
Unfortunately, this selector did not exist in very old versions of rustdoc.
Instead, this now adds the `vendored.css` stylesheet at the start of the `<head>` tag.

Closes https://github.com/rust-lang/docs.rs/issues/1005
r? @cynecx

### DIFF
--- a/src/utils/html.rs
+++ b/src/utils/html.rs
@@ -64,7 +64,8 @@ pub(crate) fn rewrite_lol(
         Ok(())
     };
 
-    // Append `vendored.css` before the first stylesheet (rustdoc's first stylesheet is `normalize.css`).
+    // Append `vendored.css` before `rustdoc.css`, so that the duplicate copy of
+    // `normalize.css` will be overridden by the later version.
     let first_stylesheet_handler = |head: &mut Element| {
         head.before(&tera_vendored_css, ContentType::Html);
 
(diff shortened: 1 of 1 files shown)


jyn514:
Ugh, this brings back the issues in https://github.com/rust-lang/docs.rs/issues/992.

jyn514:
Ok, I think the latest fix might have done it.
image

cynecx:
I am not quite sure about order of elements inside the `head` element, in terms of semantic differences (I'd rather not put css before the viewport/charset meta tags).

An alternative fix would be to change the selector to `link[type='text/css'][href*='rustdoc']`, that would put the vendored css after normalize.css. That would work too because the vendored css does also contain a version of normalize.css.

@jyn514 Thoughts?

jyn514:
This page does not have a `rustdoc.css`: https://docs.rs/syntex_syntax/0.54.0/syntex_syntax/ast/enum.MetaItemKind.html

jyn514:
Oh whoops, it does have `rustdoc-20161205-1.15.0-nightly-daf8c1dfc.css` though. Yeah, that seems like a better approach.

cynecx:
LGTM!

PR #1062, opened 2020-09-20
Capture more metadata about crates and stop using json for dependencies

Kixiron:
The goals of this pr are simple: Get some more data about crates so that I can implement a features display, and stop using json to represent dependencies within the `releases` table.

The structs that parse out info from `cargo metadata`'s output got a little refactor and now get everything we need, so that part is done. The part I'm not having a fun time with is the database side of things, migrating the old json columns into the new table. There's also some only vaguely finished stuff (e.g. dependency versioning, the idea is that we'll have an enum on our side that tells us if it's the old and limited-in-data picture of the dependency or the new rich one, this is only actually mentioned in the migration query atm). The type-unsafe nature of our current database interactions is also a source of annoyance since I now have to manually hunt down points of interaction with the dependencies column, which as of now has not been done.

### DIFF
--- a/crates/metadata/build.rs
+++ b/crates/metadata/build.rs
@@ -4,6 +4,7 @@ fn main() {
         "cargo:rustc-env=DOCS_RS_METADATA_HOST_TARGET={}",
         std::env::var("TARGET").unwrap(),
     );
+
     // This only needs to be rerun if the TARGET changed, in which case cargo reruns it anyway.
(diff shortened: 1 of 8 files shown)


jyn514:
Personally I'd prefer to have https://github.com/rust-lang/docs.rs/issues/874 before making large changes like this.

jyn514:
Also, please split out the formatting changes into a separate commit so it's easier to see the logic.

PR #1068, opened 2020-09-25
Update index map

Nemo157:
Avoids rebuilding every time when using `--remap-path-prefix` (https://github.com/bluss/indexmap/issues/123)


jyn514:
This adds a new hashbrown dependency :(

Nemo157:
https://github.com/bluss/indexmap/pull/131, sounds like it's a large benefit.

PR #1072, opened 2020-09-30
Extend coverage feature

GuillaumeGomez:
Like we talked about some time ago, the doc examples have been added to rustdoc so we can now use them on docs.rs as well!

<s>For information, I decided to go for this computation: the documentation itself is worth 80% of the rating, the doc examples is the remaining 20%.</s>

After debate, for now it won't be taken into account in the percentage computation.

Here is what it looks like:

Screenshot from 2020-10-01 16-06-23

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,12 @@ To rebuild the site, run `docker-compose build`.
 Note that docker-compose caches the build even if you change the source code,
 so this will be necessary anytime you make changes.
 
+If you want to completely clean up the database, don't forget to remove the volumes too:
+
+```
+$ docker-compose down --volumes
+```
+
 #### FAQ
 
 ##### I keep getting the error `standard_init_linux.go:211: exec user process caused "no such file or directory"` when I use docker-compose.
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -438,6 +438,25 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
             ALTER TABLE owners ALTER COLUMN name DROP NOT NULL;
             ALTER TABLE crates ALTER COLUMN github_stars DROP NOT NULL;
             "
+        ),
+        migration!(
+            context,
+            // version
+            18,
+            // description
+            "Add more information into doc coverage",
+            // upgrade query
+            "
+            ALTER TABLE doc_coverage
+                ADD COLUMN total_items_needing_examples INT,
+                ADD COLUMN items_with_examples INT;
+            ",
+            // downgrade query
+            "
+            ALTER TABLE doc_coverage
+                DROP COLUMN total_items_needing_examples,
+                DROP COLUMN items_with_examples;
+            "
         )
     ];
 
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -45,6 +45,8 @@ pub struct CrateDetails {
     documentation_url: Option<String>,
     total_items: Option<f32>,
     documented_items: Option<f32>,
+    total_items_needing_examples: Option<f32>,
+    items_with_examples: Option<f32>,
 }
 
 fn optional_markdown<S>(markdown: &Option<String>, serializer: S) -> Result<S::Ok, S::Error>
@@ -100,7 +102,9 @@ impl CrateDetails {
                 releases.documentation_url,
                 releases.default_target,
                 doc_coverage.total_items,
-                doc_coverage.documented_items
+                doc_coverage.documented_items,
+                doc_coverage.total_items_needing_examples,
+                doc_coverage.items_with_examples
             FROM releases
             INNER JOIN crates ON releases.crate_id = crates.id
             LEFT JOIN doc_coverage ON doc_coverage.release_id = releases.id
@@ -156,6 +160,8 @@ impl CrateDetails {
 
         let documented_items: Option<i32> = krate.get("documented_items");
         let total_items: Option<i32> = krate.get("total_items");
+        let total_items_needing_examples: Option<i32> = krate.get("total_items_needing_examples");
+        let items_with_examples: Option<i32> = krate.get("items_with_examples");
 
         let mut crate_details = CrateDetails {
             name: krate.get("name"),
@@ -189,6 +195,8 @@ impl CrateDetails {
             documentation_url: krate.get("documentation_url"),
             documented_items: documented_items.map(|v| v as f32),
             total_items: total_items.map(|v| v as f32),
+            total_items_needing_examples: total_items_needing_examples.map(|v| v as f32),
+            items_with_examples: items_with_examples.map(|v| v as f32),
         };
 
         if let Some(repository_url) = crate_details.repository_url.clone() {
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -137,17 +137,24 @@ pub(crate) fn add_doc_coverage(
 ) -> Result<i32> {
     debug!("Adding doc coverage into database");
     let rows = conn.query(
-        "INSERT INTO doc_coverage (release_id, total_items, documented_items)
-            VALUES ($1, $2, $3)
+        "INSERT INTO doc_coverage (
+            release_id, total_items, documented_items,
+            total_items_needing_examples, items_with_examples
+        )
+            VALUES ($1, $2, $3, $4, $5)
             ON CONFLICT (release_id) DO UPDATE
                 SET
                     total_items = $2,
-                    documented_items = $3
+                    documented_items = $3,
+                    total_items_needing_examples = $4,
+                    items_with_examples = $5
             RETURNING release_id",
         &[
             &release_id,
             &doc_coverage.total_items,
             &doc_coverage.documented_items,
+            &doc_coverage.total_items_needing_examples,
+            &doc_coverage.items_with_examples,
         ],
     )?;
     Ok(rows[0].get(0))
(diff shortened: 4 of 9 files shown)


jyn514:
Hmm, I'm not sure we should count them as the same percentage - most people don't even know `missing_doc_examples` exists. Maybe we could have two separate percentages, and make doc-examples a smaller font or something?

jyn514:
cc @sunjay and @yaahc who had concerns on https://github.com/rust-lang/docs.rs/pull/939

README.md
```diff
@@ -137,6 +137,12 @@ To rebuild the site, run `docker-compose build`.
 Note that docker-compose caches the build even if you change the source code,
 so this will be necessary anytime you make changes.
 
+If want to completely clean up the database, don't forget to remove the volumes too:
```
jyn514, line 140:
```suggestion
If you want to completely clean up the database, don't forget to remove the volumes too:
```

src/db/migrate.rs
```diff
@@ -438,6 +438,25 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
+            context,
+            // version
+            18,
+            // description
+            "Add more information into doc coverage",
+            // upgrade query
+            "
+            ALTER TABLE doc_coverage
+                ADD COLUMN total_items_needing_examples INT,
+                ADD COLUMN items_with_examples INT;
```
jyn514, line 452:
```suggestion
                ADD COLUMN total_items_needing_examples INT NOT NULL,
                ADD COLUMN items_with_examples INT NOT NULL;
```

src/web/crate_details.rs
```diff
@@ -100,7 +102,9 @@ impl CrateDetails {
                 releases.documentation_url,
                 releases.default_target,
                 doc_coverage.total_items,
-                doc_coverage.documented_items
+                doc_coverage.documented_items,
+                COALESCE(doc_coverage.total_items_needing_examples, 0) AS total_items_needing_examples,
+                COALESCE(doc_coverage.items_with_examples, 0) AS items_with_examples
```
jyn514, line 107: Hmm, I don't think this should be a COALESCE. If there's no data, we shouldn't display the percentage, not say that it's 0.

src/db/migrate.rs
GuillaumeGomez: Won't it break for existing content?

sunjay:
> Hmm, I'm not sure we should count them as the same percentage - most people don't even know `missing_doc_examples` exists. Maybe we could have two separate percentages, and make doc-examples a smaller font or something?

Yeah I'm also not sure about having them as the same percentage. If that ends up being the way we do it, we should at least include an asterisk somewhere to tell people how it's calculated. Don't want to cause unnecessary confusion.

src/db/migrate.rs
jyn514: Oh you're right, there are existing docs without this info. Yeah, this needs to be nullable then.

Manishearth:
I still share @sunjay's concerns from https://github.com/rust-lang/docs.rs/pull/939#issuecomment-673047077 about this feature, but less so because it's not as prominent anymore.

I'm quite wary of coming up with an arbitrary formula for this. What's the reasoning behind it? Why 80%? And yeah we should be transparent about it, or perhaps just include two different numbers?

GuillaumeGomez:
> I'm quite wary of coming up with an arbitrary formula for this. What's the reasoning behind it? Why 80%? And yeah we should be transparent about it, or perhaps just include two different numbers?

I agree, we should at least put the explanation of the computation somewhere. Just don't know where... As for why not two numbers: because the ultimate goal is to provide this information alongside crates search results (when you look for crates on docs.rs, not when you search something in a crate).

pietroalbini:
I wouldn't include examples in the percentage, as for a lot of items examples are just noise. I think it's fine to display that number, but it shouldn't count for the %.

GuillaumeGomez:
> I wouldn't include examples in the percentage, as for a lot of items examples are just noise

Interesting argument! Do you have some examples in mind by any chance?

pietroalbini:
I think a big example is builders ([like `Command` in Rustwide](https://docs.rs/rustwide/0.10.0/rustwide/cmd/struct.Command.html)), where you have a bunch of `fn set_option(self, new_value: bool) -> Self` that don't really have any meaningful examples (but still need text documentation to explain what the option does).

GuillaumeGomez:
The example in itself isn't useful but:

1. It doesn't hurt to have it
2. It allows to copy/paste code directly

So it is definitely worth it having examples even then in my opinion.

pietroalbini:
I feel like useless examples are just noise: if a crate author wants to add them that's fine, but we shouldn't encourage people to do so.

GuillaumeGomez:
Unfortunately, we can't have shades here: it's either we enforce it on everything (there are still items that aren't required by the lint and those are fines) or on nothing... Methods, types and functions should always be documented and have a code example.

Nemo157:
I agree with that having an example on every item should not be necessary for the overall quality indicator. As another example I have [this fully public error type](https://docs.rs/bs58/0.3.1/bs58/decode/enum.Error.html) which is [shown in other examples](https://docs.rs/bs58/0.3.1/bs58/fn.decode.html#errors). Adding an example on each variant (judging by the percentage shown public fields are not required to have examples?) would be useless extra detail clogging up the documentation. I personally consider this crate to be nearly 100% covered with examples (less adding some examples for the single item `Alphabet`), but `--show-coverage` gives it just a 47% or 54% rating depending on features.

GuillaumeGomez:
You're missing my point (also, at least one short example should be on the type directly with a link to the examples, but that's another debate): we can relax the rules later on, but currently, most crates are strongly lacking documentation. For some of them, it's fine (FFI ones), but it's problematic for all others.

Nemo157:
> at least one short example should be on the type directly with a link to the examples

I very much disagree with the first half of this sentence (but the latter is definitely something I should add :grin:).

Actually testing adding empty examples to stuff it appears I was wrong about the variants being the source of low coverage. The problem is that I don't have examples on the types and instead link out to the more centralized documentation. (There does seem like 1 bug here that type aliases probably shouldn't be in the set of items that require examples?).

GuillaumeGomez:
It's open to debate. The option is still nightly (and will remain for some time I expect), so you can definitely open an issue on the rust repository so it can be discussed. But yes, fields and variants are excluded from this count. ;)

jyn514:
I think that especially since this option is experimental and changing rapidly, we should not put it front and center on the /crate page. I'd be fine with listing the number of examples:
image

But not with showing the percentage:
image

And certainly not by including the examples in the 'overall documentation' percentage:

image

@GuillaumeGomez does that first option sound reasonable to you?

GuillaumeGomez:
I'm not too strongly in favor of merging both results in the computed percentage, even though I think we should. For now we can simply show the numbers. The important part for me is to have the data available so we can use it later on. If it's fine with you, I'll then keep the display of the example numbers (not the percentage) and remove it from the total percentage computation.

src/web/crate_details.rs
GuillaumeGomez: Ok!

jyn514:
> If it's fine with you, I'll then keep the display of the example numbers (not the percentage) and remove it from the total percentage computation.

Sounds like a plan :)

GuillaumeGomez:
Updated!

Manishearth:
> As for why not two numbers: because the ultimate goal is to provide this information alongside crates search results (when you look for crates on docs.rs, not when you search something in a crate).

I am very strongly against this, for the reasons @sunjay articulated in his comment on the other thread. I do not think we should engineer a race-for-metrics via some coverage number.

I've _already_ heard people complaining about flaws in lib.rs' ranking algorithm, and of crates sometimes doing things that don't improve them to improve their "score"

If this is the direction we're going in I'd like to see a holistic plan (instead of incremental changes), and, ideally, an RFC.

sunjay:
I strongly agree with what Manish said. I am also very against putting this in the search results for every crate. 

A proposal I could get on board with:

* Make this into a badge that people can add to their README and/or Cargo.toml
* Display badges in the search results

That way people can have the metric and choose where they want it to be displayed. It's a win-win. 

Overall, I don't think this is a bad feature. It makes sense to provide metrics like this just like it makes sense to provide people with the ability to generate code coverage reports. I just think that people shouldn't be forced into having to care about a number that they might not care about and that doesn't even accurately represent the quality of their crate. Putting this number in the search results would definitely be overstating its importance. 

I think Manish put it perfectly when he said:

> I do not think we should engineer a race-for-metrics via some coverage number.

I completely agree with this.

GuillaumeGomez:
This isn't supposed to impact ranking, just a visual indication alongside the crate.

templates/rustdoc/header.html
```diff
@@ -161,7 +161,7 @@
                                         <ul class="pure-menu-list">
                                             <li>
                                                 <a href="{{ crate_url | safe }}" class="pure-menu-link">
-                                                    <b>{{ percent | round(precision=2) }}%</b>
+                                                    <b>{{ percent | round(precision=2 )}}%</b>
```
jyn514, line 164: I don't think you meant to change this.
```suggestion
                                                    <b>{{ percent | round(precision=2) }}%</b>
```

jyn514 (approved):

templates/rustdoc/header.html
GuillaumeGomez: Good catch!

Issue #52, opened 2016-09-20
help menu for doc make the whole page blur


Kerollmops:
When pressing `?` the whole page is blur, a little error in the css I supposed 😄

onur:
Thanks for reporting.

But I am not sure when this occurs, can you give a link?

Kerollmops:
go on this page https://docs.rs/keen/1.2.0/keen/ and press `?`

onur:
woah thanks for reporting, I have an addon replacing `?` content thats why I never saw it. Let me investigate what is causing this blur.

Kerollmops:
What is this addon ? 😄  And what is it for ?

onur:
Ahh it's nothing related with rust. VimFx was replacing help content for me.

This issue is caused by this CSS rules defined in rustdoc.css:

```css
body.blur > :not(#help) {
    filter: blur(8px);
    -webkit-filter: blur(8px);
    opacity: .7;
}
```

It is adding a blur filter to every child element of `body.blur` except `#help` child. Problem is, in docs.rs help is inside `body.blur > div.rustdoc > #help`. I don't know how to override this, I hope someone can help.

Kerollmops:
Why not using `-Z unstable-options --extend-css`.

https://blog.guillaume-gomez.fr/articles/2016-09-16+Generating+doc+with+rustdoc+and+a+custom+theme

onur:
Well actually docs.rs is not just extending css. It is adding some navigation and moving rustdoc content to div.rustdoc. You can see it in [rustdoc.hbs](https://github.com/onur/docs.rs/blob/444ff0cb43d915cf0c212a56812e51440cf42172/templates/rustdoc.hbs). I don't think `--extend-css` will make any difference.

Kerollmops:
Can we just erase this rule:
```
body.blur > :not(#help) {
    filter: blur(8px);
    -webkit-filter: blur(8px);
    opacity: .7;
}
```

by this one:
```
div.rustdoc.blur > :not(#help) {
    filter: blur(8px);
    -webkit-filter: blur(8px);
    opacity: .7;
}
```

and erase the previous rule: http://stackoverflow.com/questions/15901030/reset-remove-css-styles-for-element-only

This is ugly, no ?

Kerollmops:
In fact that's the javascript script !
It touch the DOM and add the `.blur` class to the `body` element.

onur:
[...]

PR #1080, opened 2020-10-08
Fix help blurring whole page

benediktwerner:
Currently, when pressing the `?` key or the little `?` button next to the search bar that was recently added to rustdoc, the whole page is blurred. This makes the help unreadable.

This is caused by [doc.rs](doc.rs)'s "sandboxing"/"containerization" of rustdoc which messes up rustdoc's style. This was previously discussed in #52 and fixed in a8a3ece but a recent refactoring of the stylesheet in #943 messed up the fix.

From a cursory glance, it looks like all other changes in that PR were correct but the rule for `body.blur` was changed to `body { .blur { ... } }` in SASS which gets translated to `body .blur` (with a space in between, which has a different meaning).

This PR fixes this problem by changing the rule to `body { &.blur { ... } }` which SASS now again correctly translates to `body.blur { ... }`, therefore restoring the correct behavior of the previous fix, causing only the background to be blurred again.

### DIFF
--- a/templates/style/base.scss
+++ b/templates/style/base.scss
@@ -41,17 +41,19 @@ body {
 
     // this is a super nasty override for help dialog in rustdocs
     // see #52 for details
-    .blur > :not(#help) {
-        filter: none;
-        -webkit-filter: none;
-    }
+    &.blur {
+        > :not(#help) {
+            filter: none;
+            -webkit-filter: none;
+        }
 
-    .blur > div.nav-container > *,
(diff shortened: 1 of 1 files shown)


jyn514:
Nice, yes this looks like the right fix.

r? @GuillaumeGomez - I like this fix better than https://github.com/rust-lang/rust/pull/77651, I had a feeling this was something on the docs.rs side.

GuillaumeGomez:
I'm fine with it too. :)

Issue #1077, opened 2020-10-07
Clicking menu title changes pages


jsha:
Steps to reproduce:

1. Visit a crate page, like https://docs.rs/ureq/1.5.1/ureq/.
2. Hover over the second item in the top bar, that has the crate name and version ("ureq-1.5.1").
3. A menu pops up on hover.
4. Click the second item in the top bar ("ureq-1.5.1").

Expected result:

Menu goes away.

Actual result:

Page navigates to a different page, https://docs.rs/crate/ureq/1.5.1.

Note that this problem is worse on touchscreen-based devices, since the user has to tap to open the menu in the first place, so they would reasonably expect that tapping in the same place would make the menu go away again.

jyn514:
The menu taking you to a separate page is intended behavior - where else would you suggest to put the link? FWIW approximately the same information is available in the dropdown and on the new page.

It's strange that the menu opens for you on click, on my android it redirects me immediately without opening the dropdown.

Nemo157:
IIRC iOS Safari has special handling to support `:hover` elements on first click.

jsha:
Yes, I should have said: The touchscreen device I tested was an iPad. On Android I see the behavior @jyn514 describes. Although actually, it flashes the menu for a moment first before navigating to the crate page.

As a user, this navigation was confusing for a few reasons. Probably the main one was that the design of the crate page is so different from the documentation page that I though I'd wound up on an older part of the site, possibly by accident.

![image](https://user-images.githubusercontent.com/220205/95359503-13bdb280-087f-11eb-914a-0768dd34765d.png)

![image](https://user-images.githubusercontent.com/220205/95359480-0b657780-087f-11eb-9cb2-6772570efcd8.png)

For example, the big grey bar switches from vertical to horizontal orientation; the Rust logo goes away; The menu bar moves from left-anchored to centered; the items on the menu change; the logo on the "docs.rs" homepage link changes.

It was also confusing because the content on the crate page looked very similar to the content on the documentation page. In the case of the crate I was looking at, the top of the crate README replicates a lot of what's at the top of the documentation. I think that's probably a common pattern: crate authors want new visitors to see certain intro information whether they came to the crate page or the documentation page.

[...]

PR #1081, opened 2020-10-08
Make crate menu onclick rather than onhover.

jsha:
This makes for a more consistent experience on mobile and desktop.

It also, in my opinion, makes for better usability. It's always clear
how to close a menu - the same way as you opened it, by clicking. It
means that menus aren't opened by stray mouse movements. There are more
details on the usability of hover menus at
https://uxmovement.com/navigation/why-hover-menus-do-users-more-harm-than-good/.

Followup from #1077. I realize we hadn't concluded that changing to onclick
was the desired outcome, so consider this a speculative PR to demonstrate
the concept and see how it feels.

### DIFF
--- a/templates/rustdoc/header.html
+++ b/templates/rustdoc/header.html
@@ -21,8 +21,8 @@
                 </a>
 
                 <ul class="pure-menu-list">
-                    <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
-                        <a href="{{ crate_url | safe }}" class="pure-menu-link" title="{{ krate.description }}">
+                    <li class="pure-menu-item pure-menu-has-children">
+                        <a href="#" class="pure-menu-link" title="{{ krate.description }}">
                             {{ "cube" | fas(fw=true) }}
                             <span class="title"> {{ krate.name }}-{{ krate.version }}</span>
                         </a>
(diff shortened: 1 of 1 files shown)


templates/rustdoc/header.html
```diff
@@ -96,6 +96,10 @@
                                             <a href="https://crates.io/crates/{{ krate.name }}" class="pure-menu-link" title="See {{ krate.name }} in crates.io">
                                                 {{ "cube" | fas(fw=true) }} Crates.io
                                             </a>
+
+                                            <a href="{{ crate_url | safe }}" class="pure-menu-link" title="See {{ krate.name }} in docs.rs">
+                                                {{ "cube" | fas(fw=true) }} Crate page on docs.rs
```
jyn514, line 101: Hmm, I don't think this is a great message because we're already on docs.rs. Maybe something like this?
```suggestion
                                                {{ "cube" | fas(fw=true) }} More information
```

GuillaumeGomez:
Looks good to me as well!

jyn514:
This has a slight bug: when you hit space to close the menu (navigating with keyboard instead of mouse) it will scroll down the page, not just close the menu. Do you know how to fix that?

jyn514:
Actually, this is a pre-existing bug, no need to fix it here.

PR #1086, opened 2020-10-10
Show READMEs even if the build fails

jyn514:
r? @Kixiron
cc @devsnek

This makes the build page much more useful for libraries where the build will _never_ succeed (usually due to depending on libraries that can't be packaged).


Kixiron (approved):
LGTM

PR #1090, opened 2020-10-12
Refactor static file handling

### LINKED ISSUE #55: "The requested crate does not exist" message if a version does not exist

Nemo157:
* Moved the static files out of templates/ and src/ (both as static files that got included in the build, and as static strings in the source) into the static/ directory

 * Copied the static/ directory into the docker image as static/

 * Removed the `staticfile` handler serving from $CRATESFYI_PREFIX/public_html/ since those files are now in static/

 * Add metrics tracking DatabaseFileHandler usage 

Fixes #55


src/web/routes.rs
```diff
@@ -10,9 +10,18 @@ pub(super) const DOC_RUST_LANG_ORG_REDIRECTS: &[&str] =
 pub(super) fn build_routes() -> Routes {
     let mut routes = Routes::new();
 
-    routes.static_resource("/robots.txt", super::sitemap::robots_txt_handler);
-    routes.static_resource("/sitemap.xml", super::sitemap::sitemap_handler);
-    routes.static_resource("/opensearch.xml", super::opensearch_xml_handler);
+    routes.static_resource("/robots.txt", PermanentRedirect("/-/static/robots.txt"));
+    routes.static_resource("/favicon.ico", PermanentRedirect("/-/static/favicon.ico"));
```
jyn514, line 14: Are you sure crawlers/browsers will follow this redirect? I guess there's no reason they wouldn't.

jyn514 (commented):
Looks reasonable, I want to play around with it a little more and make sure it behaves the same. Thanks for working on this!

src/web/routes.rs
pietroalbini: [Seems like Google supports up to 5 redirections](https://developers.google.com/search/reference/robots_txt#handling-http-result-codes)

Nemo157: Firefox and mobile Safari followed the favicon redirect in my testing.

src/web/routes.rs
```diff
@@ -10,9 +10,18 @@ pub(super) const DOC_RUST_LANG_ORG_REDIRECTS: &[&str] =
...
+
+    // These should not need to be served from the root as we reference the inner path in links,
+    // but clients might have cached the url and need to update it.
+    routes.static_resource(
+        "/opensearch.xml",
+        PermanentRedirect("/-/static/opensearch.xml"),
+    );
+    routes.static_resource("/sitemap.xml", PermanentRedirect("/-/sitemap.xml"));
```
pietroalbini, line 22: The sitemap has to live at the top level:

> A sitemap can be posted anywhere on your site, but a sitemap affects only descendants of the parent directory. Therefore, a sitemap posted at the site root can affect all files on the site, which is where we recommend posting your sitemaps.

https://support.google.com/webmasters/answer/183668?hl=en

Nemo157: Ok, that was one I was wondering about, I'll change this back and put in a comment.

PR #1096, opened 2020-10-14
Make right-side menus onclick.

jsha:
This is followup from #1081 and #1077.

This takes the link that was the click action for each menu, and moves it to the first item in the menu.

### DIFF
--- a/templates/header/topbar_begin.html
+++ b/templates/header/topbar_begin.html
@@ -50,7 +50,7 @@
                         </a>
 
                         <ul class="pure-menu-children">
-                            {{ macros::menu_link(href="/releases", text="Releases by Recency") }}
+                            {{ macros::menu_link(href="/releases", text="Recent Releases") }}
                             {{ macros::menu_link(href="/releases/stars", text="Releases by Stars") }}
                             {{ macros::menu_link(href="/releases/recent-failures", text="Recent Build Failures") }}
                             {{ macros::menu_link(href="/releases/failures", text="Build Failures by Stars") }}
@@ -111,4 +111,3 @@
                         </ul>
                     </li>
                 </ul>
-
--- a/templates/header/topbar_begin.html
+++ b/templates/header/topbar_begin.html
@@ -29,7 +29,7 @@
 
                 <ul class="pure-menu-list pure-menu-right">
                     <li class="pure-menu-item pure-menu-has-children pure-menu-opt-children">
-                        <a href="#" class="pure-menu-link">
+                        <a href="/about" class="pure-menu-link">
                             <span title="About">{{ "info-circle" | fas }}</span>
                             <span class="title">About</span>
                         </a>
@@ -44,7 +44,7 @@
                     </li>{#
 
                     #}<li class="pure-menu-item pure-menu-has-children pure-menu-opt-children">
-                        <a href="#" class="pure-menu-link">
+                        <a href="/releases" class="pure-menu-link">
                             <span title="Releases">{{ "leaf" | fas }}</span>
                             <span class="title">Releases</span>
                         </a>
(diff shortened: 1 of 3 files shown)


Nemo157:
One issue here is that the `pure-menu-opt-children` class suppresses showing the children on mobile, turning the "about" menu item into just a direct link to the main about page; I'm not sure how that will behave now that it doesn't link.

jsha:
Ah, good catch. I tested this using the mobile simulator in Chrome, and tapping the "(i)" icon (which is what the "about" menu becomes) does nothing.

So, this change is not ready to merge yet. I'm not sure how to make the item act as a link on mobile (or small screens), while acting to open the menu on larger screens. I'm open to ideas!

jyn514:
Unfortunately I don't know enough about CSS to help you here, but @GuillaumeGomez might have ideas.

templates/header/topbar_begin.html
```diff
@@ -28,27 +28,29 @@
                 </a>
 
                 <ul class="pure-menu-list pure-menu-right">
-                    <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover pure-menu-opt-children">
-                        <a href="/about" class="pure-menu-link">
+                    <li class="pure-menu-item pure-menu-has-children pure-menu-opt-children">
+                        <a href="#" class="pure-menu-link">
```
GuillaumeGomez, line 32: Why keeping this element then?

templates/header/topbar_begin.html
```diff
@@ -28,27 +28,29 @@
...
                             {{ macros::menu_link(href="/about/builds", text="Builds") }}
                             {{ macros::menu_link(href="/about/metadata", text="Metadata") }}
                             {{ macros::menu_link(href="/about/redirections", text="Shorthand URLs") }}
                         </ul>
                     </li>{#
 
-                    #}<li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover pure-menu-opt-children">
-                        <a href="/releases" class="pure-menu-link">
+                    #}<li class="pure-menu-item pure-menu-has-children pure-menu-opt-children">
+                        <a href="#" class="pure-menu-link">
```
GuillaumeGomez, line 47: Why keeping this element?

templates/header/topbar_begin.html
jsha: When I removed the `<a>` tag and just had the text "About", clicks on "About" didn't do anything.

jsha:
I though of a solution to the mobile problem: Right now `menu.js` has a handler that ignores clicks on `#` links. I could change that to also ignore clicks on links that are part of `pure-menu-opt-children`, _unless_ the children are hidden due to being on a small screen, in which case it would allow those clicks through.

jyn514:
> I though of a solution to the mobile problem: Right now `menu.js` has a handler that ignores clicks on `#` links. I could change that to also ignore clicks on links that are part of `pure-menu-opt-children`, _unless_ the children are hidden due to being on a small screen, in which case it would allow those clicks through.

Sounds like a plan :)

templates/header/topbar_begin.html
```diff
@@ -28,27 +28,29 @@
...
                             <span title="Releases">{{ "leaf" | fas }}</span>
                             <span class="title">Releases</span>
                         </a>
 
                         <ul class="pure-menu-children">
+                            {{ macros::menu_link(href="/releases", text="Releases by Recency") }}
```
jyn514, line 53:
```suggestion
                            {{ macros::menu_link(href="/releases", text="Recent Releases") }}
```

templates/header/topbar_begin.html
```diff
@@ -58,12 +60,18 @@
...
-                        <a href="https://www.rust-lang.org/" target="_blank" class="pure-menu-link">
+                    #}<li class="pure-menu-item pure-menu-has-children pure-menu-opt">
+                        <a href="#" target="_blank" class="pure-menu-link">
                             Rust
                         </a>
 
                         <ul class="pure-menu-children">
+                            {{ macros::menu_link(
+                                href="https://www.rust-lang.org/",
+                                text="Rust-Lang.org",
```
jyn514, line 71:
```suggestion
                                text="The Rust Programming Language",
```
or maybe
```suggestion
                                text="Official website of the Rust programming languge",
```

jsha:
I'm going to close this for now; it has conflicts and I'm not likely to pick it up again soon.

PR #1097, opened 2020-10-14
menus: Tweak JS handling of clicks.

jsha:
Prior to this change, if you clicked a menu to open it, then clicked to
close it, the menu heading would still have a grey background
highlighting it. That's due to a CSS rule in pure-menu that causes
focused elements to have a highlighted color. To fix: unfocus the menu
when it is closed via mouse (but leave it focused when it is closed via
keyboard).

Remove a mouseover handler that is not needed anymore.

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -54,6 +54,7 @@
         }
         if (this.parentNode === currentMenu) {
             closeMenu();
+            this.blur();
         } else {
             if (currentMenu) closeMenu();
             openMenu(this.parentNode);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Need to check locally to confirm it but the code looks good.

jyn514:
Tested it out and it works great, thanks :) It does keep the highlighting when you use space to close, but I think that's intentional because it shows the tag that's in focus.

PR #1098, opened 2020-10-14
Force the navbar to be left-aligned on rustdoc pages

Nemo157:
Rather than relying on a different navbar class, I added a `rustdoc-page` class to the body so the navbar style can be overridden. Also moved all the rustdoc overrides into the `_rustdoc.css` file (`.nav-container-rustdoc` is replaced by the `.rustdoc-page .nav-container > .container` style, `.cratesfyi-package-container-rustdoc` was unused).


jyn514:
Looks great, thanks!

Issue #918, opened 2020-07-29
Please provide a dark theme for non-rustdoc pages on docs.rs


joshtriplett:
The rustdoc pages on docs.rs have all the theme options rustdoc normally provides; however, the search pages and the front page always use a light theme, which can be jarring. I'd love to see an adaptation of `ayu.css` for the non-rustdoc portions of docs.rs.

Nemo157:
There is a userstyle for this <https://github.com/Cldfire/ayu-rs>, and @Cldfire has expressed interest in getting it integrated into docs.rs proper.

Cldfire:
I would be more than happy to work on this if we can come up with a plan that is satisfactory for the docs.rs maintainers.

I know @pietroalbini expressed concerns with the idea of styling the front page and such.

jyn514:
More discussion on this around https://discordapp.com/channels/442252698964721669/541978667522195476/731948537591365642. The main thing that was hard was syncing the theme preference to the rustdoc selection, because we don't control rustdoc CSS or the way it stores preferences. If we started with a less ambitious goal of having a dark theme at all, I think this would be more doable.

@Cldfire are you interested in making a PR for adding a dark theme? The relevant code is https://github.com/rust-lang/docs.rs/blob/731d03183d1c65636cc98dc8be15768fe4725969/templates/style.scss#L6.

I guess another issue is how to switch between themes; right now docs.rs doesn't have a settings page at _all_. So maybe we should sync with rustdoc after all? I'm not sure how hard that would be.

joshtriplett:
@jyn514 
> The main thing that was hard was syncing the theme preference to the rustdoc selection, because we don't control rustdoc CSS or the way it stores preferences.

Would it be simpler, rather than tying it to the rustdoc theme selection, to just use the user's browser setting for `prefers-color-scheme`? That would solve the problem for many people.

jyn514:
Ooh, `prefers-color-scheme` is a great idea. Yes, that wouldn't be too hard to implement I think.

Nemo157:
I think having an explicit toggle still is very important for browsers that don't support setting the colorscheme (preferably with a default "use `prefers-color-scheme`" option).

Cldfire:
> Would it be simpler, rather than tying it to the rustdoc theme selection, to just use the user's browser setting for prefers-color-scheme? That would solve the problem for many people.

[...]

PR #1100, opened 2020-10-15
Theme detection

Nemo157:
First step of #918 

Moves all the colours out of the style files and uses CSS variables instead, applying a default set on `html`, then adds a couple of snippets that detect the current rustdoc theme and copies it to a `data-theme` variable on `html`, allowing overriding the CSS variables to change the theming.

Next step after this is to add overrides for all the `pure` specified colours, using CSS variables, so that we can change them; then adding colour sets matching the current `dark` and `ayu` rustdoc themes.

### DIFF
--- a/static/storage-change-detection.html
+++ b/static/storage-change-detection.html
@@ -2,6 +2,12 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8">
+        <!--
+            Only other windows get notified when we change local storage, so
+            this is used in an invisible iframe to send a message to JS in
+            ../templates/rustdoc/body.html when local storage changes so we can
+            detect rustdoc changing the theme
+        -->
         <script type="text/javascript">
             onstorage = function(ev) {
                 parent.postMessage({
--- a/static/storage-change-detection.html
+++ b/static/storage-change-detection.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <script type="text/javascript">
+            onstorage = function(ev) {
+                parent.postMessage({
+                    storage: {
+                        key: ev.key,
+                        value: ev.newValue,
+                    }
+                })
+            }
+        </script>
+    </head>
+</html>
--- a/templates/rustdoc/body.html
+++ b/templates/rustdoc/body.html
@@ -29,4 +29,16 @@
       window.addEventListener("scroll", maybeFixupViewPortPosition, {"once": true});
     }
   }
+
+  window.addEventListener('message', function (ev) {
+    if (ev.data && ev.data.storage && ev.data.storage.key === 'rustdoc-theme') {
+      applyTheme(ev.data.storage.value);
+    }
+  });
 </script>
+<!--
+    Only other windows get notified when we change local storage, so we have an
+    invisible iframe that sends us a message when local storage changes so we
+    can detect rustdoc changing the theme
+-->
+<iframe src="/-/static/storage-change-detection.html" width="0" height="0" style="display: none"></iframe>
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -5,15 +5,15 @@ div.nav-container {
     // Nothing is supposed to be over or hovering the top navbar. Maybe add a few others '('? :)
     z-index: 999;
     height: $top-navbar-height;
-    border-bottom: 1px solid $color-border;
-    background-color: #fff;
+    border-bottom: 1px solid var(--color-border);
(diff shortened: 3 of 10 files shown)


Nemo157:
cc @Cldfire if you have an thoughts

Cldfire:
From my perspective it looks great! This'll make it really easy to add themes. Thank you :)

templates/rustdoc/body.html
```diff
@@ -29,4 +29,16 @@
...
+  window.addEventListener('message', function (ev) {
+    if (ev.data && ev.data.storage && ev.data.storage.key === 'rustdoc-theme') {
+      applyTheme(ev.data.storage.value);
+    }
+  });
 </script>
+<!--
+    Only other windows get notified when we change local storage, so we have an
+    invisible iframe that sends us a message when local storage changes so we
+    can detect rustdoc changing the theme
```
jyn514, line 42: > Only other windows get notified when we change local storage

Who came up with that idea??? 🤦 

Well, if that's the case this looks fine.

templates/style/_navbar.scss
```diff
@@ -5,15 +5,15 @@ div.nav-container {
     // Nothing is supposed to be over or hovering the top navbar. Maybe add a few others '('? :)
     z-index: 999;
     height: $top-navbar-height;
-    border-bottom: 1px solid $color-border;
-    background-color: #fff;
+    border-bottom: 1px solid var(--color-border);
+    background-color: var(--color-background);
```
jyn514, line 9: What's the difference between `$color-border` and `var(--color-border)`? If the change is only because you renamed them in `_themes.scss`, I think I'd prefer to have the old name.

templates/style/_themes.scss
```diff
@@ -0,0 +1,25 @@
+// Standard white theme
+html {
+  --color-background-code: #f5f5f5;
+  --color-background: #fff;
+  --color-border-light: lighten(#ddd, 5%);
+  --color-border: #ddd;
+  --color-doc-link-background: #333;
```
jyn514, line 7: Can you keep the comments from `_vars.scss`? I don't know any of these hex numbers without looking them up.

templates/style/_themes.scss
```diff
@@ -0,0 +1,25 @@
+// Standard white theme
+html {
+  --color-background-code: #f5f5f5;
+  --color-background: #fff;
+  --color-border-light: lighten(#ddd, 5%);
```
jyn514, line 5: Should probably be 
```suggestion
  --color-border-light: lighten(var(--color-border), 5%);
```
although not sure if that syntax works.

templates/style/_navbar.scss
Nemo157: `$color-border` is an SCSS variable, so gets converted into a hard-coded color in the CSS generated. `var(--color-border)` is a CSS variable applied by the browser at runtime, so when we change the theme it changes the value of `--color-border` on the `html` element, which cascades down to all the other elements and updates any styles using `var(--color-border)`.

templates/style/_themes.scss
Nemo157: Nope, `lighten` is an SCSS function which needs the value at compile time, _but_ lucky you mentioned it cause it turns out that function doesn't run in this context for some reason and instead `lighten` is put into the output CSS, I think we need to just precalculate the values here.

templates/style/_themes.scss
Nemo157: Firefox has a handy little display:

image

I think for any style editing it makes sense to prototype it with tools in the browser then copy the values over to the source once it's looking the way you want it to, and use screenshots of the elements in actual usage for review instead of trying to imagine what any changes will look like.

templates/rustdoc/body.html
Nemo157: Yeah, and I was _very_ surprised to find out that neither dedicated or shared workers get the events either, even though they should be treated as a separate context, requiring going all the way to loading another document in an `iframe`.

static/storage-change-detection.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
Kixiron, line 1: Can you duplicate the comment [here](https://github.com/rust-lang/docs.rs/blob/819f5dc8aac441d057b5fe6fdb39d7c1a3c52afe/templates/rustdoc/body.html#L40-L42) so that the purpose of this file is clear?

templates/style/_themes.scss
jyn514: Ok, that makes sense.

jyn514:
Thanks for working on this :)

Issue #1103, opened 2020-10-20
/crate is no longer reachable without javascript


jyn514:
This is a regression introduced by https://github.com/rust-lang/docs.rs/pull/1081 - since dropdowns open using javascript, and clicking on the dropdown itself no longer goes to a new page, there's no longer a way to go to /crate without javascript enabled (or typing the URL in manually).

Rather than reverting the PR, I think the fix should be to make dropdowns open using CSS instead. That may not be feasible - if so, I think I would prefer to revert the PR so the site is still usable without JS.

cc @jsha, do you have time to follow-up on this?

jsha:
Oops, sorry for the regression! I agree that it's best to revert and then roll-forward later with a fix. I'll send that PR.

jyn514:
@jsha not your fault, we really need better UI tests (#996).

Nemo157:
It seems like the new approach in #1096 could solve this too. Have the elements be links, but block them using JS, so without JS they act as normal links.

jsha:
That's possible! I have had a little trouble getting it to work right, but with a few more tries, I think it's pretty feasible.

But, another thought: Can we simplify instead of adding complexity? It seems like each of these menus has a corresponding page that serves the same purpose as the menu, and on mobile, we're satisfied just sending people directly to that page. Why not make that the implementation on desktop as well? Would save some of the HTML used to render menus, and would result in less overall JS and CSS to maintain.

jyn514:
Hmm, I'm a little nervous people will want the old functionality. _Personally_ I think that would be a good change but one of the downsides of having a website is that people can't choose whether or not to update.

PR #1104, opened 2020-10-20
Revert onclick menus.

jsha:
This reverts the changes in #1081 and #1097.

As reported in #1103, those changes broke navigation to the crate page with JavaScript turned off.


jyn514:
FWIW, I'd be ok with leaving this broken temporarily until we get the proper fix in, non-JS users are a small minority and very few people visit /crate.

Issue #1105, opened 2020-10-21
Hyphenated standard library crate does not redirect


dtolnay:
https://docs.rs/serde-json correctly redirects to the same location as https://docs.rs/serde_json but https://docs.rs/proc-macro does not redirect to the same location as https://docs.rs/proc_macro.

PR #1106, opened 2020-10-22
Redirect /proc-macro to /proc_macro

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/1105.

r? @Kixiron

### DIFF
--- a/src/web/routes.rs
+++ b/src/web/routes.rs
@@ -142,6 +142,11 @@ pub(super) fn build_routes() -> Routes {
             super::rustdoc::RustLangRedirector::new(redirect),
         );
     }
+    // redirect proc-macro to proc_macro
+    routes.internal_page(
+        "/proc-macro",
+        super::rustdoc::RustLangRedirector::new("proc_macro"),
+    );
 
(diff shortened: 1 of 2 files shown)


src/web/routes.rs
```diff
@@ -142,6 +142,11 @@ pub(super) fn build_routes() -> Routes {
             super::rustdoc::RustLangRedirector::new(redirect),
         );
     }
+    // redirect proc-macro to proc_macro
+    routes.internal_page(
```
Kixiron, line 146: Couldn't this instead be put with the [rest of the rust-lang redirects](https://github.com/rust-lang/docs.rs/blob/master/src/web/routes.rs#L6-7)?

Kixiron (changes requested):

src/web/routes.rs
jyn514: No, because those hard-code that the source and destination are the same. This redirects `/proc-macro` to /proc_macro`.

Kixiron (approved):
LGTM

Issue #1070, opened 2020-09-30
404 for favicon on recent docs


Nemo157:
e.g. https://docs.rs/chrono/0.4.18/chrono/

Since nightly-2020-09-26 it appears that rustdoc is using a `.svg` favicon, which isn't being stored as part of the essential files.

icp1994:
Is the change not live yet? The chrono favicon still doesn't load for me.

jyn514:
@icp1994 that's #464. I'll see if I can rebuild chrono manually sometime today.

PR #1109, opened 2020-10-22
Add favicon to essential files

jyn514:
This can be tested with `cargo run build add-essential-files && cargo build chrono 0.4.18`.

Closes https://github.com/rust-lang/docs.rs/issues/1070.

r? @Nemo157


Nemo157:
So when I was working on the static files I realised that the per-version favicon.ico is redirected to our favicon.ico, should we do the same and add a static favicon.svg that we redirect to?

Nemo157 (approved):

jyn514:
From our conversation on discord:
- The complexity of special-casing the favicon doesn't seem worth the benefit (~100 copies of the favicon is trivial compared to the rest of our costs)
- Storing a single static favicon would miss changes in new versions and we'd have to update manually.

It might be nice long term to look into de-duplicating files in general, which I imagine would have *enormous* storage cost reductions, but sounds hard to do.

PR #1118, opened 2020-10-23
Add .workspace into .gitignore

GuillaumeGomez:


Nemo157:
The default for `CRATESFYI_RUSTWIDE_WORKSPACE` was changed in https://github.com/rust-lang/docs.rs/commit/5fa1950f7b2c914c5786a75bc6d75df83b6a5719, and `.env.sample` doesn't override it so this is what most developers will be using locally.

PR #1123, opened 2020-10-26
Detect theme changes on all docs.rs pages

Nemo157:
This allows us to change non-rustdoc pages if another open tab is looking at a rustdoc page and changes the theme.

### DIFF
--- a/templates/theme.js
+++ b/templates/theme.js
@@ -1,7 +1,18 @@
-// This is a global function also called from a script in ./rustdoc/body.html
-// which detects when the rustdoc theme is changed
 function applyTheme(theme) {
   document.documentElement.dataset.theme = theme;
 }
+
+window.addEventListener('storage', function (ev) {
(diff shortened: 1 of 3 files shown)


templates/theme.js
```diff
@@ -1,7 +1,18 @@
...
+    applyTheme(ev.newValue);
+  }
+});
+
+// see ../static/storage-change-detection.html for details
+window.addEventListener('message', function (ev) {
+  if (ev.data && ev.data.storage && ev.data.storage.key === 'rustdoc-theme') {
+    applyTheme(ev.data.storage.value);
+  }
+});
```
jyn514, line 16: Do we need both of these listeners? Could we only listen to `storage` and get rid of the iframe?

Nemo157: Then we miss out on knowing when this current window changes the theme, we would only update when another window changes it.

Issue #590, opened 2020-02-06
Feature request: list crate features at the top of package index.html


kespindler:
When learning Rust, it's surprising to not find a namespace available within a package. Having crate features available at the top of the docs.rs package would be a great indication that some namespaces are gated behind features.

jyn514:
I think listing all the features of the crate is useful and in scope (maybe under the `crate` tab). However, we can't put it on the page for every module since those pages are generated by rustdoc.

jyn514:
Or in other words: just finding which features are available is simple, we just parse `Cargo.toml`; finding what changes when a given feature is enabled/disabled requires expanding macros and parsing the code and also lies on a weird boundary between cargo and rustdoc.

jyn514:
There's an extension that adds this already and I like their approach - they put an extra `Features` drop-down in the header at the top.

https://rust-search-extension.now.sh/feature-flags.png

jyn514:
Relevant code:
- [header template](https://github.com/rust-lang/docs.rs/blob/master/templates/navigation_rustdoc.hbs#L103)
- [adding releases to the database](https://github.com/rust-lang/docs.rs/blob/master/src/db/add_package.rs#L28)
- [`Cargo.toml` parsing](https://github.com/rust-lang/docs.rs/blob/master/src/utils/cargo_metadata.rs#L26) - note that this is `CargoMetadata`, not `Metadata`. `Metadata` is only used for controlling the build parameters.

Nemo157:
One change I would suggest for the example drop-down would be to highlight the `default` feature, maybe by moving it to the top and adding a little whitespace/horizontal rule between it and the rest.

jyn514:
This was added in https://github.com/rust-lang/docs.rs/pull/1144.

Issue #1111, opened 2020-10-22
"Platform" button unresponsive with no targets


XAMPPRocky:
The platform button in the navbar currently gives no feedback if there are no platforms. I feel it would more appropriate to display something or hide the button entirely.

- Example crate: https://docs.rs/crate/timeago-cli/0.1.3

jyn514:
Mentoring instructions: In https://github.com/rust-lang/docs.rs/blob/5f4f6cf533e8e5db478df55fc6918855c1b135fd/templates/header/package_navigation.html#L43-L44
check whether `platforms` is empty, not whether it is truthy. https://tera.netlify.app/docs/#control-structures

PR #1129, opened 2020-10-27
Store feature flags per release in database

almusil:
Hello,

I have attempted to implement feature requested in #590. I am not sure if I have managed to find every reference correctly.

### DIFF
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use crate::{
+    db::types::Feature,
     docbuilder::{BuildResult, DocCoverage},
     error::Result,
     index::api::{CrateData, CrateOwner, ReleaseData},
@@ -42,6 +43,7 @@ pub(crate) fn add_package_into_database(
     let dependencies = convert_dependencies(metadata_pkg);
     let rustdoc = get_rustdoc(metadata_pkg, source_dir).unwrap_or(None);
     let readme = get_readme(metadata_pkg, source_dir).unwrap_or(None);
+    let features = get_features(metadata_pkg);
     let is_library = metadata_pkg.is_library();
 
     let rows = conn.query(
@@ -52,12 +54,12 @@ pub(crate) fn add_package_into_database(
             homepage_url, description, description_long, readme,
             authors, keywords, have_examples, downloads, files,
             doc_targets, is_library, doc_rustc_version,
-            documentation_url, default_target
+            documentation_url, default_target, features
          )
          VALUES (
             $1,  $2,  $3,  $4,  $5,  $6,  $7,  $8,  $9,
             $10, $11, $12, $13, $14, $15, $16, $17, $18,
-            $19, $20, $21, $22, $23, $24, $25
+            $19, $20, $21, $22, $23, $24, $25, $26
          )
          ON CONFLICT (crate_id, version) DO UPDATE
             SET release_time = $3,
@@ -82,7 +84,8 @@ pub(crate) fn add_package_into_database(
                 is_library = $22,
                 doc_rustc_version = $23,
                 documentation_url = $24,
-                default_target = $25
+                default_target = $25,
+                features = $26
          RETURNING id",
         &[
             &crate_id,
@@ -110,6 +113,7 @@ pub(crate) fn add_package_into_database(
             &res.rustc_version,
             &metadata_pkg.documentation,
             &default_target,
+            &features,
         ],
     )?;
 
@@ -210,6 +214,30 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)> 
 
             (name, version, kind)
         })
+        .collect()
+}
+
+/// Reads features and converts them to Vec<Feature> with default being first
+fn get_features(pkg: &MetadataPackage) -> Vec<Feature> {
+    let mut features = Vec::with_capacity(pkg.features.len());
+    if let Some(subfeatures) = pkg.features.get("default") {
+        features.push(Feature::new("default".into(), subfeatures.clone()));
+    };
+    features.extend(
+        pkg.features
+            .iter()
+            .filter(|(name, _)| *name != "default")
+            .map(|(name, subfeatures)| Feature::new(name.clone(), subfeatures.clone())),
+    );
+    features.extend(get_optional_dependencies(pkg));
+    features
+}
+
+fn get_optional_dependencies(pkg: &MetadataPackage) -> Vec<Feature> {
+    pkg.dependencies
+        .iter()
+        .filter(|dep| dep.optional)
+        .map(|dep| Feature::new(dep.name.clone(), Vec::new()))
         .collect()
 }
 
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -457,7 +457,24 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
                 DROP COLUMN total_items_needing_examples,
                 DROP COLUMN items_with_examples;
             "
-        )
+        ),
+        migration!(
+            context,
+            // version
+            19,
+            // description
+            "Add features that are available for given release",
+            // upgrade query
+            "
+                CREATE TYPE feature AS (name TEXT, subfeatures TEXT[]);
+                ALTER TABLE releases ADD COLUMN features feature[];
+            ",
+            // downgrade query
+            "
+                ALTER TABLE releases DROP COLUMN features;
+                DROP TYPE feature;                         
+            "
+        ),
     ];
 
     for migration in migrations {
--- a/src/test/fakes.rs
+++ b/src/test/fakes.rs
@@ -5,6 +5,7 @@ use crate::storage::Storage;
 use crate::utils::{Dependency, MetadataPackage, Target};
 use chrono::{DateTime, Utc};
 use failure::Error;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 #[must_use = "FakeRelease does nothing until you call .create()"]
@@ -48,11 +49,21 @@ impl<'a> FakeRelease<'a> {
                     name: "fake-dependency".into(),
                     req: "^1.0.0".into(),
                     kind: None,
+                    optional: false,
                 }],
                 targets: vec![Target::dummy_lib("fake_package".into(), None)],
                 readme: None,
                 keywords: vec!["fake".into(), "package".into()],
                 authors: vec!["Fake Person <fake@example.com>".into()],
+                features: [
+                    ("default".into(), vec!["feature1".into(), "feature3".into()]),
+                    ("feature1".into(), Vec::new()),
+                    ("feature2".into(), vec!["feature1".into()]),
+                    ("feature3".into(), Vec::new()),
+                ]
+                .iter()
+                .cloned()
+                .collect::<HashMap<String, Vec<String>>>(),
             },
             build_result: BuildResult {
                 rustc_version: "rustc 2.0.0-nightly (000000000 1970-01-01)".into(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -265,14 +265,16 @@ checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
 
 [[package]]
 name = "chrono"
-version = "0.4.11"
+version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "80094f509cf8b5ae86a4966a39b3ff66cd7e2a3e594accec3743ff3fabeab5b2"
+checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
 dependencies = [
+ "libc",
  "num-integer",
  "num-traits",
  "serde",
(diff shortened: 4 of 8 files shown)


src/db/migrate.rs
```diff
@@ -457,7 +457,22 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
+        ),
+        migration!(
+            context,
+            // version
+            19,
+            // description
+            "Add features that are available for given release",
+            // upgrade query
+            "
+                ALTER TABLE releases ADD COLUMN features JSON;
```
jyn514, line 469: I'd prefer not to store JSON in the database if possible. Instead, could you make a many-to-many table, like this?
```sql
CREATE TABLE features (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) UNIQUE NOT NULL,
);
CREATE TABLE feature_rels (
    feature_id INTEGER REFERENCES features(id) NOT NULL,
    release_id INTEGER REFERENCES releases(id) NOT NULL,
    CONSTRAINT no_duplicate_features UNIQUE (feature_id, crate_id)
);
```
(cc @Nemo157 to double check my SQL)

templates/rustdoc/topbar.html
```diff
@@ -226,7 +226,25 @@
...
+        <li class="pure-menu-item pure-menu-has-children">
+            <a href="#" class="pure-menu-link">
+                <span class="title">Feature flags</span>
+            </a>
+
+            {# Build the dropdown list showing available features #}
+            <ul class="pure-menu-children">
+                {%- for feature in krate.features -%}
+                    <li class="pure-menu-item">
+                        <a href="#" class="pure-menu-link">{{ feature }}</a>
```
jyn514, line 242: Is there a reason to make these links? It seems weird that clicking on them does nothing.

templates/rustdoc/topbar.html
```diff
@@ -226,7 +226,25 @@
                 </li>
             {%- endfor -%}
         </ul>
-    </li>
+    </li>{#
+    Display the features available in current build
+    #}
+    {% if 'krate' in __tera_context %}
```
jyn514, line 232: What does this do? I would prefer to use `{% if krate.features %}` or things like that instead, that will also avoid bugs like https://github.com/rust-lang/docs.rs/issues/1111.

src/web/crate_details.rs
```diff
@@ -187,6 +189,7 @@ impl CrateDetails {
             total_items: total_items.map(|v| v as f32),
             total_items_needing_examples: total_items_needing_examples.map(|v| v as f32),
             items_with_examples: items_with_examples.map(|v| v as f32),
+            features: krate.get("features"),
```
jyn514, line 192: We should also strip out 'private' features with underscores and optional dependencies. See some of the discussion in https://internals.rust-lang.org/t/docs-rs-crate-features-display/12895/3.

jyn514 (commented):
Thank you so much for working on this!

jyn514:
Screenshots:
image
image
highlighted

templates/rustdoc/topbar.html
jyn514: Also, I think this should be shown on every page that has `Platform`, there's no reason to special case the documentation page.

src/db/migrate.rs
Nemo157: I think we probably want to go further and store the feature and the subfeatures it enables separately so that we can change rendering of them easily.

I'd be tempted to just store them inline since we don't want to query on them:

```sql
create type feature as (name text, subfeatures text[]);
alter table releases add column features feature[];
```

(hopefully the postgres libraries give a good way to work with custom types :crossed_fingers:)

jyn514:
Also, this currently panics on crates built before the feature was added:
```
2020/10/27 12:39:39 [INFO] docs_rs::web: Running docs.rs web server on http://0.0.0.0:3000
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to render \'rustdoc/topbar.html\'"), source: Some(Error { kind: Msg("Tried to iterate on a container (`krate.features`) that has a unsupported type"), source: None }) }', src/utils/html.rs:24:77
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Could you add a test for that?

src/db/migrate.rs
almusil: Done, thanks

templates/rustdoc/topbar.html
almusil: Frankly I was not sure about style of other things, but yes it looks weird, done.

templates/rustdoc/topbar.html
almusil: The first one is done. I am not sure how to address the second one. Can you please give me a hint?

src/web/crate_details.rs
almusil: Added, hopefully I did understand the requirements.

almusil:
I have added two test cases but I am not sure where can I test the add_release.rs part. Can you please help me out?

src/test/fakes.rs
```diff
@@ -48,11 +49,21 @@ impl<'a> FakeRelease<'a> {
...
                 keywords: vec!["fake".into(), "package".into()],
                 authors: vec!["Fake Person <fake@example.com>".into()],
+                features: [
+                    ("default".into(), vec!["feature1".into(), "feature3".into()]),
+                    ("feature1".into(), Vec::new()),
+                    ("feature2".into(), vec!["feature1".into()]),
+                    ("feature3".into(), Vec::new()),
+                ]
+                .iter()
+                .cloned()
```
jyn514, line 65: :cry: https://github.com/rust-lang/rust/pull/65819

src/web/crate_details.rs
```diff
@@ -278,6 +281,15 @@ fn map_to_release(conn: &mut Client, crate_id: i32, version: semver::Version) ->
     }
 }
 
+fn get_features(features: Option<Vec<Feature>>) -> Option<Vec<String>> {
+    features.map(|vec| {
+        vec.iter()
+            .filter(|feature| !feature.is_private())
+            .map(|feature| format!("{}", feature))
```
jyn514, line 288:
```suggestion
            .map(|feature| feature.to_string())
```

src/web/crate_details.rs
```diff
@@ -741,4 +754,42 @@ mod tests {
...
+        });
+    }
+
+    #[test]
+    fn feature_private_feature_flags_are_hidden() {
+        wrapper(|env| {
+            let features = [("_private".into(), Vec::new())]
+                .iter()
+                .cloned()
+                .collect::<HashMap<String, Vec<String>>>();
```
jyn514, line 781: You use this later, so I think the type annotation can be removed.
```suggestion
                .collect()
```

templates/rustdoc/topbar.html
```diff
@@ -226,7 +226,26 @@
                 </li>
             {%- endfor -%}
         </ul>
-    </li>
+    </li>{#
+    Display the features available in current build
+    #}
+    {% if krate.features %}
```
jyn514, line 232: You need `{%-` to remove the whitespace (which @Nemo157 said broke things for some reason).
```suggestion
    {%- if krate.features -%}
```

jyn514 (commented):
> I am not sure where can I test the add_release.rs part

Do you mean how to test documentation for an old crate? You'll have to insert a crate into the database which has `NULL` for features. I think the easiest way would be to make `features` an `Option` in tests only (with `cfg(test)`). Then you could explicitly set it to `None` for the test.

> I am not sure how to address the second one.

`krate` is provided by `WebPage` for rustdoc pages here: https://github.com/rust-lang/docs.rs/blob/750061b0619b0f6b12d4bc6ed4ccd47a25b74fc3/src/web/rustdoc.rs#L437-L446
But on /crate it's called `details` instead: https://github.com/rust-lang/docs.rs/blob/750061b0619b0f6b12d4bc6ed4ccd47a25b74fc3/src/web/crate_details.rs#L302
I think you'll need to rename them both to be consistent (I like `krate` better than `details`) so you can use them both in the header. Otherwise, you'll only ever have `krate` or `details` but not both and it will be annoying to work with.

src/db/add_package.rs
```diff
@@ -213,6 +217,35 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)>
...
+}
+
+fn get_optional_features(pkg: &MetadataPackage) -> Vec<Feature> {
+    pkg.dependencies
+        .iter()
+        .filter(|dep| dep.optional)
+        .filter(|dep| {
+            pkg.features
+                .values()
+                .any(|features| features.iter().any(|sub| sub.eq(&dep.name)))
```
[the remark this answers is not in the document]
Nemo157: You can't have a feature with the same name as a dependency.

Nemo157:
> I think you'll need to rename them both to be consistent (I like `krate` better than `details`) so you can use them both in the header. Otherwise, you'll only ever have `krate` or `details` but not both and it will be annoying to work with.

The header (other than the crate details dropdown) uses `metadata` for the values, which is populated on all the `/crate/.../` subpages, so you would want to put the features on that instead of `CrateDetails` (and update the other pages to pull the details from the database too) https://github.com/rust-lang/docs.rs/blob/7109f405ba56d289f74ad51d9949385ef6cf5a51/src/web/mod.rs#L513

templates/rustdoc/topbar.html
Nemo157: Not _breakage_, just an extra space gets rendered making the spacing between elements in the header inconsistent :grin:

src/db/add_package.rs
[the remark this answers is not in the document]
Nemo157: We probably want to not filter anything out when storing the features into the database, and instead filter them during display so that we can change the heuristics and have them apply to existing data.

jyn514:
If possible it would be nice to also make the display look more like rust-search-extension: https://rust.extension.sh/feature-flags.png. In particular I like making the `feature` look the same as `doc(cfg)` with the blue border, and the alignment of the equal signs. But we can fix those in follow-ups, they don't have to block the initial implementation.

Kixiron:
I'm not really sold on this layout, it looks super mobile-unfriendly and a wee claustrophobic. I think it'd be better as a page alongside the builds and crate page, with a link on the top bar going to it. I'm also not really a fan of the "toml dump" it has currently (the `key = ["value"]`), I'd rather it involve tables or something along those lines so things look pretty

jyn514:
@Kixiron sure, if you want we can make this just collect the features in the database and we can work out the frontend later, like we did for #1100 . But I'd like to start collecting features quickly since otherwise we'll have to rebuild crates for the data to show up.

Kixiron:
Sounds good to me, more data is always good

almusil:
> @Kixiron sure, if you want we can make this just collect the features in the database and we can work out the frontend later, like we did for #1100 . But I'd like to start collecting features quickly since otherwise we'll have to rebuild crates for the data to show up.

@jyn514  So should I skip the frontend part entirely and just put the features into database?

jyn514:
@almusil for now, let's do that. If you want to avoid throwing away your work you could make a follow-up PR with the design changes and we can bikeshed that.

src/db/add_package.rs
```diff
@@ -213,6 +217,30 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)>
...
+/// Reads features and converts them to Vec<Feature> with default being first
+fn get_features(pkg: &MetadataPackage) -> Vec<Feature> {
+    let mut features = Vec::with_capacity(pkg.features.len());
+    if let Some(subfeatures) = pkg.features.get("default") {
+        features.push(Feature::new("default".into(), subfeatures.clone()));
+    };
+    features.extend(
+        pkg.features
+            .iter()
+            .filter(|(name, _)| name != "default")
```
jyn514:
```suggestion
            .filter(|(name, _)| *name != "default")
```
```
 src/db/add_package.rs#L229
can't compare `&std::string::String` with `str`
```

almusil:
> @almusil for now, let's do that. If you want to avoid throwing away your work you could make a follow-up PR with the design changes and we can bikeshed that.

Ok, I will create additional WIP PR with the design that was here to continue it. Thanks

src/db/add_package.rs
```diff
@@ -213,6 +217,30 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)>
...
+        pkg.features
+            .iter()
+            .filter(|(name, _)| name != &"default")
+            .map(|(name, subfeatures)| Feature::new(name.clone(), subfeatures.clone())),
+    );
+    features.extend(get_optional_features(pkg));
+    features
+}
+
+fn get_optional_features(pkg: &MetadataPackage) -> Vec<Feature> {
```
jyn514, line 236: This is getting optional *dependencies* - can you put that in the name?
```suggestion
fn get_optional_dependencies(pkg: &MetadataPackage) -> Vec<Feature> {
```

jyn514 (commented):
LGTM with nit fixed

src/db/add_package.rs
almusil: Done

src/db/add_package.rs
almusil: Sure, done

PR #1134, opened 2020-10-27
Improve warnings display

GuillaumeGomez:
ayu:

Screenshot from 2020-10-27 16-36-04

dark:

Screenshot from 2020-10-27 16-46-07

### DIFF
--- a/templates/style/_themes.scss
+++ b/templates/style/_themes.scss
@@ -20,6 +20,7 @@ html {
   --color-type: #e57300;
   --color-url: #4d76ae;
   --color-warn-background: #ffe5cc;
+  --color-warn-msg: #e57300;
   --color-warn-hover: #b25900;
   --color-warn: #e57300;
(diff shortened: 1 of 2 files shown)


Nemo157:
Does this affect the "not latest version" warning too? Can we get screenshots of that?

GuillaumeGomez:
Good point, checking!

jyn514:
Default color: 
image
Color with hover: 
hover

We might want to make the color without hover a yellow?

GuillaumeGomez:
For me it's yellow/orange by default?

GuillaumeGomez:
Ah no you're right! Since we changed the color in the display, it's the same in the topbar haha. So this'll be the same problem for ayu. Need to think about that...

GuillaumeGomez:
Updated!

jyn514:
The warnings look a lot better :)

image
image
image
image

PR #1136, opened 2020-10-27
Fix platform handling

jyn514:
- Remove unused platforms tab on /crate page. We always show platforms, even on rustdoc pages.
- Don't show the platform dropdown if it's empty.
- Add a test

Fixes https://github.com/rust-lang/docs.rs/issues/1111.

### DIFF
--- a/templates/rustdoc/topbar.html
+++ b/templates/rustdoc/topbar.html
@@ -213,7 +213,9 @@
     </li>{#
 
     Display the platforms that the release has been built for
-    #}<li class="pure-menu-item pure-menu-has-children">
+    #}
+    {% if metadata.doc_targets %}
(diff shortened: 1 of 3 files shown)


templates/rustdoc/topbar.html
```diff
@@ -210,7 +210,9 @@
     </li>{#
 
     Display the platforms that the release has been built for
-    #}<li class="pure-menu-item pure-menu-has-children">
+    #}
+    {% if metadata.doc_targets %}
```
Nemo157, line 214:
```suggestion
    {%- if metadata.doc_targets -%}
```

IIRC it's important to avoid whitespace here to avoid extra spacing between each menu element.

jyn514:
This is ready for review.

Nemo157 (approved):

PR #1137, opened 2020-10-27
Fix top navbar resize

GuillaumeGomez:
Without the "not latest version":
not-latest

With the "latest version":
latest

### DIFF
--- a/templates/header/topbar_begin.html
+++ b/templates/header/topbar_begin.html
@@ -9,7 +9,8 @@
 <div class="nav-container">
     <div class="container">
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
-            <form action="/releases/search" method="GET" class="landing-search-form-nav">
+            <form action="/releases/search" method="GET"
+                class="landing-search-form-nav {% if not is_latest_version %}not-latest{% endif %}">
                 {# The search bar #}
                 <div id="search-input-nav" class="pure-menu-right">
                     <label for="nav-search">
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -37,6 +37,12 @@ div.nav-container {
             color: var(--color-standard);
             background-color: inherit;
         }
+
+        &.crate-name {
+            text-overflow: ellipsis;
+            overflow: hidden;
+            max-width: 200px;
+        }
     }
 
     .pure-menu-right {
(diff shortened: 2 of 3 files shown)


jyn514 (approved):
This is what it looked like previously:

image
image

Nemo157:
Having a fixed-width name element looks weird with short names, could that be `max-width` instead?

As well as latest version, there's a warning message for yanked crates, it should probably take that into account too.

Nemo157:
Did you try using the right-hand dropdowns when they're in their icon+dropdown form? IIRC they go off the right edge of the screen, which is why I made them disable the dropdown when shrunk.

templates/header/topbar_begin.html
```diff
@@ -9,7 +9,8 @@
 <div class="nav-container">
     <div class="container">
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
-            <form action="/releases/search" method="GET" class="landing-search-form-nav">
+            <form action="/releases/search" method="GET"
+                {% if not is_latest_version %}class="landing-search-form-nav not-latest"{% else %}class="landing-search-form-nav"{% endif %}>
```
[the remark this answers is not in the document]
GuillaumeGomez: I'm still sad that this form doesn't work: `{{ 'not-latest' if not is_latest_version }}` :'(

GuillaumeGomez:
> Having a fixed-width name element looks weird with short names, could that be `max-width` instead?

I tried with max-width but honestly, having an element with a fixed size was much more comfortable (for me at least).

templates/style/_navbar.scss
```diff
@@ -25,6 +25,9 @@ div.nav-container {
         }
         &:after {
             font-size: 0.8em;
+            position: absolute;
```
Nemo157, line 28: This makes the menu feel very cramped to me, and especially there is not enough padding on the right-hand side of the arrow.

GuillaumeGomez: Do you have a screenshot by any chance?

Nemo157: Current look:

image

On this branch:

image

GuillaumeGomez: Yes, that's exactly how I intended to have it. Do you prefer having the arrow right next to the text maybe?

Nemo157: (that screenshot also shows off the extra whitespace after short crate names that I dislike).

Nemo157: IMO the whitespace from the left-border to the icon/text and from the drop-down-arrow to the right border should be the same to make it feel balanced.

GuillaumeGomez: If you have a solution which allows to have a way to have just a max-width and that the items will remove the display of the text by themselves, I'd be glad to hear it (a lot of CSS remains unknown to me).

Nemo157: Changing `width: 200px;` to `max-width: 200px;` appears to work to me.

GuillaumeGomez: It "works". However, the items on the right won't remove the text at the "right" time when the windows size is changing on smaller crate titles.

Nemo157: You mean that there's extra whitespace available they could use when they shrink? Doesn't seem like a problem to me, the exact amount of whitespace there will be influenced by things like the font selected anyway, and not many people drag window sizes around and notice if the responsiveness is _exactly_ right. As long as the media queries target the maximum that could be used it seems fine.

GuillaumeGomez: I guess this is good enough. I'll do that then!

Nemo157 (approved):

PR #1139, opened 2020-10-28
Use a minimal stylesheet for the navbar on rustdoc pages

### LINKED ISSUE #1126: Old docs look bad if current theme is ayu
Since we apply a base set of styles to the body based on the current theme, going to old docs that don't have that theme and fallback to the default `light` theme can interact badly.

Repro:
 * Go to a new crate and switch to ayu
 * Go to an old crate (https://docs.rs/url/2.1.1/url/)

image

Nemo157:
If for some reason rustdoc fails to apply the current theme (e.g. it's set to `ayu` and it's old docs that don't have that theme) then this will result in the navbar applying the theme, but the rustdoc output will fallback to the light theme without interference.

We also don't bother loading the rustdoc specific overrides on non-rustdoc pages.

image

fixes #1126

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -56,7 +56,6 @@ fn compile_sass_file(
         }
     }
 
-    // Compile base.scss
     let mut context = Context::new_file(format!("{}/{}.scss", STYLE_DIR, name))?;
     context.set_options(Options {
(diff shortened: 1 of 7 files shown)


jyn514:
Seems reasonable.

PR #1141, opened 2020-10-29
Do not scroll when closing the menu with space

### LINKED ISSUE #1082: Pressing space to close a header menu (navigating with keyboard instead of mouse) will scroll down the page, not just close the menu
_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/pull/1081#issuecomment-705666955_

notriddle:
Fixes #1082

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -140,6 +140,28 @@
                     // space closes the menu, and activates the current link
                     // this behavior is identical between both the WAI example, and GitHub's
                     if (document.activeElement instanceof HTMLAnchorElement && !document.activeElement.hasAttribute("aria-haspopup")) {
+                        // It's supposed to copy the behaviour of the WAI Menu Bar
+                        // page, and of GitHub's menus. I've been using these two
+                        // sources to judge what is basically "industry standard"
+                        // behaviour for menu keyboard activity on the web.
+                        //
+                        // On GitHub, here's what I notice:
+                        //
+                        // 1 If you click open a menu, the menu button remains
+                        //   focused. If, in this stage, I press space, the menu will
+                        //   close.
+                        //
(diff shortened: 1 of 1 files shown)


static/menu.js
```diff
@@ -128,14 +128,25 @@
...
+                    break;
                 case "space":
                 case " ":
-                    // enter, return, and space have the default browser behavior,
-                    // but they also close the menu
+                    // space closes the menu, and activates the current link
                     // this behavior is identical between both the WAI example, and GitHub's
+                    if (document.activeElement instanceof HTMLAnchorElement && !document.activeElement.hasAttribute("aria-haspopup")) {
+                        document.activeElement.click();
+                    }
```
jyn514, line 144: What does this do?

notriddle: It's supposed to copy the behaviour of the [WAI Menu Bar page](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html), and of GitHub's menus. I've been using these two sources to judge what is basically "industry standard" behaviour for menu keyboard activity on the web.

On GitHub, here's what I notice:

1. If you click open a menu, the menu button remains focused. If, in this stage, I press space, the menu will close.
2. If I use the arrow keys to focus a menu item, and then press space, the menu item will be activated. For example, clicking "+", then pressing down, then pressing space will open the New Repository page.

Behaviour 1 is why the `!document.activeElement.hasAttribute("aria-haspopup")` condition is there. It's to make sure the menu-link on things like the About dropdown don't get activated. Behaviour 2 is why this code is required at all; I want to activate the currently highlighted menu item.

jyn514: Makes sense. Thanks for the explanation - can you add it as a comment?

PR #1142, opened 2020-10-30
Improve incremental compile times

jyn514:
r? @Nemo157 

As a follow-up I want to switch from our own cargo data structures in `cargo_metadata` to https://docs.rs/cargo_metadata/0.12.0/cargo_metadata/ and see if that helps at all.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,3 @@
+[profile.debug]
+# rustc spends an inordinate amount of time on LLVM
+rustflags = ["-C", "no-prepopulate-passes"]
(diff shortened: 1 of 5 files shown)


.cargo/config
```diff
@@ -0,0 +1,3 @@
+[profile.debug]
+# rustc spends an inordinate amount of time on LLVM
+rustflags = ["-C", "no-prepopulate-passes"]
```
Nemo157, line 3: I'm a little worried about how profile rustflags + my global rustflags will interact, but lets see.

Nemo157 (approved):

PR #1144, opened 2020-10-30
frontend: Show feature flags in topbar and separate page

almusil:
As discussed in #1129. Opening the frontend part for discussion. 

Current implementation look:
Feature flags

### DIFF
--- a/src/db/types.rs
+++ b/src/db/types.rs
@@ -1,7 +1,7 @@
 use postgres_types::{FromSql, ToSql};
 use serde::Serialize;
 
-#[derive(Debug, Clone, PartialEq, Serialize, FromSql, ToSql)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, FromSql, ToSql)]
 #[postgres(name = "feature")]
 pub struct Feature {
     name: String,
@@ -11,5 +11,9 @@ pub struct Feature {
 impl Feature {
     pub fn new(name: String, subfeatures: Vec<String>) -> Self {
         Feature { name, subfeatures }
+    }
+
+    pub fn is_private(&self) -> bool {
+        self.name.starts_with('_')
     }
 }
--- a/src/db/types.rs
+++ b/src/db/types.rs
@@ -4,8 +4,8 @@ use serde::Serialize;
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, FromSql, ToSql)]
 #[postgres(name = "feature")]
 pub struct Feature {
-    name: String,
-    subfeatures: Vec<String>,
+    pub(crate) name: String,
+    pub(crate) subfeatures: Vec<String>,
 }
 
 impl Feature {
--- a/src/test/fakes.rs
+++ b/src/test/fakes.rs
@@ -210,6 +210,15 @@ impl<'a> FakeRelease<'a> {
         self
     }
 
+    pub(crate) fn features(mut self, opt_features: Option<HashMap<String, Vec<String>>>) -> Self {
+        if let Some(features) = opt_features {
+            self.package.features = features;
+        } else {
+            self.package.features = HashMap::new();
+        }
+        self
+    }
+
     /// Returns the release_id
     pub(crate) fn create(self) -> Result<i32, Error> {
         use std::fs;
--- a/src/test/fakes.rs
+++ b/src/test/fakes.rs
@@ -210,12 +210,8 @@ impl<'a> FakeRelease<'a> {
         self
     }
 
-    pub(crate) fn features(mut self, opt_features: Option<HashMap<String, Vec<String>>>) -> Self {
-        if let Some(features) = opt_features {
-            self.package.features = features;
-        } else {
-            self.package.features = HashMap::new();
-        }
+    pub(crate) fn features(mut self, features: HashMap<String, Vec<String>>) -> Self {
+        self.package.features = features;
         self
     }
 
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -103,6 +103,7 @@ impl CrateDetails {
                 releases.license,
                 releases.documentation_url,
                 releases.default_target,
+                releases.features,
                 doc_coverage.total_items,
                 doc_coverage.documented_items,
                 doc_coverage.total_items_needing_examples,
@@ -148,6 +149,7 @@ impl CrateDetails {
             default_target: krate.get("default_target"),
             doc_targets: MetaData::parse_doc_targets(krate.get("doc_targets")),
             yanked: krate.get("yanked"),
+            features: MetaData::parse_features(krate.get("features")),
         };
 
         let documented_items: Option<i32> = krate.get("documented_items");
@@ -325,6 +327,7 @@ mod tests {
     use crate::test::{wrapper, TestDatabase};
     use failure::Error;
     use kuchiki::traits::TendrilSink;
+    use std::collections::HashMap;
 
     fn assert_last_successful_build_equals(
         db: &TestDatabase,
@@ -738,6 +741,81 @@ mod tests {
                 vec![("barfoo".into(), "https://example.org/barfoov2".into())]
             );
 
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn feature_flags_report_empty() {
+        wrapper(|env| {
+            env.fake_release()
+                .name("library")
+                .version("0.1.0")
+                .binary(false)
+                .features(None)
+                .create()?;
+
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/crate/library/0.1.0/features")
+                    .send()?
+                    .text()?,
+            );
+            assert!(page.select_first(r#"p[data-id="empty-features"]"#).is_ok());
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn feature_private_feature_flags_are_hidden() {
+        wrapper(|env| {
+            let features = [("_private".into(), Vec::new())]
+                .iter()
+                .cloned()
+                .collect::<HashMap<String, Vec<String>>>();
+            env.fake_release()
+                .name("library")
+                .version("0.1.0")
+                .binary(false)
+                .features(Some(features))
+                .create()?;
+
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/crate/library/0.1.0/features")
+                    .send()?
+                    .text()?,
+            );
+            assert!(page.select_first(r#"p[data-id="empty-features"]"#).is_ok());
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn feature_flags_without_default() {
+        wrapper(|env| {
+            let features = [("feature1".into(), Vec::new())]
+                .iter()
+                .cloned()
+                .collect::<HashMap<String, Vec<String>>>();
+            env.fake_release()
+                .name("library")
+                .version("0.1.0")
+                .binary(false)
+                .features(Some(features))
+                .create()?;
+
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/crate/library/0.1.0/features")
+                    .send()?
+                    .text()?,
+            );
+            assert!(page.select_first(r#"p[data-id="empty-features"]"#).is_err());
+            let def_len = page
+                .select_first(r#"b[data-id="default-feature-len"]"#)
+                .unwrap();
+            assert_eq!(def_len.text_contents(), "0");
             Ok(())
         });
     }
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -103,7 +103,6 @@ impl CrateDetails {
                 releases.license,
                 releases.documentation_url,
                 releases.default_target,
-                releases.features,
                 doc_coverage.total_items,
                 doc_coverage.documented_items,
                 doc_coverage.total_items_needing_examples,
@@ -149,7 +148,6 @@ impl CrateDetails {
             default_target: krate.get("default_target"),
             doc_targets: MetaData::parse_doc_targets(krate.get("doc_targets")),
             yanked: krate.get("yanked"),
-            features: MetaData::parse_features(krate.get("features")),
         };
 
         let documented_items: Option<i32> = krate.get("documented_items");
@@ -818,13 +816,44 @@ mod tests {
     }
 
     #[test]
+    fn feature_flags_with_nested_default() {
+        wrapper(|env| {
+            let features = [
+                ("default".into(), vec!["feature1".into()]),
+                ("feature1".into(), vec!["feature2".into()]),
+                ("feature2".into(), Vec::new()),
+            ]
+            .iter()
+            .cloned()
+            .collect::<HashMap<String, Vec<String>>>();
+            env.fake_release()
+                .name("library")
+                .version("0.1.0")
+                .features(features)
+                .create()?;
+
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/crate/library/0.1.0/features")
+                    .send()?
+                    .text()?,
+            );
+            assert!(page.select_first(r#"p[data-id="empty-features"]"#).is_err());
+            let def_len = page
+                .select_first(r#"b[data-id="default-feature-len"]"#)
+                .unwrap();
+            assert_eq!(def_len.text_contents(), "3");
+            Ok(())
+        });
+    }
+
+    #[test]
     fn feature_flags_report_null() {
         wrapper(|env| {
             let id = env
                 .fake_release()
                 .name("library")
                 .version("0.1.0")
-                .features(HashMap::new())
                 .create()?;
 
             env.db()
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -751,8 +751,7 @@ mod tests {
             env.fake_release()
                 .name("library")
                 .version("0.1.0")
-                .binary(false)
-                .features(None)
+                .features(HashMap::new())
                 .create()?;
 
             let page = kuchiki::parse_html().one(
@@ -776,8 +775,7 @@ mod tests {
             env.fake_release()
                 .name("library")
                 .version("0.1.0")
-                .binary(false)
-                .features(Some(features))
+                .features(features)
                 .create()?;
 
             let page = kuchiki::parse_html().one(
@@ -801,8 +799,7 @@ mod tests {
             env.fake_release()
                 .name("library")
                 .version("0.1.0")
-                .binary(false)
-                .features(Some(features))
+                .features(features)
                 .create()?;
 
             let page = kuchiki::parse_html().one(
@@ -816,6 +813,31 @@ mod tests {
                 .select_first(r#"b[data-id="default-feature-len"]"#)
                 .unwrap();
             assert_eq!(def_len.text_contents(), "0");
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn feature_flags_report_null() {
+        wrapper(|env| {
+            let id = env
+                .fake_release()
+                .name("library")
+                .version("0.1.0")
+                .features(HashMap::new())
+                .create()?;
+
+            env.db()
+                .conn()
+                .query("UPDATE releases SET features = NULL WHERE id = $1", &[&id])?;
+
+            let page = kuchiki::parse_html().one(
+                env.frontend()
+                    .get("/crate/library/0.1.0/features")
+                    .send()?
+                    .text()?,
+            );
+            assert!(page.select_first(r#"p[data-id="null-features"]"#).is_ok());
             Ok(())
         });
     }
--- a/src/web/features.rs
+++ b/src/web/features.rs
@@ -1,0 +1,30 @@
+use crate::{
+    db::Pool,
+    impl_webpage,
+    web::{page::WebPage, MetaData},
+};
+use iron::{IronResult, Request, Response};
+use router::Router;
+use serde::Serialize;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+struct FeaturesPage {
+    metadata: MetaData,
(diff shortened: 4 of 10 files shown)


pietroalbini:
I'm not sure showing the additional dependencies is much useful, we could simplify the layout like this (showing the features on two columns on desktop):

photo_2020-10-30_11-41-39

almusil:
> I'm not sure showing the additional dependencies is much useful, we could simplify the layout like this (showing the features on > two columns on desktop):

From my point of view it's useful to know which feature gets enabled by which flag. Especially for default.

jyn514:
I also prefer having the features that are enabled shown.

Kixiron:
I think we should show full detail, but on a separate page (placed alongside `/builds`) with the top bar linking to it

Kixiron:
Made a very rough and out of proportion sketch of what I'm thinking, display the available features on the left, clicking on each "maximizes" it to show all the features and dependencies it activates. This allows us to give as much info as possible (ex. saying if the feature is actually a feature or just an optional dependency) without immediately overwhelming the user. 

image

almusil:
Ok, I have been able to come up with this (I'm really bad at frontend, so please excuse some rough edges):
Screenshot_2020-11-05 regex 1 3 1 - Docs rs
Screenshot_2020-11-05 regex 1 3 1 - Docs rs(1)
Screenshot_2020-11-05 regex 1 3 1 - Docs rs(2)

jyn514:
I like it! It could be nice to split it into columns like @pietroalbini  suggested, but I don't think we need to block on that.

templates/crate/features.html
```diff
@@ -0,0 +1,70 @@
...
+                        <h3 id="{{ feature.name }}">{{ feature.name }}</h3>
+                        <ul class="pure-menu-list">
+                            {%- if feature.subfeatures -%}
+                                {%- for subfeature in feature.subfeatures -%}
+                                    <li class="pure-menu-item">
+                                        <span>{{ subfeature }}</span>
+                                    </li>
+                                {%- endfor -%}
+                            {%- else -%}
+                                <p>This feature flag do not enable additional features.</p>
```
jyn514, line 60:
```suggestion
                                <p>This feature flag does not enable additional features.</p>
```

src/db/types.rs
```diff
@@ -12,4 +12,16 @@ impl Feature {
...
+        self.name.starts_with('_')
+    }
+
+    pub fn is_default(&self) -> bool {
+        self.name == "default"
+    }
+
+    pub fn subfeature_len(&self) -> usize {
+        self.subfeatures.len()
+    }
```
jyn514, line 26: Is this function necessary? It seems like it would be more clear inlined.

src/web/crate_details.rs
```diff
@@ -741,4 +744,42 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    fn feature_flags_is_hidden_when_empty() {
```
jyn514, line 749:
```suggestion
    fn feature_flags_are_hidden_when_empty() {
```
or just
```suggestion
    fn feature_flags_hidden_when_empty() {
```

templates/crate/features.html
```diff
@@ -0,0 +1,70 @@
...
+                            {%- for feature in features -%}
+                                <li class="pure-menu-item">
+                                    <a href="#{{ feature.name }}" class="pure-menu-link" style="text-align:center;">
+                                        {{ feature.name }}
+                                    </a>
+                                </li>
+                            {%- endfor -%}
+                        {%- else -%}
+                            <li class="pure-menu-item">
+                                <span style="font-size: 13px;">Feature flags are not available for this version.</span>
```
jyn514, line 39: It's possible they may not be available for any version.
```suggestion
                                <span style="font-size: 13px;">Feature flags are not available for this release.</span>
```

templates/rustdoc/topbar.html
```diff
@@ -226,7 +226,27 @@
...
+
+            {# Build the dropdown list showing available features #}
+            <ul class="pure-menu-children">
+                {%- for feature in metadata.features -%}
+                {%- set target_url = "/crate/" ~ metadata.name ~ "/" ~ metadata.version ~ "/features#" ~ feature.name -%}
+                    <li class="pure-menu-item">
+                        <a href="{{ target_url | safe }}" class="pure-menu-link">{{ feature.name }}</a>
+                    </li>
+                {%- endfor -%}
+            </ul>
```
jyn514, line 247: This gets a little overwhelming with a lot of features. Can you have just a single link to `/crate/.../features` instead, the way you have it on /crate?

image

templates/crate/features.html
```diff
@@ -0,0 +1,70 @@
...
+                            </li>
+                        {%- endif -%}
+                    </ul>
+                </div>
+            </div>
+
+            <div class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24 package-details" id="main">
+                <h1>{{ metadata.name }}</h1>
+                {%- if features -%}
+                    <p>This version has <b>{{ feature_len }}</b> feature flags, <b>{{ default_feature_len }}</b> of them being enabled by <b>default</b>.</p>
```
jyn514, line 49:
```suggestion
                    <p>This version has <b>{{ features | length }}</b> feature flags, <b>{{ features[0] | length }}</b> of them being enabled by <b>default</b>.</p>
```
Then you can remove `feature_len` and `default_feature_len` from the template data. You may have to handle the case where `features | length` is 0 specially to avoid out of bounds errors.

src/web/crate_details.rs
```diff
@@ -741,4 +744,42 @@ mod tests {
...
+        wrapper(|env| {
+            env.fake_release()
+                .name("binary")
+                .version("0.1.0")
+                .binary(true)
```
jyn514, line 754: Rather than use this hack, could you change `Fakes::features` to take an `Option`?

src/web/crate_details.rs
```diff
@@ -741,4 +744,42 @@ mod tests {
...
+        wrapper(|env| {
+            let features = [("_private".into(), Vec::new())]
+                .iter()
+                .cloned()
+                .collect::<HashMap<String, Vec<String>>>();
+            env.fake_release()
+                .name("binary")
+                .version("0.1.0")
+                .binary(true)
+                .features(features)
```
jyn514, line 776: Because you hid the features with `binary(true)`, this doesn't test anything. Please change this to a library instead.

src/web/features.rs
```diff
@@ -0,0 +1,63 @@
...
+             FROM builds
+             INNER JOIN releases ON releases.id = builds.rid
+             INNER JOIN crates ON releases.crate_id = crates.id
+             WHERE crates.name = $1 AND releases.version = $2
+             ORDER BY releases.id DESC",
+            &[&name, &version]
+        )
+    );
+    let row = cexpect!(req, query.get(0));
+    let features = MetaData::parse_features(row.get("features"));
```
jyn514, line 44: This is already a field on `MetaData`, no need to parse and `SELECT` it twice. You can change everything using this to use `metadata.features` instead, then remove it from the SQL query.

Alternatively, since the /features page is the only one using this data, you can remove it from `MetaData` altogether and only have it for `/features`. But it definitely shouldn't be in both.

jyn514 (commented):
Thanks for sticking with this :)

src/db/types.rs
almusil: To be honest I didn't know that you can do something like that, thanks

src/web/crate_details.rs
almusil: Done

templates/crate/features.html
almusil: Done

templates/rustdoc/topbar.html
almusil: Done (Hopefully I understood it well)

src/web/crate_details.rs
almusil: Done

src/web/features.rs
almusil: Good idea, thanks

templates/crate/features.html
almusil: Done

src/web/crate_details.rs
almusil: Done

jyn514:
@almusil it's not a big deal, but in the future it would be nice to add changes as additional commits instead of force-pushing. The diff github shows also includes the changes from master. Fortunately here there weren't many changes, but in a big repo like rust-lang/rust it would be unreadable.

src/web/crate_details.rs
```diff
@@ -741,4 +744,79 @@ mod tests {
...
         });
     }
+
+    #[test]
+    fn feature_flags_report_empty() {
+        wrapper(|env| {
+            env.fake_release()
+                .name("library")
+                .version("0.1.0")
+                .binary(false)
```
jyn514, line 754: This is the default.
```suggestion
```

almusil:
> @almusil it's not a big deal, but in the future it would be nice to add changes as additional commits instead of force-pushing. The diff github shows also includes the changes from master. Fortunately here there weren't many changes, but in a big repo like rust-lang/rust it would be unreadable.
> 
> (FWIW I consider this a bug in github [isaacs/github#1834](https://github.com/isaacs/github/issues/1834))

Ah right, the ugly thing about that is that every repo has different policy about this. I'll keep it in mind for future changes in this repo, thanks :)

src/test/fakes.rs
```diff
@@ -210,6 +210,15 @@ impl<'a> FakeRelease<'a> {
         self
     }
 
+    pub(crate) fn features(mut self, opt_features: Option<HashMap<String, Vec<String>>>) -> Self {
+        if let Some(features) = opt_features {
+            self.package.features = features;
+        } else {
+            self.package.features = HashMap::new();
+        }
```
jyn514, line 218: Wait, I'm confused - if the `features` field on package isn't optional, then how does this render pages for old docs? Does it confuse 'no features' with 'no data'? `[features]` with no entries is also valid, please add a test for that.

I ask because this definitely looks wrong - now there's two ways to say `HashMap::new()`, one by passing it in and one by passing in `None`.

templates/crate/features.html
```diff
@@ -0,0 +1,76 @@
...
+            <div class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24 package-details" id="main">
+                <h1>{{ metadata.name }}</h1>
+                {%- if metadata.features -%}
+                    <p>This version has <b>{{ metadata.features | length }}</b> feature flags, <b data-id="default-feature-len">
+                        {%- if metadata.features[0].name == 'default' -%}
+                            {{ metadata.features[0].subfeatures | length }}
+                        {%- else -%}
+                            0
+                        {%- endif -%}
+                    </b> of them being enabled by <b>default</b>.</p>
```
jyn514, line 55:
```suggestion
                    </b> of them enabled by <b>default</b>.</p>
```

templates/header/package_navigation.html
```diff
@@ -85,6 +86,15 @@ <h1 id="crate-title">
                                 <span class="title"> Builds</span>
                             </a>
                         </li>
+
+                        {# The features tab #}
+                        <li class="pure-menu-item">
+                            <a href="/crate/{{ crate_path | safe }}/features"
```
jyn514, line 92: `| safe` is the default, you don't need to restate it. https://tera.netlify.app/docs/#auto-escaping
```suggestion
                            <a href="/crate/{{ crate_path }}/features"
```

templates/rustdoc/topbar.html
```diff
@@ -229,6 +229,14 @@
                 </li>
             {%- endfor -%}
         </ul>
+    </li>{#
+    Display the features available in current build
+    #}
+    <li class="pure-menu-item">
+        <a href="{{ crate_url | safe }}/features" title="Browse available feature flags of {{ metadata.name }}-{{ metadata.version }}" class="pure-menu-link">
```
jyn514, line 236:
```suggestion
        <a href="{{ crate_url }}/features" title="Browse available feature flags of {{ metadata.name }}-{{ metadata.version }}" class="pure-menu-link">
```

src/test/fakes.rs
almusil: They are not optional, package will always have features that are passed from `cargo metadata`: `"features": []`. The option that I have enabled on the last version is that it will show it even for old docs. And feature flags page without features will state it. That seems to be desirable IMO.

templates/header/package_navigation.html
Nemo157: `safe` declares the string to have already been escaped and _disables_ the auto-escaping.

src/test/fakes.rs
jyn514: @almusil what about crates built before this feature was added? Won't it be `NULL` in the database?

templates/header/package_navigation.html
jyn514: Hmm, ok - @almusil why did you disable the escaping?

almusil: I'll be honest, it was copied from the builds link.

src/test/fakes.rs
almusil: @jyn514 It will but the frontend part of this is optional. It is kinda confusing. Backend metadata are parsed as non-optional for the reason I have mentioned earlier. Frontend is optional so if the database part is NULL or simply empty it should result in the same thing AFAIK. Maybe I have wrong understanding of the templates and their evaluation of variables.

templates/header/package_navigation.html
jyn514: Ok, it was probably switched in https://github.com/rust-lang/docs.rs/pull/906 then. No need to change it.

src/test/fakes.rs
jyn514: Something weird is going on here, the tests don't match the live version. Notice here the features are optional: https://github.com/rust-lang/docs.rs/blob/2df55efcfde3541e83d0ebe2647a60cb6b266b5c/src/web/mod.rs#L525
but yet you're setting it here to a `HashMap` directly. I don't know why that's possible, but it's not right - we should be testing `None`, not just an empty HashMap.

> Maybe I have wrong understanding of the templates and their evaluation of variables.

It's not _wrong_ but you're not distinguishing between the two I think. `{% if metadata.features %}` is false *either* if the array is empty or if it doesn't exist. It would be nice to say 'no features' if it's empty and 'data not available' if it doesn't exist.

almusil: That's the thing I don't know how to fake the NULL without changing the backend. The change in backend to have option would require some filtering if the features are just empty list from `cargo metadata`.

jyn514: Got it, that makes sense. In that case you can use a direct SQL query - create the package like normal, then `UPDATE releases SET features = NULL WHERE release_id = ?`.

almusil: Ah right, I will add the test and update the frontend to match what you have described NULL vs empty. Thanks

almusil: Should be done

templates/crate/features.html
almusil: Done

src/web/crate_details.rs
almusil: Done

src/web/crate_details.rs
```diff
@@ -819,4 +816,29 @@ mod tests {
...
     }
+
+    #[test]
+    fn feature_flags_report_null() {
+        wrapper(|env| {
+            let id = env
+                .fake_release()
+                .name("library")
+                .version("0.1.0")
+                .features(HashMap::new())
```
jyn514, line 827: No need to add this call, since you delete the features later.
```suggestion
```

jyn514 (approved):
This already looks really good and I'm happy to merge it as is :) Some improvements I'd like to see at some point:

- Right now the features look like they're unordered: 
image
Could we change it to put features that enable other features first? Or we could get really fancy and do a tree structure, where we put `default -- [std, perf, unicode]`, then std, then perf, then unicode, then, recursively, features that are enabled by std/perf/unicode.
- We don't distinguish between optional dependencies and features. Could we add that?
- @Nemo157 I think you mentioned a while back you wouldn't want this for your crates - do you still want to add an opt-out to `package.metadata.docs.rs`?
- Could we add some color? Maybe green for `default`, red for features that enable other features, black otherwise, something like that?

Nemo157:
> * 
> 
> * @Nemo157 I think you mentioned a while back you wouldn't want this for your crates - do you still want to add an opt-out to `package.metadata.docs.rs`?

I would, but I wouldn't want to block this on that being added.

I did run across a bug in the feature recording while looking at it though, `async-compression 0.3.6` has a `tokio-02` and a `tokio-03` optional dependency, but these both get recorded and displayed as `tokio`.

templates/crate/features.html
```diff
@@ -0,0 +1,82 @@
...
+                    </ul>
+                </div>
+            </div>
+
+            <div class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24 package-details" id="main">
+                <h1>{{ metadata.name }}</h1>
+                {%- if metadata.features -%}
+                    <p>This version has <b>{{ metadata.features | length }}</b> feature flags, <b data-id="default-feature-len">
+                        {%- if metadata.features[0].name == 'default' -%}
+                            {{ metadata.features[0].subfeatures | length }}
```
Nemo157, line 55: This count is wrong when one of the default features activates more features. In `bs58 0.4.0` I have `default = ["std"]` and `std = ["alloc"]`.

Also I think the previous count should be decreased by 1 (if the crate has a `default` feature), otherwise a crate which activates all its features by default will still say something like "This version has 5 feature flags, 4 of them enabled by default" because `default` is not counted.

almusil:
@jyn514 @Nemo157 What should be addressed by this PR? The default counting to follow the dependencies?

jyn514:
@almusil if you can fix the bugs Nemo noticed and address my nit I think this is good to merge. The rest can be fixed in follow-ups.

templates/crate/features.html
almusil: > async-compression 0.3.6

I was trying it locally and I can see both `tokio-02` and `tokio-03`.

Nemo157: Since the renamed dependency issue is unrelated to this UI I opened https://github.com/rust-lang/docs.rs/issues/1175 for it.

almusil:
@jyn514 Hopefully got it right this time, thanks for all the patience

src/web/features.rs
```diff
@@ -0,0 +1,88 @@
...
+    let mut default_len = 0;
+    let features = row
+        .get::<'_, usize, Option<Vec<Feature>>>(0)
+        .map(|raw| {
+            raw.into_iter()
+                .filter(|feature| !feature.is_private())
+                .map(|feature| (feature.name.clone(), feature))
+                .collect::<HashMap<String, Feature>>()
+        })
+        .map(|mut feature_map| {
```
jyn514, line 50:
```suggestion
            let feature_map: HashMap<_> = raw.into_iter()
                .filter(|feature| !feature.is_private())
                .map(|feature| (feature.name.clone(), feature))
                .collect();
```

src/web/features.rs
```diff
@@ -22,9 +26,63 @@ pub fn build_features_handler(req: &mut Request) -> IronResult<Response> {
...
+            let mut features = get_tree_structure_from_default(&mut feature_map);
+            let mut remaining = feature_map
+                .into_iter()
+                .map(|(_, feature)| feature)
+                .collect::<Vec<Feature>>();
+            remaining.sort_by_key(|feature| feature.subfeatures.len());
```
jyn514, line 56: Can you add some tests for this behavior? Ideally unit tests, but if web tests are easier to write, those are fine too.

src/web/features.rs
almusil: Done

src/web/features.rs
almusil: Sure, I have added a few unit tests

src/web/features.rs
```diff
@@ -38,28 +40,14 @@ pub fn build_features_handler(req: &mut Request) -> IronResult<Response> {
...
-            default_len = features.len();
-
-            features.extend(remaining.into_iter().rev());
-            features
-        });
+
+    if let Some(raw) = row.get(0) {
+        let result = order_features_and_count_default_len(raw);
+        features = Some(result.0);
+        default_len = result.1;
```
jyn514, line 49: https://github.com/rust-lang/rust/pull/71156 would help with this :) 'coming soon to rust versions near you'

src/web/features.rs
```diff
@@ -0,0 +1,220 @@
...
+            feature1.clone(),
+        ];
+        let mut feature_map = get_feature_map(raw);
+        let default_tree = get_tree_structure_from_default(&mut feature_map);
+
+        assert_eq!(feature_map.len(), 1);
+        assert_eq!(default_tree.len(), 4);
+        assert!(feature_map.contains_key(&non_default.name));
+        assert!(!feature_map.contains_key(&default.name));
+        assert_eq!(default_tree.get(0), Some(&default));
```
jyn514, line 146: `get` just adds clutter here IMO.
```suggestion
        assert_eq!(default_tree[0], default);
```
and the same for the rest.

src/web/features.rs
almusil: That will be nice :)

src/web/features.rs
almusil: Done

jyn514:
I think further improvements can be done in follow-ups. Thanks so much for working on this, I've wanted this in docs.rs for a while :heart:

almusil:
Happy to help :) Thank you for your patience. If you don't mind I would like to continue with improvements to this feature.

jyn514:
Sounds great! If you want to talk about ideas we have a [discord channel](https://discord.gg/f7mTXPW) for docs.rs

PR #1145, opened 2020-10-30
Add a way to not build tier one targets by default

jyn514:
This is still opt-in and off by default. However, it's useful for local
development and custom registries.

This is all of https://github.com/rust-lang/docs.rs/pull/532 except for actually switching it on. It does turn it on for local development, so that builds will be faster.

r? @pietroalbini 
cc @l4l

### DIFF
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -20,7 +20,7 @@
 //! let metadata = Metadata::from_crate_root(&source_root)?;
 //!
 //! // Next, learn what arguments we need to pass to `cargo`.
-//! let targets = metadata.targets();
+//! let targets = metadata.targets(/* include_default_targets: */ true);
 //! let mut cargo_args = metadata.cargo_args(&[], &[]);
 //! cargo_args.push(targets.default_target.into());
 //!
@@ -190,7 +190,10 @@ impl Metadata {
     /// Return the targets that should be built.
     ///
     /// The `default_target` will never be one of the `other_targets`.
-    pub fn targets(&self) -> BuildTargets<'_> {
+    /// If `include_default_targets` is `true` and `targets` is unset, this also includes
+    /// [`DEFAULT_TARGETS`]. Otherwise, if `include_default_targets` is `false` and `targets`
+    /// is unset, `other_targets` will be empty.
+    pub fn targets(&self, include_default_targets: bool) -> BuildTargets<'_> {
         let default_target = self
             .default_target
             .as_deref()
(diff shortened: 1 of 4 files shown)


crates/metadata/lib.rs
```diff
@@ -202,12 +205,16 @@ impl Metadata {
             })
             .unwrap_or(HOST_TARGET);
 
-        // Let people opt-in to only having specific targets
-        let mut targets: HashSet<_> = self
+        let user_targets = self
```
l4l, line 208: Looks quite ambiguous, I first thought that these are coming from cli or similar input, but it seems that it comes from crate's cargo.toml, right? Then might be call it `crate_targets`?

l4l (commented):
Maybe it's better to specify `DEFAULT_TARGETS` instead? Then it's possible to specify a list of targets rather than limiting to the host one

crates/metadata/lib.rs
jyn514: Yeah, everything in this file is about Cargo.toml. I'll change the variable name.

jyn514:
> Maybe it's better to specify DEFAULT_TARGETS instead? Then it's possible to specify a list of targets rather than limiting to the host one

Do you need this flexibility? We certainly don't in docs.rs and it would make the implementation more complicated.

l4l:
Well, that sounds like a fairly natural need. I'm not sure that I personally need this but I think it can be helpful.

pietroalbini:
I would invert the variable, something like `include_default_targets: true`.

jyn514:
Ok, I updated it in the metadata library but not in docs.rs, since I'd like 'no variable set' to match `false`.

src/config.rs
```diff
@@ -87,6 +87,7 @@ impl Config {
             local_docker_image: maybe_env("DOCS_RS_LOCAL_DOCKER_IMAGE")?,
             toolchain: env("CRATESFYI_TOOLCHAIN", "nightly".to_string())?,
             build_cpu_limit: maybe_env("DOCS_RS_BUILD_CPU_LIMIT")?,
+            only_build_default_target: env("DOCSRS_ONLY_BUILD_DEFAULT_TARGET", false)?,
```
Nemo157, line 90:
```suggestion
            include_default_targets: env("DOCSRS_INCLUDE_DEFAULT_TARGETS", true)?,
```

And an example of `DOCSRS_INCLUDE_DEFAULT_TARGETS=false` in `.env.sample` seems like it'd support defaulting to on without the double-negative of `!only_build_default_target`.

jyn514: @Nemo157 well this is what I mean by I want the env variable to default to `false` - otherwise it's confusing what the default is. If we change it to `INCLUDE_DEFAULT_TARGETS` the default will be true.

Nemo157: 🤷 I don't find it that confusing when things default to on and you have to explicitly disable them in config, but I don't really care that much.

jyn514: Ok, I changed this to `INCLUDE_DEFAULT_TARGETS`, set it to be on by default in docs.rs, but off by default in env.sample.

Nemo157 (approved):

PR #1146, opened 2020-10-30
Remove rustdoc CSS files import from non-rustdoc pages

GuillaumeGomez:
Normally, if everything was done properly, no UI changes should appear.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -12,8 +12,6 @@
         {# Docs.rs styles #}
         <link rel="stylesheet" href="/-/static/vendored.css?{{ docsrs_version() | slugify }}" type="text/css" media="all" />
         <link rel="stylesheet" href="/normalize-{{ rustc_resource_suffix() }}.css" type="text/css" media="all" />
-        <link rel="stylesheet" href="/rustdoc-{{ rustc_resource_suffix() }}.css" type="text/css" media="all" />
-        <link rel="stylesheet" href="/light-{{ rustc_resource_suffix() }}.css" type="text/css" media="all" />
         <link rel="stylesheet" href="/-/static/style.css?{{ docsrs_version() | slugify }}" type="text/css" media="all" />
 
         {%- block css -%}{%- endblock css -%}
--- a/templates/base.html
+++ b/templates/base.html
@@ -11,7 +11,6 @@
 
         {# Docs.rs styles #}
         <link rel="stylesheet" href="/-/static/vendored.css?{{ docsrs_version() | slugify }}" type="text/css" media="all" />
-        <link rel="stylesheet" href="/normalize-{{ rustc_resource_suffix() }}.css" type="text/css" media="all" />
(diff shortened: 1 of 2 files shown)


templates/base.html
```diff
@@ -12,8 +12,6 @@
         {# Docs.rs styles #}
         <link rel="stylesheet" href="/-/static/vendored.css?{{ docsrs_version() | slugify }}" type="text/css" media="all" />
         <link rel="stylesheet" href="/normalize-{{ rustc_resource_suffix() }}.css" type="text/css" media="all" />
```
jyn514, line 14: Can you try to remove this file too?

templates/style/base.scss
```diff
@@ -25,12 +75,68 @@ body {
...
+
+    h1 {
+        font-size: 1.5em;
+    }
+    h1, h2, h3, h4 {
+        font-family: "Fira Sans",sans-serif;
+    }
+    h1, h2,
+    h3:not(.impl):not(.method):not(.type):not(.tymethod):not(.notable),
+    h4:not(.method):not(.type):not(.tymethod):not(.associatedconstant) {
```
Nemo157, line 92: We shouldn't need all these special cases since we don't use these classes, do we? (same below)

templates/base.html
jyn514: I'm pretty sure we already vendor it, there have been a lot of bugs related to it.
https://github.com/rust-lang/docs.rs/issues/992
https://github.com/rust-lang/docs.rs/issues/1005
https://github.com/rust-lang/docs.rs/pull/1059

templates/style/base.scss
GuillaumeGomez: Indeed! Forgot to remove all the rules...

templates/base.html
GuillaumeGomez: Yes!

GuillaumeGomez:
Updated! I also removed `normalize.css` and couldn't spot any difference...

templates/style/base.scss
```diff
@@ -1,6 +1,56 @@
 // FIXME: Use modules
 @import "vars", "utils", "navbar", "themes", "fa";
 
+/* See FiraSans-LICENSE.txt for the Fira Sans license. */
+@font-face {
+    font-family: 'Fira Sans';
+    font-style: normal;
+    font-weight: 400;
+    src: local('Fira Sans'), url("FiraSans-Regular.woff") format('woff');
```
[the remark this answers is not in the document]
GuillaumeGomez: I think it's the most stable part of rustdoc. They were here before I came in and they haven't changed since then. So moving them in a "static" location seems safe enough.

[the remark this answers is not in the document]
jyn514: @GuillaumeGomez I think pietro means that the files won't appear at all unless you run a build first. It would be nice to be able to use Fira Sans without having to first run a rustdoc build.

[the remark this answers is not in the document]
GuillaumeGomez: Oh I see! Sending a PR shortly then.

PR #1157, opened 2020-11-04
Add fonts to static folder

GuillaumeGomez:
Following discussion on #1146.

cc @pietroalbini @jyn514

PR #1158, opened 2020-11-04
Don't serve non-rustdoc resources from the shared resource handler

Nemo157:


Nemo157:
Actually, we're relying on this behaviour right now for the font files, @GuillaumeGomez can you roll this commit in with #1157?

PR #1159, opened 2020-11-04
Fix tabbing through the giant crate menu

notriddle:
Without this tabindex attribute, the scroll view itself gets selected.

With this attribute added, focus jumps straight to the menu item within, which is what we want, since it contains nothing but menu items, and most of the keyboard shortcuts you would use to interact with the scroll view are trapped by JavaScript anyway.

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -187,7 +187,7 @@
                 case "pageup":
                     // page up: jump five items up, stopping at the top
                     // the number 5 is used so that we go one page in the
-                    // inner-scrolled Depedencies and Versions fields
+                    // inner-scrolled Dependencies and Versions fields
                     switchTo = currentLink;
                     for (var n = 0; n < 5 && switchTo; ++n) {
                         switchTo = previous(allItems, switchTo);
--- a/static/menu.js
+++ b/static/menu.js
@@ -188,24 +188,22 @@
                     // page up: jump five items up, stopping at the top
                     // the number 5 is used so that we go one page in the
                     // inner-scrolled Dependencies and Versions fields
-                    switchTo = currentLink;
-                    for (var n = 0; n < 5 && switchTo; ++n) {
-                        switchTo = previous(allItems, switchTo);
-                    }
-                    if (!switchTo) {
-                        switchTo = allItems[0];
+                    switchTo = currentItem || allItems[0];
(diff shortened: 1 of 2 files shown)


jyn514:
Can you describe the problem this is solving? I found that if I tab through to 'Dependencies' it will highlight the whole menu before going to the first dependency, but this PR doesn't seem to fix that.

image

notriddle:
> I found that if I tab through to 'Dependencies' it will highlight the whole menu before going to the first dependency, but this PR doesn't seem to fix that

Without the PR, when I try it, it will highlight the whole dependencies list. With the PR, it will immediately highlight the first dependency, without highlighting the whole dependencies list first.

I'm using Firefox. What browser are you using?

static/menu.js
```diff
@@ -171,21 +171,43 @@
...
                     // end: focus last menu item.
                     // This is the behavior of WAI, while GitHub scrolls,
                     // but it's unlikely that a user will try to scroll the page while the menu is open,
                     // so they won't do it on accident.
                     switchTo = last(allItems);
                     break;
+                case "pageup":
+                    // page up: jump five items up, stopping at the top
+                    // the number 5 is used so that we go one page in the
+                    // inner-scrolled Depedencies and Versions fields
```
jyn514, line 190:
```suggestion
                    // inner-scrolled Dependencies and Versions fields
```

static/menu.js
```diff
@@ -171,21 +171,43 @@
...
+                    switchTo = currentLink;
+                    for (var n = 0; n < 5 && switchTo; ++n) {
+                        switchTo = previous(allItems, switchTo);
+                    }
```
jyn514, line 194: This causes page-up to go _between_ the lists, which is not exactly the behavior I'd expect. e.g. if you have the first dependency highlighted it will behave as if you hit shift+tab 5 times, taking you to 'Github Stars'. I'd expect it to go to the top of the dependency list and stop there, requiring you to hit shift+tab if you want to change focus.

jyn514 (commented):
Huh, it's working now ... maybe I was on the wrong branch.

static/menu.js
notriddle: Sure, it can be fixed to obey menu sections.

jyn514:
Thanks for the fix!

PR #1160, opened 2020-11-05
Remove whitespace beside docs.rs text

pickfire:
image

See whitespace beside the docs.rs text which is focused. Next thing I want to fix is the weird focus which is set the title background to `#eee`, it still looks fine on light theme but works bad on dark theme.


pickfire:
Looks like the issue still isn't fixed. This removed whitespace but on some other parts.

PR #1162, opened 2020-11-05
Remove focus background color on pure-css

pickfire:
On `focus`

Before

image

After

image

This works well only on light theme, but bad on dark or ayu theme. Maybe it would be good to leave a comment that we removed this.

Removed code
```css
.pure-menu-active>.pure-menu-link,.pure-menu-link:hover,.pure-menu-link:focus{background-color:#eee}
```


Nemo157:
Rather than changing the vendored css, can we just add an overriding rule in `_navbar.scss`? Our style is loaded after the vendored styles so should take precedence. (It's a little bit of a waste of bytes, but it'd be nice to not have to remember about this if we ever update the vendored styles).

pickfire:
But still, even if we add it to our side, if vendored css happened to change it, the line will be left there forever until someone removed it.

Nemo157:
Actually, do we not want to override this with theme specific colors? Without it the only focus indication during keyboard nav of the menu is the outline, having the background color change as well is quite nice.

pickfire:
Ah, but just the foreground text is unreadable, maybe I should add that?

Nemo157:
I was thinking something like

```css
.pure-menu-active>.pure-menu-link,.pure-menu-link:hover,.pure-menu-link:focus{
  background-color:var(--color-menu-header-background);
}
```

Which is a decent color in the light theme, and not overridden in the other themes since afaict it's not actually displayed anywhere yet :grin:; but we could choose a similar "slightly lighter/darker" color for that variable.

jyn514:
@pickfire in the future, if you end up picking a different approach to fix the same issue, do you mind force-pushing to this PR instead of opening a new one? That makes it easier for me to track what's going on since all the discussion is in one place.

pickfire:
@jyn514 I know I could force push but it is way faster if I can just work it directly (lazy). I didn't even need to open rust project and change stuff for this. You will know if the branch name is `patch-N`, I did it in github. But at least I will leave a note to say it superseeds the other pull request.

Issue #1172, opened 2020-11-09
docs.rs downloads the build image before starting the web server


jyn514:
This is a recent regression; I haven't bisected the change yet. It caused this mornings ~5 minute downtime.

```
Nov 09 16:52:37 docsrs sudo[21231]:   ubuntu : TTY=pts/0 ; PWD=/home/ubuntu/docs.rs ; USER=root ; COMMAND=/bin/systemctl restart docs.rs
Nov 09 16:52:37 docsrs cratesfyi[21281]: 2020/11/09 16:52:37 [INFO] rustwide::cmd::sandbox: pulling image rustops/crates-build-env from Docker Hub
Nov 09 16:52:38 docsrs cratesfyi[21281]: 2020/11/09 16:52:38 [INFO] rustwide::cmd: [stdout] 25fa93f27f05: Pulling fs layer
Nov 09 16:52:38 docsrs cratesfyi[21281]: 2020/11/09 16:52:38 [INFO] rustwide::cmd: [stdout] 25fa93f27f05: Waiting
Nov 09 16:54:34 docsrs cratesfyi[22776]: 2020/11/09 16:54:34 [INFO] rustwide::cmd: [stdout] 25fa93f27f05: Download complete
Nov 09 16:56:09 docsrs cratesfyi[22776]: 2020/11/09 16:56:09 [INFO] rustwide::cmd: [stdout] Status: Downloaded newer image for rustops/crates-build-env:latest
Nov 09 16:56:12 docsrs cratesfyi[22776]: 2020/11/09 16:56:12 [INFO] docs_rs::web: Running docs.rs web server on http://0.0.0.0:3000
```

jyn514:
This regressed in 1aa36cc2010d68903ec38726b3a2a3c7e073eea8; cc @pietroalbini

jyn514:
I think this is the issue: it blocks on `init` in the current thread instead of running it in the build queue reader.

https://github.com/rust-lang/docs.rs/blob/1aa36cc2010d68903ec38726b3a2a3c7e073eea8/src/utils/daemon.rs#L65

PR #1173, opened 2020-11-09
Start the webserver before trying to download the crates-build-env image

jyn514:
I tested locally that this fixes https://github.com/rust-lang/docs.rs/issues/1172.

r? @Nemo157

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -116,7 +116,8 @@ impl CommandLine {
                 socket_addr,
                 reload_templates,
             } => {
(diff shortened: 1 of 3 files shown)


jyn514:
I realized - a side effect of this is that errors in `init` and the build reader won't ever be reported, because it will block on dropping the server first. I'm not sure what to do about that - maybe switch all the error reporting to logging instead?

Nemo157:
Hmmm, this means that if any of the other components fail to start and attempt to return an error, the web server blocking on drop will just block that error and leave the service running.

Nemo157 (approved):

Issue #1175, opened 2020-11-11
Renamed optional dependencies get the wrong feature name recorded


Nemo157:
`async-compression 0.3.6` has a `tokio-02` and a `tokio-03` optional dependency, but these both get recorded and displayed as `tokio`. Dumping out of the database I see:

```
{
  ...
  "(all-implementations,\"{futures-io,stream,tokio-02,tokio-03}\")",
  ...
  "(tokio,{})",
  "(tokio,{})",
  ...
}
```

So `all-implementations` depends on `tokio-02` and `tokio-03`, but the optional dependencies themselves show `tokio`.

https://github.com/rust-lang/docs.rs/blob/b527df80a7332b35294c02b2922a2899dee43aaa/src/db/add_package.rs#L240

Here it needs to take the renamed name (but unfortunately our little cargo-metadata dataset doesn't record this currently).

almusil:
Since we need the rename, it would be also wise to extend the Feature in database with optional dependency indicator. It will be easier that way to filter etc IMO. WDYT?

jyn514:
Sure, if you want to do both at once I'd be fine with that.

jyn514:
Fixed by https://github.com/rust-lang/docs.rs/pull/1180 I think.

Nemo157:
Yep, https://docs.rs/crate/async-compression/0.3.7/features shows `tokio-02` and `tokio-03` headings.

PR #1180, opened 2020-11-16
features: Distinguish between optional dependencies and features

almusil:
Hi,

as discussed in #1175, features that are taken from optional dependencies are marked in database and
stored with rename, if provided.

### DIFF
--- a/src/db/add_package.rs
+++ b/src/db/add_package.rs
@@ -221,13 +221,13 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)> 
 fn get_features(pkg: &MetadataPackage) -> Vec<Feature> {
     let mut features = Vec::with_capacity(pkg.features.len());
     if let Some(subfeatures) = pkg.features.get("default") {
-        features.push(Feature::new("default".into(), subfeatures.clone()));
+        features.push(Feature::new("default".into(), subfeatures.clone(), false));
     };
     features.extend(
         pkg.features
             .iter()
             .filter(|(name, _)| *name != "default")
-            .map(|(name, subfeatures)| Feature::new(name.clone(), subfeatures.clone())),
+            .map(|(name, subfeatures)| Feature::new(name.clone(), subfeatures.clone(), false)),
     );
     features.extend(get_optional_dependencies(pkg));
     features
@@ -237,7 +237,14 @@ fn get_optional_dependencies(pkg: &MetadataPackage) -> Vec<Feature> {
     pkg.dependencies
         .iter()
         .filter(|dep| dep.optional)
-        .map(|dep| Feature::new(dep.name.clone(), Vec::new()))
+        .map(|dep| {
+            let name = if let Some(rename) = &dep.rename {
+                rename.clone()
+            } else {
+                dep.name.clone()
+            };
+            Feature::new(name, Vec::new(), true)
+        })
         .collect()
 }
 
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -488,6 +488,20 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
             "
                 ALTER TABLE queue DROP COLUMN registry;
             "
+        ),
+        migration!(
+            context,
+            21,
+            // description
+            "Add mark for features that are derived from optional dependencies",
+            // upgrade query
+            "
+                ALTER TYPE feature ADD ATTRIBUTE optional_dependency BOOL;
+            ",
+            // downgrade query
+            "
+                 ALTER TYPE feature DROP ATTRIBUTE optional_dependency;
+            "
         )
     ];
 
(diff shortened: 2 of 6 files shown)


almusil:
@Nemo157 and @jyn514 
Frontend now shows correct features after this patch. 
async-compression 0.3.6

src/db/add_package.rs
```diff
@@ -237,7 +237,16 @@ fn get_optional_dependencies(pkg: &MetadataPackage) -> Vec<Feature> {
...
         .filter(|dep| dep.optional)
-        .map(|dep| Feature::new(dep.name.clone(), Vec::new()))
+        .map(|dep| {
+            let name = if let Some(rename) = &dep.rename {
+                rename.clone()
+            } else {
+                dep.name.clone()
+            };
+            let mut feature = Feature::new(name, Vec::new());
+            feature.set_optional_dependency(true);
```
jyn514, line 247: Rather than having this as a modifier, can you pass in an `optional_dependency` argument? That way it's not possible to forget to deal with it.

src/db/migrate.rs
```diff
@@ -488,6 +488,20 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
             "
+        ),
+        migration!(
+            context,
+            21,
+            // description
+            "Add mark for features that are derived from optional dependencies",
+            // upgrade query
+            "
+                ALTER TYPE feature ADD ATTRIBUTE optional_dependency BOOL;
```
jyn514, line 499: Where can I learn more about postgres data types? Will `optional_dependency` be added as NULL? It should be.

Nemo157: Yes, which is an issue when loading the features page for existing crates:

```
thread '<unnamed>' panicked at 'error retrieving column 0: error deserializing column 0: a Postgres value was `NULL`'
```

Probably the type needs to be changed to `optional_dependency: Option<bool>`.

almusil: Yes, I have realized that later. IMO it would make sense to make is false as default. WDYT?

src/db/add_package.rs
almusil: Done

src/db/migrate.rs
jyn514: I think we shouldn't say something we don't know for sure. We should show an optional dependency/feature tag if the info is there, but just show nothing if we don't know.

src/db/add_package.rs
```diff
@@ -221,13 +221,19 @@ fn convert_dependencies(pkg: &MetadataPackage) -> Vec<(String, String, String)>
 fn get_features(pkg: &MetadataPackage) -> Vec<Feature> {
     let mut features = Vec::with_capacity(pkg.features.len());
     if let Some(subfeatures) = pkg.features.get("default") {
-        features.push(Feature::new("default".into(), subfeatures.clone()));
+        features.push(Feature::new(
+            "default".into(),
+            subfeatures.clone(),
+            Some(false),
```
jyn514, line 227: So, this is making me regret using postgres types. `is_optional_dependency` should be non-null when being added, but nullable when being retrieved from the database. Do you know a good way to do that?

src/web/features.rs
```diff
@@ -206,8 +209,8 @@ mod tests {
 
     #[test]
     fn test_order_features_and_get_len_single_default() {
-        let default = Feature::new(DEFAULT_NAME.into(), Vec::new());
-        let non_default = Feature::new("non-default".into(), Vec::new());
+        let default = Feature::new(DEFAULT_NAME.into(), Vec::new(), Some(false));
+        let non_default = Feature::new("non-default".into(), Vec::new(), Some(false));
```
jyn514, line 214: Can you add some tests for the new behavior?

src/db/add_package.rs
Nemo157: `fn new` at least could take `bool` and implicitly wrap it into `Some`, so to set it to `None` you'd have to bypass the function and construct a `Feature` instance directly.

src/web/features.rs
jyn514: There's no changes to the frontend, so no need for tests, I was confused. Testing this requires build tests (#822).

src/db/types.rs
```diff
@@ -6,11 +6,16 @@ use serde::Serialize;
 pub struct Feature {
     pub(crate) name: String,
     pub(crate) subfeatures: Vec<String>,
+    pub(crate) optional_dependency: Option<bool>,
```
jyn514, line 9:
```suggestion
    /// [`None`] when this crate was built before optional dependencies were tracked
    pub(crate) optional_dependency: Option<bool>,
```

jyn514 (approved):

src/db/types.rs
almusil: Done :)

jyn514:
Thanks for working on this!

PR #1183, opened 2020-11-17
Remove whitespace beside docs.rs text

pickfire:
Last time I removed whitespace somewhere else, this time in menu.

Re-fix #1160 

r? @jyn514

### DIFF
--- a/templates/rustdoc/topbar.html
+++ b/templates/rustdoc/topbar.html
@@ -3,9 +3,9 @@
 {# The url of the current release, `/crate/:name/:version` #}
 {%- set crate_url = "/crate/" ~ metadata.name ~ "/" ~ metadata.version -%}
 
-{%- include "header/topbar_begin.html" -%}
+{%- include "header/topbar_begin.html" -%}{#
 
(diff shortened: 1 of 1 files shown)


pickfire:
I wonder if there is some sort of templating that automatically removes all these unnecessary whitespaces.

jyn514:
Can you post a screenshot of the issue this is fixing?

pickfire:
https://github.com/Keats/tera/pull/430/files maybe we could use `{{-`? Not sure if `{%-` is doing what I think it is.

pickfire:
Should be the same as the old one since I probably remove another whitespace last time.

image

jyn514:
Seems fine, then.

pickfire:
How long does it take for the change to be reflected in the current docs.rs?

jyn514:
@pickfire the deployments are irregular; 'when I find time for it'. For smaller PRs I usually wait until a few larger ones land, since each deployment has about 10 seconds of downtime until we fix https://github.com/rust-lang/docs.rs/issues/1150.

Issue #1177, opened 2020-11-11
Use cargo's rustdoc-map instead to link to dependencies


Nemo157:
Currently we're generating `--extern-html-root-url` flags to pass to rustdoc ourselves:

https://github.com/rust-lang/docs.rs/blob/b527df80a7332b35294c02b2922a2899dee43aaa/src/docbuilder/rustwide_builder.rs#L521-L529

This has issues (https://github.com/rust-lang/docs.rs/issues/1166, https://github.com/rust-lang/docs.rs/issues/1165).

Cargo has an [unstable `rustdoc-map` feature](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#rustdoc-map) https://github.com/rust-lang/cargo/issues/8296 intended to allow users to do a similar thing easily via a simple url template. We can likely use this instead of doing it manually.

jyn514:
This needs rustdoc to crates.io to docs.rs automatically before we can use it - right now you have to set a field in `.cargo/config`, but rustwide removes `config` before building the crate.

Issue #1166, opened 2020-11-05
Wrong --extern-html-root-url key used


Nemo157:
https://github.com/rust-lang/docs.rs/blob/9ca26817b38453ee4fec7e1330bbd734174a87ed/src/docbuilder/rustwide_builder.rs#L525

This should almost certainly be `dep.library_name()` or something similar, to handle dependencies which rename their library (I don't know of any example build logs to see if it's an issue though). See https://github.com/rust-lang/rust/issues/76296 as well, we can't actually give a unique name here when things like duplicate renamed dependencies come into play.

jyn514:
Rustdoc expects the library name anyway, we can't do anything about that until https://github.com/rust-lang/rust/issues/76296 is fixed. We definitely shouldn't be using `.replace()` though.

Issue #1165, opened 2020-11-05
Missing --extern-html-root-url for optional dependencies


Nemo157:
If a dependency is optional it will not be picked up in the root dependencies cause this is not taking the `[package.metadata.docs.rs]` into account:

https://github.com/rust-lang/docs.rs/blob/78bbdede0df5390a169491d5296955be8130b2c3/src/utils/cargo_metadata.rs#L52

Example: it has `chrono` active, but no `--extern-html-root-url` for it: https://docs.rs/crate/rusqlite/0.24.1/builds/301100

Issue #1055, opened 2020-09-18
Use semver compatible links to dependencies


Nemo157:
Currently when we build documentation we pass a flag like `--extern-html-root-url foobar=https://docs.rs/foobar/1.0.0` for each dependency with the exact version of the dependency used. This means if the dependency releases a new version compatible with the version constraint, all the links to it from the dependent crate will still go to the old version.

To avoid this we could pass semver-constraint urls like `--extern-html-root-url foobar=https://docs.rs/foobar/^1.0.0` (based on the actual constraint specified), then links will automatically go to the latest compatible version of the docs.

The biggest blocker to doing so is getting guarantees from Rustdoc that intra-doc-links generated for one version of a crate will still be valid in all semver-compatible later versions of the crate. A couple of issues currently cause this to not be true:

 1. The exact kind of a type is part of the url, so while e.g. changing a `struct` to an `enum` is a non-breaking change to code, it will result in a link to the item becoming a 404 (https://github.com/rust-lang/rust/issues/55160)
 2. Re-exports don't exist in the file tree, as an example consider these two versions of a crate:

```rust
// Version 1.0.0
pub struct Foo;
```
```rust
// Version 1.0.1 - A semver compatible change to move the canonical location and leave a re-export behind
pub mod foo { pub struct Foo; }
pub use foo::Foo;
```

Looking at the filetree of the built docs we can see that a link to the old canonical location of `foo/struct.Foo.html` will be a 404 in the new version:

```
target/doc/foo
├── all.html
├── index.html
├── sidebar-items.js
└── struct.Foo.html

target/doc/foo
├── all.html
├── foo
│   ├── index.html
│   ├── sidebar-items.js
│   └── struct.Foo.html
├── index.html
└── sidebar-items.js
```

(There may be more issues, these are just what was linked to me by @jyn514 and the first thought I had of what might go wrong. The first step of actually working on this if it's worth it would be identifying all the potential issues and working out whether it's possible/worthwhile trying to get this guarantee out of Rustdoc).

jyn514:
I posted an RFC for the rustdoc side of this: https://github.com/rust-lang/rfcs/pull/2988

jyn514:
Since https://github.com/rust-lang/docs.rs/pull/1188 this will also require changes to rustdoc-map to allow semver versions in the links.

PR #1188, opened 2020-11-22
Use `-Z rustdoc-map` instead of rewriting it

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/1177, closes https://github.com/rust-lang/docs.rs/issues/1165. Fixes https://github.com/rust-lang/docs.rs/issues/1166.

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -583,7 +583,7 @@ impl RustwideBuilder {
         cargo_args.push("--verbose".into());
         // We know that `metadata` unconditionally passes `-Z rustdoc-map`.
         // Don't copy paste this, since that fact is not stable and may change in the future.
-        // Normally we would need -Z unstable-options, but we currently pass that too.
+        cargo_args.push("-Zunstable-options".into());
         cargo_args.push(r#"--config=doc.extern-map.registries.crates-io="https://docs.rs""#.into());
         if let Some(cpu_limit) = self.config.build_cpu_limit {
             cargo_args.push(format!("-j{}", cpu_limit));
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -518,16 +518,6 @@ impl RustwideBuilder {
 
         let mut rustdoc_flags = Vec::new();
 
-        for dep in &cargo_metadata.root_dependencies() {
-            rustdoc_flags.push("--extern-html-root-url".to_string());
-            rustdoc_flags.push(format!(
-                "{}=https://docs.rs/{}/{}",
-                dep.name.replace("-", "_"),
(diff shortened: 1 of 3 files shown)


Nemo157:
A crate to test #1166 with is `futures-util-preview 0.2.2`, it depends on `futures-core-preview` which has the lib name `futures_core`.

Nemo157:
Tested with this little patch applied on top:

```patch
diff --git src/docbuilder/rustwide_builder.rs src/docbuilder/rustwide_builder.rs
index 83592dc..1a80aea 100644
--- src/docbuilder/rustwide_builder.rs
+++ src/docbuilder/rustwide_builder.rs
@@ -579,6 +579,9 @@ impl RustwideBuilder {

         // Add docs.rs specific arguments
         let mut cargo_args = Vec::new();
+        cargo_args.push("--verbose".into());
+        cargo_args.push(r#"--config=doc.extern-map.registries.crates-io="https://docs.rs""#.into());
+        cargo_args.push("-Zunstable-options".into());
         if let Some(cpu_limit) = self.config.build_cpu_limit {
             cargo_args.push(format!("-j{}", cpu_limit));
         }
```

On `master` I see us passing:

```
"--extern-html-root-url" "futures_core_preview=https://docs.rs/futures-core-preview/0.2.3"
```

And on this branch I see `cargo` passing:

```
--extern-html-root-url 'futures_core=https://docs.rs/futures-core-preview/0.2.3/'
```

So this definitely fixes #1166

jyn514:
Thanks, that definitely looks like the right approach since it means the hyperlinks will work even when built with old versions of nightly :)

This is ready for re-review.

src/docbuilder/rustwide_builder.rs
```diff
@@ -589,6 +579,12 @@ impl RustwideBuilder {
 
         // Add docs.rs specific arguments
         let mut cargo_args = Vec::new();
+        // Show the command line Rustdoc is invoked with
+        cargo_args.push("--verbose".into());
+        // We know that `metadata` unconditionally passes `-Z rustdoc-map`.
+        // Don't copy paste this, since that fact is not stable and may change in the future.
+        // Normally we would need -Z unstable-options, but we currently pass that too.
```
Nemo157, line 586: We didn't in my testing, I think it's only when the crate has `package.metadata.docs.rs.rustc-args` set that we do.

pietroalbini:
Random note on this: it'd be nice if `rustdoc-map` allowed using the semver pattern instead of the version, since docs.rs would then redirect to the latest semver-compatible docs automatically:

```
https://docs.rs/rustwide/^0.10/...
```

jyn514:
That would be nice, yeah - it needs rustdoc changes first, though.
https://github.com/rust-lang/docs.rs/issues/1055

PR #1191, opened 2020-11-22
Fix benchmarks

jyn514:
- compression just needed to use the new crate name
- `html_parsing` no longer makes sense, since we aren't
  parsing+serializing the HTML page. Rather than rewriting it from
  scratch, just remove it for now.

r? @Kixiron


Nemo157:
Are the benchmarks not being compiled in CI?

Kixiron (approved):

jyn514:
@Nemo157 no, I never set it up. Do you think I should?

Nemo157:
It'd be nice to ensure they're at least minimally updated when the interface changes. I normally have them caught via `cargo clippy --all-targets` checking.

Nemo157 (approved):

PR #1192, opened 2020-11-22
Remove --verbose when building with cargo

jyn514:
This causes the logs to be too large and get truncated.

r? @Nemo157


Nemo157 (approved):

Nemo157:
(It'd be nice to reintroduce it once the logs get uploaded to S3 and we can afford to bump the default limit).

PR #1195, opened 2020-11-28
Regression prevention

GuillaumeGomez:
Just realized that I could add a test to prevent to have to send other PRs like https://github.com/rust-lang/docs.rs/pull/1194 :)


jyn514:
I don't know why I didn't think of that earlier :laughing: great idea!

GuillaumeGomez:
Arf hold on, seems like github didn't realize the first commit was already in.

GuillaumeGomez:
After a rebase, github realized its mistake and made the commit disappear. :3

PR #1203, opened 2020-12-09
Fix behavior on menu bar with focus and active

### LINKED ISSUE #1201: #1164 introduces weird navbar item focus behavior
After https://github.com/rust-lang/docs.rs/pull/1164 I am seeing:

Peek 2020-12-09 00-18

caused by:

https://github.com/rust-lang/docs.rs/blob/b728426b65e67dc84c568da6bab4ebe59c4ad9cc/templates/style/_navbar.scss#L47-L52

pickfire:
Weird behavior is caused due to hover using color and focus
using background color as hover background-color is inherit,
so flipping will make both situation taking from focus, better.

Fix #1201

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -39,16 +39,16 @@ div.nav-container {
         font-weight: 400;
         color: var(--color-navbar-standard);
 
-        &:hover {
-            color: var(--color-standard);
-            background-color: inherit;
-        }
-
         // Improves menu link readability when inverting the colors on focus.
         // Vendor do background-color #eee, looks weird on different theme.
         &:focus {
             color: var(--color-background);
             background-color: var(--color-doc-link-background);
+        }
+
+        &:hover {
+            color: var(--color-standard);
+            background-color: inherit;
         }
 
         &.crate-name {


cynecx:
I still can reproduce #1201 locally with these changes.

It seems that the `div.nav-container .pure-menu-has-children>.pure-menu-link:hover` rule is considered before the other ones.

issue

(The screenshot is missing the cursor, which is currently hovered above the navbar item.)

pickfire:
I changed the pos from css style sheet itself but it took the `:focus` `color` first, did you swap the position if you had done it manually?

cynecx:
@pickfire I reordered the css rule. I also compiled the scss and updated the according css region in chromium devtools. Both didn't show any effects. I think we shouldn't rely on reorderings and perhaps use `!important`, which is more or less a hack but that should do it.

pickfire:
It's fixed though, for me.

Hover

image

Focus

image

Hover + Focus

image

dumbcoder7 (approved):

cynecx:
I actually did not build docs.rs from scratch (So I am applying these changes to the production build on docs.rs and not a build from master). I applied the css changes directly through devtools (chromium/firefox). By just reordering those two rules I don't see any differences in both firefox and chromium and light and dark theme.

Maybe my expectations are just different? (Or I am doing something fundamentally wrong) Could you take a gif recording on your build with these changes like I did in the original issue? (That would be great.)

pickfire:
> I actually did not build docs.rs from scratch (So I am applying these changes to the production build on docs.rs and not a build from master). I applied the css changes directly through devtools (chromium/firefox). By just reordering those two rules I don't see any differences in both firefox and chromium and light and dark theme.

I think you might have reordered the wrong css rule. IIRC the production one does not have `active` or `hover`, it is just like `:`. I reordered that and it worked.

output-opt

Github can't upload big gif so I made it smaller, also harder to see.

cynecx:
Well, sorry I am currently unable to get a dev build, but I guess that looks okay? I feel like I am blocking the PR. So please go ahead :)

pickfire:
Let's see if @jyn514 can reproduce the issue?

pickfire:
r? @jyn514 

Not sure who is the right reviewers.

jyn514:
This is fixed for me after the changes in this PR. @pickfire thanks for the fix. @cynecx I'm not sure why you can't reproduce with devtools, but if you still have trouble after this is merged feel free to re-open the issue.

Issue #1176, opened 2020-11-11
Make fewer github requests


jyn514:
This is a meta tracking issue, but for convenience I'm going to put the steps for solving it here too.

Currently, docs.rs makes way too many requests to the GitHub API:
![image](https://user-images.githubusercontent.com/23638587/98851007-9fe56b80-2423-11eb-88cd-1ccaeb763a84.png)

Note how bors and crater are only making around 20 requests per hour - docs.rs makes _5000_, and only stops because it hits the rate limit. We should fix this.

1. The first and most pressing issue is that we're making requests for repos that no longer exist:

```
 [DEBUG] docs_rs::utils::github_updater: Starting update of all crates
 [DEBUG] docs_rs::utils::github_updater: Updating BrewStillery
 [WARN] docs_rs::utils::github_updater: Failed to update BrewStillery: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/MonkeyLog/BrewStillery)
 [DEBUG] docs_rs::utils::github_updater: Updating FPS
 [WARN] docs_rs::utils::github_updater: Failed to update FPS: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/leeopop/FPS)
 [DEBUG] docs_rs::utils::github_updater: Updating Lattice
 [WARN] docs_rs::utils::github_updater: Failed to update Lattice: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/andrew-lucker/Lattice)
...
```

We should cache this failure and not retry the next time, to avoid wasting calls on deleted repos. The cache should be invalidated then next time the crate is published, in case the repository URL is changed, and we should fix the code to deal with a renamed URL: https://github.com/rust-lang/docs.rs/blob/master/src/utils/github_updater.rs#L68-L69

2. We should stop updating so frequently. Right now we update github stars once per day; we could probably turn that down to 1 week without trouble.

There may be other improvements lying around - I had the idea to fetch multiple repositories at once, but looking at https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#detailed-representations I don't think it's possible.

jyn514:
> I had the idea to fetch multiple repositories at once

I think this would be possible if we used the GraphQL API instead of REST. Thanks @nickmccurdy for the suggestion!

nickmccurdy:
> I had the idea to fetch multiple repositories at once
[...]

PR #1208, opened 2020-12-17
Switch GitHub updates to GraphQL, part 1

pietroalbini:
This PR is the first of two PRs that are going to switch our GitHub updater to use GitHub's GraphQL API, reducing the amount of API requests docs.rs makes to GitHub and fixing https://github.com/rust-lang/docs.rs/issues/1176.

We can't just switch the current code to use GraphQL, as the GraphQL API relies on "node IDs" (for example `MDEwOlJlcG9zaXRvcnkyMzg1MzQ4Mg==`) to query multiple items at once. Because of that we'll need to have a transition period where we use the old code but also collect the node IDs we'll use in the future to query the new information in batches. The plan is the following:

1. **<We're here!>** Land a code change that updates the old process to also store the node IDs in the database.
2. Keep that code change deployed for at least a day, to collect the node IDs of all the repositories we currently have.
3. Land a code change that replaces the old updater code with the new one, based on GraphQL.
4. Before deploying that code change, disable automatic GitHub updates (to ensure we have API quota), lock the build queue (to ensure no more crates are being added) and manually call `cratesfyi database update-github-fields` to ensure newly built crates are also present in the new schema.
5. Deploy the changes landed in step 3!

Since we needed to change the database schema to also store node IDs, I took this opportunity to define a new schema for storing the GitHub information: there will be a `github_repos` table keyed by the repository node ID which contains the information, and each release will have a field called `github_repo` pointing to the right row in the `github_repos` table. This will have a few really nice side-effects:

* It will be easy to handle deleted repos: when GitHub returns a 404 on a repository we'll just need to delete the row from `github_repos`, and the foreign key will set all the affected releases's `github_repo` to `NULL`.
* It will be easy to handle renamed repositories; since the node ID stays the same across renames, we only need to update the name when we get a response from the API.
* We will only fetch the stats for each repository once, even if multiple crates are pointing to the same repositories.
* We will support different releases pointing to different repositories.

This PR includes the code to store information fetched during the update on both the old `crates` table and the new `github_repos` one, and to link each release with the right repository. The rest of the changes will be done in a followup PR. There are also two vaguely related changes in this PR:

* The `CRATESFYI_GITHUB_USERNAME` environment variable has been removed, as that was not really useful.
* I added a **temporary** environment variable to disable the github updater, used for step 4 above. I'll remove it in the next PR.

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -502,6 +502,33 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
             "
                  ALTER TYPE feature DROP ATTRIBUTE optional_dependency;
             "
+        ),
+        migration!(
+            context,
+            22,
+            // description
+            "Add the github_repositories table to contain GitHub information",
+            // upgrade query
+            "
+                CREATE TABLE github_repos (
+                    id VARCHAR PRIMARY KEY NOT NULL,
+                    name VARCHAR NOT NULL,
+                    description VARCHAR,
+                    last_commit TIMESTAMP,
+                    stars INT NOT NULL,
+                    forks INT NOT NULL,
+                    issues INT NOT NULL,
+                    updated_at TIMESTAMP NOT NULL
+                );
+
+                ALTER TABLE releases ADD COLUMN github_repo VARCHAR
+                    REFERENCES github_repos(id) ON DELETE SET NULL;
+            ",
+            // downgrade query
+            "
+                ALTER TABLE releases DROP COLUMN github_repo;
+                DROP TABLE github_repos;
+            "
         )
     ];
 
--- a/src/utils/daemon.rs
+++ b/src/utils/daemon.rs
@@ -94,15 +94,19 @@ pub fn start_daemon(context: &dyn Context, enable_registry_watcher: bool) -> Res
     )?;
 
     // update github stats every hour
-    let github_updater = GithubUpdater::new(&config, context.pool()?)?;
-    cron(
-        "github stats updater",
-        Duration::from_secs(60 * 60),
-        move || {
-            github_updater.update_all_crates()?;
-            Ok(())
-        },
-    )?;
+    if config.tmp_disable_github_updater {
+        log::warn!("the github stats updater was disabled!");
+    } else {
+        let github_updater = GithubUpdater::new(&config, context.pool()?)?;
+        cron(
+            "github stats updater",
+            Duration::from_secs(60 * 60),
+            move || {
+                github_updater.update_all_crates()?;
+                Ok(())
+            },
+        )?;
+    }
 
     // Never returns; `server` blocks indefinitely when dropped
     // NOTE: if a failure occurred earlier in `start_daemon`, the server will _not_ be joined -
--- a/src/utils/github_updater.rs
+++ b/src/utils/github_updater.rs
@@ -34,18 +34,15 @@ pub struct GithubUpdater {
 
 impl GithubUpdater {
     pub fn new(config: &Config, pool: Pool) -> Result<Self> {
-        let github_auth = config.github_auth();
-
-        let mut headers = HeaderMap::with_capacity(2 + github_auth.is_some() as usize);
+        let mut headers = HeaderMap::new();
         headers.insert(USER_AGENT, HeaderValue::from_static(APP_USER_AGENT));
         headers.insert(ACCEPT, HeaderValue::from_static("application/json"));
 
-        if let Some((username, accesstoken)) = github_auth {
-            let basicauth = format!(
-                "Basic {}",
-                base64::encode(format!("{}:{}", username, accesstoken))
+        if let Some(token) = &config.github_accesstoken {
+            headers.insert(
+                AUTHORIZATION,
+                HeaderValue::from_str(&format!("token {}", token))?,
             );
-            headers.insert(AUTHORIZATION, HeaderValue::from_str(&basicauth).unwrap());
         } else {
             warn!("No GitHub authorization specified, will be working with very low rate limits");
         }
--- a/src/utils/github_updater.rs
+++ b/src/utils/github_updater.rs
@@ -26,7 +26,7 @@ struct GitHubFields {
     stars: i64,
     forks: i64,
     issues: i64,
-    last_commit: DateTime<Utc>,
+    last_commit: Option<DateTime<Utc>>,
 }
 
 pub struct GithubUpdater {
@@ -143,7 +143,7 @@ impl GithubUpdater {
                 &fields.node_id,
                 &fields.full_name,
                 &fields.description,
-                &fields.last_commit.naive_utc(),
+                &fields.last_commit.as_ref().map(|lc| lc.naive_utc()),
                 &(fields.stars as i32),
                 &(fields.forks as i32),
                 &(fields.issues as i32),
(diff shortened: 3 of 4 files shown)


Turbo87:
> as the GraphQL API relies on "node IDs"

I tried this out on the GraphQL explorer of GitHub a few days/weeks ago, aber I don't remember needing to specify any node IDs. It probably still makes sense to use them for all of the other reasons that you mentioned, but I'm curious why they appear to be a requirement in your query.

pietroalbini:
> I tried this out on the GraphQL explorer of GitHub a few days/weeks ago, aber I don't remember needing to specify any node IDs. It probably still makes sense to use them for all of the other reasons that you mentioned, but I'm curious why they appear to be a requirement in your query.

My understanding is that node IDs are required when you need to fetch a list of repositories in the same query.

Turbo87:
```graphql
query { 
  crates: repository(owner:"rust-lang", name:"crates.io") {
    stargazerCount
  }
  docs: repository(owner:"rust-lang", name:"docs.rs") {
    stargazerCount
  }
}
```

this was the query that I ran and it appeared to work fine without node IDs, but maybe I missed something 😅

pietroalbini:
Well that's true, if we want to build the query dynamically (which could get messy). Using node IDs allows this query instead:

```graphql
query($ids: [ID!]!) {
  nodes(ids: $ids) {
    ... on Repository {
      stargazerCount
    }
  }
}
```

...which natively loops over the repositories and returns a list of them.

Turbo87:
I see, thanks! :)

src/utils/github_updater.rs
```diff
@@ -141,12 +167,21 @@ impl GithubUpdater {
             ],
         )?;
 
+        // Temporary statement to migrate production data over to the new table. Will be removed by
+        // Pietro's next PR.
+        conn.execute(
+            "UPDATE releases SET github_repo = $1 WHERE crate_id = $2;",
```
jyn514, line 173: This inherits the confusion between crates and releases from the existing codebase. Are you planning to change that in a follow-up? (This is fine for now to be clear.)

src/utils/daemon.rs
```diff
@@ -94,15 +94,19 @@ pub fn start_daemon(context: &dyn Context, enable_registry_watcher: bool) -> Res
...
+        cron(
+            "github stats updater",
+            Duration::from_secs(60 * 60),
+            move || {
+                github_updater.update_all_crates()?;
+                Ok(())
+            },
+        )?;
+    } else {
+        log::warn!("the github stats updater was disabled!");
```
jyn514, line 108: nit: double negatives are confusing, can you say `if disable { warn }` instead?

src/utils/github_updater.rs
```diff
@@ -34,18 +34,15 @@ pub struct GithubUpdater {
...
-            let basicauth = format!(
-                "Basic {}",
-                base64::encode(format!("{}:{}", username, accesstoken))
+        if let Some(token) = &config.github_accesstoken {
+            headers.insert(
+                AUTHORIZATION,
+                HeaderValue::from_str(&format!("token {}", token))?,
             );
-            headers.insert(AUTHORIZATION, HeaderValue::from_str(&basicauth).unwrap());
         } else {
```
jyn514, line 48: How can I test that Github accepts this new auth format?

src/db/migrate.rs
```diff
@@ -502,6 +502,33 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
+                    description VARCHAR,
+                    last_commit TIMESTAMP,
+                    stars INT NOT NULL,
+                    forks INT NOT NULL,
+                    issues INT NOT NULL,
+                    updated_at TIMESTAMP NOT NULL
+                );
+
+                ALTER TABLE releases ADD COLUMN github_repo VARCHAR
+                    REFERENCES github_repos(id) ON DELETE SET NULL;
```
jyn514, line 525: This will be NULL by default, and then set to the right repository as github_updater runs? That doesn't have a way to distinguish between 'not yet updated' and 'deleted repo', will we need to know that? Or is the plan to have everything converted before switching this off?

In other words, I think this doesn't give us a way to know if the updater has run on a given repo, so we don't know whether we need to let it run longer.

jyn514 (commented):
The plan sounds great :) I have a few questions on the implementation.

src/utils/github_updater.rs
pietroalbini: [That's the documented and recommended way to authenticate with the GitHub API](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#authentication).

src/utils/github_updater.rs
pietroalbini: The second PR will remove all GitHub data from crates.

src/db/migrate.rs
pietroalbini: My plan is to have the updater go through every repository before removing the "legacy" code. For crates published after the switch to GraphQL, my plan is to gather their GitHub stats at the end of the build process.

src/db/migrate.rs
```diff
@@ -502,6 +502,33 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
...
+            22,
+            // description
+            "Add the github_repositories table to contain GitHub information",
+            // upgrade query
+            "
+                CREATE TABLE github_repos (
+                    id VARCHAR PRIMARY KEY NOT NULL,
+                    name VARCHAR NOT NULL,
+                    description VARCHAR,
+                    last_commit TIMESTAMP,
```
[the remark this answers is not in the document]
pietroalbini: ...oh, we're not actually using the last commit anywhere in the application :sweat_smile: 

Still, pushed a commit to make it optional.

pietroalbini:
Addressed or responded to all the review feedback.

src/db/migrate.rs
jyn514: > My plan is to have the updater go through every repository before removing the "legacy" code. For crates published after the switch to GraphQL, my plan is to gather their GitHub stats at the end of the build process.

Right, but how do you tell when it's gone through every repository? You can't tell by looking at the database because it's valid normally for the `github_repo` to be NULL.

pietroalbini: > Right, but how do you tell when it's gone through every repository?

It goes through every repository each day. The number of crates that weren't updated in a day (query 1) is just 1369, and when randomly checking their repositories (clicking random links of query 2) all of them were 404s. If we leave it running over the weekend we'll surely convert all of the crates to the new schema.

```sql
-- Query 1
SELECT COUNT(*) FROM crates WHERE github_last_update < NOW() - INTERVAL '1 day 1 hour';
-- Query 2:
SELECT DISTINCT (crates.id) crates.name, releases.repository_url FROM crates, releases WHERE crates.id = releases.crate_id AND github_last_update < NOW() - INTERVAL '1 day 1 hour';
```

The only thing we'd be missing is newly published crates, but I'll handle them in a followup PR.

src/utils/github_updater.rs
jyn514: Right, but that's not my question - how can I test that it works?

src/db/migrate.rs
jyn514: Ok, seems reasonable.

src/utils/github_updater.rs
pietroalbini:
```
CRATESFYI_GITHUB_ACCESSTOKEN=yourtoken cargo run database update-github-fields
```

PR #1209, opened 2020-12-19
Switch GitHub updates to GraphQL, part 2

### LINKED ISSUE #1176: Make fewer github requests

pietroalbini:
This PR finishes the migration to GraphQL for fetching GitHub repository statistics started in #1208, and fixes #1176 in the process.

There are mainly four things changed by this PR:

* The code that periodically updates the GitHub metadata was changed to use GraphQL to fetch updates in batches, and to only update the new `github_repos` table. When a repository is deleted or made private the line in the `github_repos` table will be removed, preventing further updates to that repository.
* The documentation builder was changed to fetch the GitHub statistics for new releases right after the build (or to use already cached statistics). This means new crates will get their stats as soon as they're available, and there is no need to distinguish "crates that don't have a repository" vs "crates that were not scraped yet" in the schema.
* The command  `cratesfyi database backfill-github-stats` was added to manually try and gather stats for every crate that links to GitHub but doesn't have a repository attached. This will be useful during the migration to ensure we don't lose crates.
* The frontend was changed to fetch the stats from the new `github_repos` table, and the old data is dropped from the database.

Another notable change is that the daily updater now stops *before* the rate limit is reached (the threshold is configured through an environment variable). This was done both to avoid monitoring yelling that the rate limit was reached, and to leave room for the documentation builder to fetch statistics after builds. The default threshold is 2.5k requests, which will be able to update 250k repositories in one go.

The revised deployment plan for GraphQL is:

| # | Status | Task |
| --- | --- | --- |
| 1 | :green_circle: Done! | Land a code change that updates the old process to also store the node IDs in the database. |
| 2 | :green_circle: Done! | Keep that code change deployed for at least a day, to collect the node IDs of all the repositories we currently have. |
| 3 | :wave: This! | Land a code change that replaces the old updater code with the new one, based on GraphQL. |
| 4 | - | Before deploying that code change, disable automatic GitHub updates (to ensure we have API quota), lock the build queue (to ensure no more crates are being added) and manually call `cratesfyi database backfill-github-stats` to ensure newly built crates are also present in the new schema. |
| 5 | - | Deploy the changes landed in step 3! |

r? @jyn514 

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -8,7 +8,7 @@ use crate::docbuilder::{crates::crates_from_path, Limits};
 use crate::error::Result;
 use crate::index::api::ReleaseData;
 use crate::storage::CompressionAlgorithms;
-use crate::utils::{copy_doc_dir, parse_rustc_version, CargoMetadata};
+use crate::utils::{copy_doc_dir, parse_rustc_version, CargoMetadata, GithubUpdater};
 use crate::{Config, Context, Index, Metrics, Storage};
 use docsrs_metadata::{Metadata, DEFAULT_TARGETS, HOST_TARGET};
 use failure::ResultExt;
@@ -392,9 +392,25 @@ impl RustwideBuilder {
                     }
                 };
 
+                let cargo_metadata = res.cargo_metadata.root();
+
+                // Fetch repository statistics from GitHub
+                let github_updater = GithubUpdater::new(self.config.clone(), self.db.clone())?;
+                let github_repo = if let Some(repo) = &cargo_metadata.repository {
+                    match github_updater.load_repository(&mut conn, repo) {
+                        Ok(repo) => repo,
+                        Err(err) => {
+                            warn!("failed to get github stats: {}", err);
+                            None
+                        }
+                    }
+                } else {
+                    None
+                };
+
                 let release_id = add_package_into_database(
                     &mut conn,
-                    res.cargo_metadata.root(),
+                    cargo_metadata,
                     &build.host_source_dir(),
                     &res.result,
                     &res.target,
@@ -404,6 +420,7 @@ impl RustwideBuilder {
                     has_docs,
                     has_examples,
                     algs,
+                    github_repo,
                 )?;
 
                 if let Some(doc_coverage) = res.result.doc_coverage {
--- a/src/utils/github_updater.rs
+++ b/src/utils/github_updater.rs
@@ -1,7 +1,7 @@
 use crate::error::Result;
 use crate::{db::Pool, Config};
 use chrono::{DateTime, Utc};
-use log::{info, trace, warn};
+use log::{debug, info, trace, warn};
 use once_cell::sync::Lazy;
 use postgres::Client;
 use regex::Regex;
@@ -32,6 +32,18 @@ const GRAPHQL_UPDATE: &str = "query($ids: [ID!]!) {
     }
     rateLimit {
         remaining
+    }
+}";
+
+const GRAPHQL_SINGLE: &str = "query($owner: String!, $repo: String!) {
+    repository(owner: $owner, name: $repo) {
+        id
+        nameWithOwner
+        pushedAt
+        description
+        stargazerCount
+        forkCount
+        issues { totalCount }
     }
 }";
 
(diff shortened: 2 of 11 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -392,9 +392,25 @@ impl RustwideBuilder {
                     }
                 };
 
+                let cargo_metadata = res.cargo_metadata.root();
+
+                // Fetch repository statistics from GitHub
+                let github_updater = GithubUpdater::new(self.config.clone(), self.db.clone())?;
+                let github_repo = if let Some(repo) = &cargo_metadata.repository {
+                    match github_updater.load_repository(&mut conn, repo) {
```
jyn514, line 400: this is lazy because `load_repository` is lazy, nice :) (lazy -> uses cached data from the DB if available)

src/web/crate_details.rs
```diff
@@ -139,6 +145,16 @@ impl CrateDetails {
                 .collect()
         };
 
+        let github_metadata = if krate.get::<_, Option<String>>("github_repo").is_some() {
```
jyn514, line 148: This is a change in behavior, right? Before it would have `github = true` whenever you had `http://github.com` as the URL in Cargo.toml, now it will only have it when the repository exists (not private or deleted).

I think that's a good change, just wanted to double check.

src/web/releases.rs
```diff
@@ -82,13 +82,14 @@ pub(crate) fn get_releases(conn: &mut Client, page: i64, limit: i64, order: Orde
...
-            crates.github_stars
+            github_repos.stars
         FROM crates
         INNER JOIN releases ON crates.id = releases.crate_id
+        LEFT JOIN github_repos ON releases.github_repo = github_repos.id
         WHERE
             ((NOT $3) OR (releases.build_status = FALSE AND releases.is_library = TRUE))
             AND crates.latest_version_id = releases.id
-        ORDER BY {} DESC
+        ORDER BY {} DESC NULLS LAST
```
jyn514, line 92: Can you add a test that crates without github metadata are shown last on the 'releases by stars' page?

jyn514 (commented):
This looks awesome :D I want to play with the changes some more locally but this definitely looks like the right approach.

src/utils/github_updater.rs
```diff
@@ -140,71 +164,22 @@ impl GithubUpdater {
...
-            full_name: response.full_name,
-            description: response.description.unwrap_or_default(),
-            stars: response.stargazers_count,
-            forks: response.forks_count,
-            issues: response.open_issues,
-            last_commit: response.pushed_at,
-        })
+    fn delete_repository(&self, conn: &mut Client, id: &str) -> Result<()> {
+        trace!("removing GitHub repository stats for ID {}", id);
+        conn.execute("DELETE FROM github_repos WHERE id = $1;", &[&id])?;
```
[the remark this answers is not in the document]
pietroalbini: Yes!

src/db/migrate.rs
```diff
@@ -74,12 +74,6 @@ pub fn migrate(version: Option<Version>, conn: &mut Client) -> CratesfyiResult<(
                  latest_version_id INT DEFAULT 0,
                  versions JSON DEFAULT '[]',
                  downloads_total INT DEFAULT 0,
-                 github_description VARCHAR(1024),
-                 github_stars INT DEFAULT 0,
-                 github_forks INT DEFAULT 0,
-                 github_issues INT DEFAULT 0,
-                 github_last_commit TIMESTAMP,
-                 github_last_update TIMESTAMP,
```
Nemo157, line 82: These columns still need to exist for later migrations to modify, until migration 23 is run.

pietroalbini: Removed that migration, will put it as its own separate PR later.

src/utils/github_updater.rs
```diff
@@ -51,81 +61,95 @@ impl GithubUpdater {
...
+            .map(|row| row.get(0))
+            .collect::<Vec<String>>();
+
+        if needs_update.is_empty() {
+            info!("no GitHub repository stats needed to be updated");
+        } else {
+            for chunk in needs_update.chunks(UPDATE_CHUNK_SIZE) {
+                if let Err(err) = self.update_repositories(&mut conn, &chunk) {
+                    if err.downcast_ref::<RateLimitReached>().is_some() {
+                        warn!("rate limit reached, blocked the GitHub repository stats updater");
```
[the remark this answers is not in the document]
pietroalbini: Honestly I'm not sure whether that's needed... I'd add the message in the future if we end up needing it.

src/utils/github_updater.rs
```diff
@@ -51,81 +61,95 @@ impl GithubUpdater {
...
+            .json()?;
+
+        // The error is returned *before* we reach the rate limit, to ensure we always have an
+        // amount of API calls we can make at any time.
+        trace!(
+            "GitHub GraphQL rate limit remaining: {}",
+            response.data.rate_limit.remaining
+        );
+        if response.data.rate_limit.remaining < self.config.github_updater_min_rate_limit {
+            return Err(RateLimitReached.into());
```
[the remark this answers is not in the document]
pietroalbini: Grafana's rate limit stats are gathered by [monitorbot](https://github.com/rust-lang/monitorbot), so we don't really care about gathering the same stats from docs.rs.

src/web/crate_details.rs
pietroalbini: Yes.

src/utils/github_updater.rs
```diff
@@ -183,22 +184,17 @@ impl GithubUpdater {
...
-        Some(cap) => {
-            let username = cap.get(1).unwrap().as_str();
-            let reponame = cap.get(2).unwrap().as_str();
-
-            let reponame = if reponame.ends_with(".git") {
-                reponame.split(".git").next().unwrap()
-            } else {
-                reponame
-            };
+fn extract_owner_and_repo(url: &str) -> Option<(&str, &str)> {
```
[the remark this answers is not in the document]
pietroalbini: I don't really like putting that information in doc comments. I agree it should be more explicit though, so I switched the return type to be a struct.

pietroalbini:
Addressed all the review comments!

jyn514:
This looks great! I don't have time to do the deploy steps right now, but @pietroalbini feel free to deploy yourself if you have time.

PR #1213, opened 2020-12-22
Use `clip-path` instead of `mask` to make round corners

### LINKED ISSUE #1212: The round corners of badge are blurry on iOS Safari
image

The first and the last badges are from shields.io, and the second is from docs.rs.

This is probably because shields.io makes the round corners with `clip-path`

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="102" height="20" role="img" aria-label="crates.io: v0.5.2">
    ...
    <clipPath id="r">
        <rect width="102" height="20" rx="3" fill="#fff"/>
    </clipPath>
    <g clip-path="url(#r)">
        <rect width="57" height="20" fill="#555"/>
        <rect x="57" width="45" height="20" fill="#fe7d37"/>
        <rect width="102" height="20" fill="url(#s)"/>
    </g>
    ...
</svg>
```

but docs.rs makes it with `mask`

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="73" height="20">
    ...
    <mask id="round">
        <rect width="73" height="20" rx="3" fill="#fff"/>
    </mask>
    <g mask="url(#round)">
        <rect width="34" height="20" fill="#555"/>
        <rect x="34" width="39" height="20" fill="#4d76ae"/>
        <rect width="73" height="20" fill="url(#smooth)"/>
    </g>
    ...
</svg>
```

balthild:
Before and after:

image

Fixes #1212

### DIFF
--- a/crates/badge/src/lib.rs
+++ b/crates/badge/src/lib.rs
@@ -78,11 +78,11 @@ impl Badge {
                 <stop offset="1" stop-opacity=".1"/>
               </linearGradient>
 
-              <mask id="round">
+              <clipPath id="round">
                 <rect width="{badge_width}" height="20" rx="3" fill="#fff"/>
-              </mask>
+              </clipPath>
 
-              <g mask="url(#round)">
+              <g clip-path="url(#round)">
                 <rect width="{left_width}" height="20" fill="#555"/>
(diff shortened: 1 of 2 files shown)


jyn514:
Thanks for the fix!

PR #1215, opened 2020-12-23
Switch GitHub updates to GraphQL, part 3

pietroalbini:
*Episode 3 of our beloved serie of "let's not hammer the GitHub API anymore". You can find previous episodes here: https://github.com/rust-lang/docs.rs/pull/1208 https://github.com/rust-lang/docs.rs/pull/1209*

This PR finishes the work of switching GitHub updates over to GraphQL:

* It's not possible anymore to use the updater without a GitHub token, as the GraphQL API straight up rejects unauthenticated requests. Warnings will be emitted instead, both at the start of the daemon and every time the builder tries to load stats.
* The temporary flag to disable the updater has been removed.
* The old database columns have been removed.

r? @jyn514

### DIFF
--- a/src/utils/github_updater.rs
+++ b/src/utils/github_updater.rs
@@ -59,6 +59,8 @@ pub struct GithubUpdater {
 }
 
 impl GithubUpdater {
+    /// Returns `Err` if the access token has invalid syntax (but *not* if it isn't authorized).
+    /// Returns `Ok(None)` if there is no access token.
     pub fn new(config: Arc<Config>, pool: Pool) -> Result<Option<Self>> {
         let mut headers = HeaderMap::new();
         headers.insert(USER_AGENT, HeaderValue::from_static(APP_USER_AGENT));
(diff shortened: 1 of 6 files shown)


src/utils/github_updater.rs
```diff
@@ -59,7 +59,7 @@ pub struct GithubUpdater {
 }
 
 impl GithubUpdater {
-    pub fn new(config: Arc<Config>, pool: Pool) -> Result<Self> {
+    pub fn new(config: Arc<Config>, pool: Pool) -> Result<Option<Self>> {
```
jyn514, line 62:
```suggestion
    /// Returns `Err` if the access token has invalid syntax (but *not* if it isn't authorized).
    /// Returns `Ok(None)` if there is no access token.
    pub fn new(config: Arc<Config>, pool: Pool) -> Result<Option<Self>> {
```

jyn514 (approved):

PR #1216, opened 2020-12-23
Ensure tests don't leak database connections

Nemo157:
Even if a pool instance is leaked (such as when the web server is leaked), internally drop the actual pool at the end of the test to ensure all connections are closed.

### DIFF
--- a/src/db/pool.rs
+++ b/src/db/pool.rs
@@ -10,6 +10,9 @@ const DEFAULT_SCHEMA: &str = "public";
 
 #[derive(Debug, Clone)]
 pub struct Pool {
+    #[cfg(test)]
+    pool: Arc<std::sync::Mutex<Option<r2d2::Pool<PostgresConnectionManager<NoTls>>>>>,
+    #[cfg(not(test))]
     pool: r2d2::Pool<PostgresConnectionManager<NoTls>>,
     metrics: Arc<Metrics>,
     max_size: u32,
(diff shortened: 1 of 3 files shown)


jyn514:
I still can't run all tests because we leak file descriptors, but I guess there's not much to do about that ... ideally we'd switch to warp or hyper so this wasn't an issue :/

Nemo157:
I might have fixed that when reducing the number of threads that the tests take, I could pull that commit into this PR too if you want?

jyn514:
> I might have fixed that when reducing the number of threads that the tests take, I could pull that commit into this PR too if you want?

That would be great!

jyn514:
This still leaks file descriptors for me :/ I guess if it fixes CI it's worth it, though.

```
thread 'r2d2-worker-1' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Other, message: "Too many open files" }', /home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/postgres-0.17.5/src/config.rs:322:14
```

jyn514:
Oh wait, I forgot to set `ulimit -n 4096` first. Now it works :)

jyn514:
r? @pietroalbini 

This seems ok to me, but I'm still a little hesitant about adding hacks on top of hacks when @Nemo157's been working on the warp switch: https://github.com/rust-lang/docs.rs/compare/master...Nemo157:warp-speed-iron

Nemo157:
Yeah, it doesn't _stop_ leakage, but it does reduce it by a multiplier of something like 4*num_cpus.

The warp switch is going to take a _long_ time to finish, and all these hacks can be deleted once it's completed.

pietroalbini:
I hate this, let's merge it :+1:

Issue #1073, opened 2020-10-04
Pagination links are broken when viewing releases by author


connec:
When viewing the releases for an author at `https://docs.rs/releases/{author}/` the pagination links are incorrectly set to `/releases/author/{n}`.

This would seem to be a bug in the construction of `page_link` [here](https://github.com/rust-lang/docs.rs/blob/38ba774e14eb4f658d834218d003583e423c5f37/templates/releases/releases.html#L60), which is then interpolated into the pagination links [here](https://github.com/rust-lang/docs.rs/blob/38ba774e14eb4f658d834218d003583e423c5f37/templates/releases/releases.html#L66) and [here](https://github.com/rust-lang/docs.rs/blob/38ba774e14eb4f658d834218d003583e423c5f37/templates/releases/releases.html#L72).

Presumably, if the `release_type` is `"author"`, then `page_link` should instead be set to `"/releases/" ~ author`.

jyn514:
@connec are you interested in working on this yourself? It looks like you already have a good idea of how to fix it :)

connec:
@jyn514 I haven't started anything yet as I don't have time right now to establish a build environment to validate my suggested fix. I may get around to it, but I wouldn't want to put off anyone else's efforts!

PR #1219, opened 2020-12-24
Interpolate author login into releases url for pagination

toshokan:
The issue is as described in #1073
I mostly followed the schema suggested there. I also had to change the handler to add the author username to the template context rather than the author name, but this is safe because the value wasn't used before anyway.

Fixes rust-lang/docs.rs#1073


jyn514:
The fixes look right, thanks. Can you add a test case for this? It would go in `src/web/releases.rs`, see the bottom of that file for examples.

toshokan:
Is this the sort of test you had in mind?

jyn514:
Looks great, thanks so much!

PR #1221, opened 2020-12-25
Highlight code blocks according to the current theme

### LINKED ISSUE #1127: Themes don't update codeblocks, so they look a little blinding in night mode
This doesn't update codeblocks, so they look a little blinding in night mode:

image

Could we fix that?

_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/pull/1116#issuecomment-716594541_

jyn514:
r? @Nemo157 

Closes #1127

### DIFF
--- a/templates/macros.html
+++ b/templates/macros.html
@@ -30,6 +30,7 @@
             case "ayu":
                 stylesheet = "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/styles/dark.min.css";
                 break;
+            case "null": // The user is visiting docs.rs for the first time and hasn't set a theme yet.
             case "light":
                 stylesheet = "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/styles/github.min.css";
                 break;
--- a/templates/macros.html
+++ b/templates/macros.html
@@ -22,8 +22,29 @@
 {# Makes the appropriate CSS imports for highlighting #}
 {% macro highlight_css() %}
     {# Load the highlighting theme css #}
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/styles/github.min.css"
-        type="text/css" media="all" />
+    <script>
+        // Choose which highlight.js theme to load based on the user theme
+        var stylesheet;
+        switch(document.documentElement.dataset.theme) {
+            case "dark":
+            case "ayu":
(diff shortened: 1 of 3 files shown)


templates/macros.html
```diff
@@ -22,8 +22,29 @@
...
-        type="text/css" media="all" />
+    <script>
+        // Choose which highlight.js theme to load based on the user theme
+        var stylesheet;
+        switch(document.documentElement.dataset.theme) {
+            case "dark":
+            case "ayu":
+                stylesheet = "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.4.0/styles/dark.min.css";
+                break;
+            case "light":
```
Nemo157, line 33: A `switch`/`case` in javascript 😰 

I think this either needs `theme.js` to set the theme to `light` explicitly when there is none, or include `null` in this case (and I have no idea if `case null` works).

jyn514:
Dark mode: 
image
Ayu: 
image
Light (stays the same, but just in case): 
image

Readme (Ayu): 
image
Readme (Dark): 
image

Source (Ayu): 
image

So I think what I'm seeing is it would be nice to have another theme for Ayu mode, but I'm not sure if highlight.js has one. @Cldfire do you happen to know of one that matches?

templates/macros.html
jyn514: The `default` was intentional, in case someone adds a theme without updating this switch it will give an error and they'll have to update it. I'm ok with defaulting to 'light' but `case null` doesn't seem right.

Nemo157: This isn't for unknown cases, it's for when you initially load docs.rs before rustdoc has stored a theme to use. If you check in a private browser it doesn't set it to `light`.

Actually, it sets it to `data-theme="null"`, so `case "null":` would work 🤣

jyn514: Oh! Good catch, thanks :) I hadn't thought about private mode.

Cldfire:
> So I think what I'm seeing is it would be nice to have another theme for Ayu mode, but I'm not sure if highlight.js has one. @Cldfire do you happen to know of one that matches?

There isn't a pre-existing one that I'm aware of, but I'd be happy to put one together tomorrow night, it won't take long :smile:

jyn514:
Oh wow! That would be awesome if you have time :heart: I was just hoping for something somewhat close.

jyn514:
I think it makes sense to merge this as is and then work on the custom theme in a follow-up - this is already a big improvement even on ayu and it would make it easier for @Cldfire to PR against the main codebase instead of my fork I think.

@Nemo157 does this look good to you?

Nemo157 (approved):

PR #1226, opened 2020-12-28
Add ayu-highlight.css

Cldfire:
Follow-up to #1221.

I began putting a theme together and thought "hmm, I feel like I've done this before." Turns out... I did! https://github.com/rust-lang/mdBook/blob/master/src/theme/ayu-highlight.css

I copied that CSS into `static/ayu-highlight.css` and added an exception for it to the `.gitignore`; if there's a better way to include it, let me know.

Here's what it looks like:

image

Note that a nice, easy way to see what it looks like for a large variety of input is to go to https://highlightjs.org/static/demo/ and replace the default stylesheet with the contents of whatever theme you'd like to test:

<details>
<summary>Example</summary>

image

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ target
 *.css
 *.css.map
 !vendor/**/*.css
+!static/ayu-highlight.css
 .sass-cache
 .vagrant
 .rustwide
--- a/static/ayu-highlight.css
+++ b/static/ayu-highlight.css
@@ -1,0 +1,79 @@
+/*
+Based off of the Ayu theme
+Original by Dempfi (https://github.com/dempfi/ayu)
+*/
+
+.hljs {
+    display: block;
+    overflow-x: auto;
+    background: #191f26;
+    color: #e6e1cf;
+    padding: 0.5em;
+}
+
+.hljs-comment,
+.hljs-quote {
+    color: #5c6773;
+    font-style: italic;
+}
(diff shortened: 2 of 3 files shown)


jyn514:
> I copied that CSS into static/ayu-highlight.css and added an exception for it to the .gitignore; if there's a better way to include it, let me know.

I think that's fine; ideally it would get upstreamed to highlight.js but since it's a single CSS file maintaining it in tree doesn't seem like a burden. Doing fancy git things with subtrees or submodules doesn't seem worth it (and you'd have to have a repo for a single file, which also seems silly).

This looks really good in source mode, which was pretty lacking before. Great job!

Before: 
image

After: image

The one thing I would change is that I think you made `///` dark green for rustdoc - could you add the same color here?

Cldfire:
> ideally it would get upstreamed to highlight.js

I agree, that would be neat to do someday. First the theme would need some cleanup and modernization; ayu has changed in the years since I made it for mdbook.

> The one thing I would change is that I think you made /// dark green for rustdoc - could you add the same color here?

Unfortunately I don't believe I can. highlight.js doesn't differentiate between doc comments and normal comments when highlighting rust code; they both get the `hljs-comment` class. This means they have to share a style.

jyn514:
Oh, that's unfortunate :( ok, this is still great then :)

Issue #1225, opened 2020-12-28
'I am feeling lucky' gives an internal error


neetdai:
when i use https://docs.rs/releases/search?query=&i-am-feeling-lucky=0, will found

```
    db error: ERROR: column "github_stars" does not exist 
```

jyn514:
This is a regression from https://github.com/rust-lang/docs.rs/pull/1208. cc @pietroalbini, do you have time to do the follow-up work?

pietroalbini:
Opened https://github.com/rust-lang/docs.rs/pull/1227 to fix this.

PR #1227, opened 2020-12-28
Fix "I am feeling lucky" after the GraphQL switch

pietroalbini:
Fixes https://github.com/rust-lang/docs.rs/issues/1225

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -530,7 +530,9 @@ pub fn search_handler(req: &mut Request) -> IronResult<Response> {
                      FROM crates
                      INNER JOIN releases
                          ON crates.latest_version_id = releases.id
-                     WHERE github_stars >= 100 AND rustdoc_status = true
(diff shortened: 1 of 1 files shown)


src/web/releases.rs
```diff
@@ -530,7 +530,9 @@ pub fn search_handler(req: &mut Request) -> IronResult<Response> {
                      FROM crates
                      INNER JOIN releases
                          ON crates.latest_version_id = releases.id
-                     WHERE github_stars >= 100 AND rustdoc_status = true
+                     LEFT JOIN github_repos
+                         ON releases.github_repo = github_repos.id
+                     WHERE github_repos.stars >= 100 AND rustdoc_status = true
```
jyn514, line 535: The `unwrap()` on line 540 panics if there are no results, but I think that's an existing bug that shouldn't happen in production because we have lots of crates with over 100 stars. Looks good otherwise.

PR #1233, opened 2021-01-01
Fix sidebar rendering bugs

### LINKED ISSUE #1121: Mobile: Opening rustdoc hamburger menu overlaps rustdoc header
Steps to reproduce:

1. On a mobile device (or in Chrome's mobile simulation mode), visit a crate page, for instance https://docs.rs/ureq/1.5.1/ureq/index.html.
2. Tap the hamburger menu in the top left.

Expected result:

Rustdoc's crate header is not obscured by the menu.

Actual result:

The header is obscured by the menu.

https://docs.rs/ureq/1.5.1/ureq/index.html

image

vs https://doc.rust-lang.org/std/vec/struct.Vec.html

image

(noticed while writing up #1120 )

arusahni:
This corrects two bugs with the sidebar's presentation:

* The CSS selector to reposition the top offset of the rustdoc sidebar menu was not being applied to the element. It also updates the magic number to be less magical :sparkles: (Fixes #1121)
* Due to the padding being applied to the rustdoc container, the sidebar nav's sizing and position was offset, resulting in some jank. This addresses it.
  * Before: Before
  * After: After

The fixes are grouped by commit so they can be separately reviewed.

Thanks, and Happy New Year! :fireworks: :crossed_fingers:

### DIFF
--- a/templates/style/rustdoc.scss
+++ b/templates/style/rustdoc.scss
@@ -29,8 +29,9 @@ body.blur {
 
 // rustdoc overrides
 div.rustdoc {
+    $padding-x: 15px;
     font-family: $font-family-serif;
-    padding: 10px 15px 20px 15px;
+    padding: 10px $padding-x 20px;
     position: relative;
 
     @media (max-width: 700px) {
@@ -47,8 +48,13 @@ div.rustdoc {
         }
 
         @media (max-width: 700px) {
+            margin-left: -2 * $padding-x; // offset the additional padding added by the parent containers
+            width: calc(100% + #{4 * $padding-x});
+
             &.mobile {
                 top: $top-navbar-height;
+                margin-left: 0; // since the sidebar is now fixed position, remove the padding workaround
+                width: 100%;
 
                 .sidebar-elems.show-it {
                     top: 45px + $top-navbar-height;
--- a/templates/style/rustdoc.scss
+++ b/templates/style/rustdoc.scss
@@ -50,8 +50,8 @@ div.rustdoc {
             &.mobile {
                 top: $top-navbar-height;
 
-                .sidebar-elems .show-it {
-                    top: 77px;
+                .sidebar-elems.show-it {
+                    top: 45px + $top-navbar-height;
                 }
 
                 #sidebar-filler {


jyn514:
Please drop the clippy commits, those were fixed in https://github.com/rust-lang/docs.rs/pull/1235.

templates/style/rustdoc.scss
```diff
@@ -50,8 +50,8 @@ div.rustdoc {
             &.mobile {
                 top: $top-navbar-height;
 
-                .sidebar-elems .show-it {
-                    top: 77px;
+                .sidebar-elems.show-it {
+                    top: 45px + $top-navbar-height;
```
jyn514, line 59: Where does 45px come from?

arusahni:
Clippy commits removed, and this PR has been rebased on master.

templates/style/rustdoc.scss
arusahni: That's the height of the rustdocs sidebar nav.

jyn514:
> Due to the padding being applied to the rustdoc container, the sidebar nav's sizing and position was offset, resulting in some jank. This addresses it. 

I still see some issues with this (I have it open on desktop, taking up half of the screen).

[Before](https://user-images.githubusercontent.com/23638587/103442069-df108b80-4c20-11eb-978d-54c9a8d19731.png)
[After](https://user-images.githubusercontent.com/23638587/103442072-e33ca900-4c20-11eb-8640-ee55b7c4cf0e.png)

Notice how 'Regex' and 'Version' move over to the right when you open the sidebar. I'm ok with fixing that in a follow-up, but do you think it would be easy to fix here?

arusahni:
I'm in favor of getting it sorted out in this PR so we don't have to revisit it. I'll try to replicate and address. :+1:

arusahni:
Alright - this was a fun difference in how Firefox (which I was developing and testing in) and Chrome handled scrollbars and overflow. I've pushed an update that still results in a jump, but I confirmed locally that the jump is also present with rustdocs.  [Here's a comparison screencap](https://user-images.githubusercontent.com/139487/103443127-06b82180-4c2a-11eb-8204-37463aa13803.mp4). Given consistency is our goal, this feels like the most-correct course of action.

jyn514:
I agree it's better to fix this in rustdoc than special case docs.rs. If you're interested in working on that let me know :)

Thanks for working on this!

arusahni:
@jyn514 I'd love to tackle the root cause in rustdoc! :hammer_and_wrench:

jyn514:
@arusahni awesome! I'd take a look around https://github.com/rust-lang/rust/blob/master/src/librustdoc/html/static/rustdoc.css. There's instructions for building rustdoc at https://rustc-dev-guide.rust-lang.org/rustdoc.html. Feel free to ask for help [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc) if you need it :)

PR #1236, opened 2021-01-02
Validate requested crate version when rendering the build and feature views

### LINKED ISSUE #1206: "Internal Server Error" for /builds and /features when release does not exist
If I visit `https://docs.rs/crate/{name}/{version}/builds` or `https://docs.rs/crate/{name}/{version}/features` with an invalid combination of `name` and `version` (i.e. a release that doesn't exist) I get an `Internal server error`. I only got there by manually typing in a version number for a crate that never existed.

## Expected Behavior

An error page with `The requested crate does not exist`.

## Actual Behavior

An error page with `Internal Server Error`

## Example

<https://docs.rs/crate/serde/2.0.0/builds>

## Probable Cause

The `crate_details` handler wraps most of its logic into a `match_version` call that early-returns `Err(Nope::CrateNotFound)` when no crate is found. That turns into an IronError and produces the expected error page while the rest of the handler can assume that there is an actual release to work with.

The `features` handler does not use such a wrapper and just fails on an assertion in [/src/web/features.rs](https://github.com/rust-lang/docs.rs/blob/b728426b65e67dc84c568da6bab4ebe59c4ad9cc/src/web/features.rs#L41). This also has the effect that
`https://docs.rs/crate/serde/1.0.118`, `https://docs.rs/crate/serde/latest` and `https://docs.rs/crate/serde/1.0.118/features` work, but `https://docs.rs/crate/serde/latest/features` does not.

arusahni:
This brings the functionality of the crate `/builds` and `/features` routes in-line with the rest of the crate-specific views for handling the requested version.

### Testing
(assuming an environment bootstrapped according to the README)

1. Visit http://localhost:3000/crate/regex/1.3.1/builds
2. Verify the route resolves as it used to
3. Visit http://localhost:3000/crate/regex/latest/builds
4. Verify that the the view is redirected to the latest version of the regex crate
5. Visit http://localhost:3000/crate/regex/10.2.1/builds
6. Verify that a crate not found error is supplied
7. Do steps 1-6, but with `/features`

Fixes #1206

### DIFF
--- a/src/web/builds.rs
+++ b/src/web/builds.rs
@@ -1,3 +1,4 @@
+use super::{match_version, redirect_base, MatchSemver};
 use crate::{
     db::Pool,
     docbuilder::Limits,
@@ -9,7 +10,7 @@ use iron::{
     headers::{
         AccessControlAllowOrigin, CacheControl, CacheDirective, ContentType, Expires, HttpDate,
     },
-    status, IronResult, Request, Response,
+    status, IronResult, Request, Response, Url,
 };
 use router::Router;
 use serde::Serialize;
@@ -37,10 +38,29 @@ impl_webpage! {
 pub fn build_list_handler(req: &mut Request) -> IronResult<Response> {
     let router = extension!(req, Router);
     let name = cexpect!(req, router.find("name"));
-    let version = cexpect!(req, router.find("version"));
+    let req_version = router.find("version");
 
     let mut conn = extension!(req, Pool).get()?;
     let limits = ctry!(req, Limits::for_crate(&mut conn, name));
+
+    let version =
+        match match_version(&mut conn, name, req_version).and_then(|m| m.assume_exact())? {
+            MatchSemver::Exact((version, _)) => version,
+
+            MatchSemver::Semver((version, _)) => {
+                let url = ctry!(
+                    req,
+                    Url::parse(&format!(
+                        "{}/crate/{}/{}/builds",
+                        redirect_base(req),
+                        name,
+                        version
+                    )),
+                );
+
+                return Ok(super::redirect(url));
+            }
+        };
 
     let query = ctry!(
         req,
@@ -102,6 +122,7 @@ mod tests {
     use crate::test::{wrapper, FakeBuild};
     use chrono::{DateTime, Duration, Utc};
     use kuchiki::traits::TendrilSink;
+    use reqwest::StatusCode;
 
     #[test]
     fn build_list() {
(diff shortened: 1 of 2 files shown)


src/web/builds.rs
```diff
@@ -37,63 +38,82 @@ impl_webpage! {
...
-                builds.build_time
-             FROM builds
-             INNER JOIN releases ON releases.id = builds.rid
-             INNER JOIN crates ON releases.crate_id = crates.id
-             WHERE crates.name = $1 AND releases.version = $2
-             ORDER BY id DESC",
-            &[&name, &version]
-        )
-    );
+    match match_version(&mut conn, name, req_version).and_then(|m| m.assume_exact())? {
```
Nemo157, line 45: This might be simpler as an early-return to avoid nesting everything

```rust
let version = match match_version(&mut conn, name, req_version).and_then(|m| m.assume_exact())? {
    MatchSemver::Exact((version, _)) => version,
    MatchSemver::Semver((version, _)) => {
        let url = ctry!(
            req,
            Url::parse(&format!(
                "{}/crate/{}/{}/builds",
                redirect_base(req),
                name,
                version
            )),
        );

        return Ok(super::redirect(url));
    }
};
```

arusahni: :facepalm: I like that a lot better. I have updated the PR to use early returns. The diff is much cleaner, now.

src/web/builds.rs
```diff
@@ -37,11 +38,30 @@ impl_webpage! {
 pub fn build_list_handler(req: &mut Request) -> IronResult<Response> {
     let router = extension!(req, Router);
     let name = cexpect!(req, router.find("name"));
-    let version = cexpect!(req, router.find("version"));
+    let req_version = router.find("version");
```
jyn514, line 41: I'm confused: Where does `find` come from? I don't see it on https://docs.rs/router/0.6.0/router/struct.Router.html. I do see a find() method on https://docs.rs/router/0.6.0/router/struct.Params.html#method.find, is that it?

src/web/builds.rs
```diff
@@ -37,11 +38,30 @@ impl_webpage! {
...
+                        "{}/crate/{}/{}/builds",
+                        redirect_base(req),
+                        name,
+                        version
+                    )),
+                );
+
+                return Ok(super::redirect(url));
+            }
+        };
```
jyn514, line 63: This is repeated a few places in the code base now. At some point I'd like to reduce the duplication, but it doesn't need to block this PR.

jyn514 (commented):
Can you add a few tests for the new behavior? You can find examples at the bottom of `web/builds.rs` and `web/features.rs`. See `web/mod.rs` for ideas of tests to add (but that has a lot more than you'll need here, just testing that it works and doesn't crash is fine).

src/web/builds.rs
arusahni: Yeah, this is a bit of magic introduced by Iron's router middleware (accessed via the `extension!` macro). `router::Params` is actually a re-exported [`route_recognizer::Params`](https://docs.rs/route-recognizer/0.3.0/route_recognizer/struct.Params.html).

arusahni:
@jyn514 I've pushed an update that adds unit tests to validate the behavior for both routes.

src/web/builds.rs
Nemo157: Specifically it's [this `impl Key for Router { type Value = Params; }`](https://docs.rs/router/0.6.0/router/struct.Router.html#impl-Key) that means `extension!(req, Router): Params`..

Nemo157 (approved):

PR #1241, opened 2021-01-06
Add print to show on which address the server has been started

GuillaumeGomez:
It happens to me everytime: I forget which port is being used when I start the server... So instead of looking in the code every time, I think it'd be better to just display it.

It looks like this:

```
$ cargo run -- start-web-server --reload-templates
   Compiling docs-rs v0.6.0 (/home/imperio/rust/docs.rs)
    Finished dev [unoptimized + debuginfo] target(s) in 11.16s
     Running `target/debug/cratesfyi start-web-server --reload-templates`
2021/01/06 17:05:04 [INFO] cratesfyi: Starting server on `0.0.0.0:3000`... (reload templates: true)
```

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -117,6 +117,11 @@ impl CommandLine {
                 reload_templates,
             } => {
                 // Blocks indefinitely
+                log::info!(
+                    "Starting server on `{}`... (reload templates: {})",
+                    socket_addr,
+                    reload_templates,
+                );
                 let _ = Server::start(Some(&socket_addr), reload_templates, &ctx)?;
             }
             Self::Daemon {


Nemo157 (approved):
:shipit: (though, could be a `log::info!` line)

GuillaumeGomez:
As you prefer!

pietroalbini:
The message is already posted through logging though.

GuillaumeGomez:
Weird... Why it's not showing up in my case? Maybe it's only print in one specific config?

Nemo157:
If you override the `DOCSRS_LOG` setting to not show `info` level it won't show, but the default should show it.

GuillaumeGomez:
I don't override anything. So I guess there is an issue here. :)

jyn514:
@GuillaumeGomez please paste your .env file. If you didn't set one up, that's why you aren't seeing logging.

GuillaumeGomez:
Ah you're right:

```
CRATESFYI_PREFIX=ignored/cratesfyi-prefix
CRATESFYI_DATABASE_URL=postgresql://cratesfyi:password@localhost:15432
DOCSRS_LOG=cratesfyi,rustwide=info
AWS_ACCESS_KEY_ID=cratesfyi
AWS_SECRET_ACCESS_KEY=secret_key
S3_ENDPOINT=http://localhost:9000
```

Nemo157:
> @GuillaumeGomez please paste your .env file. If you didn't set one up, that's why you aren't seeing logging.

if you don't set one up it defaults everything to info level.

jyn514:
I think this can be closed then.

PR #1242, opened 2021-01-06
Make it more obvious that this is the repository link

GuillaumeGomez:
Before:

Screenshot from 2021-01-06 15-48-06

After:

Screenshot from 2021-01-06 21-09-31

And the menu "repository link" will always look like this:

Screenshot from 2021-01-13 14-09-47

What do you think?

### DIFF
--- a/src/web/crate_details.rs
+++ b/src/web/crate_details.rs
@@ -53,6 +53,7 @@ struct GitHubMetadata {
     stars: i32,
     forks: i32,
     issues: i32,
+    name: Option<String>,
 }
 
 fn optional_markdown<S>(markdown: &Option<String>, serializer: S) -> Result<S::Ok, S::Error>
@@ -102,6 +103,7 @@ impl CrateDetails {
                 github_repos.stars AS github_stars,
                 github_repos.forks AS github_forks,
                 github_repos.issues AS github_issues,
+                github_repos.name AS github_name,
                 releases.is_library,
                 releases.yanked,
                 releases.doc_targets,
@@ -150,6 +152,7 @@ impl CrateDetails {
                 issues: krate.get("github_issues"),
                 stars: krate.get("github_stars"),
                 forks: krate.get("github_forks"),
+                name: krate.get("github_name"),
             })
         } else {
             None
--- a/src/web/page/templates.rs
+++ b/src/web/page/templates.rs
@@ -301,9 +301,9 @@ impl tera::Filter for IconType {
         let mut aria_hidden = true;
         let icon_name = tera::escape_html(value.as_str().expect("Icons only take strings"));
         let class = if args.get("fw").and_then(|fw| fw.as_bool()).unwrap_or(false) {
-            "class=\"fa-svg fa-svg-fw\""
+            "fa-svg fa-svg-fw"
         } else {
-            "class=\"fa-svg\""
+            "fa-svg"
         };
         let aria_label = args
             .get("aria-label")
@@ -332,9 +332,12 @@ impl tera::Filter for IconType {
         let icon_file_string = font_awesome_as_a_crate::svg(type_, &icon_name[..]).unwrap_or("");
 
         let icon = format!(
-            r#"<span {class} aria-hidden="{aria_hidden}"{aria_label}{id}>{icon_file_string}</span>"#,
+            "<span \
+                class=\"{class} {class_extra}\" \
+                aria-hidden=\"{aria_hidden}\"{aria_label}{id}>{icon_file_string}</span>",
             icon_file_string = icon_file_string,
             class = class,
+            class_extra = args.get("extra").and_then(|s| s.as_str()).unwrap_or(""),
             aria_hidden = aria_hidden,
             aria_label = aria_label,
             id = id,
--- a/templates/crate/details.html
+++ b/templates/crate/details.html
@@ -75,12 +75,21 @@
                                     {# If the repo link is for github, show some github stats #}
                                     {# TODO: add support for hosts besides github (#35) #}
                                     {%- if details.github_metadata -%}
-                                        {{ "star" | fas(fw=true) }} {{ details.github_metadata.stars }}
+                                        {{ "github" | fab(fw=true) }}
+                                        {% if details.github_metadata.name %}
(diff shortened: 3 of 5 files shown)


pietroalbini:
Hmm, I would probably do something like this.

GuillaumeGomez:
Looks better indeed!

templates/crate/details.html
```diff
@@ -75,15 +75,12 @@
...
-                                        {{ "exclamation-circle" | fas(fw=true) }} {{ details.github_metadata.issues }}
-
-                                    {# If the repo link is unknown, just show a normal link #}
-                                    {%- else -%}
-                                        {{ "code-branch" | fas(fw=true) }} Repository
+                                        {{ "exclamation-circle" | fas(fw=true) }} {{ details.github_metadata.issues }}<br>
                                     {%- endif -%}
+
+                                    {{ "code-branch" | fas(fw=true) }} Repository
                                 </a>
```
pietroalbini, line 84: This code is duplicated two times: one in the dropdown of rustdoc pages and one here. It would probably be good to extract it into a macro, so changes can be shared between the two places.

GuillaumeGomez: Oh wow, didn't even pay attention to that. I'll do that too then.

GuillaumeGomez: Ok, tested and the result is pretty bad. It gets even worse on long names. So in the end, I did the opposite: I simplified it like when you don't have github information.

src/web/crate_details.rs
```diff
@@ -145,11 +146,15 @@ impl CrateDetails {
...
 
+        let repository_url: Option<String> = krate.get("repository_url");
         let github_metadata = if krate.get::<_, Option<String>>("github_repo").is_some() {
             Some(GitHubMetadata {
                 issues: krate.get("github_issues"),
                 stars: krate.get("github_stars"),
                 forks: krate.get("github_forks"),
+                name: repository_url
+                    .as_ref()
+                    .and_then(|u| u.splitn(2, "/github.com/").nth(1).map(|s| s.to_owned())),
```
pietroalbini, line 157: The up to date repository name is already present in the `github_repos` table, we should fetch it from there.

templates/rustdoc/topbar.html
```diff
@@ -62,10 +62,7 @@
                             {%- if krate.github_metadata -%}
                                 <li class="pure-menu-item">
                                     <a href="{{ krate.repository_url }}" class="pure-menu-link">
-                                        {{ "github" | fab(fw=true) }}
-                                        {{ "star" | fas(fw=true) }} {{ krate.github_metadata.stars }}
-                                        {{ "code-branch" | fas(fw=true) }} {{ krate.github_metadata.forks }}
-                                        {{ "exclamation-circle" | fas(fw=true) }} {{ krate.github_metadata.issues }}
+                                        {{ "code-branch" | fas(fw=true) }} Repository
```
pietroalbini, line 65: Now that it's more clear I would still show the GitHub metadata.

templates/crate/details.html
pietroalbini: Missed this comment... how did it look?

GuillaumeGomez: The github logo is on its own line and the name is on a second one. And the repository info are just too much. (Don't have the screenshot anymore, urg)

pietroalbini: We can use some rules to have ellipsis on the repository name if it's too long:

```css
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

GuillaumeGomez: True, but in my opinion, this still remains a bit too much.

jyn514: Personally I would be ok with just showing 'Repository'. I've heard lots of feedback saying that it's not very obvious right now.

If you want to show the stars/PRs/issues underneath that seems fine, but showing the repository name too seems like overkill.

templates/rustdoc/topbar.html
GuillaumeGomez: And I answer the same: it looks kinda weird in the menu which is already quite "heavy". Just having "repository" looks better in my opinion. If people want more information, the docs.rs crate page is a much better place.

src/web/crate_details.rs
GuillaumeGomez: Indeed, thanks!

GuillaumeGomez:
Updated!

src/web/page/templates.rs
```diff
@@ -332,9 +332,12 @@ impl tera::Filter for IconType {
         let icon_file_string = font_awesome_as_a_crate::svg(type_, &icon_name[..]).unwrap_or("");
 
         let icon = format!(
-            r#"<span {class} aria-hidden="{aria_hidden}"{aria_label}{id}>{icon_file_string}</span>"#,
+            "<span \
+                class=\"{class} {class_extra}\" \
```
jyn514, line 336: What happens for `class=" "`? Does the browser just ignore it?

jyn514 (approved):
This looks good to me. @pietroalbini what do you think?

src/web/page/templates.rs
Nemo157: yep

GuillaumeGomez: It's just an attribute with no values. :)

templates/rustdoc/topbar.html
GuillaumeGomez: (I added a screenshot with the current menu look to show what it looks like 😃)

pietroalbini:
I still prefer having the metadata everywhere, as having it shows me at a glance what the maintenance status of the crate might be. Still, it seems like I'm the only holdout, so feel free to merge anyway.

GuillaumeGomez:
Then let's merge! I think overall that reducing the size of the menu is the good way forward.

lnicola:
Just to make sure, is the repository info (stars etc.) supposed to be gone now? This doesn't match the screenshots above:

image

GuillaumeGomez:
Yes, the stars, forks and name info have been replaced by "Repository". Although, not sure which screenshot you're referring to.

lnicola:
Ah, I didn't see the third image in the first post here.

PR #1244, opened 2021-01-08
Note that docker service must be running in README.

valentinewallace:
I'm not super familiar with docker, so this would've saved me a bit of time. Feel free to close if it seems too obvious though. Thank you!


jyn514 (approved):
Looks useful, thanks!

PR #1252, opened 2021-01-18
Use a vendored copy of highlight.js

pietroalbini:
This PR vendors the JavaScript and CSS of highlight.js into docs.rs instead of fetching it from cdnjs, which will make it easier to implement a Content Security Policy in the future. Two other changes are present in this PR:

* The `/-/static` handler has been updated to also support loading files from subdirectories (like `/-/static/highlightjs/styles/dark.min.css`). The check to prevent reading arbitrary files has been updated accordingly.
* Pure CSS has been removed from the list of directories served by `/-/static`, as it's now included in `vendored.css`.


src/web/routes.rs
```diff
@@ -29,7 +29,8 @@ pub(super) fn build_routes() -> Routes {
         PermanentRedirect("/-/static/opensearch.xml"),
     );
 
-    routes.static_resource("/-/static/:file", super::statics::static_handler);
+    routes.static_resource("/-/static/:single", super::statics::static_handler);
+    routes.static_resource("/-/static/*", super::statics::static_handler);
```
jyn514, line 33: Can you use `/-/static/*` for both of these? I would expect it to work for single files too.

src/web/statics.rs
```diff
@@ -16,10 +15,11 @@ const RUSTDOC_CSS: &str = include_str!(concat!(env!("OUT_DIR"), "/rustdoc.css"))
 const STATIC_SEARCH_PATHS: &[&str] = &["vendor/pure-css/css", "static"];
 
 pub(crate) fn static_handler(req: &mut Request) -> IronResult<Response> {
-    let router = extension!(req, Router);
-    let file = cexpect!(req, router.find("file"));
+    let mut file = req.url.path();
+    file.drain(..2).for_each(std::mem::drop);
+    let file = file.join("/");
```
jyn514, line 20: What does the `/` do?

src/web/statics.rs
```diff
@@ -28,15 +28,21 @@ pub(crate) fn static_handler(req: &mut Request) -> IronResult<Response> {
...
     // Find the first path that actually exists
     let path = STATIC_SEARCH_PATHS
         .iter()
-        .map(|p| Path::new(p).join(file))
+        .filter_map(|root| {
+            let path = Path::new(root).join(file);
+
+            // Prevent accessing static files outside the root. This could happen if the path
+            // contains `/` or `..`. The check doesn't outright prevent those strings to be present
+            // to allow accessing files in subdirectories.
```
jyn514, line 39: Are there tests for this?

src/web/statics.rs
```diff
@@ -252,7 +252,7 @@ mod tests {
         wrapper(|env| {
             let web = env.frontend();
 
-            let files = &[("pure-min.css", "text/css")];
+            let files = &[("highlightjs/styles/dark.min.css", "text/css")];
```
jyn514, line 255: Maybe test both files instead of only highlight.js?

src/web/routes.rs
pietroalbini: I'm not sure why but just `/-/static/*` did not work when I tried it :/

I can check again...

src/web/statics.rs
pietroalbini: `/-/static/pure-min.css` has been removed in the last commit since it was not used anywhere.

src/web/statics.rs
pietroalbini: `req.url.path()` is a `Vec<&str> of path segments, so the join turns it back into a path.

src/web/routes.rs
Nemo157: It's probably hitting the bugs around priority of the different specificity levels in the old version of `route-recogniser`

src/web/statics.rs
pietroalbini: Well, it turns out there *is already* a test about this (`directory_traversal`), but the test itself was broken :sweat_smile: 

Fixed both my code (I forgot to canonicalize) and the test itself.

src/web/routes.rs
pietroalbini: Chatted with @Nemo157 on this a bit on Discord. It's a bug of the router library, and this workaround seems to work.

pietroalbini:
All review comments have been addressed.

src/web/statics.rs
jyn514: > Fixed both my code (I forgot to canonicalize) and the test itself.

Did you forget to push your changes?

pietroalbini: Woops, yes, pushed!

src/web/statics.rs
```diff
@@ -23,29 +23,37 @@ pub(crate) fn static_handler(req: &mut Request) -> IronResult<Response> {
...
                 None
             }
         })
-        .find(|p| p.exists())
+        .next()
         .ok_or(Nope::ResourceNotFound)?;
-    let contents = ctry!(req, fs::read(&path));
+    let contents = fs::read(&path).map_err(|e| {
+        log::error!("failed to read static file {}: {}", path.display(), e);
+        Nope::InternalServerError
```
jyn514, line 55: This doesn't look correct; it will give internal server errors when you try to request a file that doesn't exist. Can you either

a) add back the `exists()` call, or
b) check if `e.kind() == ErrorKind::NotFound`

and either way add a test?

src/web/statics.rs
```diff
@@ -272,23 +280,22 @@ mod tests {
...
+        ];
+
+        for path in PATHS {
+            // This doesn't test an actual web request as the web framework used at the time of
+            // writing this test (iron 0.5) already resolves `..` before calling any handler.
+            //
+            // Still, the test ensures the underlying function called by the request handler to
+            // serve the file also includes protection for path traversal, in the event we switch
+            // to a framework that doesn't include builtin protection in the future.
+            assert!(serve_file(path).is_err(), "{} was served", path);
```
jyn514, line 298: Can you check that these are 404s instead of internal errors?

src/web/statics.rs
pietroalbini: The `exists()` call is already there, on line 35.

src/web/statics.rs
pietroalbini: Pushed a test!

jyn514:
Thanks!

PR #1255, opened 2021-01-21
Switch to chart.js for the release activity page

### LINKED ISSUE #921: self-host all static content
Currently we use some libraries hosted on cdnjs (run by Cloudflare)

https://github.com/rust-lang/docs.rs/blob/664577aa55756d77124efc29e891ccbe44cf7897/templates/base.html#L12-L17

This has [caused issues at least once](https://discord.com/channels/442252698964721669/541978667522195476/738691962642890793). To reduce our points of failure it might be worth self-hosting these files.

pietroalbini:
This PR switches the charts in the [`/releases/activity`](https://docs.rs/releases/activity) page from [Highcharts](https://www.highcharts.com/) to [chart.js](https://www.chartjs.org/), improving multiple things:

* Highcharts is proprietary, and I'm not actually sure docs.rs fits into the cases where we don't need a commercial license for it.
* Highcharts doesn't fit well with CSP, requiring `style-src 'unsafe-inline'` to work.
* This removes the last dependency we fetch from cdnjs, so this fixes #921.

Rendered example of the page with chart.js:

Screenshot_2021-01-21 Releases - Docs rs


Nemo157 (approved):
Code LGTM. I haven't tested it locally and I assume we have no tests covering this page, but I trust your screenshot.

Nemo157:
Test failure is probably #1251 making the test flaky, probably needs reworking to inject more test crates to ensure there is a result.

pietroalbini:
If anyone wants to test it locally I recommend running this query to fill some data in:

```sql
UPDATE config SET value = '{"counts":[233,281,431,564,390,188,230,291,320,352,458,332,349,338,358,410,467,579,841,397,565,298,407,532,511,456,484,426,326,295],"dates":["19 Dec","20 Dec","21 Dec","22 Dec","23 Dec","24 Dec","25 Dec","26 Dec","27 Dec","28 Dec","29 Dec","30 Dec","31 Dec","01 Jan","02 Jan","03 Jan","04 Jan","05 Jan","06 Jan","07 Jan","08 Jan","09 Jan","10 Jan","11 Jan","12 Jan","13 Jan","14 Jan","15 Jan","16 Jan","17 Jan"],"failures":[16,9,6,7,29,8,6,8,4,13,18,12,11,4,19,18,28,44,23,17,14,8,10,28,19,14,7,13,8,17]}' WHERE name = 'release_activity';
```

> I assume we have no tests covering this page

Yeah, we don't really have a way to do UI tests as of today.

templates/releases/activity.html
```diff
@@ -9,51 +9,63 @@
...
-            },
-            title: {
-                text: 'Monthly release activity',
-                x: -20 //center
-            },
-            xAxis: {
-                categories: [
+    <script type="text/javascript">
+        // We're including the CSS file manually to avoid issues with the CSP.
+        Chart.platform.disableCSSInjection = true;
```
jyn514, line 25: By default this will load the CSS from the internet?

jyn514 (approved):
LGTM but I would like to deploy before merging this, there are quite a few PRs queued up to be deployed.

templates/releases/activity.html
pietroalbini: No, without the option it will inject the styling directly into the DOM. [Documentation](https://www.chartjs.org/docs/latest/getting-started/integration.html#content-security-policy)

PR #1256, opened 2021-01-21
Add JSON output for badges

### LINKED ISSUE #170: /crate/:crate/latest/builds.json strips the .json suffix

coolshaurya:
Resolves #170. 

The JSON matches [the shields.io endpoint](https://shields.io/endpoint).


coolshaurya:
Hmm, I'm not sure if this is a duplicate of #809 😄

jyn514:
I have questions about the use case I'll ask on the issue. I don't think it makes sense to review the implementation until the design issues are settled.

jyn514:
As per https://github.com/rust-lang/docs.rs/issues/170#issuecomment-764837701, I would prefer we fixed the `latest` bug rather than introducing a completely new API.

PR #1259, opened 2021-01-22
Improve topbar crate dropdown menu by removing uneeded information

GuillaumeGomez:
I always found the link to go to the crate page very misleading because you don't expect a description to be this link. Instead, I removed this description from the menu completely (I think the menu is already very loaded) and made it clear what the link was about.

Screenshot from 2021-01-22 15-01-20

### DIFF
--- a/templates/rustdoc/topbar.html
+++ b/templates/rustdoc/topbar.html
@@ -23,7 +23,7 @@
                     </li>
 
                     <li class="pure-menu-item">
-                        <a href="{{ crate_url | safe }}" class="pure-menu-link description">
+                        <a href="{{ crate_url | safe }}" class="pure-menu-link description" title="See {{ krate.name }} in docs.rs">
                             {{ "cube" | fas(fw=true) }} Docs.rs crate page
                         </a>
                     </li>
@@ -86,12 +86,6 @@
                                 <a href="{{ crate_url | safe }}/source/" title="Browse source of {{ metadata.name }}-{{ metadata.version }}" class="pure-menu-link">
                                     {{ "folder-open" | fas(fw=true) }}
                                     <span class="title">Source</span>
-                                </a>
-                            </li>
-
-                            <li class="pure-menu-item">
-                                <a href="{{ crate_url | safe }}" class="pure-menu-link" title="See {{ krate.name }} in docs.rs">
-                                    {{ "cube" | fas(fw=true) }} More information
                                 </a>
                             </li>
                         </ul>
--- a/templates/rustdoc/topbar.html
+++ b/templates/rustdoc/topbar.html
@@ -23,8 +23,8 @@
                     </li>
 
                     <li class="pure-menu-item">
-                        <a href="{{ crate_url | safe }}" class="pure-menu-link" class="description">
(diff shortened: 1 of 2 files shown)


jyn514:
@GuillaumeGomez instead of adding a new link text that's always the same, what do you think about making 'Regex 1.3.1' a link?

GuillaumeGomez:
Because it's not "obvious" when reading it. The problem I had with most of the original items is that I needed to look at the `href` to know what they were about.

jyn514:
@GuillaumeGomez this already has a link to /crate under 'more information' on the sidebar:

image

So I don't think it makes sense to add another link, and I kind of like having the description. I would be ok with making the title a link instead of the description, but otherwise I don't see much point to this.

GuillaumeGomez:
I didn't even realize there was a "more information" link. What about I remove it? :)

Nemo157:
We definitely do need to do _something_ about the embedded description, it gets really over the top sometimes e.g. https://docs.rs/chess/3.1.1/chess/. Removing it does seem like the easiest approach, the majority of the time it seems like this will just be a duplicate of the first few sentences of the crate-level docs.

As another point towards minimizing our exposure of the description, on crates.io you get to see the description for about a second before it is replaced by the readme; afaict there is no way to then go back to viewing the description.

I feel like the "More Information" link is a bit hidden, I have personally been using the description as the link, but I think it makes more sense than adding another link under the crate name. Maybe we could have _both_ the "More Information" link and make the name a link?

GuillaumeGomez:
I think the menu should be minimalist and that the docs.rs crate page link should be easier to find. So I'm fine with removing the "More information" link.

Now, the big remaining quest is either we turn the crate name into this link (but then, it's not explicit where it's leading to so I'm not a big fan) or keeping the "docs.rs crate page" I suggested.

jyn514:
I think we should get rid of 'More information' and make the crate name a link. I think that's clear enough, you only have to click it once to find what it does.

GuillaumeGomez:
I agree for removing "more information", but why someone would expect that clicking on the crate name would bring to another page? It's not obvious at all.

GuillaumeGomez:
I removed the "More information" link which was now a duplicate.

templates/rustdoc/topbar.html
```diff
@@ -23,8 +23,8 @@
                     </li>
 
                     <li class="pure-menu-item">
-                        <a href="{{ crate_url | safe }}" class="pure-menu-link" class="description">
-                            {{ "cube" | fas(fw=true) }} {{ krate.description }}
+                        <a href="{{ crate_url | safe }}" class="pure-menu-link description" title="See {{ krate.name }} in docs.rs">
```
jyn514, line 26: What does the `description` class do here?

jyn514 (approved):
Well, I'm still not convinced the separate link is better than making the crate name a link, but it's certainly better than the current behavior.

templates/rustdoc/topbar.html
GuillaumeGomez: Before we add two `class` attributes, so I simply merged them. :)

Issue #1273, opened 2021-02-11
Paddings in about page


MisterFISHUP:
I was using Docs.rs and came across the about page. I think it might be nice to add some paddings/margins for the contents of tabs about, badges, builds, etc :)

Liberatys:
@jyn514 I'd like to work on this if possible?

jyn514:
@Liberatys sounds great! The relevant HTML is in `templates/about` and the CSS is in [`templates/style`](https://github.com/rust-lang/docs.rs/tree/master/templates/style).

jyn514:
Fixed by https://github.com/rust-lang/docs.rs/pull/1275 (but not deployed yet, it's waiting on #1278).

PR #1275, opened 2021-02-13
Add left and right margin to center about container

Liberatys:
This pull request adds left/right margins to the content on the about pages. It centers the page content to make it easier to read on smaller devices.

<details>
<summary>Tablet</summary>

image

</details>

<details>
<summary>Phone</summary>

image

</details>

I opted to use the pure-css classes rather than altering the css.

Addresses #1273


templates/core/about/badges.html
```diff
@@ -4,50 +4,51 @@
...
-		<tr>
-			<td>
-			Version 0.1.0: <a href="{{ mio_badge | safe }}?version=0.1.0">{{ mio_badge }}?version=0.1.0</a>
-			</td>
-			<td><img src="{{ mio_badge | safe }}?version=0.1.0" alt="mio" /></td>
-		</tr>
-		</tbody>
-	</table>
+	<div class="purge-g">
+		<div class="container pure-u-5-6 about">
```
jyn514, line 8: Could you please avoid re-indenting with tabs?

templates/core/about/badges.html
```diff
@@ -4,50 +4,51 @@
...
-
-		<tr>
-			<td>
-			Version 0.1.0: <a href="{{ mio_badge | safe }}?version=0.1.0">{{ mio_badge }}?version=0.1.0</a>
-			</td>
-			<td><img src="{{ mio_badge | safe }}?version=0.1.0" alt="mio" /></td>
-		</tr>
-		</tbody>
-	</table>
+	<div class="purge-g">
```
jyn514, line 7: Is this class documented somewhere? I don't remember it being a custom docs.rs class.

Liberatys: It's a pure-css grid class: https://purecss.io/grids/

jyn514: I think you meant to add `pure-g` then, `purge-g` doesn't have any results. Which makes me think this class isn't actually doing anything - does the page look any different if you remove it?

Liberatys: 🤦 reading is very difficult ^^ the only thing needed is the div.

jyn514:
@Liberatys this looks good to me once you revert the whitespace changes :)

PR #1279, opened 2021-02-15
Don't execute a build when not enough memory is available

### LINKED ISSUE #657: docs.rs starts building a crate even when there is not enough RAM
While investigating #656 we discovered that docs.rs starts building crates even when the host system doesn't have the amount of RAM needed for the build. To avoid making the situation worse in OOM situations we should tweak docs.rs to avoid starting builds when we don't have enough RAM.

jyn514:
Example error:

```
2021/02/15 01:43:12 [INFO] docs_rs::docbuilder::rustwide_builder: building package regex 1.3.1
Error: Building documentation failed

Caused by:
    not enough memory to build regex 1.3.1: needed 3072 MiB, have 2909 MiB
help: set DOCSRS_DISABLE_MEMORY_LIMIT=true to force a build
```

Fixes #657.

r? @pietroalbini or @Nemo157

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -11,6 +11,7 @@ use crate::utils::{copy_doc_dir, parse_rustc_version, CargoMetadata, GithubUpdat
 use crate::{db::blacklist::is_blacklisted, utils::MetadataPackage};
 use crate::{Config, Context, Index, Metrics, Storage};
 use docsrs_metadata::{Metadata, DEFAULT_TARGETS, HOST_TARGET};
+use failure::bail;
 use failure::ResultExt;
 use log::{debug, info, warn, LevelFilter};
 use postgres::Client;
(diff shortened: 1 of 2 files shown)


Nemo157:
> Don't _queue_ a build [...]

Or "_dequeue_ a build" ?

src/docbuilder/rustwide_builder.rs
```diff
@@ -309,6 +310,26 @@ impl RustwideBuilder {
         }
 
         let limits = Limits::for_crate(&mut conn, name)?;
+        #[cfg(target_os = "linux")]
+        if !self.config.disable_memory_limit {
+            let mem_info = procfs::Meminfo::new().context("failed to read /proc/meminfo")?;
+            let available = mem_info.mem_available.unwrap_or(
+                // Assume about half of cached memory can be reclaimed without thrashing.
+                // This is a terrible hack.
+                mem_info.mem_free + mem_info.cached / 2,
```
Nemo157, line 319: On my system at least `free + cached` is ~half of `available`, I think because of how ZFS is implemented it ends up being attributed as `k_reclaimable` instead of `cached`.

But, this branch should never be taken, `available` is 7 years old at this point, I'd check it's available in production then change this to `expect("Your kernel is too old")`.

jyn514: Yup, this is available in prod:
```
$ cat /proc/meminfo | grep -i available
MemAvailable:   14260052 kB
```
Thanks.

Nemo157 (approved):

PR #1283, opened 2021-02-23
Fix the size of top-bar triangle

### LINKED ISSUE #1276: The top-bar triangles are too small
Hello, I am new to Rust. I had a suboptimal very-first-user-visit on docs.rs. I typed "luajit rust" into google and this turned up a docs.rs page for a crate, [https://docs.rs/luajit/0.1.1/luajit/](https://docs.rs/luajit/0.1.1/luajit/). Once I got to the page I wanted to find the source for the crate. The correct way to do this is to click "luajit-0.1.1", which brings down a menu with all kinds of helpful content like a link to crates.io and a link to the gitlab repository. However, I did not realize that this was a menu. It didn't jump out to me as a menu or even interactable, and it didn't occur to me (maybe because it was in the top bar with the whole-site navigation "stuff" and not below with the crate "stuff") that "luajit-0.1.1" would show me crate-specific material. I had to ask for help on "Twitter".

Now, given, I *did* get help, and now that I've gotten help I know where to look and I won't make that mistake on docs.rs again. So maybe it is on me for spacing out and not really looking at the site UI, and maybe there is not really anything wrong with your layout. But, I think there is one thing you could/should change. This is a screenshot of docs.rs on Windows 10 on Chrome.

image

If you expand this image, you might realize that the triangles, the only thing indicating that text at the top is a drop down menu and not a label or link, barely look like triangles at all.

image

I have also a Macbook with a Retina screen. If I look at this same page on the Retina screen, also in Chrome, the triangles are clearly legible as triangles. But on a 1x monitor they are more like small smudges.

"Expected behavior": I think you should consider making those triangles bigger. In the current design they are not only so unobtrusive they are possible to overlook, they are arguably illegible on 1x monitors. (Another solution might be to use CSS to detect non-retina/1x resolution and use a slightly different, clearer graphic there, if that is possible.)

chansuke:
Fixes #1276.

Changed the size of the top bar triangle.

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -31,6 +31,7 @@ div.nav-container {
         }
         &:after {
             font-size: 0.8em;
+            content: "\25BC"
         }
     }
 


jyn514:
Can you post a screenshot of what this looks like?

chansuke:
@jyn514 Sure! I will post that.

chansuke:
sample_one
sample_two

@jyn514 These are the screenshots, thanks.

jyn514:
Looks great, thanks!

PR #1288, opened 2021-02-26
Remove possible whitespace in title

pickfire:
image

It should fix the whitespace after regex and after platform, the first whitespace I thought I fixed it but it wasn't fixed.


jyn514:
The current header looks like this:
image
I don't really see a difference between the two ...

pickfire:
@jyn514 Right click on it, you will notice there is a bit of whitespace next to the bar, try selecting it (clicking or right clicking) since it will flip the color, then will see an obvious bar with some space next to it.

jyn514:
It still looks the same to me:
Currently: 
image

Your changes: image

pickfire:
This time should work except the first one still not fixed since tera have no way to remove the whitespace https://github.com/Keats/tera/issues/610, changing html is probably the only option right now.

image

templates/rustdoc/topbar.html
```diff
@@ -162,16 +162,16 @@
...
             <a href="{{ crate_url | safe }}" class="pure-menu-link crate-name" title="{{ metadata.description }}">
                 {{ "cube" | fas }}
                 <span class="title">{{ metadata.name }}-{{ metadata.version }}</span>
             </a>
         </li>
-    {%- endif -%}
-
-    {# If this is the latest release and it's been yanked, just display a warning #}
+    {%- endif -%}{#
+
```
jyn514, line 173: Can you remove this extra whitespace at the end of the line?

templates/rustdoc/topbar.html
```diff
@@ -202,9 +202,9 @@
             </a>
         </li>
     {%- endif -%}
-
+
```
jyn514, line 205: Same here.

jyn514 (commented):
LGTM with the whitespace fixed.

templates/rustdoc/topbar.html
pickfire: Done, thanks.

PR #1308, opened 2021-03-06
Fix spelling of FiraSans-*.woff2

Nemo157:
Fixes the underlying issue causing #1305 to fail.


Nemo157:
I've confirmed this works with a local build.

Issue #294, opened 2019-02-10
Preserve .css files in target/doc


japaric:
### Background / use case

For the `cortex-m-rtfm` crate I ship [an mdbook][book] with [the API
documentation][api] and host both on GH pages. This setup works OK for now but
I'll start having problems once I release a new minor version: people looking at
older minor versions (e.g. on docs.rs) will be directed to the latest version of
the book on GH pages and get confused.

[book]: https://japaric.github.io/cortex-m-rtfm/book/
[api]: https://japaric.github.io/cortex-m-rtfm/api/rtfm/index.html

Ideally, I'd like to move away from GH pages and host both the API docs and the
mdbook on docs.rs. That way people will get the right version of the mdbook
regardless of the minor version of the crate they are looking at.

---

I experimented a bit with getting an mdbook rendered on docs.rs and managed to
place the mdbook output in the target/doc directory using build scripts. You can
see the results at
https://docs.rs/cautious-octo-funicular/0.1.3/cautious_octo_funicular/ -- the
mdbook is [there] but the layout is all wrong. According to @quietmisdreavus the
mdbook CSS files are not being preserved and that's why the mdbook shows
unstyled.

[there]: https://docs.rs/cautious-octo-funicular/0.1.3/cautious_octo_funicular/book/index.html

This issue is to request that .css files placed in target/doc are preserved to
support the above use case.

jyn514:
Hmm, I would like to support this, but I also don't want to copy .css files generated by rustdoc, since that would result in a lot of duplicated files for every crate.

jyn514:
This change was reverted because it broke search: https://github.com/rust-lang/docs.rs/pull/1329. I hope https://github.com/rust-lang/rust/pull/83478 would allow relanding the change.

jyn514:
> I experimented a bit with getting an mdbook rendered on docs.rs and managed to
place the mdbook output in the target/doc directory using build scripts.

FYI we are hoping to add first-class support for mdbook at some point: https://github.com/rust-lang/docs.rs/issues/1293

PR #1312, opened 2021-03-10
Don't hard-code essential rustdoc files

jyn514:
This avoids having to constantly update docs.rs when new files are
added.

Closes https://github.com/rust-lang/docs.rs/issues/1302, closes https://github.com/rust-lang/docs.rs/issues/1240, closes https://github.com/rust-lang/docs.rs/issues/294.
r? @Nemo157

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -14,47 +14,16 @@ use docsrs_metadata::{Metadata, DEFAULT_TARGETS, HOST_TARGET};
 use failure::ResultExt;
 use log::{debug, info, warn, LevelFilter};
 use postgres::Client;
-use rustwide::cmd::{Command, SandboxBuilder, SandboxImage};
+use rustwide::cmd::{Binary, Command, SandboxBuilder, SandboxImage};
 use rustwide::logging::{self, LogStorage};
 use rustwide::toolchain::ToolchainError;
 use rustwide::{Build, Crate, Toolchain, Workspace, WorkspaceBuilder};
 use serde_json::Value;
 use std::collections::{HashMap, HashSet};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 const USER_AGENT: &str = "docs.rs builder (https://github.com/rust-lang/docs.rs)";
-const ESSENTIAL_FILES_VERSIONED: &[&str] = &[
-    "brush.svg",
-    "favicon.svg",
-    "wheel.svg",
-    "down-arrow.svg",
-    "dark.css",
-    "light.css",
-    "ayu.css",
-    "main.js",
-    "normalize.css",
-    "rustdoc.css",
-    "settings.css",
-    "settings.js",
-    "storage.js",
-    "theme.js",
-    "source-script.js",
-    "noscript.css",
-    "rust-logo.png",
-];
-const ESSENTIAL_FILES_UNVERSIONED: &[&str] = &[
-    "FiraSans-Medium.woff",
-    "FiraSans-Medium.woff2",
-    "FiraSans-Regular.woff",
-    "FiraSans-Regular.woff2",
-    "SourceCodePro-Regular.woff",
-    "SourceCodePro-Semibold.woff",
-    "SourceSerifPro-Bold.ttf.woff",
-    "SourceSerifPro-Regular.ttf.woff",
-    "SourceSerifPro-It.ttf.woff",
-];
-
 const DUMMY_CRATE_NAME: &str = "empty-library";
 const DUMMY_CRATE_VERSION: &str = "1.0.0";
 
@@ -213,6 +182,20 @@ impl RustwideBuilder {
         let krate = Crate::crates_io(DUMMY_CRATE_NAME, DUMMY_CRATE_VERSION);
         krate.fetch(&self.workspace)?;
 
+        // TODO: remove this when https://github.com/rust-lang/rustwide/pull/53 lands.
+        struct Rustdoc<'a> {
+            toolchain_version: &'a str,
+        }
+        impl rustwide::cmd::Runnable for Rustdoc<'_> {
+            fn name(&self) -> Binary {
+                Binary::ManagedByRustwide(PathBuf::from("rustdoc"))
+            }
+
+            fn prepare_command<'w, 'pl>(&self, cmd: Command<'w, 'pl>) -> Command<'w, 'pl> {
+                cmd.args(&[format!("+{}", self.toolchain_version)])
+            }
+        }
+
         build_dir
             .build(&self.toolchain, &krate, self.prepare_sandbox(&limits))
             .run(|build| {
@@ -229,19 +212,29 @@ impl RustwideBuilder {
                     .prefix("essential-files")
                     .tempdir()?;
 
-                let files = ESSENTIAL_FILES_VERSIONED
+                let toolchain_version = self.toolchain.as_dist().unwrap().name();
+                let output = build.cmd(Rustdoc { toolchain_version })
+                    .args(&["-Zunstable-options", "--print=unversioned-files"])
+                    .run_capture()
+                    .context("failed to learn about unversioned files - make sure you have nightly-2021-03-07 or later")?;
+                let essential_files_unversioned = output
+                    .stdout_lines()
                     .iter()
-                    .map(|f| (f, true))
-                    .chain(ESSENTIAL_FILES_UNVERSIONED.iter().map(|f| (f, false)));
-                for (&file, versioned) in files {
-                    let segments = file.rsplitn(2, '.').collect::<Vec<_>>();
-                    let file_name = if versioned {
-                        format!("{}-{}.{}", segments[1], rustc_version, segments[0])
-                    } else {
-                        file.to_string()
-                    };
+                    .map(PathBuf::from);
+                let resource_suffix = format!("-{}", parse_rustc_version(&self.rustc_version)?);
+                let essential_files_versioned: Vec<_> = source.read_dir()?
+                    .collect::<std::result::Result<Vec<_>, _>>()?
+                    .into_iter()
+                    .filter_map(|entry| {
+                        entry.file_name().to_str().and_then(|name| if name.contains(&resource_suffix) {
+                            Some(entry.file_name().into())
+                        } else { None })
+                    })
+                    .collect();
+                for file_name in essential_files_unversioned.chain(essential_files_versioned) {
                     let source_path = source.join(&file_name);
                     let dest_path = dest.path().join(&file_name);
+                    debug!("copying {} to {}", source_path.display(), dest_path.display());
                     ::std::fs::copy(&source_path, &dest_path).with_context(|_| {
                         format!(
                             "couldn't copy '{}' to '{}'",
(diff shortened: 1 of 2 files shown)


jyn514:
Example build:
image
Generated with rustc 1.52.0-nightly (3a5d45f68 2021-03-09).

syphar:
this is awesome :)

Nemo157:
Code changes look good to me, running a local build to confirm it (though I guess we have actual tests covering this code path since #1309 now 🥳).

src/docbuilder/rustwide_builder.rs
```diff
@@ -230,19 +211,29 @@ impl RustwideBuilder {
                     .prefix("essential-files")
                     .tempdir()?;
 
-                let files = ESSENTIAL_FILES_VERSIONED
+                let toolchain_version = self.toolchain.as_dist().unwrap().name();
```
Nemo157, line 214:
```suggestion
                let toolchain_version = self.toolchain.as_dist().expect("we only use the Toolchain::dist constructor").name();
```

maybe worth it 🤷. It'd be nice to have it statically known that it is a dist toolchain, but I guess that'd require rustwide changes.

Nemo157:
So, fun bug, this breaks searching because the "essential files" version of `search-index.js` is preferred over the local one. Can see the effect in https://docs.rs.dev.nemo157.com/bs58/0.4.0/search-index-20210312-1.52.0-nightly-b3e19a221.js returning the search index for `empty_library` (IPv6 only).

jyn514:
> So, fun bug, this breaks searching because the "essential files" version of search-index.js is preferred over the local one. Can see the effect in https://docs.rs.dev.nemo157.com/bs58/0.4.0/search-index-20210312-1.52.0-nightly-b3e19a221.js returning the search index for empty_library (IPv6 only).

This seems like the same rustdoc bug as https://github.com/rust-lang/rust/pull/83094 - I don't think rustdoc should be using --static-root-path for search-index.js.

jyn514:
Hmm actually that might be necessary. rustdoc uses `root_path` to load it:
```
$ rg search-index src/librustdoc/
src/librustdoc/html/layout.rs
117:       data-search-js=\"{root_path}search-index{suffix}.js\"></div>
```
but writes it to the shared file cache: https://github.com/rust-lang/rust/blob/e7e1dc158c3de232750b568163f6941a184ee8be/src/librustdoc/html/render/write_shared.rs#L381. I think that's because it re-uses the search index for all crates documented at the same time. I guess we could special-case it in docs.rs?

Nemo157:
I wonder whether rustdoc shouldn't be applying the suffix to it? What is the purpose of putting a per-toolchain suffix on a dynamically generated file?

jyn514:
> I wonder whether rustdoc shouldn't be applying the suffix to it? What is the purpose of putting a per-toolchain suffix on a dynamically generated file?

@GuillaumeGomez do you know why rustdoc uses a resource suffix for the search index?

GuillaumeGomez:
Because it allows you to have multiple search-index files in the same location I assume? I remember that it was done for docs.rs, but I don't remember why at all...

jyn514:
Well, docs.rs isn't using it, so if that's the only reason I think we should remove the suffix.

GuillaumeGomez:
Let's make sure no one is actually using it first. I'll ask that on twitter and users.rust-lang.org.

GuillaumeGomez:
> I think that's because it re-uses the search index for all crates documented at the same time.

I didn't understand what you meant with this. `rustdoc` uses only one `search-index.js`, one `source-files.js` and one `crates.js` and update them if they already exists and always has been. Is it an issue for docs.rs?

Nemo157:
The issue here is that the existence of the suffix is used to detect the "essential files" that are shared between all builds, so because it generates `search-index-20210312-1.52.0-nightly-b3e19a221.js` that file is picked up from the initial `empty-library` build and stored for the toolchain. When the browser then does a request for `https://docs.rs/<crate>/<version>/search-index-20210312-1.52.0-nightly-b3e19a221.js ` to get the search index for that doc-build, the "essential file" from `empty-library` is then served instead (it detects the presence of what looks like a rustdoc version suffix on the url and then serves the essential file, ignoring the subdirectory that file is in, currently because we don't store the `search-index` that is a 404 and it _then_ fallsback to serving the correct build-specific file).

Nemo157:
Big alternative: Can we just stop storing the essential files? It appears that we're saving the per-build versions of all these files anyway, so if we stop passing `--static-root-path` etc. it seems like it would Just Work:tm:.

The main downside I see is that would probably affect browser caching, but given the likelihood of different docs you're looking at being built from the same nightly that doesn't seem like a big downside. Also if `rustdoc` started emitting subresource integrity hashes those would match cross-build for the static files and might be used by browsers as a cache key.

Nemo157:
https://github.com/rust-lang/docs.rs/blob/07396f11bbfd720c0b3c9149a6e1ac8a72e8e35c/src/utils/copy.rs#L20-L23

This also seems like it should be updated to take into account the known shared static files, instead of being a hardcoded regex.

jyn514:
> Big alternative: Can we just stop storing the essential files?

GuillaumeGomez:
That would work too I guess. Just a question: if the file already exists, wouldn't it cost money to copy it again? I guess we can prevent it by checking if the file already exists?

jyn514:
> Big alternative: Can we just stop storing the essential files? It appears that we're saving the per-build versions of all these files anyway, so if we stop passing --static-root-path etc. it seems like it would Just Worktm. EDIT: Nope, I just hit the fact that the files are available in any directory when testing this, we don't actually store them all for each build, just a somewhat random subset of recently added ones.

Can we hold off on this for now and just fix the essential files to be tracked properly? I think the search-index thing could be fixed by special casing that file not to be copied.

syphar:
> The main downside I see is that would probably affect browser caching, but given the likelihood of different docs you're looking at being built from the same nightly that doesn't seem like a big downside. 

Not only browser-caching, but also CDN caching for us europeans :) But your point stands: the cache is only by nightly version. 

> Also if `rustdoc` started emitting subresource integrity hashes those would match cross-build for the static files and might be used by browsers as a cache key.

That would IMHO be the best way forward. Then we could really cache the static files across a longer timespan in the CDN. In the end, we do host many docs, and they do share static files to an extend. So I would keep the shared-resource logic and try to improve it.

jyn514:
These are the files that are different between the ones we hard-coded and the ones rustdoc adds a resource-suffix to:
```
> diff hard-coded.txt rustdoc-generated.txt 
2a3
> crates.js
4a6,7
> favicon-16x16.png
> favicon-32x32.png
11a15
> search-index.js
13a18
> source-files.js
```
The new favicon files are correct; merging this will fix https://github.com/rust-lang/docs.rs/issues/1240. The crates.js and search-index.js files are not - I think it's a rustdoc bug that these files have a resource-suffix appended, because they can't be shared across crates. Here's `source-files.js` for `empty_library`:

```
var N = null;var sourcesIndex = {};
sourcesIndex["empty_library"] = {"name":"","files":["lib.rs"]};
createSourceSidebar();
```

so I expect it can't be shared across crates either. I'll see about fixing that in rustdoc.

jyn514:
> I think it's a rustdoc bug that these files have a resource-suffix appended, because they can't be shared across crates.

Yeah, https://github.com/rust-lang/rust/pull/57011/ explicitly says

> I made sure to only change links for the static files, those that don't change between crates. Files like the search index, aliases, the source files listing, etc, are still linked with relative links.

jyn514:
I think https://github.com/rust-lang/rust/pull/59776 was incorrect, it means that files with a resource-suffix will be shared across crates even if they're incompatible. That links to https://github.com/rust-lang/docs.rs/issues/316#issuecomment-482274857 which says

> That PR suggests to start adding the resource suffix to the search index, which will sidestep the issue by making newer builds request a different URL for it. That will solve the Rocket issue, and this specific problem of main.js and search-index.js using different versions of the index structure.
> However, now that i'm typing this up here, i see that [using redirects on the docs.rs side] doesn't fix the Rocket problem, so maybe the complete solution is to do both this proposal and that PR.

I don't know what 'the rocket problem' was by reading that - I would expect it to be the same as whatever's going on with tokio in that issue? @QuietMisdreavus sorry to bother you - do you still remember what you meant back then?

I think either way it *shouldn't* matter because docs.rs never added those files to `ESSENTIAL_FILES_VERSIONED`, so whatever the rustdoc change did, it didn't have an impact until now.

jyn514:
Also, it seems *really* strange to me that rustdoc would emit a file with a version suffix, but load it from `root_path` instead of `static_root_path` - I don't know why that would ever be useful, the version suffix is explicitly so you can share it across different crates.

Nemo157:
My reading of the "rocket problem" is that it is essentially an issue of inconsistent caching. Imagine that we build the documentation for a crate with one version of rustdoc, that loads `/main-<rustdoc version 1>.js` and `/<crate>/<version>/search-index.js`. If we then rebuild those docs with a later nightly that will start loading `/main-<rustdoc version 2>.js` and `/<crate>/<version>/search-index.js`. If someone had a cached copy of the original `search-index.js` then they would be loading that with a newer version of `main.js`, which could fail because the format has changed.

(I think the original issue was with semver-constraint-paths rather than rebuilt documentation, but I think that's no longer an issue because we never serve any files under semver-constraint-paths, just redirect to the current matching version).

jyn514:
> If someone had a cached copy of the original search-index.js then they would be loading that with a newer version of main.js, which could fail because the format has changed.

Hmm, when would search-index.js be cached but not main.js? Aren't they all loaded at the same time?

Nemo157:
If we didn't have resource suffixes on `search-index.js` the load dependency tree would be:

```
/<crate>/<version>/index.html
├── /main-<suffix>.js
└── /<crate>/<version>/search-index.js
```

and after a rebuild:

```
/<crate>/<version>/index.html
├── /main-<suffix 2>.js
└── /<crate>/<version>/search-index.js
```

The browser when loading `/<crate>/<version>/index.html` would (almost) always do a conditional request with the ETag to check that the cached copy is still valid, the condition would fail and it would be sent the new version of the html which referenced the new path for the `main.js` and the same path for `search-index.js`. Since the path for `main.js` changed it would then have to do a new request for that path, while `search-index.js` appears to be the same file and it could simply use the cached copy (assuming that we have cache-directives on it).

jyn514:
I think another possibility rather than messing with this in rustdoc is to invalidate the CloudFront cache whenever we rebuild a crate - that seems like good practice anyway since otherwise you'll see older versions of docs, and we rebuild infrequently enough it shouldn't be too expensive.

Nemo157:
Invalidating the CDN wouldn't be enough, when navigating to a doc page I see the `search-index.js` served directly from cache without any conditional request.

One idea would be to have a _different_ suffix that we could tell rustdoc to stick on dynamic subresources, we could then fill that suffix with something that is per-build so that the `search-index.js` for each build would actually be different paths and cached separately (that would also mean we could mark the subresources as truly `immutable` for caching purposes).

jyn514:
I mean, at that point I don't know if it's worth the effort - I would rather just store the extra 3 files per toolchain, it's trivial compared to our other costs (or we could special-case those three if @pietroalbini thinks the savings are worth it).

The immediate problem should be solved by https://github.com/rust-lang/docs.rs/pull/1324 - even if we do copy files that shouldn't be shared, rustdoc will still load them per-crate, which should avoid any errors.

jyn514:
So I don't forget: this is waiting on https://github.com/rust-lang/docs.rs/pull/1312#issuecomment-798798642.

jyn514:
> This also seems like it should be updated to take into account the known shared static files, instead of being a hardcoded regex (e.g. it's copying the .woff2 fonts, and the .png favicons).

Done - I don't know a great way to test it, but it doesn't have any fewer tests than the rest of docbuilder.

Nemo157 (approved):

PR #1314, opened 2021-03-10
Cleanup unneeded JS

notriddle:


templates/core/home.html
```diff
@@ -81,16 +81,6 @@ <h1 class="brand">{{ "cubes" | fas(fw=true) }} Docs.rs</h1>
...
 
-        document.getElementById("i-am-feeling-lucky-button").onclick = function () {
-            var form = document.getElementsByClassName("landing-search-form");
-            var input = document.createElement('input');
-            input.type = 'hidden';
-            input.name = 'i-am-feeling-lucky';
-            input.value = 1;
-            document.getElementsByClassName("landing-search-form")[0].appendChild(input);
-            return true;
-        };
```
jyn514, line 92: Are you sure just adding the `name` and `value` has the same behavior? That doesn't create the form or hide the element. Do you know what the form was for?

notriddle: It's not creating a form. It's adding a new `input` element to the existing form.

Yes, adding the name does the same thing. The resulting URL, `/search?query=regex&i-am-feeling-lucky=1`, is the same before and after this change.

templates/core/home.html
```diff
@@ -14,7 +14,7 @@ <h1 class="brand">{{ "cubes" | fas(fw=true) }} Docs.rs</h1>
 
             <div class="buttons">
                 <button type="submit" class="pure-button pure-button-normal">Search</button>
-                <button type="submit" class="pure-button pure-button-normal" id="i-am-feeling-lucky-button">
+                <button type="submit" class="pure-button pure-button-normal" id="i-am-feeling-lucky-button" name="i-am-feeling-lucky" value="1">
```
jyn514, line 17: I think you can remove the `id` now that you're not using `getElementById`.
```suggestion
                <button type="submit" class="pure-button pure-button-normal" name="i-am-feeling-lucky" value="1">
```

jyn514 (approved):
LGTM with the nit fixed, sorry for the delay reviewing.

notriddle:
I fixed your nit. Is it mergeable now?

jyn514:
Yes, thank you!

PR #1316, opened 2021-03-11
Remove redundant call

GuillaumeGomez:
The folder is created in the function called recursively anyway, there is no point to call this function there.


jyn514:
@GuillaumeGomez could you run a local build and make sure this still has all the files we need? If so this looks good to me :)

GuillaumeGomez:
`docker pull` times out every time for the moment so I'll retry later on.

GuillaumeGomez:
Just re-built the regex crate. No issue when navigating files.

Issue #1300, opened 2021-03-05
Search within crate is broken: "Uncaught TypeError: crates is undefined"


dtolnay:
https://docs.rs/syn/1.0.61/syn/?search=Expr fails to load, with the following in the JS console. In comparison, https://docs.rs/syn/1.0.60/syn/?search=Expr works fine as before. https://github.com/rust-lang/rust/pull/82310 seems relevant. @jsha

---

![Screenshot from 2021-03-04 19-58-25](https://user-images.githubusercontent.com/1940490/110065454-fef92400-7d23-11eb-935b-8343a5b18d1e.png)

jyn514:
Hmm, I rebuilt syn 1.0.61 after deploying https://github.com/rust-lang/docs.rs/pull/1301, but it didn't seem to help. I think the list of essential files might only be calculated once per toolchain version?

jyn514:
This has been fixed.

There have been a lot of changes to essential files recently, but the most recent ones were https://github.com/rust-lang/docs.rs/issues/1302 and https://github.com/rust-lang/docs.rs/pull/1329. I rebuilt syn 1.0.61 after 1329, but because the new file has the same name as the old one, some CDNs have cached the old version even though I invalidated the CloudFlare cache. You might run into the search showing no results until that cache expires, but there's nothing we can do about it on our end.

jyn514:
Also, if I could ask a favor, could you please avoid linking directly to docs.rs issues in the changelog for big crates like syn? Normally I don't mind, but since depend-a-bot includes the changelog in its PRs it leads to lots of github spam, which makes it hard to see what's actually relevant.

puzzlewolf:
Hi, I think I'm seeing the same issue for `sequoia-openpgp`. https://docs.rs/sequoia-openpgp/1.1.0/sequoia_openpgp/?search=foo is stuck at "Loading search results...", with the following in the console:

```
This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”. sequoia_openpgp
GEThttps://docs.rs/crates-20210304-1.52.0-nightly-45b3c2851.js
[HTTP/2 404 Not Found 188ms]

Loading failed for the <script> with source “https://docs.rs/crates-20210304-1.52.0-nightly-45b3c2851.js”. sequoia_openpgp:3:1
Uncaught TypeError: crates is undefined
[...]

PR #1317, opened 2021-03-11
Fix unwanted "empty-library" appearance

### LINKED ISSUE #1313: Publishing a crate without dependencies causes a `empty_library` dependency.
When publishing a no_std crate without dependencies, it will add an `empty_library` to the crates on [docs.rs].

Reproduction steps:
1. Publish a no_std library without dependencies.
2. The docs.rs page will now have a `empty_library` in the crates.

This seems like a rather awkward way to tell the reader the crate has no dependencies. Now if you click any deeper into the documentation, on the left (under the crates header) the empty_library will be a clickable link that leads to a dead link.

This is what I mean: [Link to CRC8-RS Docs](https://docs.rs/crc8-rs/1.0.0/crc8_rs/fn.fetch_crc8.html)

GuillaumeGomez:
Fixes #1313.

cc @Nemo157 @jyn514 

We will need to remove the files into the existing dirs too I guess?

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -42,7 +42,6 @@ const ESSENTIAL_FILES_VERSIONED: &[&str] = &[
     "source-script.js",
     "noscript.css",
     "rust-logo.png",
-    "crates.js",
 ];
 const ESSENTIAL_FILES_UNVERSIONED: &[&str] = &[
     "FiraSans-Medium.woff",


src/docbuilder/rustwide_builder.rs
```diff
@@ -237,6 +237,14 @@ impl RustwideBuilder {
                 for (&file, versioned) in files {
                     let segments = file.rsplitn(2, '.').collect::<Vec<_>>();
                     let file_name = if versioned {
+                        if segments[0] == "js"
+                            && ["crates", "search-index"].iter().any(|x| *x == segments[1])
```
Nemo157, line 241: Couldn't you just remove `crates.js` from the `ESSENTIAL_FILES_VERSIONED` list above?

Nemo157: This will also cause https://github.com/rust-lang/docs.rs/issues/1300 to reoccur, we might need to generate a `crates.js` that sets an empty array or something (or really it should come from each doc build, not the essential files, to support multi-crate packages in the future).

Nemo157:
Looking back into #1300 more, what I don't understand is why it is trying to load a global `crates.js` instead of a local one for each crate: https://docs.rs/crates-20210304-1.52.0-nightly-45b3c2851.js does not exist but https://docs.rs/syn/1.0.61/crates-20210304-1.52.0-nightly-45b3c2851.js does. Are we incorrectly rewriting this url? (note that `search-index.js` _is_ loaded locally from the current version)

src/docbuilder/rustwide_builder.rs
GuillaumeGomez: It should definitely come from each doc build.

GuillaumeGomez:
@Nemo157 This fix works too. Now the problem is that the `crates.js` path is invalid. I don't how we generate it correctly for `search-index.js` and `source-files.js`. Any idea maybe?

Nemo157:
This can be merged once https://github.com/rust-lang/rust/pull/83094 is in a nightly, that on its own should fix the issue, this should just avoid copying a file that shouldn't actually be part of the essential files (merging it before that would cause the errors from #1300 again, the symptom in #1313 is less bad).

jyn514:
> This can be merged once [rust-lang/rust#83094](https://github.com/rust-lang/rust/pull/83094) is in a nightly, that on its own should fix the issue, this should just avoid copying a file that shouldn't actually be part of the essential files (merging it before that would cause the errors from #1300 again, the symptom in #1313 is less bad).

@Nemo157 https://docs.rs/crc8-rs/1.1.0/crc8_rs/ was built with rustc 1.52.0-nightly (36f1f04f1 2021-03-17), but it still shows empty-library and loads crates.js from the suffix path. That makes me worry that deploying this will break all builds again.

jyn514:
Hmm, it does seem to work locally though.

GuillaumeGomez:
I think a file needs to be removed somewhere on our end.

PR #1318, opened 2021-03-13
Allow unauthenticated users to set basic status labels

camelid:
More labels can be permitted later, but these labels are useful to
indicate the status of a PR. E.g., if someone addresses a review,
there should be a way for them to switch `S-waiting-on-author` to
`S-waiting-on-review`.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -1,0 +1,6 @@
+[relabel]
+allow-unauthenticated = [
+    "S-waiting-on-review",
+    "S-waiting-on-author",
+    "S-blocked",
+]


jyn514 (commented):
I have the same comments as in https://github.com/rust-lang/rustc-dev-guide/pull/910. I think

> Should we also add an "auto-labeler" to add waiting-on-review to all PRs when opened?

is the only one that's actionable.

Thanks for the PR!

camelid:
> I have the same comments as in [rust-lang/rustc-dev-guide#910](https://github.com/rust-lang/rustc-dev-guide/pull/910). I think
> 
> > Should we also add an "auto-labeler" to add waiting-on-review to all PRs when opened?
> 
> is the only one that's actionable.

Hmm, what do you mean? Is it necessary to add autolabeling before allowing unauthenticated users to change labels? Also, I think you would need to set up rust-highfive to get autolabeling.

jyn514:
> Is it necessary to add autolabeling before allowing unauthenticated users to change labels? Also, I think you would need to set up rust-highfive to get autolabeling.

Oh good point, I forgot the bots are separate. Wasn't there a plan to unify them at some point ... ?

camelid:
> Wasn't there a plan to unify them at some point ... ?

You might be thinking of rust-lang/triagebot#433, which IIRC is blocked.

PR #1320, opened 2021-03-17
Sort crate features by name

apiraino:
This patch changes one detail implemented in #1144. The feature flags
sorting was by "number of subfeatures". Now it is by feature alphabetically
which I find suitable to easily discover the feature I'm looking for in a crate.
It also ensures that the special "default" feature stays on top of the list.

If we don't want to touch this behaviour, I'm also 100% ok to decline.
If there is a better way to implement this I'm happy to improve it.

cc @almusil since author of the original patch
super-thanks to @GuillaumeGomez for mentoring me on this!

r? @GuillaumeGomez or @jyn514

### DIFF
--- a/src/web/features.rs
+++ b/src/web/features.rs
@@ -60,6 +60,7 @@ pub fn build_features_handler(req: &mut Request) -> IronResult<Response> {
     let row = cexpect!(req, rows.get(0));
 
     let mut features = None;
+    // features enabled by the default feature flag
     let mut default_len = 0;
 
     if let Some(raw) = row.get(0) {
@@ -79,15 +80,22 @@ pub fn build_features_handler(req: &mut Request) -> IronResult<Response> {
 fn order_features_and_count_default_len(raw: Vec<Feature>) -> (Vec<Feature>, usize) {
     let mut feature_map = get_feature_map(raw);
     let mut features = get_tree_structure_from_default(&mut feature_map);
-    let mut remaining: Vec<_> = feature_map
+    let remaining: Vec<_> = feature_map
         .into_iter()
         .map(|(_, feature)| feature)
         .collect();
-    remaining.sort_by_key(|feature| feature.subfeatures.len());
 
     let default_len = features.len();
 
     features.extend(remaining.into_iter().rev());
+    features.sort_by(|a, b| {
+        // 'default' feature is special and should always stay on top
+        if b.name == "default" {
+            std::cmp::Ordering::Greater
+        } else {
+            a.name.partial_cmp(&b.name).unwrap()
+        }
+    });
     (features, default_len)
 }
 


src/web/features.rs
```diff
@@ -79,15 +80,22 @@ pub fn build_features_handler(req: &mut Request) -> IronResult<Response> {
...
 
     features.extend(remaining.into_iter().rev());
+    features.sort_by(|a, b| {
+        // 'default' feature is special and should always stay on top
+        if b.name == "default" {
+            std::cmp::Ordering::Greater
+        } else {
+            a.name.partial_cmp(&b.name).unwrap()
+        }
+    });
```
apiraino, line 98: at first I've tried to simply change the sorting at the SQL level on [line 55](https://github.com/rust-lang/docs.rs/pull/1320/files#diff-0c19324803df42be471baabdc27c2f8f371245f7626c7534bfb17de43872dc5eR55) but I think this sorting in this function would neutralize it anyway

apiraino:
How the sorting looks before this patch

<details>
old
</details>

How the sorting looks after this patch
<details>
new
</details>

GuillaumeGomez (approved):
Nice work!

almusil:
Thank you for including me.

I don't mind this change, I am not sure if the requirement has changed since the initial implementation.
However if we go with this, there is actually no need to compute the whole tree structure etc. 
This is maintainers call but I don't see a point in keeping code that is not used.

apiraino:
> I don't mind this change, I am not sure if the requirement has changed since the initial implementation.

@almusil good point, I forgot to give some context. Example: I went to the tokio crate feature list (see my screenshots) looking for a specific one and I was confused because I expected them in alphabetical order, so I had to run `CTRL+F`and find it.

This may be a subjective "feeling" on how I see the sorting and this PR is just a small question+proposal, this is why I mentioned it's perfectly fine to decline this PR :)

Meta-thought: it's often hard to please every and each user, often there are conflicting thoughts about a UX feature :) The expectations of a single user (in this case: me) should not put anyone under pressure to accept a PR.

almusil:
> > I don't mind this change, I am not sure if the requirement has changed since the initial implementation.
> 
> @almusil good point, I forgot to give some context. Example: I went to the tokio crate feature list (see my screenshots) looking for a specific one and I was confused because I expected them in alphabetical order, so I had to run `CTRL+F`and find it.

The idea was to have flat tree structure of features that are enabled by default followed by features that enable any other feature. 
But having them sorted by name also makes sense in wider context. 

> 
> This may be a subjective "feeling" on how I see the sorting and this PR is just a small question+proposal, this is why I mentioned it's perfectly fine to decline this PR :)

I understand that, this is worth a discussion.

> 
> Meta-thought: it's often hard to please every and each user, often there are conflicting thoughts about a UX/UI :) The expectations of a single user (in this case: me) should not put anyone under pressure to accept a PR.

jyn514:
Personally I like the tree structure better, I think it shows the features you'd want most often towards the top. If the crate has really complicated features, I think it should be documenting those itself anyway.

apiraino:
> Personally I like the tree structure better, I think it shows the features you'd want most often towards the top. If the crate has really complicated features, I think it should be documenting those itself anyway.

Based on your reasoning, I think you have a stronger case to keep it as it is, so I'll close my proposal.

(thanks @jyn514 for your thoughts on the matter)

PR #1322, opened 2021-03-19
stop displaying and serving authorship information

hi-rustin:
See: https://github.com/rust-lang/rust/issues/83227

Part of https://github.com/rust-lang/docs.rs/issues/1323

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -11,6 +11,7 @@ pub enum Nope {
     ResourceNotFound,
     BuildNotFound,
     CrateNotFound,
+    OwnerNotFound,
     VersionNotFound,
     NoResults,
     InternalServerError,
@@ -22,6 +23,7 @@ impl fmt::Display for Nope {
             Nope::ResourceNotFound => "Requested resource not found",
             Nope::BuildNotFound => "Requested build not found",
             Nope::CrateNotFound => "Requested crate not found",
+            Nope::OwnerNotFound => "Requested owner not found",
             Nope::VersionNotFound => "Requested crate does not have specified version",
             Nope::NoResults => "Search yielded no results",
             Nope::InternalServerError => "Internal server error",
@@ -39,6 +41,7 @@ impl From<Nope> for IronError {
             Nope::ResourceNotFound
             | Nope::BuildNotFound
             | Nope::CrateNotFound
+            | Nope::OwnerNotFound
             | Nope::VersionNotFound
             | Nope::NoResults => status::NotFound,
             Nope::InternalServerError => status::InternalServerError,
@@ -79,6 +82,13 @@ impl Handler for Nope {
                 }
                 .into_response(req)
             }
+
+            Nope::OwnerNotFound => ErrorPage {
+                title: "The requested owner does not exist",
+                message: Some("no such owner".into()),
+                status: Status::NotFound,
+            }
+            .into_response(req),
 
             Nope::VersionNotFound => {
                 // user tried to navigate to a crate with a version that does not exist
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -109,61 +109,11 @@ pub(crate) fn get_releases(conn: &mut Client, page: i64, limit: i64, order: Orde
         .collect()
 }
 
-fn get_releases_by_author(
-    conn: &mut Client,
-    page: i64,
-    limit: i64,
-    author: &str,
-) -> (String, Vec<Release>) {
-    let offset = (page - 1) * limit;
-
-    let query = "
-        SELECT crates.name,
-               releases.version,
-               releases.description,
-               releases.target_name,
-               releases.release_time,
-               releases.rustdoc_status,
-               github_repos.stars,
-               authors.name
-        FROM crates
-        INNER JOIN releases ON releases.id = crates.latest_version_id
-        INNER JOIN author_rels ON releases.id = author_rels.rid
-        INNER JOIN authors ON authors.id = author_rels.aid
-        LEFT JOIN github_repos ON releases.github_repo = github_repos.id
-        WHERE authors.slug = $1
-        ORDER BY github_repos.stars DESC NULLS LAST
-        LIMIT $2 OFFSET $3";
-    let query = conn.query(query, &[&author, &limit, &offset]).unwrap();
-
-    let mut author_name = None;
-    let packages = query
-        .into_iter()
-        .map(|row| {
-            if author_name.is_none() {
-                author_name = Some(row.get(7));
-            }
-
-            Release {
-                name: row.get(0),
-                version: row.get(1),
-                description: row.get(2),
-                target_name: row.get(3),
-                release_time: row.get(4),
-                rustdoc_status: row.get(5),
-                stars: row.get::<_, Option<i32>>(6).unwrap_or(0),
-            }
-        })
-        .collect();
-
-    (author_name.unwrap_or_default(), packages)
-}
-
 fn get_releases_by_owner(
     conn: &mut Client,
     page: i64,
     limit: i64,
-    author: &str,
+    owner: &str,
 ) -> (String, Vec<Release>) {
     let offset = (page - 1) * limit;
 
@@ -184,14 +134,14 @@ fn get_releases_by_owner(
                  WHERE owners.login = $1
                  ORDER BY github_repos.stars DESC NULLS LAST
                  LIMIT $2 OFFSET $3";
-    let query = conn.query(query, &[&author, &limit, &offset]).unwrap();
+    let query = conn.query(query, &[&owner, &limit, &offset]).unwrap();
 
-    let mut author_name = None;
+    let mut owner_name = None;
     let packages = query
         .into_iter()
         .map(|row| {
-            if author_name.is_none() {
-                author_name = Some(if !row.get::<usize, String>(7).is_empty() {
+            if owner_name.is_none() {
+                owner_name = Some(if !row.get::<usize, String>(7).is_empty() {
                     row.get(7)
                 } else {
                     row.get(8)
@@ -210,7 +160,7 @@ fn get_releases_by_owner(
         })
         .collect();
 
-    (author_name.unwrap_or_default(), packages)
+    (owner_name.unwrap_or_default(), packages)
 }
 
 /// Get the search results for a crate search query
@@ -336,7 +286,7 @@ struct ViewReleases {
     show_next_page: bool,
     show_previous_page: bool,
     page_number: i64,
-    author: Option<String>,
+    owner: Option<String>,
 }
 
 impl_webpage! {
@@ -350,7 +300,7 @@ pub(super) enum ReleaseType {
     Stars,
     RecentFailures,
     Failures,
-    Author,
+    Owner,
     Search,
 }
 
@@ -369,8 +319,8 @@ fn releases_handler(req: &mut Request, release_type: ReleaseType) -> IronResult<
             Order::FailuresByGithubStars,
         ),
 
-        ReleaseType::Author | ReleaseType::Search => panic!(
-            "The authors and search page have special requirements and cannot use this handler",
+        ReleaseType::Owner | ReleaseType::Search => panic!(
+            "The owners and search page have special requirements and cannot use this handler",
         ),
     };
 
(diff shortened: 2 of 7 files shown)


jyn514:
This is wrong, it also prevents viewing pages by the owner. It should only remove `get_releases_by_author`.

hi-rustin:
@jyn514 Do we need to rename these api and routing parameters?

templates/crate/details.html
```diff
@@ -39,15 +39,6 @@
                             </li>
                         {%- endif -%}
                         {# List the release author's names and a link to their docs.rs profile #}
```
Nemo157, line 41: This comment can go too.

Nemo157 (commented):
I think it would be good to rename where appropriate, but that seems less important to be done straight away. We also need to stop recording the data we're no longer displaying, and then delete the existing data from the database. I think it would make sense to open a local tracking issue here so those steps can be done independently.

src/web/releases.rs
```diff
@@ -439,7 +389,8 @@ pub fn author_handler(req: &mut Request) -> IronResult<Response> {
                 cexpect!(req, author.nth(1)),
             )
         } else {
-            get_releases_by_author(&mut conn, page_number, RELEASES_IN_RELEASES, author)
+            // TODO: Accurate error here, the author name is invalid
+            return Err(Nope::CrateNotFound.into());
```
hi-rustin, line 393: I would prefer to fix all the error message issues in that method in a separate PR.

templates/crate/details.html
hi-rustin: Removed.

templates/releases/header.html
```diff
@@ -9,8 +9,7 @@
         * `failures`
         * `activity`
         * `queue`
-        * `author`
-    * `author` A string, used for the authors page
+        * `author` A string, used for the authors page
 #}
 {% macro header(title, description, tab, author=false) %}
```
jyn514, line 14: Is this ever used?

src/web/releases.rs
jyn514: This seems like a really bad error message for a page that used to work ... could you add a new error variant to `Nope` that has a more descriptive message?

templates/releases/header.html
hi-rustin: I think it's still working for owner.

src/web/releases.rs
hi-rustin: There are a lot of TODOs about messages here, and I'd prefer to submit another pr to fix them all.

templates/releases/header.html
hi-rustin: image

src/web/releases.rs
jyn514: Ok, do you mind submitting that PR first then?

templates/releases/header.html
jyn514: Great, could you rename it from "author" to "owner" then?

hi-rustin: > Great, could you rename it from "author" to "owner" then?

Maybe can do it in https://github.com/rust-lang/docs.rs/issues/1323.

src/web/releases.rs
hi-rustin: OK. I will add the Nope.

src/web/error.rs
```diff
@@ -11,6 +11,8 @@ pub enum Nope {
     ResourceNotFound,
     BuildNotFound,
     CrateNotFound,
+    AuthorNotFound,
```
jyn514, line 14: This should be 'owner', not 'author'.
```suggestion
    OwnerNotFound,
```

src/web/releases.rs
```diff
@@ -439,13 +386,12 @@ pub fn author_handler(req: &mut Request) -> IronResult<Response> {
                 cexpect!(req, author.nth(1)),
             )
         } else {
-            get_releases_by_author(&mut conn, page_number, RELEASES_IN_RELEASES, author)
+            return Err(Nope::AuthorNameInvalid.into());
```
jyn514, line 389: Do you think it would be better to show the owner even without an `@` sign? That seems more helpful.

If not, the error should be more accurate and say that only owners are shown, not authors. "Invalid author" isn't very helpful for finding what went wrong.

jyn514:
Can you add a test for what happens when you try to request an owner instead of an author?

src/web/error.rs
hi-rustin: Renamed.

src/web/releases.rs
hi-rustin: > Do you think it would be better to show the owner even without an `@` sign? That seems more helpful.

Make sense.

hi-rustin:
> Can you add a test for what happens when you try to request an owner instead of an author? There are some examples near the end of src/web/releases.rs.

I think we already have `owners_page` and `owners_pagination` tests.

hi-rustin:
@ jyn514  Thank you for your review.

jyn514:
> > Can you add a test for what happens when you try to request an owner instead of an author? There are some examples near the end of src/web/releases.rs.
> 
> I think we already have `owners_page` and `owners_pagination` tests.

Sorry, I meant to say "an author instead of an owner". The existing tests are only for when the owner is found, not when there's an error.

hi-rustin:
> > > Can you add a test for what happens when you try to request an owner instead of an author? There are some examples near the end of src/web/releases.rs.
> > 
> > 
> > I think we already have `owners_page` and `owners_pagination` tests.
> 
> Sorry, I meant to say "an author instead of an owner". The existing tests are only for when the owner is found, not when there's an error.

Add `nonexistent_owner_page` for it.

src/web/releases.rs
```diff
@@ -1446,29 +1393,75 @@ mod tests {
...
+            );
+
+            assert_eq!(page.select("#crate-title").unwrap().count(), 1);
+            assert_eq!(
+                page.select("#crate-title")
+                    .unwrap()
+                    .next()
+                    .unwrap()
+                    .text_contents(),
+                "The requested crate does not exist",
```
hi-rustin, line 1421: The message is still this because of our BlockBlacklistedPrefixes.

See: https://github.com/rust-lang/docs.rs/blob/e6b702bf8fd434da63108b0ee725ab6c3df1cb2c/src/web/routes.rs#L325

hi-rustin: @jyn514 Do you have any suggestions to solve the problem? Or is this acceptable?

hi-rustin: Sorry I misunderstood this blacklist routing process, I think this should be correct.

hi-rustin:
@jyn514 I think it's ready for review.

jyn514:
> Add nonexistent_owner_page for it.

Thanks. That shows the error message was wrong: you returned the right error, but the handler ignores it, which was the long-standing bug the comments didn't describe very well. I opened https://github.com/rust-lang/docs.rs/pull/1326 fixing it, I'd prefer to wait for that before merging.

src/web/releases.rs
jyn514: This is fixed by https://github.com/rust-lang/docs.rs/pull/1326.

hi-rustin:
> Thanks. That shows the error message was wrong: you returned the right error, but the handler ignores it, which was the long-standing bug the comments didn't describe very well. I opened #1326 fixing it, I'd prefer to wait for that before merging.

Got it. Besides this issue, is there anything else that needs to be changed in this PR?

src/web/releases.rs
```diff
@@ -413,39 +363,35 @@ pub fn releases_failures_by_stars_handler(req: &mut Request) -> IronResult<Respo
...
     // page number of releases
     let page_number: i64 = router
         .find("page")
         .and_then(|page_num| page_num.parse().ok())
         .unwrap_or(1);
-    let author = router
-        .find("author")
-        // TODO: Accurate error here, the author wasn't provided
-        .ok_or(Nope::CrateNotFound)?;
+    let owner = router.find("owner").ok_or(Nope::OwnerNameNotProvided)?;
```
jyn514, line 373: This is how `owner_handler` is used in the router: https://github.com/rust-lang/docs.rs/blob/63647614a4b12d06b1b8ddf344e1cc439db05b8c/src/web/routes.rs#L44-L45
That should always have an `owner` field, or otherwise this function wouldn't be called. So I think this can just be an `unwrap()` and you can remove the `OwnerNameNotProvided` variant.

src/web/releases.rs
```diff
@@ -413,39 +363,35 @@ pub fn releases_failures_by_stars_handler(req: &mut Request) -> IronResult<Respo
...
                 page_number,
                 RELEASES_IN_RELEASES,
                 // TODO: Is this fallible?
-                cexpect!(req, author.nth(1)),
+                cexpect!(req, owner.nth(1)),
             )
         } else {
-            get_releases_by_author(&mut conn, page_number, RELEASES_IN_RELEASES, author)
+            get_releases_by_owner(&mut conn, page_number, RELEASES_IN_RELEASES, owner)
         }
```
jyn514, line 390: You can simplify this quite a bit:
```
         if owner.starts_with('@') {
             owner = &owner[1..];
         }
         get_releases_by_owner(&mut conn, page_number, RELEASES_IN_RELEASES, owner)
```

hi-rustin: Refactored.

src/web/releases.rs
hi-rustin: Removed.

jyn514:
@hi-rustin I don't plan to review this again until https://github.com/rust-lang/docs.rs/pull/1326 is merged. Feel free to ping me if I forget :)

src/web/releases.rs
hi-rustin: Fixed testing issue.

templates/rustdoc/topbar.html
```diff
@@ -90,21 +90,6 @@
                             </li>
                         </ul>
                     </div>
-
-                    {# Show the crate authors #}
```
jyn514, line 94: Could you show the owners instead?

templates/crate/details.html
```diff
@@ -38,16 +38,6 @@
                                 {%- endif -%}
                             </li>
                         {%- endif -%}
-                        {# List the release author's names and a link to their docs.rs profile #}
```
jyn514, line 41: Could you show the owners instead?

Nemo157: The owners are already shown at the bottom of this sidebar.

hi-rustin: Check https://github.com/rust-lang/docs.rs/blob/696ff69d1d8ab9b3f2794e88fa543284ad640d6b/templates/crate/details.html#L123

templates/rustdoc/topbar.html
hi-rustin: I am currently unable to check the topbar related pages because I am having problems with add-essential-files.
See: https://github.com/rust-lang/rustwide/issues/41#issuecomment-804198425

So I don't know if the whole topbar is currently displayed properly. Can you help check it? Or do you have any suggestions for solving the above problem.

Nemo157: If you have working IPv6 then https://docs.rs.dev.nemo157.com/ is running this PR currently, looks like it's working to me.

hi-rustin: Looking at the code, I should have used the correct field. https://github.com/rust-lang/docs.rs/blob/a0438c4bd81459c43640acde8db4f57e8e40db1c/src/web/crate_details.rs#L19

hi-rustin: > If you have working IPv6 then https://docs.rs.dev.nemo157.com/ is running this PR currently, looks like it's working to me.

Thanks!

jyn514 (approved):
This is great, thanks!

templates/rustdoc/topbar.html
hi-rustin: > I am having problems with add-essential-files.
> See: [rust-lang/rustwide#41 (comment)](https://github.com/rust-lang/rustwide/issues/41#issuecomment-804198425)

@jyn514  @Nemo157  Do you have any suggestions to solving this?

PR #1324, opened 2021-03-20
Give precedence to local shared files over global ones

### LINKED ISSUE #1313: Publishing a crate without dependencies causes a `empty_library` dependency.

jyn514:
When serving 'essential files', we can either serve the global one,
created when building `empty_library`, or the local one, created when
building the local crate. Currently we default to the global one, but
this causes issues when the file should never have been global in the
first place (such as recently for `crates.js`). This is why https://github.com/rust-lang/docs.rs/pull/1317 needed to be deployed to fix #1313.

This gives precedence to the local file so that the bug will be fixed
when rustdoc fixes it, even if we forget to update
`ESSENTIAL_FILES_UNVERSIONED`.

r? @Nemo157

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -171,12 +171,23 @@ impl Handler for CratesfyiHandler {
             }
         }
 
-        // try serving shared rustdoc resources first, then db/static file handler and last router
-        // return 404 if none of them return Ok. It is important that the router comes last,
-        // because it gives the most specific errors, e.g. CrateNotFound or VersionNotFound
-        self.shared_resource_handler
+        // This is kind of a mess.
+        //
+        // Almost all files should be served through the `router_handler`; eventually
+        // `shared_resource_handler` should go through the router too, and `database_file_handler`
+        // could be removed altogether.
+        //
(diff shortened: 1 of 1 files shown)


Nemo157:
The comment above was already out of date, but now it's even more out of date 😁

jyn514:
lol, thanks, I fixed the comment :) while reading it I think `database_handler` can be removed right away, it's only for the favicon but we already handle that in the router. Making a separate PR for that now.

Nemo157 (approved):

PR #1326, opened 2021-03-21
Remove `database_file_handler`; give precendence to router error messages

jyn514:
The database handler was unused; the router already serves the favicon and it
doesn't do anything else. Note that 'database handler' is different
from 'files served from storage', the router already handles those.

This also fixes the long standing bug that all 404 errors are either
'crate not found' or 'resource not found'.

Closes https://github.com/rust-lang/docs.rs/issues/55. This builds on https://github.com/rust-lang/docs.rs/pull/1324.

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -163,9 +163,6 @@ mod tests {
 
     #[test]
     fn check_404_page_content_resource() {
-        // Resources with a `.js` and `.ico` extension are special cased in the
-        // routes_handler which is currently run last. This means that `get("resource.exe")` will
-        // fail with a `no so such crate` instead of 'no such resource'
         wrapper(|env| {
             let page = kuchiki::parse_html().one(
                 env.frontend()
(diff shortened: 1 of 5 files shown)


Nemo157:
According to the `database success` route metric I see 26 requests over the last hour from the database handler. I was really hoping when I added it that it would just display 0 and we could trivially get rid of it, but I think there needs to be some more investigation into what requests these are that it is serving.

jyn514:
@Nemo157 it's serving pages like `/rustdoc/regex/1.4.3/regex/index.html`. I don't think we need to keep them around - none of the redirects work and you can get the pages just as easily by removing `/rustdoc/` at the front.

jyn514:
It also lets you have programmatic access to the build-logs (e.g. https://docs.rs/build-logs/334832/x86_64-unknown-linux-gnu.txt), but I don't think we should keep it around just for that, if we think it's useful it should be part of the router, not a completely separate handler.

Nemo157:
But how are people discovering those URLs and going to them in production? Maybe the easiest way to see if removing this breaks someone's workflow is just to merge it and see who complains 😁

Nemo157 (approved):

PR #1329, opened 2021-03-23
Revert "Don't hard-code essential files in `copy_doc_dir`"

jyn514:
This meant we stopped copying *all* JS and CSS files, which was
incorrect. Possibly there's a better way to do this, but this fixes the
immediate problem that search is broken.

Here's an example build which is broken because of the change to
`copy_doc_dir`:
```
> aws s3 ls s3://rust-docs-rs/rustdoc/tokio-eventfd/0.2.0/
                           PRE i686-unknown-linux-gnu/
                           PRE implementors/
                           PRE src/
                           PRE tokio_eventfd/
                           PRE x86_64-apple-darwin/
2021-03-23 14:25:56          9 .lock
2021-03-23 14:25:56       1691 settings.html
```

Re-opens https://github.com/rust-lang/docs.rs/issues/294.

r? @Nemo157

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -182,6 +182,20 @@ impl RustwideBuilder {
         let krate = Crate::crates_io(DUMMY_CRATE_NAME, DUMMY_CRATE_VERSION);
         krate.fetch(&self.workspace)?;
 
+        // TODO: remove this when https://github.com/rust-lang/rustwide/pull/53 lands.
+        struct Rustdoc<'a> {
+            toolchain_version: &'a str,
+        }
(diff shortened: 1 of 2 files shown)


jyn514:
I expect this to fix https://github.com/rust-lang/docs.rs/issues/1328, but haven't tested it yet.

jyn514:
(The diff looks big, but it's a line-for-line revert of 6f4477ba8350f6154085961fc0bdcf9fe661da9a.)

Nemo157 (approved):

jyn514:
> I expect this to fix #1328, but haven't tested it yet.

Oh weird, it does *not*. Looking into that now. This does fix search being broken though. 
image

PR #1330, opened 2021-03-23
Serve rust-logo.png as a shared file

jyn514:
These used to be served by the database handler. Now that it no longer
exists, handle them in the shared resource handler directly.

Fixes https://github.com/rust-lang/docs.rs/issues/1328.

r? @Nemo157

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -18,6 +18,7 @@ use iron::{
 use lol_html::errors::RewritingError;
 use router::Router;
 use serde::Serialize;
+use std::path::Path;
 
 #[derive(Clone)]
 pub struct RustLangRedirector {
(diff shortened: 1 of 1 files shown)


Nemo157:
Feels like it should also handle `.woff2`, except those already work in production somehow 🤷

Nemo157:
Ah, caching, `curl -I 'https://docs.rs/FiraSans-Regular.woff2?test'` gives a 404 while `curl -I 'https://docs.rs/FiraSans-Regular.woff?test'` is 200.

jyn514:
> Feels like it should also handle `.woff2`, except those already work in production somehow shrug

Good catch, thanks. Done.

Nemo157 (approved):

Issue #1327, opened 2021-03-22
Docs for releases during 2018-{10-12} appear unstyled


inejge:
The most recent docs.rs displays the contents for [rouille-3.0.0](https://docs.rs/rouille/3.0.0/rouille/) with no styles; only the top nav bar is styled. Tested with FF 86 and Chromium 89. When I last visited the page, in the past week or so, it displayed correctly.

Inspector shows the stylesheet links in `<head>` as relative, e.g.,

```
<link rel="stylesheet" type="text/css" href="../rustdoc-20181129-1.32.0-nightly-3e90a12a8.css" id="mainThemeStyle">
```

...and obviously doesn't  read valid CSS, since the console is full of "Ruleset ignored due to bad selector" messages. If I tweak the links to absolute like this:

```
<link rel="stylesheet" type="text/css" href="/rustdoc-20181129-1.32.0-nightly-3e90a12a8.css" id="mainThemeStyle">
```

...the styling appears, but will of course again be absent if I navigate away to any subpage.

jyn514:
I can't replicate this - can you post a screenshot of what the page looks like for you?

![image](https://user-images.githubusercontent.com/23638587/112060882-2299d800-8b34-11eb-957f-623206bf2f47.png)

inejge:
![image](https://user-images.githubusercontent.com/1049870/112063198-4e7f8280-8b61-11eb-8513-dc98b964577e.png)

jyn514:
Can you pull up the network console and show what file it's trying to load? Mine is loading `https://docs.rs/rouille/3.0.0/rustdoc-20181129-1.32.0-nightly-3e90a12a8.css` which succeeds.

inejge:
For that particular URL (the others behave similarly), first I get a 302 which pastes an extra `/rouille` on the response in `Location`, resulting in a subsequent 404. Copy/paste from the network console:

```
{
	"GET": {
		"scheme": "https",
		"host": "docs.rs",
		"filename": "/rouille/3.0.0/rustdoc-20181129-1.32.0-nightly-3e90a12a8.css",
		"remote": {
			"Address": "52.84.109.107:443"
		}
	}
}
{
	"Status": "302Found",
	"Version": "HTTP/2",
	"Transferred": "2.79 KB (8.98 KB size)",
	"Referrer Policy": "strict-origin-when-cross-origin"
}
{
	"Response Headers (460 B)": {
		"headers": [
			{
				"name": "content-length",
[...]

PR #1331, opened 2021-03-23
Add hack so that docs from 2018-11-29 are styled correctly

jyn514:
See comments in code. Fixes https://github.com/rust-lang/docs.rs/issues/1327.

r? @Nemo157


jyn514:
Ugh. The failure is real, this approach won't work.

```
thread 'web::rustdoc::test::default_target_redirects_to_base' panicked at '/dummy/0.3.0/index.html: expected redirect to /dummy/0.3.0/dummy/, got 404 Not Found', src/test/mod.rs:81:13
```

Issue #1303, opened 2021-03-05
`go run` fails


joshlf:
**Crate name:** mundane
**Build failure link:** https://docs.rs/crate/mundane/0.4.4/builds/357678

**Additional details:** Based on the following log output, I believe what's happening is that, since the `GOCACHE` environment variable is not set, `go run` is defaulting to `$HOME/.cache`, and I'm guessing that `$HOME` is set to `/`. This means that `go run` tries creating `/.cache`, which it doesn't have permission to do. See [here](https://golang.org/cmd/go/#hdr-Build_and_test_caching) for more details on Go's build cache.

```text
[INFO] [stderr]   cd /opt/rustwide/workdir/boringssl/boringssl/crypto/err && /usr/bin/go run err_data_generate.go > /opt/rustwide/target/debug/build/mundane-6275d256bd1e392c/out/boringssl/build_1/crypto/err_data.c
[INFO] [stderr]   failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
```

jyn514:
Most of the build environment is read-only. You could try setting GOCACHE to a subdirectory of the target directory in your build script.

joshlf:
My thinking is that it would make sense to set GOCACHE to some reasonable value for all docs.rs builds; I can't think of a reason that it would need to be specific to Mundane?

jyn514:
Hmm, I wonder if it makes sense rather than fixing GOCACHE specifically to set HOME to a temporary directory.

joshlf:
That makes sense too, although I'd be worried about that having more wide-ranging implications (much more software uses `HOME` than `GOCACHE`), so I feel like just setting `GOCACHE` would be safer.

jyn514:
Note that this is blocked on a rustwide update one way or another: https://github.com/rust-lang/docs.rs/pull/1278

joshlf:
I'm not sure I understand why the change we're discussing here is blocked on #1278? Is it just that all docs.rs deployments are blocked on it?

jyn514:
Docs.rs doesn't set any of these variables itself, they're set through rustwide. Rustwide 0.12 has switched to tokio 1.0, and I'm not willing to have over 600 dependencies just to fix this bug.

joshlf:
Ah, I understand, gotcha.

jyn514:
@joshlf go fails even with `HOME` set to a writeable directory:
```
[...]

PR #1264, opened 2021-01-24
Upgrade rusoto (and tokio)

coolreader18:
Fixes #1102

The new rusoto version also updates tokio, so this is a WIP of upgrading that.

Currently waiting on rust-lang/rustwide#51

PR #1332, opened 2021-03-24
Reland rustwide and tokio upgrade

jyn514:
The cause of the leak was that `Regex::new` leaked 2.5 MB of memory on
every build (in `parse_rustc_version`). Searching around I found
https://github.com/rust-lang/regex/issues/742, which said this was a
regression in thread_local.  Updating thread_local to 1.1.3 fixed the
issue.

For future reference: I debugged the leak as follows.

1. Build the crate: `cargo build`
2. Run the daemon under massif without registry watching: `valgrind --tool=massif target/debug/cratesfyi daemon --registry-watcher disabled`
3. Add a whole bunch of crates to the queue: `for i in $(seq 1 100); do cargo run -q queue add serde 1.0.$i; done`
4. Stop the daemon with SIGINT when you get tired of waiting (I ran about 30 builds).
5. Run `massif-visualizer` on the generated file.

Successor to https://github.com/rust-lang/docs.rs/pull/1264. Fixes https://github.com/rust-lang/docs.rs/issues/1102. Unblocks https://github.com/rust-lang/docs.rs/issues/1303.

This is a WIP because I tested the leaks with an ancient version of the lockfile, not the one in this PR; I need to test again before this merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,7 +122,7 @@ dependencies = [
 name = "badge"
 version = "0.3.0"
 dependencies = [
- "base64 0.12.1",
+ "base64 0.13.0",
  "once_cell",
  "rusttype",
 ]
@@ -677,7 +677,7 @@ dependencies = [
  "arc-swap",
  "backtrace",
  "badge",
- "base64 0.12.1",
+ "base64 0.13.0",
  "chrono",
  "comrak",
  "crates-index",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -727,6 +727,7 @@ dependencies = [
  "systemstat",
  "tempfile",
  "tera",
+ "thread_local",
  "time 0.1.43",
  "tokio",
  "toml",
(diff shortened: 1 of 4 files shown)


jyn514:
Ok cool, I ran about 80 builds with no spikes.
image

Nemo157:
Code looks good, I'm setting up an overnight test running builds to see if I get any leaks.

Nemo157:
I ran 200 builds on the server running in the docker-compose setup, checking the rss usage after each build:

```sh
for i in $(seq 1 200); do
  cargo run -- queue add bs58 0.4.0 2>/dev/null
  sleep 180
  ps -C cratesfyi -o rss --no-headers
done
```

image

There is what looks like a slight leak, but it might just be heap fragmentation or something. It's the same as what I previously saw when testing e8a6303, I didn't record the full data the same as this, but at that version after 10 builds it was at 69MB and after 100 builds was at 173M, which are right on the line.

Comparatively when testing 48444ef at 10 builds it had reached 109MB and 100 builds it had reached 530MB, so significantly more leakage.

Nemo157 (approved):

PR #1333, opened 2021-03-26
Add a Content Security Policy to non-rustdoc pages

pietroalbini:
This PR adds a Content Security Policy to all pages except for rustdoc pages, increasing the security of the website. This *should* not break anything, but in case something breaks there is no need to revert the PR: setting the `DOCSRS_CSP_REPORT_ONLY` environment variable to `true` will switch the CSP to be "report-only", notifying breakages in the console without actually blocking any content.

There are a couple of changes here that will affect how docs.rs is developed (cc @rust-lang/docs-rs):

* Inline styles with `style=""` are not allowed anymore. Using CSS classes is required instead.
* Any JavaScript will need to have the `nonce="{{ csp_nonce }}"` HTML attribute for it to work.
* `eval()` inside JavaScript doesn't work anymore (we weren't using it before, but listing that just to be sure).

The best way to test this is to start the server locally and see if anything breaks while visiting pages.

r? @jyn514


jyn514:
```
error: name `IO` contains a capitalized acronym
Error:   --> crates/metadata/lib.rs:74:5
   |
74 |     IO(#[from] io::Error),
   |     ^^ help: consider making the acronym lowercase, except the initial letter (notice the capitalization): `Io`
   |
   = note: `-D clippy::upper-case-acronyms` implied by `-D warnings`
```
Globally allowing that error in `crates/metadata` seems fine to me.

jyn514:
@pietroalbini this causes `utils.js` to fail to load:

image

jyn514:
There's also an error on /features pages (e.g. http://localhost:3000/crate/konst/0.2.0/features).
image
I can't seem to figure out how to get firefox to tell me what line it's on, though.

pietroalbini:
Fixed the features page, I'm not exactly sure where you see the `utils.js` error though.

jyn514:
> Fixed the features page, I'm not exactly sure where you see the `utils.js` error though.

I think this might be something injected by firefox devtools, it doesn't seem to be affecting the site at all.

syphar:
> > Fixed the features page, I'm not exactly sure where you see the `utils.js` error though.
> 
> I think this might be something injected by firefox devtools, it doesn't seem to be affecting the site at all.

Short note here (without having read the code)
- chrome and safari don't show the error, so it seems to be related to FF (and IMHO dev tools)
- I do see an error when checking rustdoc pages.

src/web/rustdoc.rs
```diff
@@ -253,6 +253,12 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     let metrics = extension!(req, Metrics).clone();
     let mut rendering_time = RenderingTimesRecorder::new(&metrics.rustdoc_rendering_times);
 
+    // Pages generated by Rustdoc are not ready to be served with a CSP yet.
+    req.extensions
+        .get_mut::<Csp>()
+        .expect("missing CSP")
+        .suppress(true);
```
jyn514, line 260: This is shared between all threads, isn't it? Won't this surpress the CSP policy for all requests? Or is it per-request because you used `link_before`?

src/web/csp.rs
```diff
@@ -0,0 +1,189 @@
...
+        result.push_str("; style-src 'self'");
+
+        // Allow loading any font from the current origin.
+        result.push_str("; font-src 'self'");
+
+        // Only allow scripts with the random nonce attached to them.
+        //
+        // We can't just allow 'self' here, as users can upload arbitrary .js files as part of
+        // their documentation and 'self' would allow their execution. Instead, every allowed
+        // script must include the random nonce in it, which an attacker is not able to guess.
```
jyn514, line 66: Ahhh, I just realized while reading this that there's a new nonce *for each request*, it's not a hash of the file or anything like that. Maybe add a comment to the CSP saying that? And a test that two sequential requests get different nonces?

jyn514 (commented):
I clicked around some more and everything seems to be working now :)

Nemo157:
I see a few errors from extensions injected scripts, that might be where `utils.js` is coming from too, I don't get that specific one in FF devtools.

src/web/csp.rs
```diff
@@ -0,0 +1,189 @@
...
+
+pub(super) struct Csp {
+    nonce: String,
+    suppress: bool,
+}
+
+impl Csp {
+    fn new() -> Self {
+        let mut random = [0u8; 36];
+        getrandom::getrandom(&mut random).expect("failed to generate a nonce");
```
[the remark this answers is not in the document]
pietroalbini: Yes.

src/web/csp.rs
pietroalbini: [Hmm, the test is already there](https://github.com/rust-lang/docs.rs/pull/1333/files#diff-6f288f9159b4a7665a0894ea5bbfb4de699bebe2de84a24b423895e0e0daae09R138-R143). I'll add a comment.

src/web/rustdoc.rs
pietroalbini: Well, all the extensions [are added in each request](https://github.com/rust-lang/docs.rs/blob/master/src/web/extensions.rs). Most of the others are just cloned `Arc<_>`s, instead here it's a fresh instance every time.

src/web/csp.rs
```diff
@@ -0,0 +1,189 @@
...
+        Self {
+            nonce: base64::encode(&random),
+            suppress: false,
+        }
+    }
+
+    pub(super) fn suppress(&mut self, suppress: bool) {
+        self.suppress = suppress;
+    }
```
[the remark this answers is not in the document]
pietroalbini: Mostly out of habit, and because this looks weird:

```rust
     req.extensions
        .get_mut::<Csp>()
        .expect("missing CSP")
        .suppress = true; 
```

pietroalbini:
Rebased and addressed all review commentd.

PR #1335, opened 2021-03-26
Fix 1.51 Clippy lints

pietroalbini:
r? @jyn514

### DIFF
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -67,6 +67,7 @@ pub const DEFAULT_TARGETS: &[&str] = &[
 
 /// The possible errors for [`Metadata::from_crate_root`].
 #[derive(Debug, Error)]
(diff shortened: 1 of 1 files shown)


crates/metadata/lib.rs
```diff
@@ -67,6 +67,7 @@ pub const DEFAULT_TARGETS: &[&str] = &[
 
 /// The possible errors for [`Metadata::from_crate_root`].
 #[derive(Debug, Error)]
+#[allow(clippy::upper_case_acronyms)]
```
jyn514, line 70: There's a PR open fixing this in clippy, I think I'd rather wait for that:
https://github.com/rust-lang/rust-clippy/pull/6981

jyn514: Oh wait, we use the stable toolchain, not nightly. That's unfortunate, I guess we'll have to add this for the 12 weeks until this hits stable.

PR #1343, opened 2021-04-04
Improve error message when feature flags aren't available

jyn514:
Suggested by `@malaire` on discord.

### DIFF
--- a/templates/crate/features.html
+++ b/templates/crate/features.html
@@ -40,7 +40,7 @@
                             </li>
                         {%- else -%}
                             <li class="pure-menu-item">
-                                <span class="documented-info">Feature flags data are not available for this release.</span>
+                                <span class="documented-info">Feature flags are not available for this release because it was built before features were collected by docs.rs.</span>
                             </li>
                         {%- endif -%}
                     </ul>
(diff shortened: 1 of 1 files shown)


pietroalbini:
The change itself is great, but there is another copy of the same paragraph below that needs to be updated.

Ideally we should remove the duplicate text, but that can be done in a later PR.

templates/crate/features.html
```diff
@@ -40,7 +40,7 @@
                             </li>
                         {%- else -%}
                             <li class="pure-menu-item">
-                                <span style="font-size: 13px;">Feature flags data are not available for this release.</span>
+                                <span style="font-size: 13px;">Feature flags data are not available for this release because it was built before features were collect by docs.rs.</span>.
```
Nemo157, line 43:
```suggestion
                                <span style="font-size: 13px;">Feature flags data are not available for this release because it was built before features were collected by docs.rs.</span>.
```

(The usage of plural data really threw me off, but I think it's actually correct).

jyn514: Happy to remove "data", I don't think it adds anything.

Nemo157 (approved):

Issue #1347, opened 2021-04-08
500 internal error viewing https://docs.rs/crate/mbedtls/0.6.0/source/


jyn514:
```
 note: while attempting to fetch the route Url { generic_url: "http://docs.rs/crate/mbedtls/0.6.0/source/" }
    0: docs_rs::web::page::web_page::WebPage::into_response
    1: <docs_rs::web::error::Nope as iron::middleware::Handler>::handle
    2: iron::middleware::Chain::continue_from_before
    3: <iron::middleware::Chain as iron::middleware::Handler>::handle
    4: <docs_rs::web::CratesfyiHandler as iron::middleware::Handler>::handle
    5: <iron::iron::RawHandler<H> as hyper::server::Handler>::handle
    6: hyper::server::Worker<H>::handle_connection
    7: hyper::server::listener::spawn_with::{{closure}}
    8: std::sys_common::backtrace::__rust_begin_short_backtrace
    9: core::ops::function::FnOnce::call_once{{vtable.shim}}
   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
              at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/liballoc/boxed.rs:1076
       <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
              at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/liballoc/boxed.rs:1076
       std::sys::unix::thread::Thread::new::thread_start
              at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/sys/unix/thread.rs:87
   11: start_thread
   12: __clone
```

Issue #1344, opened 2021-04-05
500 internal error on docs.rs source view with semver


jyn514:
https://docs.rs/crate/regex/*/source/Cargo.toml gives a 500 internal error. Probably it's either not calling match_version or handling the return value wrong.

_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/pull/1342#discussion_r607136006_

PR #1350, opened 2021-04-13
Handle 404s and semver in source file viewer

jyn514:
This also changes `ctry!` to print the debug version of errors so it's possible to see what goes wrong when a template fails to render.

Closes https://github.com/rust-lang/docs.rs/issues/1344. Closes https://github.com/rust-lang/docs.rs/issues/1347.

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -16,7 +16,7 @@ macro_rules! ctry {
                 let request: &::iron::Request = $req;
 
                 ::log::error!(
-                    "called ctry!() on an `Err` value: {}\nnote: while attempting to fetch the route {:?}\n{:?}",
+                    "called ctry!() on an `Err` value: {:?}\nnote: while attempting to fetch the route {:?}\n{:?}",
                     error,
                     request.url,
                     ::backtrace::Backtrace::new(),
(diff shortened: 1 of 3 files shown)


src/web/source.rs
```diff
@@ -217,8 +243,8 @@ pub fn source_browser_handler(req: &mut Request) -> IronResult<Response> {
         (None, false)
     };
 
-    let file_list =
-        FileList::from_path(&mut conn, &name, &version, &req_path).ok_or(Nope::NoResults)?;
+    let file_list = FileList::from_path(&mut conn, &crate_name, &version, &req_path)
+        .ok_or(Nope::ResourceNotFound)?;
```
jyn514, line 247: This is what fixed the 500 error; before it was treating this as a search page and panicking because variables weren't set.

src/web/mod.rs
```diff
@@ -16,7 +16,7 @@ macro_rules! ctry {
                 let request: &::iron::Request = $req;
 
                 ::log::error!(
-                    "called ctry!() on an `Err` value: {}\nnote: while attempting to fetch the route {:?}\n{:?}",
+                    "called ctry!() on an `Err` value: {:?}\nnote: while attempting to fetch the route {:?}\n{:?}",
```
jyn514, line 19: Without this it only shows the name of the template page that failed to render, not what the error actually was.

Nemo157 (approved):

PR #1352, opened 2021-04-14
remove css font override for div.rustdoc

### LINKED ISSUE #1349: Incorrect font used on latest docs
At some point recently (within a few weeks) rustdoc updated from using _Source Serif Pro_ to _Source Serif 4_ for the body text. This is being overridden by docs.rs' css on the `rustdoc_body_wrapper` element resulting it in needing to fallback to system fonts, rather than using any _Source Serif_ font.

tspiteri:
Fixes #1349.

### DIFF
--- a/templates/style/rustdoc.scss
+++ b/templates/style/rustdoc.scss
@@ -30,7 +30,6 @@ body.blur {
 // rustdoc overrides
 div.rustdoc {
     $padding-x: 15px;
-    font-family: $font-family-serif;
     padding: 10px $padding-x 20px;
     position: relative;
 


tspiteri:
"Source Serif Pro"  is retained after "Source Serif 4" because `$font-family-serif` may be used in other files which make use of the files inside the docs.rs project's /static, which still contain Source Serif Pro.

Nemo157:
Another option might be to delete the override from our `rustdoc.css`

https://github.com/rust-lang/docs.rs/blob/f3a3740a8f60e000f041ebd05d973001cd2d4de4/templates/style/rustdoc.scss#L33

I'm not sure why we have this, it might be necessary for older rustdoc output, but checking some random docs from 2016 and 2018 deleting that line works fine.

Nemo157:
It comes from the initial commit of the UI

https://github.com/rust-lang/docs.rs/commit/f59bc5f030ef9d5e220bd6503c9e5fa5efbb6264#diff-1ba7be47f68c69933566610bc4a27b54149989a893f733c0a233fd315ad2a19dR61

I think it's a good time to check whether it's needed by deleting it and seeing if we get any complaints :sweat_smile:

tspiteri:
Should I change the PR to remove the override instead?

Nemo157:
@tspiteri yep, I think it's worth trying out to simplify issues if rustdoc changes its fonts again.

tspiteri:
Done.

Nemo157 (approved):

PR #1358, opened 2021-04-17
Put back warning when rate limit has been reached on a specific updater but do not discard other updaters

GuillaumeGomez:


src/repositories/updater.rs
```diff
@@ -152,7 +152,19 @@ impl RepositoryStatsUpdater {
             // FIXME: The collect can be avoided if we use Itertools::chunks:
             // https://docs.rs/itertools/0.10.0/itertools/trait.Itertools.html#method.chunks.
             for chunk in needs_update.chunks(updater.chunk_size()) {
-                let res = updater.fetch_repositories(chunk)?;
+                let res = match updater.fetch_repositories(chunk) {
+                    Ok(r) => r,
+                    Err(err) => {
+                        if err.downcast_ref::<RateLimitReached>().is_some() {
+                            warn!(
+                                "rate limit reached, blocked the `{}` repository stats updater",
```
jyn514, line 160: "Blocked" doesn't seem correct, it will still collect stats the next time you run it. Maybe "ignored" is better?

GuillaumeGomez: I took the original message. Maybe "stopped"?

jyn514: "skipping" maybe - that seems to imply it's temporary.

GuillaumeGomez: Let's go for skipping then!

GuillaumeGomez:
Updated!

jyn514 (approved):

PR #1367, opened 2021-04-18
Add footer

### LINKED ISSUE #75: Advertising and tracking policy
In the "About docs.rs" page, please:
1. Document your tracking and advertising policy on pages that are generated from crates.
2. Document how a crate author can avoid advertising and other tracking appearing on docs.rs pages generated from their content.

GuillaumeGomez:
Fixes #75 

Screenshot from 2021-04-21 14-55-57
Screenshot from 2021-04-21 15-28-09

As you can see, whatever the size of the page, it's always at the bottom and not in the middle of the screen but if the content if bigger that the height, you have to scroll to the bottom of the page to be able to see the footer.

Little sidenode for the rustdoc pages in mobile mode. In such case, the footer takes the full width and goes under the sidebar in case it's displayed:

Screenshot from 2021-04-21 15-27-56
Screenshot from 2021-04-21 15-28-00

### DIFF
--- a/templates/footer.html
+++ b/templates/footer.html
@@ -1,0 +1,5 @@
+<div class="docs-rs-footer">
+    <a href="/about">About docs.rs</a>
+    <a href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">Privacy policy</a>
+    <a href="/releases/queue">Build queue</a>
+</div>
--- a/templates/style/rustdoc.scss
+++ b/templates/style/rustdoc.scss
@@ -1,5 +1,10 @@
 // FIXME: Use modules
-@import "vars", "navbar", "themes", "fa";
+@import "vars", "navbar", "themes", "fa", "footer";
+
+// This rule is needed to be sure that the footer will always be at the bottom of the page.
+body.rustdoc-page {
+    min-height: 100vh;
+}
 
 // Force the navbar to be left-aligned on rustdoc pages
 body.rustdoc-page > .nav-container > .container {
@@ -8,6 +13,34 @@ body.rustdoc-page > .nav-container > .container {
 
 div.container-rustdoc {
     text-align: left;
+
+    > .docs-rs-footer {
+        bottom: -20px;
+    }
+}
+
+div.container-rustdoc:not(.source) {
+    > .docs-rs-footer {
+        right: -15px;
+        width: calc(100vw - 212px);
+    }
+
+    // This is when the rustdoc sidebar "disappears" (for mobile mode).
+    @media (max-width: 700px) {
+        > .docs-rs-footer:not(.source) {
+            width: 100vw;
+        }
+    }
+}
+
+div.container-rustdoc.source {
+    > .docs-rs-footer {
+        width: 100vw;
+        left: -15px;
+        // This is needed because even though the sidebar only contains the header, it still takes
+        // all the height.
+        z-index: 1;
+    }
 }
 
 // this is a super nasty override for help dialog in rustdocs
--- a/templates/base.html
+++ b/templates/base.html
@@ -29,6 +29,7 @@
         {%- block header %}{% endblock header -%}
 
         {%- block body -%}{%- endblock body -%}
+        {%- include "footer.html" -%}
 
         <script type="text/javascript" nonce="{{ csp_nonce }}" src="/-/static/menu.js?{{ docsrs_version() | slugify }}"></script>
         <script type="text/javascript" nonce="{{ csp_nonce }}" src="/-/static/index.js?{{ docsrs_version() | slugify }}"></script>
--- a/templates/core/about/badges.html
+++ b/templates/core/about/badges.html
@@ -5,7 +5,7 @@
 {%- block body -%}
 	<h1>Badges</h1>
 
-	<div>
+	<div class="about-page">
 	<div class="container pure-u-5-6 about">
 	<p>
 		You can use badges to show state of your documentation to your users.
--- a/templates/core/about/builds.html
+++ b/templates/core/about/builds.html
@@ -5,7 +5,7 @@
 {%- block body -%}
     {%- set docsrs_repo = "https://github.com/rust-lang/docs.rs" -%}
     <h1>Builds</h1>
-    <div>
+    <div class="about-page">
     <div class="container pure-u-5-6 about">
     <p>
         Docs.rs automatically builds documentation for crates released on <a href="https://crates.io/">crates.io</a>.
--- a/templates/core/about/index.html
+++ b/templates/core/about/index.html
@@ -6,7 +6,7 @@
     {%- set docsrs_repo = "https://github.com/rust-lang/docs.rs" -%}
 
     <h1 id="crate-title">About Docs.rs</h1>
-    <div>
+    <div class="about-page">
     <div class="container pure-u-5-6 about">
         <p>
             Docs.rs is an
--- a/templates/core/about/metadata.html
+++ b/templates/core/about/metadata.html
@@ -5,7 +5,7 @@
 {%- block body -%}
 	<h1>Metadata for custom builds</h1>
 
-	<div>
+	<div class="about-page">
 	<div class="container pure-u-5-6 about">
 	<p>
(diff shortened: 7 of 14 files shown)


jyn514:
I wonder if it makes sense to move the 'About' and homepage links in the bottom. 'About' because it seems common to put that in the footer, and the homepage because to be frank there's not much on our homepage other than search, which we already show in the top-right. That would also free up some space in the header.

templates/footer.html
```diff
@@ -0,0 +1,3 @@
+<div class="docs-rs-footer">
+    Who wants some content?
```
jyn514, line 2:
```suggestion
    <a href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">Privacy</a>
```

templates/style/rustdoc.scss
```diff
@@ -1,13 +1,29 @@
 // FIXME: Use modules
 @import "vars", "navbar", "themes", "fa";
 
+body.rustdoc-age {
```
Nemo157, line 4:
```suggestion
body.rustdoc-page {
```

I assume by the next rule?

GuillaumeGomez: Woups, indeed!

templates/footer.html
GuillaumeGomez: It can't be the only thing we will put in there, right? Do we also want to add something like "docs.rs is owned by the Rust foundation" or something along the line?

jyn514: I don't think we need to? I would only put things there that are useful things people might want to click, I don't think we should add links just to take up space.

GuillaumeGomez: Ok!

Nemo157:
Is there a reason to not make the footer fullwidth on rustdoc pages?

The other page it'd be useful to verify it works on is rustdoc-src pages, since they ~~don't have the sidebar~~ have a collapsible sidebar.

GuillaumeGomez:
@Nemo157 You'll see in a bit. ;)

GuillaumeGomez:
Updated the images and the PR. With this, I think it's now ready for review.

Nemo157:
Ah, so the footer scrolls into view on long rustdoc pages, so that's why it can't be across both the page and sidebar. Still seems a bit weird in that it implies that it's part of the rustdoc output; maybe having some kind of `DOCS.RS` wordmark in the footer would help avoid that.

On rustdoc-src pages it doesn't extend all the way left:

image

templates/footer.html
```diff
@@ -0,0 +1,4 @@
+<div class="docs-rs-footer">
+    <a href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">Privacy</a>
```
Nemo157, line 2: I think _Privacy_ on its own looks a bit weird, _Privacy policy_ would probably be better (and is what crates.io uses).

GuillaumeGomez: :+1:

GuillaumeGomez:
Fixed the position on the source code page and updated "Privacy" to "Privacy policy".

templates/footer.html
```diff
@@ -0,0 +1,4 @@
+<div class="docs-rs-footer">
+    <a href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">Privacy policy</a>
+    <a href="https://foundation.rust-lang.org/">Rust Foundation</a>
```
pietroalbini, line 3: Hmm, is there a reason to link to the foundation's home page here?

GuillaumeGomez: Just policy seemed a bit weird. So I just said to myself: "why not?".

jyn514: @GuillaumeGomez I would rather you moved the 'about' links to the footer instead: https://github.com/rust-lang/docs.rs/pull/1367#issuecomment-822051232

GuillaumeGomez: Fine by me! Gonna need a little reword on the top navbar resize then.

GuillaumeGomez: After taking a look, I don't think we should move "about" from the topbar: it provides information that are required when you want to customize your "docs.rs experience". It's not "extra info" like what we put in the footer.

GuillaumeGomez: If the issue is the "Rust foundation" link, I can simply remove it. :)

pietroalbini: Honestly I would remove both the "About" and "Releases" links from the navbar, and have the footer be:

```
About docs.rs | Privacy Policy | Build queue
```

I looked at the stats we have on Grafana for the past 7 days:

* The whole about section got 0.0045% of the visits
* The whole release section except for the build queue got 0.0023% of the visits
* The build queue page got 0.0008% of the visits

I don't think either of the links in the navbar are that used to warrant their positioning.

pietroalbini:
Another thing I noticed is that the footer on rustdoc pages uses the rustdoc serif font, could it be switched to use docs.rs's main font?

GuillaumeGomez:
@pietroalbini Wow, you have good eye. Didn't spot this one at all!

templates/footer.html
GuillaumeGomez: Ok, let's go with that then.

GuillaumeGomez:
Updated the links, added icons and updated the screenshots.

templates/footer.html
```diff
@@ -0,0 +1,14 @@
+<div class="docs-rs-footer">
+    <a href="/about">
+        <span title="About">{{ "info-circle" | fas }}</span>
+        <span class="title">About docs.rs</span>
+    </a>
+    <a href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">
+        <span title="Privacy">{{ "shield-alt" | fas }}</span>
+        <span class="title">Privacy policy</span>
```
jyn514, line 8: Why add a separate span instead of using `<a title="Privacy policy>`?
```suggestion
    <a title="Privacy policy" href="https://foundation.rust-lang.org/policies/privacy-policy/#docs.rs">
        {{ "shield-alt" | fas }} Privacy policy
```

GuillaumeGomez: It's done like this in the navbar so I replicated the same behavior.

pietroalbini:
Hmm, looking at the screenshots the icons feels too prominent? I think I prefer the icon-less version.

GuillaumeGomez:
So before removing them again, what do you think @rust-lang/docs-rs ? :)

GuillaumeGomez:
Removed the icons and updated the screenshots.

pietroalbini:
The font in rustdoc pages still seems to be serif. Otherwise that looks great!

GuillaumeGomez:
@pietroalbini You're absolutely right! I updated the font so it's now sans-serif. :)

Nemo157:
Found one fun bug, the footer renders at the bottom of the initial page load and scrolls with the content on non-rustdoc pages.

image

GuillaumeGomez:
Nice!

GuillaumeGomez:
So there were actually two issues:
1. CSS `position` is never `relative` by default (uuuurg) so I had to set it as such on `<body>`.
2. Some docs.rs pages were using `margin-bottom` or didn't have bottom margin at all, so I used `padding-bottom` in both cases, so on pages that didn't use bottom margin, the end of the page won't be right under the text. It's actually a nice improvement, I never paid attention to that before. :)

Nemo157:
Looks like the about pages could do with a little extra `padding-bottom` too:

image

GuillaumeGomez:
I don't have the issue on the `/builds` page:

Screenshot from 2021-04-27 10-50-53

Nemo157:
`/<crate>/builds` is different from `/about/builds`

GuillaumeGomez:
Indeed, also I confirmed that this issue only appeared on mobile.

GuillaumeGomez:
Fixed! So I applied @Nemo157's suggestion and added bottom padding on the about pages (which forced me to update the DOM and create a new CSS class to correctly handle it).

Nemo157 (approved):
Styling all looks good to me now, I don't have much of an opinion on what links are in it.

syphar (approved):
I did a round of tests, in Firefox/Chrome and in mobile sizes. 

All seems fine to me

Issue #244, opened 2018-10-19
Rename the remaining mentions of cratesfyi to docs-rs, docs_rs or similar.


eddyb:

Nemo157:
Also, there's an inconsistency between some environment variables being `DOCS_RS_*` and others being `DOCSRS_*`, would be good to standardize these.

alb:
Which one should be used? Is it `DOCS_RS_*` or `DOCSRS_*`?

jyn514:
I don't have a strong preference. I think more uses DOCS_RS currently so let's go with that.

Nemo157:
I think it's actually the opposite

```console
> rg --hidden '(DOCSRS|DOCS_RS)_[A-Z_]+' --only-matching --no-heading --no-filename --no-line-number | LC_ALL=C sort -u
DOCSRS_BUILD_ATTEMPTS
DOCSRS_LOG
DOCSRS_MAX_FILE_SIZE
DOCSRS_MAX_FILE_SIZE_HTML
DOCSRS_MAX_POOL_SIZE
DOCSRS_MIN_POOL_IDLE
DOCS_RS_BUILD_CPU_LIMIT
DOCS_RS_DOCKER
DOCS_RS_LOCAL_DOCKER_IMAGE
```

PR #1373, opened 2021-04-20
 Standardize environment variables on DOCSRS_

jyn514:
This changes the following env variables used in prod:

- CRATESFYI_PREFIX
- CRATESFYI_DATABASE_URL
- CRATESFYI_GITHUB_ACCESSTOKEN
- CRATESFYI_RUSTWIDE_WORKSPACE
- CRATESFYI_TOOLCHAIN
- DOCS_RS_BUILD_CPU_LIMIT

Before deploying, we should set the new names of all these variables.
After deploying, we can delete the old names.

Helps with https://github.com/rust-lang/docs.rs/issues/244.


jyn514:
@pietroalbini do you have time to review this?

pietroalbini:
I'm wondering if we should have some code that errors out if the old environment variables are present, pointing to the new ones. That would help with people's local development environments, especially ones that are used occasionally.

jyn514:
@pietroalbini we can't do that until this is deployed because the production server needs to be able to switch between the two, there's no way to switch the environment and update the code at the same time currently. Happy to turn it into a hard error in a future PR.

pietroalbini:
Sure, sounds good to have it in a followup PR!

Issue #1353, opened 2021-04-15
Allow passing custom args to cargo in builds


jyn514:
Right now, `.cargo/config` is stripped before the build, so there's no way to pass custom arguments to `cargo` during the build. It would be helpful to add `cargo-args` to the metadata in order to support flags like `-Zbuild-std=std,alloc`.

_Originally posted by @Nemo157 in https://github.com/rust-lang/docs.rs/issues/1299#issuecomment-790964243_

jyn514:
Mentoring instructions:
1. Add a new `cargo_args` field to `Metadata`, similar to rustdoc_args: https://github.com/rust-lang/docs.rs/blob/f6aef6deff4be50cc4ba2af445da5a4c016f2479/crates/metadata/lib.rs#L129-L131.
2. Add it to `cargo_args()`: https://github.com/rust-lang/docs.rs/blob/f6aef6deff4be50cc4ba2af445da5a4c016f2479/crates/metadata/lib.rs#L239
3. Document it on /about/builds: https://github.com/rust-lang/docs.rs/blob/f6aef6deff4be50cc4ba2af445da5a4c016f2479/templates/core/Cargo.toml.example

PR #1374, opened 2021-04-20
Add support for cargo-args in package metadata

jyn514:
Closes https://github.com/rust-lang/docs.rs/issues/1353

### DIFF
--- a/templates/core/Cargo.toml.example
+++ b/templates/core/Cargo.toml.example
@@ -42,3 +42,8 @@ rustc-args = ["--example-rustc-arg"]
 
 # Additional `RUSTDOCFLAGS` to set (default: [])
 rustdoc-args = ["--example-rustdoc-arg"]
+
+# List of command line arguments for `cargo`.
+#
+# These cannot be a subcommand, they may only be options.
+cargo-args = ["-Z", "build-std"]
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -129,6 +129,12 @@ pub struct Metadata {
     /// List of command line arguments for `rustdoc`.
     #[serde(default)]
     rustdoc_args: Vec<String>,
+
+    /// List of command line arguments for `cargo`.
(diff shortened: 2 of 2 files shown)


templates/core/Cargo.toml.example
```diff
@@ -42,3 +42,8 @@ rustc-args = ["--example-rustc-arg"]
 
 # Additional `RUSTDOCFLAGS` to set (default: [])
 rustdoc-args = ["--example-rustdoc-arg"]
+
+# List of command line arguments for `cargo`.
+#
+# These cannot be a subcommand, they may only be options.
```
Nemo157, line 48: Maybe mention that these will be passed to multiple subcommands so need to be generally applicable.

jyn514: Is that true? I thought we only ever use the `rustdoc` subcommand.

I can add a disclaimer that we don't guarantee the subcommand and you should make these independent, but I don't know how useful that is in practice - they can't *truly* be independent because e.g. `cargo new` won't have any useful options, and I think almost everything else is shared between `rustdoc` and `doc`.

Nemo157: We use `cargo metadata` too I think.

jyn514: Yes, but we don't pass user-defined args there: https://github.com/rust-lang/docs.rs/blob/20aa7060349175424e915368790a5f064bfc4ad7/src/utils/cargo_metadata.rs#L17-L21

jyn514: If it helps, the only place we call `cargo_args()` is for `prepare_command`: https://github.com/rust-lang/docs.rs/blob/20aa7060349175424e915368790a5f064bfc4ad7/src/docbuilder/rustwide_builder.rs#L588
which is only used for doc coverage and actually generating the documentation.

Nemo157 (approved):

PR #1388, opened 2021-04-30
Add CSP nonce on inline script in storage-change-detection.html

### LINKED ISSUE #1387: Detecting rusdoc theme change fails with CSP
As mentioned in #1386 there is a CSP error from `storage-change-detection.html` having an inline script. This results in the header/footer not instantly updating when the theme in rustdoc changes, instead it updates after a page reload.

```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“default-src”). storage-change-detection.html:11:1
```

Nemo157:
fixes #1387

### DIFF
--- a/src/web/page/web_page.rs
+++ b/src/web/page/web_page.rs
@@ -10,7 +10,7 @@ use tera::Context;
 #[macro_export]
 macro_rules! impl_webpage {
     ($page:ty = $template:literal $(, status = $status:expr)? $(, content_type = $content_type:expr)? $(,)?) => {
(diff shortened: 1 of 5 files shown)


jyn514:
```
 thread 'docbuilder::rustwide_builder::tests::test_build_crate' panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`', src/docbuilder/rustwide_builder.rs:735:13
```
Same issue as before; unrelated to this change.

PR #1394, opened 2021-05-04
Use the same icon for clipboard button that crates.io itself uses

notriddle:
image


jyn514:
Hmm, I'm not a giant fan of the icon rustdoc uses - maybe we could make it a clipboard for both instead? https://fontawesome.com/icons/clipboard?style=solid

cc @GuillaumeGomez

GuillaumeGomez:
We don't use fontawesome in rustdoc, so the only thing I can suggest is replacing the rustdoc one with a svg so then we can use the same vector image in docs.rs. How does it sound?

jyn514:
Yes, that sounds good  :) I just think it should be a clipboard icon or something similar.

GuillaumeGomez:
I don't like much the current icon so very open to changes. :)

notriddle:
I don't like the current icon very much either, but I like that it's a unicode character instead of an SVG. Less overhead.

How about `\u{1f4cb}\u{fe0e}`, the [text presentation mode of the clipboard emoji](http://unicode.org/emoji/charts/emoji-variants.html#1f4cb)? 📋︎

GuillaumeGomez:
The current font rendering on mac is quite bad and broken in fact, which is why I suggested an SVG. Simpler to have a unified rendering.

jyn514:
That seems pretty complicated for a button :/ what do you mean by bad and broken?

notriddle:
So, are we going to import an SVG copy icon into rustdoc, or use the `\u{1f4cb}\u{fe0e}` icon in both?

GuillaumeGomez:
I'd prefer an SVG icon like we do for other icons in rustdoc. Also cc @jsha

jsha:
See also https://github.com/rust-lang/rust/issues/84639, where @rukai helpfully posted a screenshot of how crates.io does it. I think it's really valuable to harmonize the icon and the interaction between crates.io, doc.rust-lang.org, and docs.rs. And the crates.io icon is an SVG! 🎉

notriddle:
@jsha @GuillaumeGomez

Okay, here's the crates.io icon.

GuillaumeGomez (approved):

GuillaumeGomez:
cc @Nemo157 (since it's front-end ;) )

templates/header/package_navigation.html
```diff
@@ -27,7 +27,7 @@ <h1 id="crate-title">
                         {{ title }}
                     {%- else -%}
                         {{ metadata.name }} {{ metadata.version }}
-                        {{ "copy" | far(id="clipboard", aria_label="Copy crate name and version information", fa=true) }}
+                        <span id="clipboard" class="fa-svg fa-svg-fw" aria-label="Copy crate name and version information">{%- include "clipboard.svg" -%}</span>
```
Nemo157, line 30: We probably want this label as a title? As mentioned on <https://github.com/rust-lang/rust/issues/84639>/<https://github.com/rust-lang/rust/issues/84986> it's not obvious what this button does. (I personally don't know enough about ARIA to know whether the `aria-label` would still be necessary as well or not).

notriddle: Okay, that's done.

PR #1418, opened 2021-06-12
Add {target} to the extern-html-root-url

jyn514:
This avoids cross-crate intra-doc links that go to the wrong platform,
and possibly to items that don't exist.

Fixes https://github.com/rust-lang/docs.rs/issues/1397. Builds on https://github.com/rust-lang/docs.rs/pull/1417 and should not be merged before.

r? @Nemo157


jyn514:
I tested locally that this fixes tokio-stream: 
image

I don't know an easy way to test this even with a full build-test.

Nemo157 (approved):

PR #1433, opened 2021-06-24
Don't spam migrating logs when running tests

jyn514:
We run these all the way from 0 to 29 and back when running tests, which
causes quite a lot of output. This changes the output to `trace` in
tests, and changes the default logging to only show `debug` and above.

r? @Nemo157

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -1,6 +1,6 @@
 //! Database migrations
 
-use log::info;
+use log::{log, Level};
 use postgres::{Client, Error as PostgresError, Transaction};
 use schemamama::{Migration, Migrator, Version};
 use schemamama_postgres::{PostgresAdapter, PostgresMigration};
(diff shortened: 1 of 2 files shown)


src/db/migrate.rs
```diff
@@ -29,15 +28,23 @@ macro_rules! migration {
         }
         impl PostgresMigration for Amigration {
             fn up(&self, transaction: &mut Transaction) -> Result<(), PostgresError> {
-                info!(
+                #[cfg(not(test))]
+                use log::info as log_level;
+                #[cfg(test)]
+                use log::trace as log_level;
+                log_level!(
```
Nemo157, line 35:
```suggestion
                log!(
                    if cfg!(test) { Level::Trace } else { Level::Info },
```

+ appropriate imports, seems a little cleaner to me.

Nemo157 (approved):

Issue #382, opened 2019-07-18
Grave errors during building do not increment attempt counter


Mark-Simulacrum:
This means that we will build crates forever if every build results in a grave error.

jyn514:
I wonder if we should be catching this panic at all? The worst that happens is no more builds finish until an admin logs in and restarts the server, but I'd expect most panics to be keep happening on every retry anyway, and at least that way we avoid wasting resources.

https://github.com/rust-lang/docs.rs/blob/55c7386b9178dc3d33e34508aad29a9b8f4f2639/src/utils/queue_builder.rs#L83-L99

PR #1434, opened 2021-06-24
Handle panics in `build_next_queue_package`

jyn514:
Previously, these would cause the builder to loop forever. Now, it locks
the queue until someone has a chance to look at it.

Fixes https://github.com/rust-lang/docs.rs/issues/382.


pietroalbini:
This looks fine, but we should then have a metrics that outputs 0 if the queue is locked. That way we can alert when this happens.

jyn514:
This got kind of big - happy to split it out into smaller PRs if you like.

Nemo157 (approved):

jyn514:
Oh oops, I forgot to push the commit adding the metrics. I'll try and get to it this week.

jyn514:
> Oh oops, I forgot to push the commit adding the metrics. I'll try and get to it this week.

Nope, I just named the commit horribly :facepalm: fixed in the latest push.

I'm going to go ahead and merge this since @Nemo157 already approved it - thanks for taking a look!

Issue #444, opened 2019-10-25
nrfxlib needs bindgen binary to build


jonathanpallant:
**Crate name:**
nrfxlib
nrfxlib-sys

**Build failure link:**
https://docs.rs/crate/nrfxlib-sys/0.2.0
https://docs.rs/crate/nrfxlib/0.2.1

**Additional details:**
I used to use bindgen as a build-dependency to convert my third-party header files to Rust source in nrfxlib-sys. Now I shell out to the bindgen command line tool, to work around Cargo 5730. Is it possible to get bindgen pre-installed in the build environment? I found it was already installed when I used the Rust action on Github.

jyn514:
@jonathanpallant if bindgen is not packaged on Ubuntu 18.04, it is unlikely this can be fixed. We use https://github.com/rust-lang/crates-build-env for the build environment and it only installs dependencies that are packaged.

thejpster:
(swapped to my personal account)

Hmm, that makes sense. It's packaged for 19.04 - https://packages.ubuntu.com/disco/bindgen.

I assume there's no mechanism to `cargo install` something?

jyn514:
> It's packaged for 19.04

Ok, in that case we can add it when 20.04 gets released in a few months.

> I assume there's no mechanism to cargo install something?

No, because the build environment doesn't have a Rust toolchain installed.

jonathanpallant:
So, it seems rust-bindgen isn't packaged for focal (20.04) yet. I presume that means someone needs to do some work to package it?

jyn514:
That's correct. I didn't realize that packages for 19.04 weren't automatically transferred over to 20.04, that seems like a shame.

jyn514:
Another alternative I just thought of is to install bindgen on the host toolchain with the rest of the toolchain components. Rustwide exposes a [cargo command](https://docs.rs/rustwide/0.5.1/rustwide/toolchain/struct.Toolchain.html#method.cargo) - cc @pietroalbini , if I run `cargo install bindgen` with that, will `bindgen` be in PATH inside the build environment?

pietroalbini:
> if I run cargo install bindgen with that, will bindgen be in PATH inside the build environment?

It should be.

jyn514:
[...]

PR #1435, opened 2021-06-24
Install bindgen in the build environment

jyn514:
I tested locally that this only installs bindgen when new versions are
published, not on every build.

Fixes https://github.com/rust-lang/docs.rs/issues/444.

r? @Nemo157

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -147,6 +147,13 @@ impl RustwideBuilder {
             log::info!("continuing anyway, since this must be the first build");
         }
 
+        let cargo =
+            Command::new(&self.workspace, self.toolchain.cargo()).args(&["install", "bindgen"]);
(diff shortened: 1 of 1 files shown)


Nemo157:
https://github.com/rust-lang/cargo/issues/5730 is fixed with the new resolver, and `nrfxlib-sys` has updated to rely on that, are there any other crates still relying on this? (Eventually https://github.com/rust-lang/rfcs/pull/3028 would allow having a `build-dependency` on the `bindgen` binary if for some reason crates want to call it as a CLI instead of a library).

jyn514:
Hmm, that's a good point - I'd forgotten the original reason they used the binary.

I don't know any way we could check this. I guess we can close the issue and see if anyone complains?

Nemo157:
Yeah, the lack of "me too" comments on the issue makes me think it's a pretty rare problem, and now that it's fixed upstream everyone should hopefully migrate to `resolver = "2"`.

PR #1437, opened 2021-06-24
Remove `<span class="title">` around `Source` link on dropdown

jyn514:
This fixes a bug where it would disappear if the screen width got too
narrow (https://github.com/rust-lang/docs.rs/issues/1422).

Hmm, r? @jsha maybe?


jyn514:
Yup, this seems to have worked: image

jsha:
Looks good to me!

PR #1440, opened 2021-06-26
Make initialization for BinContext easier to read

jyn514:
r? @Nemo157

### DIFF
--- a/src/bin/cratesfyi.rs
+++ b/src/bin/cratesfyi.rs
@@ -550,78 +550,51 @@ impl BinContext {
     }
 }
 
+macro_rules! lazy {
+    ( $(fn $name:ident($self:ident) -> $type:ty = $init:expr);+ $(;)? ) => {
+        $(fn $name(&$self) -> Result<Arc<$type>, Error> {
+            Ok($self
+                .$name
+                .get_or_try_init::<_, Error>(|| Ok(Arc::new($init)))?
+                .clone())
+        })*
+    }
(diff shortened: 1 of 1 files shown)


src/bin/cratesfyi.rs
```diff
@@ -550,39 +550,31 @@ impl BinContext {
...
-    }
+    lazy! { fn storage(self) -> Storage = Storage::new(
+            self.pool()?,
+            self.metrics()?,
+            &*self.config()?,
+    )? }
+
+    lazy! { fn config(self) -> Config = Config::from_env()? }
 
     fn pool(&self) -> Result<Pool, Error> {
```
jyn514, line 579: This one can't be part of the macro because it doesn't return `Arc`. I thought it was simpler to just leave it be than to try and make the macro more flexible.

Nemo157:
Only minor change I can think would be to make the macro support many at once, that'll reduce the need for "weird" indentation within the macro

```rust
    lazy! {
        fn build_queue(self) -> BuildQueue = BuildQueue::new(
            self.pool()?,
            self.metrics()?,
            &*self.config()?,
        );

        fn storage(self) -> Storage = Storage::new(
            self.pool()?,
            self.metrics()?,
            &*self.config()?,
        )?;
    }
```

jyn514:
You're right, that looks a lot nicer :)

Nemo157 (approved):

PR #1464, opened 2021-08-12
Add toggle mechanism for docs.rs source sidebar

### LINKED ISSUE #1445: Add a way to collapse the sidebar in docs.rs' source view
This is not a great UI: 
image

especially since horizontal scrolling only scrolls within that box:
image

GuillaumeGomez:
Fixes #1445 

Showing how it works in video:

https://user-images.githubusercontent.com/3050060/129723245-62847767-a6b3-42d9-ace8-260a235e9c27.mp4

### DIFF
--- a/static/source.js
+++ b/static/source.js
@@ -1,0 +1,41 @@
+(function() {
+    var oldLabel;
+
+    function showSourceFiles(button, sideMenu, sourceCode) {
+        button.title = oldLabel;
+        button.setAttribute("aria-label", button.title);
+        button.setAttribute("aria-expanded", "true");
+
+        sideMenu.classList.remove("collapsed");
+        sourceCode.classList.remove("expanded");
+    }
+
+    function hideSourceFiles(button, sideMenu, sourceCode) {
+        button.title = "Show source sidebar";
+        button.setAttribute("aria-label", button.title);
+        button.setAttribute("aria-expanded", "false");
+
+        sideMenu.classList.add("collapsed");
+        sourceCode.classList.add("expanded");
+    }
+
+    function toggleSource(button) {
+        var sideMenu = document.getElementById("side-menu");
+        var sourceCode = document.getElementById("source-code");
+
+        if (sideMenu.classList.contains("collapsed")) {
+            showSourceFiles(button, sideMenu, sourceCode);
+        } else {
+            hideSourceFiles(button, sideMenu, sourceCode);
+        }
+    }
+
+    document.addEventListener("DOMContentLoaded", function(event) { 
+        var toggleSourceButton = document.querySelector("li.toggle-source button");
+        oldLabel = toggleSourceButton.getAttribute("aria-label");
+
+        toggleSourceButton.addEventListener("click", function() {
+            toggleSource(toggleSourceButton);
+        });
+    });
+})();
--- a/templates/style/base.scss
+++ b/templates/style/base.scss
@@ -438,7 +438,7 @@ div.package-page-container {
             margin-top: 0;
         }
 
-        a.pure-menu-link {
+        .pure-menu-link {
             font-size: 14px;
             color: var(--color-standard);
             white-space: nowrap;
@@ -447,7 +447,7 @@ div.package-page-container {
             padding: 7px 8px;
         }
 
-        a.pure-menu-link:hover {
+        .pure-menu-link:hover {
             background-color: var(--color-background-code);
         }
 
@@ -758,4 +758,65 @@ div.search-page-search-form {
 h3 > code,
 h4 > code {
     display: inline-block;
+}
+
+ul.pure-menu-list {
+    li.toggle-source {
+        cursor: pointer;
+        border: 1px solid var(--color-border);
+        display: none;
+    }
+}
+
+@media screen and (min-width: 35.5em) {
+    ul.pure-menu-list {
+        li.toggle-source {
+            display: list-item;
+
+            button {
+                font-size: 14px;
+                color: var(--color-standard);
+                white-space: nowrap;
+                overflow: hidden;
+                text-overflow: ellipsis;
+                padding: 7px 8px;
+                background: transparent;
+                border: 0;
+                width: 100%;
+            }
+
+            .right {
+                display: none;
+            }
+        }
+    }
+
+    #side-menu.collapsed {
+        max-width: 46px;
+
+        ul {
+            li:not(.toggle-source), .text {
+                display: none;
+            }
+            li.toggle-source {
+                .left {
+                    display: none;
+                }
+                .right {
+                    display: inline-block;
+                    margin-left: -4px;
+                }
+            }
+        }
+    }
+
+    #source-code {
+        pre {
+            margin-top: 0;
+        }
+
+        &.expanded {
+            width: calc(100% - 46px);
+        }
+    }
 }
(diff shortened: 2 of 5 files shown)


jyn514:
Hmm, why do you keep the icons in the sidebar when it's collapsed? They don't seem very helpful since many of them are the same, you can't tell different icons apart. Rustdoc just shows an arrow to expand the sidebar, it doesn't still let you click things: 
image

static/source.js
```diff
@@ -0,0 +1,34 @@
+(function() {
+    function showSourceFiles(button, sideMenu, sourceCode) {
+        button.getElementsByClassName("pure-menu-link")[0].childNodes[0].textContent = "< ";
```
jyn514, line 3: Hmm, could you maybe add a title-text of "collapse" (and "expand" for the other)? This seems not great for accessibility.

Actually, @ahicks92 do you know how this is normally done for screen-readers? I assume this will read out as "left arrow" which doesn't seem ideal.

templates/style/base.scss
```diff
@@ -438,7 +438,7 @@ div.package-page-container {
             margin-top: 0;
         }
 
-        a.pure-menu-link {
+        .pure-menu-link {
```
jyn514, line 441: Hmm, why did this change?

GuillaumeGomez:
> Hmm, why do you keep the icons in the sidebar when it's collapsed? They don't seem very helpful since many of them are the same, you can't tell different icons apart. Rustdoc just shows an arrow to expand the sidebar, it doesn't still let you click things:

Because I don't know otherwise where to keep the button to "expand" the sidebar.

templates/style/base.scss
GuillaumeGomez: Because I wanted to use the style of this class in the "collapse" button.

Nemo157: The classes don't entirely work on a non-link, if you click and drag it selects the text inside instead of dragging a button around like the other links.

static/source.js
ahicks92: I can't figure out how to get a test link out of the pr; is there a way to do that?

GuillaumeGomez: @ahicks92: I can't host this PR so not possible unfortunately. :-/

@jyn514: Adding a title attribute would fix this issue I guess.

ahicks92: I think you want  `aria-label`.  You'd also want [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaExpanded) I think.  Unfortunately this is tentative: let's say that I'm 70% confident.  Fortunately, I think most screen reader users would be able to figure this out even without changes, though improving this is definitely a good idea.

Not 100% on whether or not `title` is relevant, but probably not the end of the world to add one.  I think that title ends up being togglable and counted as a tooltip or something like that, not entirely replacing the text of the button in all contexts.

ahicks92: As a followup to that note that `aria-expanded` if done right is going to make it obvious that the button expands or collapses something, and the common mistake is working the action into the label (e.g. "expand table of contents" as the label becomes "expand table of contents button collapsed" for example, which is irrelevant info).

Nemo157: It's currently being hosted on my dev instance, <https://docs.rs.dev.nemo157.com/crate/tokio-util/0.6.7/source/src/codec/decoder.rs>.

templates/style/base.scss
```diff
@@ -759,3 +759,42 @@ h3 > code,
 h4 > code {
     display: inline-block;
 }
+
+ul.pure-menu-list {
+    li.toggle-source {
+        cursor: pointer;
+        border: 1px solid var(--color-standard);
```
Nemo157, line 766: `--color-standard` doesn't really fit the theme on ayu/light, they work much better with `--color-border` (dark is better with `--color-standard`, so maybe there needs to be a separate variable for button-borders?).

static/source.js
ahicks92: That link is dead, maybe a VPN or something.  In any case we still probably want changes here, so I'll check back in tomorrow.

Nemo157: Ah, it's IPv6 only. I can maybe throw cloudflare in front of it to get an IPv4 reverse proxy :thinking:.

Nemo157: Using cloudflare as a reverse proxy is a little more annoying than I expected, but hopefully this works now: <https://ipvold-docs-rs-dev.nemo157.com/crate/tokio-util/0.6.7/source/src/codec/decoder.rs>

GuillaumeGomez: @ahicks92: Let's go for `aria-label`, `title` and `aria-expanded`. :)

templates/style/base.scss
GuillaumeGomez: Let's give it a try.

GuillaumeGomez:
Updated the UI a bit and added the aria attributes. I also updated the video to show what the new UI looks like.

Nemo157:
Updated my server, looks nicer to me 😁 

https://fontawesome.com/v5.15/icons/chevron-left?style=solid and -right might look good next to the filetype icons?

GuillaumeGomez:
Why not? I think @jyn514 didn't like just having "<"/">" but maybe I misunderstood. If they're happy with it, I will use the fontawesome icon instead.

static/source.js
ahicks92: That's not a button tag. Missed that in my review somehow.  The labeling stuff only works on items of the right role.  The easiest (read: bug-free) way here is to use a button tag outside the list instead, but if we're willing to possibly accept bugs on some platforms then adding `role="button"` might or might not work.  Unfortunately screen readers get buggy around edge cases, and "li tagged as button inside a list" seems like one. I hacked it in via the inspector with Firefox+NVDA and it seems to work for me.  If there's  a reason we need to keep it this way rather than using a proper button we can try it and I'll do some more testing on the combinations I have access to.

note also that non-button buttons etc. like this need tabindex=0 for keyboard accessibility; you can't tab to them  by default either.

GuillaumeGomez: I plan to use a `button`, for now it was mostly validating the UI (but I think it's ok now so I'll update it).

GuillaumeGomez:
I replaced with font-awesome icons and now use a button.

ahicks92:
If @Nemo157 updates their server I'll check it again.  The code looks right, though do we know if it's valid HTML to put buttons inside list elements? I didn't think so, and "isn't valid HTML" is often a good proxy for breaks with some screen reader somewhere.

GuillaumeGomez:
I tested this HTML on the [w3c validator](https://validator.w3.org/nu/#textarea):

```html
<!DOCTYPE html>
<html lang="en">
<head>
<title>hello</title>
</head>
<body>
<ul>
<li><button>yolo</button></li>
</ul>
</body>
</html>
```

No warning or error, so I guess it's fine. :smile:

Nemo157:
@ahicks92 it's updated

ahicks92:
Yeah, that works. Maybe "hide" in the label/title, because "collapse source sidebar expanded" is a bit awkward, but that's very much a NIT.

GuillaumeGomez:
@ahicks92 No it makes sense because I used "Show" in for expanding it. So I replaced "collapse" with "hide".

jyn514:
This looks good to me, but I realized after using it that I have another complaint: the "Hide files" button doesn't scroll with the content, so you can end up having to scroll back up to the top to collapse the sidebar:
image

Rustdoc does this by making the sidebar scrolling independent from the content scrolling, maybe we could copy that? Or would you prefer to leave that for a follow-up?

GuillaumeGomez:
It's easy to do but I'd prefer to do it in a follow-up for a simple reason: it's a behaviour change. So I'd prefer for people to confirm whether they want it or not on this specific part. I'll send a PR for it once this one is merged.

Issue #1474, opened 2021-08-21
"Hide files" toggle should be fixed so it isn't hidden after scrolling


jyn514:
The "Hide files" button doesn't scroll with the content, so you can end up having to scroll back up to the top to collapse the sidebar:
![image](https://user-images.githubusercontent.com/23638587/130333505-43c65eb4-c7b7-401f-a95e-8b77eb6eea04.png)

Rustdoc does this by making the sidebar scrolling independent from the content scrolling, maybe we could copy that? Or would you prefer to leave that for a follow-up?

cc @GuillaumeGomez
_Originally posted by @jyn514 in https://github.com/rust-lang/docs.rs/issues/1464#issuecomment-903169839_

GuillaumeGomez:
I also think the scrolling should be independant (in case we count votes haha).

PR #1477, opened 2021-08-25
Make source files sidebar scrolling independant

GuillaumeGomez:
Fixes https://github.com/rust-lang/docs.rs/issues/1474

https://user-images.githubusercontent.com/3050060/130795044-31ac39ea-c6e6-41ba-9b50-b2549f540483.mp4

Question now: do you want to put the "hide files" button outside of the scrolled area so it always remain on top or not?

### DIFF
--- a/templates/crate/source.html
+++ b/templates/crate/source.html
@@ -22,7 +22,7 @@
 {%- endblock header -%}
 
 {%- block body -%}
-    <div class="container package-page-container">
+    <div class="container package-page-container small-bottom-pad">
         <div class="pure-g">
             <div id="side-menu" class="pure-u-1 {% if file_content %}pure-u-sm-7-24 pure-u-md-5-24{% endif %}">
                 <div class="pure-menu package-menu">
--- a/templates/crate/source.html
+++ b/templates/crate/source.html
@@ -24,7 +24,7 @@
 {%- block body -%}
     <div class="container package-page-container small-bottom-pad">
         <div class="pure-g">
-            <div id="side-menu" class="pure-u-1 {% if file_content %}pure-u-sm-7-24 pure-u-md-5-24{% endif %}">
+            <div id="side-menu" class="pure-u-1 {% if file_content %}pure-u-sm-7-24 pure-u-md-5-24 source-view{% endif %}">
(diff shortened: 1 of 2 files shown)


Nemo157:
It doesn't look like this works to me? <https://docs.rs.dev.nemo157.com/crate/dialectic-tokio-serde/0.1.0/source/Cargo.lock>/<https://ipvold-docs-rs-dev.nemo157.com/crate/dialectic-tokio-serde/0.1.0/source/Cargo.lock> still has the list stuck to the top of the page.

Also <https://docs.rs.dev.nemo157.com/crate/dialectic-tokio-serde/0.1.0/source/> is empty.

GuillaumeGomez:
I can't see your pages for "dialectic-tokio-serde". Also: the scrollbar only appears when the source file list is longer than the source code.

GuillaumeGomez:
Ah hold on! Just understood: the issue meant that we want to have the source files list always visible whatever the scroll position in the source file. Ok, I'll update.

GuillaumeGomez:
Ok, updated. I discovered `position: sticky`. That was quite the ride. I added a video too.

Nemo157:
Is it possible to not have it scroll off the top when you scroll to the very bottom?

GuillaumeGomez:
Not that I know of. The footer is "pushing" the content up, which is why the sticky position is being disabled at that time.

Nemo157:
On <https://ipvold-docs-rs-dev.nemo157.com/crate/tokio/1.3.0/source/src/io/util/async_buf_read_ext.rs> I need to scroll the page all the way down to see the last few entries in the file list. I wonder if it should be reduced in height to account for the entire header block (which might also fix the footer pushing it up). That might leave a bit of blank space below it, but that seems better than not being able to access the files.

GuillaumeGomez:
Oh nice! That's a bug. :)

GuillaumeGomez:
Fixed the bug, great catch, thanks!

GuillaumeGomez:
Anything else to be fixed @Nemo157 ?

jyn514:
This is great, thanks :)

PR #1478, opened 2021-08-26
Fix metadata to accept quoted key table

### LINKED ISSUE #1396: Accept `[package.metadata."docs.rs"]`
I was recently confused by docs.rs not using the provided [configuration](https://docs.rs/about/metadata) from the `Cargo.toml`:
```toml
[package.metadata."docs.rs"]
# Configuration ...
```

Turns out only `[package.metadata.docs.rs]` works (note: no `""`). Not sure why I've written `"docs.rs"` in the first place, but it would be nice to support this too.

chansuke:
Fixes #1396 to accept quoted key table. 

```
[package.metadata."docs.rs"]
# Configuration ...
```

### DIFF
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -300,12 +300,19 @@ impl std::str::FromStr for Metadata {
             }
         }
 
-        let table = manifest
+        let plain_table = manifest
+            .clone()
             .and_then(|t| table(t, "package"))
             .and_then(|t| table(t, "metadata"))
             .and_then(|t| table(t, "docs"))
             .and_then(|t| table(t, "rs"));
-        let mut metadata = if let Some(table) = table {
+        let quoted_table = manifest
+            .and_then(|t| table(t, "package"))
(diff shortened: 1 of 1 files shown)


crates/metadata/lib.rs
```diff
@@ -300,12 +300,19 @@ impl std::str::FromStr for Metadata {
             }
         }
 
-        let table = manifest
+        let plane_table = manifest
```
jyn514, line 303:
```suggestion
        let plain_table = manifest
```

jyn514 (approved):
Looks good with the typo fixed :) thanks for the PR!

crates/metadata/lib.rs
chansuke: @jyn514 Thanks for your review!! I will fix this

chansuke:
Fixed and squashed in one commit

jyn514 (approved):

jyn514:
This is great, thank you!

PR #1489, opened 2021-09-12
Update to lol-html 0.3

jyn514:
This makes the RewritingError type Send + Sync, which is necessary to
convert it to an anyhow error so we can send it to sentry.

Also, it's nice to have updated dependencies.

r? @syphar


syphar (approved):
LGTM

PR #1491, opened 2021-09-12
Sort (and display) releases by build time, not release time

jyn514:
This avoids confusing situations where docs.rs will show "one hour ago"
for a crate it just built.

Fixes https://github.com/rust-lang/docs.rs/issues/1337.

r? @Nemo157

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -32,22 +32,8 @@ pub struct Release {
     description: Option<String>,
     target_name: Option<String>,
     rustdoc_status: bool,
-    pub(crate) release_time: DateTime<Utc>,
+    pub(crate) build_time: DateTime<Utc>,
     stars: i32,
-}
-
-impl Default for Release {
-    fn default() -> Release {
-        Release {
-            name: String::new(),
-            version: String::new(),
-            description: None,
-            target_name: None,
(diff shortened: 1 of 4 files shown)


src/web/releases.rs
```diff
@@ -189,13 +177,13 @@ fn get_search_results(
...
-            releases.description AS description,
-            releases.target_name AS target_name,
-            releases.release_time AS release_time,
-            releases.rustdoc_status AS rustdoc_status,
-            repositories.stars AS stars,
+            crates.name,
+            releases.version,
+            releases.description,
+            releases.target_name,
+            releases.release_time,
```
GuillaumeGomez, line 184: Why not simply renaming it as `build_time` here directly?

jyn514: Because it's not the build time, it's the release time. I wanted it to be clear that the name is wrong - it's only called `build_time` below because otherwise I'd have to add another `Option`al field to Release.

GuillaumeGomez: I meant in the request using `AS` (not sure if this is how you understood my comment so just in case :wink: ).

But with your answer, I htink it's better as it's currently done, more explicit.

Nemo157 (approved):

PR #1493, opened 2021-09-13
Fix source code file sidebar

### LINKED ISSUE #1492: Source file list is cutoff on short files
The file list stops before the bottom of the file listing, so if the file is short it gets cutoff early and needs to be scrolled even though it could completely fit on the screen.

Example: https://docs.rs/crate/tuop-ls/0.1.4/source/src/error.rs

image

cc @GuillaumeGomez

GuillaumeGomez:
Fixes #1492.

Now both elements takes the whole remaining available space (if the source code is too short to take it):

Screenshot from 2021-09-13 16-52-20

### DIFF
--- a/templates/crate/source.html
+++ b/templates/crate/source.html
@@ -21,6 +21,10 @@
     {{ navigation::package_navigation(metadata=file_list.metadata, active_tab="source") }}
 {%- endblock header -%}
 
+{%- block body_classes -%}
+    {%- if file_content -%}flex{%- endif -%}
+{%- endblock body_classes -%}
+
 {%- block body -%}
     <div class="container package-page-container small-bottom-pad">
         <div class="pure-g">
--- a/templates/base.html
+++ b/templates/base.html
@@ -21,7 +21,7 @@
         {%- block css -%}{%- endblock css -%}
     </head>
 
-    <body>
+    <body class="{% block body_classes %}{% endblock body_classes %}">
         {%- block topbar -%}
           {%- include "header/topbar.html" -%}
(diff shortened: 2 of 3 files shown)


templates/crate/source.html
```diff
@@ -21,6 +21,10 @@
     {{ navigation::package_navigation(metadata=file_list.metadata, active_tab="source") }}
 {%- endblock header -%}
 
+{% block body_classes %}
+    {% if file_content %}flex{% endif %}
+{% endblock body_classes %}
```
Nemo157, line 26:
```suggestion
{%- block body_classes -%}
    {%- if file_content %}flex{% endif -%}
{%- endblock body_classes -%}
```

to drop the extra whitespace

GuillaumeGomez: What is the difference between the two btw?

GuillaumeGomez:
Updated!

templates/crate/source.html
Nemo157: `-` makes it remove all whitespace on that side of the embedded content

GuillaumeGomez: Thanks!

Nemo157:
Do we want the file background to extend to the bottom? Maybe that should be in another column with a spacer that fills the space if it's not the full height.

GuillaumeGomez:
I think it's more coherent this way (and let's be honest, much simpler too).

Nemo157 (approved):
Tested locally and seems to work fine

GuillaumeGomez:
Thanks!

Issue #79, opened 2016-11-03
Provides rustc related crates that not in crates.io


liigo:
Docs of HIR, MIR, etc. are useful for some developers.

jyn514:
@liigo we have links to the standard library and a few other resource under the 'Rust' dropdown in the navigation panel. If you have any specific links you would like, feel free to open a new issue (or make a PR adding them yourself, mentoring instructions here: https://github.com/rust-lang/docs.rs/issues/137#issuecomment-581046680)

liigo:
@jyn514 there's no rustc related api docs listed in the 'Rust' dropdown currently, so i don't think this issue is resolved.

jyn514:
Ah sorry I misread `rustc` as `rust`.

jyn514:
Mentoring instructions:

- In [`templates/navigation.hbs`](https://github.com/rust-lang/docs.rs/blob/master/templates/navigation.hbs#L18), add links to https://doc.rust-lang.org/stable/rustc/ and any other links you find helpful.

liigo:
Thank you!

jyn514:
Also, @pietroalbini are you comfortable adding links to `rustc` specifically? I know it's currently the only Rust implementation, but `mrustc` is a thing and I'm not sure if others are planned (in the far future) ...

pietroalbini:
The link I think @liigo mentions is https://doc.rust-lang.org/nightly/nightly-rustc/rustc/.

I don't think adding the link itself is worth though: it's only needed for people working on the compiler, and it's [shown on the home page of the contributor guide](https://rust-lang.github.io/rustc-guide/).

jyn514:
Maybe it would be helpful to link to the contributor guide instead?

pietroalbini:
Hmm, I still don't see much value in that, as a rustc contributor should already know that.

jyn514:
> Hmm, I still don't see much value in that, as a rustc contributor should already know that.

Speaking personally, I've made a couple PRs to rustc and I didn't know the contributor guide existed until recently.

liigo:
> Hmm, I still don't see much value in that, as a rustc contributor should already know that.

IMO if `docs.rs` is the *home* of all crates api docs, users would expect to find `std` and `rustc` api docs at the site. And they should be provided in the same way as other crates, not just in an external links list (the ‘Rust’ dropdown menu).

jyn514:
> docs.rs is the home of all crates api docs

[...]

PR #1498, opened 2021-09-17
Add docs.rs/rustc redirect.

m-ou-se:
This adds a redirect from `docs.rs/rustc` to the nightly rustc docs.

`rustc` is a reserved crate name and will not appear on crates.io: https://github.com/rust-lang/crates.io/blob/master/migrations/20170305095748_create_reserved_crate_names/up.sql

Unlike the redirects for `std`, `core`, etc., this uses nightly rather than stable, because all development on rustc is happening on nightly. (And also `https://doc.rust-lang.org/stable/nightly-rustc/` looks a bit confusing. :P)

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -728,6 +728,12 @@ mod test {
             // with or without slash
             assert_redirect("/rustc", target, web)?;
             assert_redirect("/rustc/", target, web)?;
+
+            let target = "https://doc.rust-lang.org/nightly/nightly-rustc/rustdoc/";
+            // with or without slash
+            assert_redirect("/rustdoc", target, web)?;
+            assert_redirect("/rustdoc/", target, web)?;
+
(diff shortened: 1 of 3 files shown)


jyn514:
This would fix https://github.com/rust-lang/docs.rs/issues/79.

Can you also add rustdoc at the same time? It's also reserved: https://github.com/rust-lang/crates.io/blob/b070d14df6696d4eb868bc8d26c0c1fd6257f97c/migrations/20170305095748_create_reserved_crate_names/up.sql#L11

Ideally this would add all the tools that are hosted (clippy and rustfmt too), but they don't seem to be reserved; in fact there are crates already published for both: https://docs.rs/crate/rustfmt/0.10.0, https://docs.rs/crate/clippy/0.0.302.

jyn514 (approved):

PR #1548, opened 2021-11-08
Switch to maintained version of mmap

jyn514:
Fixes https://github.com/rust-lang/docs.rs/issues/1547

r? @syphar

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -785,7 +785,7 @@ dependencies = [
  "kuchiki",
  "log 0.4.14",
  "lol_html",
- "memmap",
+ "memmap2",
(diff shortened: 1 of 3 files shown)


jyn514:
Changelog here (there's no changes to `memmap` that aren't part of `memmap2`): https://github.com/RazrFalcon/memmap2-rs/blob/master/CHANGELOG.md

syphar (approved):
I really missed that, and for `cargo audit` it's only a warning .. 

I'll check new requirements next time. 

Thanks for the fix!

jyn514:
No problem! It would normally have been caught by the audit check, but unfortunately that's been failing for weeks ...

Issue #1556, opened 2021-11-21
linux target does not appear in docs.rs even `targets` in metadata is set


konan8205:
### Crate name

xingapi

### Build failure link

https://docs.rs/crate/xingapi/0.3.3/builds/465678

### Additional details

After I released 0.3.0, I noticed that doc for `x86_64-unknown-linux-gnu` target is missing in docs.rs. But I was able to build the documentation locally with that target. Anyway, I've released some new versions to fix these issues.

### 0.3.1
- Use `ignore` attribute for markdown code blocks to **fix doc tests** in Linux

### 0.3.2
- Do not use `ignore` attribute for markdown code blocks (now doc tests fail)
- **Make doc tests disabled** in Cargo.toml (All cargo tests except doc tests succeed)
- **Make examples compilable in Linux**

### 0.3.3
- **Remove Windows-specific markdown code blocks** in cross platform module (doc tests succeed)
- Make doc tests enabled in Cargo.toml

None of these attempts to fix the problem have been successful. `targets` and `default-target` in Cargo.toml already set. Could someone please tell me what I am doing wrong? I don't even know where and how to view the build failure log.

Nemo157:
Unfortunately build logs for non-default targets are not currently recorded (https://github.com/rust-lang/docs.rs/issues/787).

I've checked the logs on the server and it did _build_ the docs for `x86_64-unknown-linux-gnu`, but then didn't appear to add the files to the database. I wonder if there's a bug in docs.rs about building the host target as a non-default target.

Nemo157:
Yes, this is wrong:

https://github.com/rust-lang/docs.rs/blob/78e299249d292ebfba7d2ed821363f28e211cd25/src/docbuilder/rustwide_builder.rs#L659-L662

We only want to treat the host target as the `--target`-less target when we're doing a default-platform build. In fact I think we can get rid of this altogether, and always pass `--target`, since we already handle the default-platform being in a subdirectory fine for the other targets, so we should make the build server agnostic to which target it is running on.

jyn514:
> In fact I think we can get rid of this altogether, and always pass --target,

No, that breaks proc-macros. See https://github.com/rust-lang/docs.rs/issues/422.

> We only want to treat the host target as the --target-less target when we're doing a default-platform build.

Can you explain this? Why would we pass `--target` for the host? I'm not sure I understand the cause of the bug ...

Nemo157:
> No, that breaks proc-macros. See #422.

[...]

PR #1559, opened 2021-11-24
Fix compiling for x86_64-unknown-linux-gnu when it's not the default target

jyn514:
Fixes https://github.com/rust-lang/docs.rs/issues/1556 by only special casing proc-macros instead of any crate targeting the host.

r? @Nemo157

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -210,7 +210,8 @@ impl RustwideBuilder {
                     }
 
                     info!("copying essential files for {}", self.rustc_version);
-                    let source = build.host_target_dir().join("doc");
+                    assert!(!metadata.proc_macro);
+                    let source = build.host_target_dir().join(HOST_TARGET).join("doc");
                     let dest = tempfile::Builder::new()
                         .prefix("essential-files")
                         .tempdir()?;
@@ -350,14 +351,23 @@ impl RustwideBuilder {
                     if res.result.successful {
                         if let Some(name) = res.cargo_metadata.root().library_name() {
                             let host_target = build.host_target_dir();
-                            has_docs = host_target.join("doc").join(name).is_dir();
+                            has_docs = host_target
+                                .join(default_target)
+                                .join("doc")
+                                .join(name)
+                                .is_dir();
                         }
                     }
 
                     let mut algs = HashSet::new();
                     if has_docs {
                         debug!("adding documentation for the default target to the database");
-                        self.copy_docs(&build.host_target_dir(), local_storage.path(), "", true)?;
+                        self.copy_docs(
+                            &build.host_target_dir(),
+                            local_storage.path(),
+                            default_target,
+                            true,
+                        )?;
 
                         successful_targets.push(res.target.clone());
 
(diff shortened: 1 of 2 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -935,4 +944,66 @@ mod tests {
...
+                &format!("{}/{}/index.html", target, crate_path),
+            )?;
+
+            let web = env.frontend();
+            let target_url = format!(
+                "/{}/{}/{}/{}/index.html",
+                crate_, version, target, crate_path
+            );
+
+            assert!(target_docs_present);
```
jyn514, line 1003: I double checked and this does fail without the fix in this commit.

```
[2021-11-24T17:58:10Z DEBUG docs_rs::docbuilder::rustwide_builder] adding sources into database
[2021-11-24T17:58:10Z DEBUG docs_rs::db::add_package] Adding package into database
[2021-11-24T17:58:10Z DEBUG docs_rs::db::add_package] Adding doc coverage into database
[2021-11-24T17:58:10Z DEBUG docs_rs::db::add_package] Adding build into database
[2021-11-24T17:58:10Z INFO  docs_rs::db::add_package] Updating crate data for xingapi
[2021-11-24T17:58:10Z DEBUG docs_rs::docbuilder::rustwide_builder] cleaning old storage folder rustdoc/xingapi/0.3.3/
[2021-11-24T17:58:10Z DEBUG docs_rs::docbuilder::rustwide_builder] cleaning old storage folder sources/xingapi/0.3.3/
[2021-11-24T17:58:11Z ERROR docs_rs::web::page::templates] Failed to load rustc resource suffix: missing rustc version
[2021-11-24T17:58:11Z INFO  docs_rs::web] Running docs.rs web server on http://127.0.0.1:39737
thread 'docbuilder::rustwide_builder::tests::test_cross_compile_non_host_default' panicked at 'assertion failed: target_docs_present', src/docbuilder/rustwide_builder.rs:1004:13

failures:
    docbuilder::rustwide_builder::tests::test_cross_compile_non_host_default
```

Nemo157:
I think we probably want to override `Metadata::targets` to only return a single target for proc-macros too (since we don't actually obey the target setting). I found a random proc-macro without a target override and it only has docs for the default target: <https://docs.rs/datatest-derive/0.6.3/datatest_derive/>.

Nemo157:
`add_essential_files` needs to be updated to look in the target dir

```patch
-let source = build.host_target_dir().join("doc");
+let source = build.host_target_dir().join("doc").join(HOST_TARGET);

File: src/docbuilder/rustwide_builder.rs [hunk 2]
```diff
@@ -350,7 +350,11 @@ impl RustwideBuilder {
                     if res.result.successful {
                         if let Some(name) = res.cargo_metadata.root().library_name() {
                             let host_target = build.host_target_dir();
-                            has_docs = host_target.join("doc").join(name).is_dir();
+                            has_docs = if metadata.proc_macro {
+                                host_target.join("doc").join(name).is_dir()
+                            } else {
+                                host_target.join("doc").join(HOST_TARGET).join(name).is_dir()
```
Nemo157, line 356: I think this needs to be `default_target`, in case the library does not build for the host target.

jyn514:
Yay, this is finally passing tests :) this should be ready for rereview.

Nemo157 (approved):
All LGTM now. Tested locally with a variety of proc-macro/non-proc-macro/default=host/default=non-host crates and all build variants appear to work correctly.

konan8205:
@Nemo157 Could you trigger the rebuild from 0.3.0 to 0.3.3 version? I want to check if the problem is solved.

PR #1562, opened 2021-11-28
Make "Go to latest version" point to /latest/

jsha:
Still trying to get tests to pass, but figured I'd post up what I've got so far.

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -1015,13 +1015,13 @@ mod test {
             let redirect = latest_version_redirect("/dummy/0.1.0/dummy/", web)?;
             assert_eq!(
                 redirect,
-                "/crate/dummy/latest/target-redirect/x86_64-unknown-linux-gnu/dummy/index.html"
+                "/crate/dummy/0.2.1/target-redirect/x86_64-unknown-linux-gnu/dummy/index.html"
             );
 
             let redirect = latest_version_redirect("/dummy/0.2.0/dummy/", web)?;
             assert_eq!(
                 redirect,
-                "/crate/dummy/latest/target-redirect/x86_64-unknown-linux-gnu/dummy/index.html"
+                "/crate/dummy/0.2.1/target-redirect/x86_64-unknown-linux-gnu/dummy/index.html"
             );
 
             Ok(())
@@ -1815,7 +1815,7 @@ mod test {
                     "/tungstenite/0.10.0/tungstenite/?search=String%20-%3E%20Message",
                     env.frontend()
                 )?,
-                "/crate/tungstenite/0.11.0/target-redirect/x86_64-unknown-linux-gnu/tungstenite/index.html?search=String%20-%3E%20Message",
+                "/crate/tungstenite/latest/target-redirect/x86_64-unknown-linux-gnu/tungstenite/index.html?search=String%20-%3E%20Message",
             );
             Ok(())
         });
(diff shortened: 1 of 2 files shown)


templates/rustdoc/topbar.html
```diff
@@ -193,7 +193,11 @@
         {%- endif -%}
 
         <li class="pure-menu-item">
-            <a href="{{ latest_path | safe }}" class="pure-menu-link warn"
+            <a href="/crate/{{ metadata.name }}/latest/target-redirect/
+                {%- if target != "" %}{{target}}{% else %}{{ krate.metadata.default_target }}/{% endif -%}
+                {{ inner_path }}
+                {%- if query %}?{{query}}{% endif -%}
```
jyn514, line 199: Doing this in the template looks relevant prone to error. Why did you chose to do it here instead of modifying latest_path?

src/web/rustdoc.rs
```diff
@@ -447,7 +448,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
         (target, inner_path.join("/"))
     };
 
-    // Find the path of the latest version for the `Go to latest` and `Permalink` links
+    // For `Permalink`, the path of the latest version, with an explicit version in it.
     let mut latest_path = if latest_release.build_status {
```
jyn514, line 452: This looks wrong. This code is not calculating a link to the *current* page, it's finding the page that "Go to latest" should link to. So I think the Permalink feature you already implemented doesn't work properly.

jyn514 (changes requested):

templates/rustdoc/topbar.html
jsha: At first I thought it would be simple, and then I kept running into complications as I ran the tests. This does belong in the Rust code.

I wanted to avoid modifying latest_path because latest_path is now used by the permalink. But I think what we need is, e.g. latest_path and permalink_path, both generated by Rust.

src/web/rustdoc.rs
jsha: I don't understand that. This code as I understands it links (via target-redirect) to what the _current_ page would be, if you substituted the latest numeric version into the URL. That used to be useful for "Go to latest version," but now it's useful for "Permalink."

jyn514: ... right, because "Permalink" is only used when you're already on the latest page. Ok, that makes sense to me.

jyn514 (approved):
Looks great, thanks!

jsha:
Sorry, forgot to comment along with the push. I updated along the lines we discussed, but I'm still struggling with getting some of the tests to pass. I think I don't fully understand what they're doing yet. If you have a chance to take a look I'd appreciate, otherwise I can probably puzzle through it next weekend.

jyn514:
@jsha hmm, so this seems like a larger problem - `target-redirect` doesn't support `latest`, only specific version numbers. I think it's ok to change the tests to expect `latest` as long as we make sure that page doesn't 404.

jsha:
target-redirect does appear to support `latest`:

```
$ curl -i https://docs.rs/crate/rustls/latest/target-redirect/x86_64-unknown-linux-gnu/rustls/index.html
HTTP/2 302 
content-length: 0
location: https://docs.rs/rustls/latest/rustls/index.html
```

jyn514:
Ok great, I think you can just update the tests to match the new behavior then :)

jsha:
Alright, I've got almost all of them updated, except this one, which now gets a 404:

```
thread 'web::rustdoc::test::redirect_latest_with_all_yanked::_true' panicked at 'failed to GET /crate/dummy/latest/target-redirect/x86_64-unknown-linux-gnu/dummy/index.html: 404 Not Found', src/test/mod.rs:47:5
```

I don't really understand - what's the expected behavior when all versions are yanked? Do I need to add code such that target-redirect will go to a yanked version if there's no unyanked version available?

Nemo157:
Previously, the behaviour when all versions were yanked is that you could not load the crate using any of the semver/wildcard requests, only an exact version number would load; and if you were on an old version of the crate then it would show a link to the latest version still.

I'm not sure what the behaviour should be after this change, changing the latest version link on those pages to `/latest/` would presumably not work since that's not an exact version link; but not changing it would be an annoying inconsistency. One option would be to just drop the latest version redirect in those cases, a fully yanked crate is a real edge case and you probably don't want to be updating it to the latest version (especially since you would have to manually build your `Cargo.lock` to do the update).

jsha:
By "drop the latest version redirect in those cases," you mean don't generate the link in the header?

Nemo157:
Exactly.

jsha:
Alright, I checked and that's actually the current behavior - no link is generated when all versions are yanked. So I pulled out the failing redirect_latest_with_all_yanked test because it was testing a behavior that doesn't happen. Should be ready to go!

jyn514 (approved):

PR #1564, opened 2021-11-28
Make storage-change-detection.html cacheable

jsha:
This is one of the most-requested URLs on the site, because it is embedded in a lot of different pages and is not cacheable.

### DIFF
--- a/src/web/page/web_page.rs
+++ b/src/web/page/web_page.rs
@@ -1,7 +1,12 @@
 use super::TemplateData;
 use crate::ctry;
 use crate::web::csp::Csp;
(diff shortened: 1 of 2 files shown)


jsha:
Draft because we need to figure out how to make this work with CSP. Possibly a hash of the contents instead of a nonce.

jsha:
Actually, @Nemo157 pointed out that since storage-change-detection is what serves the CSP policy that we care about here, and is also what contains the nonce'd script, we're fine. Also I verified locally that CSP does not block the load.

jyn514 (approved):
Looks great, thanks!

Issue #1563, opened 2021-11-28
stm32f7xx-hal v0.6.0 docs build failure


DanHatesNumbers:
### Crate name

stm32f7xx-hal

### Build failure link

https://docs.rs/crate/stm32f7xx-hal/0.6.0/builds/457266

### Additional details

From the build logs, it appears that the build sandbox could not resolve DNS for crates.io  despite 2 retries.

jyn514:
Hmm, this happens because cargo tries to fetch sources for the dependencies when network access is blocked. But those sources should already have been fetched by rustwide before entering the sandbox. I guess cargo might make different decisions for fetching when we pass --target? I wonder if this is related to https://github.com/rust-lang/docs.rs/pull/1559 somehow.

Nemo157:
The build is from 2021-11-02, so before the #1559 changes.

Nemo157:
What's weird is that it's attempting to download `stm32-fmc 0.2.4`, while the `Cargo.lock` specifies `stm32-fmc 0.2.3`, so for some reason the lockfile got invalidated? (but at a point where it should be readonly?)

Nemo157:
The [uploaded `Cargo.lock`](https://docs.rs/crate/stm32f7xx-hal/0.6.0/source/Cargo.lock) does show 0.2.4 (which answers a question I've long had on whether the uploaded source is actually the crate contents, or docs.rs modifies it during the build).

The build was from before the currently retained logs, so no extra clues there. I have requeued the build to see if it works correctly now or if we can get some more info.

Nemo157:
Ah, the initial build fails

```
 error[E0432]: unresolved import `embedded_hal::digital::v2::PinState`
   --> src/gpio.rs:50:9
    |
 50 | pub use embedded_hal::digital::v2::PinState;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `PinState` in `digital::v2`
```

then it regenerates the `Cargo.lock` and tries again (https://github.com/rust-lang/docs.rs/pull/1490), but it doesn't refetch the newly locked dependencies; and then only uploads the second build log.

Nemo157:
Has been rebuilt and docs are available.

PR #1565, opened 2021-11-28
Generate lockfile and fetch dependencies outside sandbox on retry

Nemo157:
for #1563

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -807,18 +807,18 @@ mod tests {
             let mut conn = env.db().conn();
             let rows = conn
                 .query(
-                    "SELECT 
+                    "SELECT
                         r.rustdoc_status,
                         r.default_target,
                         r.doc_targets,
                         r.archive_storage,
                         cov.total_items
-                    FROM 
-                        crates as c 
+                    FROM
+                        crates as c
                         INNER JOIN releases AS r ON c.id = r.crate_id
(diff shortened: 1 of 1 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -1019,4 +1028,19 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    #[ignore]
+    fn test_locked_fails_unlocked_needs_new_deps() {
```
Nemo157, line 1034: This is a very slow test, it takes ~2 minutes on my machine, there's probably smaller crates that could be used instead, but I don't know an easy way to find them.

jyn514: We could just publish one - if we prefix it with `docs_rs_test` I don't think anyone will mind.

Nemo157: Mmm yeah, that should be pretty easy to setup. I'll publish one and update this test tomorrow.

jyn514: (we could also publish to staging.crates.io, but I don't think the 1-10 crates we ever end up using are a big deal compared to the ~35k that already exist)

jyn514 (approved):
LGTM other than the test, feel free to merge when you've made it faster to run :)

Nemo157:
Nice, that made the CI/build_tests about 5 minutes faster 😁

PR #1566, opened 2021-11-28
Remove rust-navigation-item style

jsha:
AFAICT the only code that generated nodes with this class was removed in
9826ca0f1ed6b373ff49f182a5e44f8fe25ba89f. I noticed the issue because
/rust-logo.png shows up as one of the top requested URLs, but is a 404.


jyn514:
That is ... a while ago :sweat_smile: excellent catch, thanks!

PR #1573, opened 2021-12-06
Make font-awesome-as-a-crate usable from a const context

notriddle:
This allows it to fit more cleanly into codebases like rustdoc that use constants for their assets.

https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc/topic/icon.20harmonization/near/263796393


jyn514 (approved):
👍 seems useful

PR #1575, opened 2021-12-06
Don't record sitemap as a 'static resource' for metrics

Nemo157:
This is very much not a static resource, and is probably one of the slowest responses we have.


jyn514 (approved):
Thanks!

jyn514:
Looks like tests are failing, but it looks good to me other than that.

PR #1579, opened 2021-12-13
Fix new sidebar changes

### LINKED ISSUE #1574: Recent docs build renders the left column oddly
https://docs.rs/imagine/0.0.6/imagine/

The gray background of the left column doesn't go all the way to the edge of the window.

GuillaumeGomez:
Fixes #1574.

Screenshot from 2021-12-13 22-13-04
Screenshot from 2021-12-13 22-21-56

This PR fixes a few issues following the sidebar layout changes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -77,6 +77,7 @@ fn compile_sass() -> Result<(), Box<dyn Error>> {
 
     // Compile rustdoc.scss -> rustdoc.css
     compile_sass_file("rustdoc", "rustdoc", &[])?;
+    compile_sass_file("rustdoc-2021-12-06", "rustdoc-2021-12-06", &[])?;
 
     // Compile vendored.scss -> vendored.css
     compile_sass_file(
--- a/src/utils/rustc_version.rs
+++ b/src/utils/rustc_version.rs
@@ -1,5 +1,7 @@
 use crate::error::Result;
 use anyhow::{anyhow, Context};
+use chrono::prelude::*;
+use once_cell::sync::Lazy;
 use regex::Regex;
 
 /// Parses rustc commit hash from rustc version string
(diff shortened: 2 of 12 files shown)


GuillaumeGomez:
Now that I think about it, I don't think docs.rs css files have versions. Do we have a way to make it work starting a given rustdoc version?

Nemo157:
Yeah, was just gonna say this causes issues on existing pages:

image

AFAIK there is no current setup to do per-version css.

GuillaumeGomez:
Indeed, good catch. I'll send a fix for the scroll bug. Any idea for the css versioning?

GuillaumeGomez:
Something kinda like this would work. Having a CSS file based on the rustdoc version would be even better.

GuillaumeGomez:
I finally did as follows: we generate a new `<link>` into the newly generated crates' documentation `<head>` so that it uses a new CSS overloading the old one. So the old crates style is still working and the new ones don't have issues anymore.

notriddle:
This does not seem like a sustainable, long-term solution.

It looks like the `rustc_version` is stored in the releases table. Could that be used to pick between two different CSS files?

GuillaumeGomez:
I'll take a look. If the date is included in the version, then we will be able to completely go around this problem!

GuillaumeGomez:
@notriddle Thanks a lot for your idea! It allowed me to determine when receiving the query for `rustdoc.css` which one I should provide depending on the rustdoc version used to generate it. Like that, the problem is fixed as soon as this fix is merged!

Nemo157:
Using the recorded date rather than changing the generated docs is definitely much nicer.

I think rather than sending the date as part of the request query I think we should just embed different `/-/static/...` urls in the html; that will cache slightly better as it only has the 2 different urls for the different css versions, instead of one url per-nightly.

GuillaumeGomez:
You mean generating the correct CSS path depending on the rustdoc version in the tera template?

Nemo157:
Yep.

GuillaumeGomez:
Makes sense! I'll update to do so then.

GuillaumeGomez:
Updated! Thanks a lot for the suggestion @Nemo157!

build.rs
```diff
@@ -77,6 +77,7 @@ fn compile_sass() -> Result<(), Box<dyn Error>> {
 
     // Compile rustdoc.scss -> rustdoc.css
     compile_sass_file("rustdoc", "rustdoc", &[])?;
+    compile_sass_file("rustdoc-2", "rustdoc-2", &[])?;
```
jyn514, line 80: This isn't super informative - could you name it after the date or PR number instead?

src/utils/rustc_version.rs
```diff
@@ -19,6 +21,38 @@ pub fn parse_rustc_version<S: AsRef<str>>(version: S) -> Result<String> {
...
+        year.parse::<i32>().unwrap(),
+        month.parse::<u32>().unwrap(),
+        day.parse::<u32>().unwrap(),
+    ))
+}
+
+/// Picks the correct "rustdoc.css" static file depending on which rustdoc version was used to
+/// generate this version of this crate.
+pub fn get_correct_docsrs_style_file(version: &str) -> String {
+    if let Ok(date) = parse_rustc_date(version) {
```
jyn514, line 45: Can this return an error instead, so we know we need to fix it instead of it silently failing?
```suggestion
    let data = parse_rustc_date(version)?;
```

jyn514 (commented):
Can you add a test or two for the new file? All the stringly-typed "rustdoc-2"s make me nervous.

build.rs
GuillaumeGomez: Done.

src/utils/rustc_version.rs
GuillaumeGomez: I did and use `.unwrap()` in two places.

GuillaumeGomez:
> Can you add a test or two for the new file?

Good idea! I will do soon.

GuillaumeGomez:
Actually I'm wondering: what tests do you have in mind?

GuillaumeGomez:
I added a test @jyn514. Hope it's the one you had in mind.

jyn514 (approved):
Yup, this is great, thanks!

syphar:
I did a manual test, I see that new doc-builds are definitely fixed, and the additional CSS file is added only in docs built with newer rustc version. So that's fine with me. 

@Nemo157 @GuillaumeGomez did one of you do a manual test of the last version with old docs? 

I recently wiped my local db & s3, so I don't have any. If you don't have them either, I would fetch some old archives from S3. 

If that's tested I'll merge and deploy.

syphar (approved):

GuillaumeGomez:
I painfully did. :')

syphar:
> I painfully did. :')

Reading my question from yesterday evening again my wording was a little off. 

I mean a manual test with the latest state in this PR, testing a release / build that was done before december 6th.

GuillaumeGomez:
I understood it this way and I confirm that's what I did. ;)

PR #1597, opened 2022-01-16
Switch from position: fixed to sticky

jsha:
This allows us to remove `padding-top: $top-navbar-height !important` on body.

Also, remove an obsolete workaround for older Chrome browsers to put scroll-padding-top on `body` rather than `html`.

Part of #1595

Tested locally with latest Chrome on Desktop in Dev Tools mobile mode, and latest Chrome on latest Android.

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -15,7 +15,7 @@ div.nav-container {
     left: 0;
     right: 0;
     top: 0;
-    position: fixed;
+    position: sticky;
     color: var(--color-navbar-standard);
     font-family: $font-family-sans;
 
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Changes look good. I'll try it locally as soon as possible.

syphar:
thanks @GuillaumeGomez 

Please comment here when the local test is fine, then I'll merge/deploy

GuillaumeGomez:
Just checked locally and the footer is not where it should be (you can see it much better in the second image):

Screenshot from 2022-01-16 17-17-40

Screenshot from 2022-01-16 17-19-04

Normally, the footer should be stuck on the bottom of the page.

syphar (changes requested):
could you check the comment by @GuillaumeGomez

jsha:
Fixed! Satisfyingly, the answer was additional simplification. :-)

GuillaumeGomez (approved):

GuillaumeGomez:
This time it's all good, thanks!

PR #1610, opened 2022-01-18
Do not duplicate `#` when using the `y` permalink shortcut

### LINKED ISSUE #1609: The `y` permalink shortcut duplicates the `#` from an anchor
Indeed, it replaces `#anchor` with `##anchor`.

  - cc @jsha & https://github.com/rust-lang/docs.rs/pull/1583

danielhenrymantilla:
Given the issue #1609, it looks like `document.location.hash` already bundles `#`?

  - ~~If my hypothesis is correct~~, this should~~ fix #1609, but I may also be wrong. Either way, the review ought to be simple, @jsha: either I'm right and this can be merged, or I'm wrong, and this can be closed~~ 😄 

  - EDIT: This does fix #1609

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -239,7 +239,7 @@
         if (ev.key == "y") {
             let permalink = document.getElementById("permalink");
             if (document.location.hash != "") {
-              permalink.href += "#" + document.location.hash;
+              permalink.href += document.location.hash;
             }
             history.replaceState({}, null, permalink.href);
         }


danielhenrymantilla:
Given

Screen Shot 2022-01-18 at 21 12 41

I now believe my fix is right 🙂

jsha:
Looks good to me! 👍🏻

PR #1618, opened 2022-01-22
Fix resize of top navbar

GuillaumeGomez:
Here what the issue looks like:

Screenshot from 2022-01-22 12-07-51

Once fixed it looks like this:

Screenshot from 2022-01-22 12-31-17

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -149,8 +149,8 @@ div.nav-container {
     // These hardcoded values are "magic", in the sense they were observed to be the lowest-possible
     // thresholds through experimentation with a package name that hit the max-width of the element.
     $full-width: 1142px;
-    $medium-width: 860px;
-    $narrow-width: 805px;
+    $medium-width: 928px;
+    $narrow-width: 872px;
 
     // use a .title span inside a menu to hide it on small screens
     span.title {


syphar (commented):
When I try a newer package I cannot reproduce the problem, is this still the right place to fix it? 

examples: 
- https://docs.rs/serde_json/1.0.76/serde_json/ (recent release, doesn't seem to have this problem?) 
- https://docs.rs/itertools/0.10.3/itertools/index.html here I can see the failure

GuillaumeGomez:
I can reproduce it on the serde_json you provided:

https://user-images.githubusercontent.com/3050060/150650884-8d0f6780-f55e-4278-a77a-e2f6da433a79.mp4

It's when reducing the screen width at a precise spot (like usual).

syphar:
Yeah, you're right. On another computer I can reproduce it now (same OS & Browser & version). 

Anyways. Manual test of the fix is fine. 

Can someone review the code-change itself? @Nemo157 ? Or @jsha ?

Nemo157:
I seem to still have the issue on <https://docs.rs.dev.nemo157.com/dialectic-tokio-serde/0.1.0/dialectic_tokio_serde/>, I've checked the css in FF devtools and it definitely has the new media-query sizes.

GuillaumeGomez:
Ok, I think I understand the problem: we need to set a maximum width to the crate name in the topbar. I'll do it in this PR too.

Nemo157:
I'm pretty sure we already do, that's a crate that exceeds it and gets `...` applied which is why I have it there as a testcase.

GuillaumeGomez:
Then we don't have the right sizes. Let me check in any case and if needed I'll update the values. Thanks for the feedback!

GuillaumeGomez:
I took the biggest possible case this time to ensure it was working in all cases.

Nemo157:
This seems to fix it now in responsive design mode, but not when I actually resize my firefox window. I think it might not be taking the scrollbar width into account for windows that actually have scrollbars instead of overlays?

GuillaumeGomez:
@Nemo157 Sorry I don't understand. Do you have a screenshot with the issue please?

Nemo157:
Screenshot by manually shrinking the window (this is with `window.innerWidth = 924` and `document.body.clientWidth = 907`, because of the 17px wide scrollbar):

image

vs screenshot using responsive design mode (giving `window.innerWidth = 919` and `document.body.clientWidth = 919`)

image

AFAICT after a little searching it's literally impossible to deal with the scrollbar width in css, the only solution would be to use JS to actually measure the width, or use a hardcoded offset of the max scrollbar width you will support. (Maybe @jsha has some other ideas? I am not at all a webdev anymore so I could be missing some new tricks 😁).

GuillaumeGomez:
Oh I see! Well then let's just always take the scrollbar in account. :)

GuillaumeGomez:
I added 6px in both sizes to take the sidebar into account.

Nemo157 (approved):

PR #1625, opened 2022-01-25
Add top and height settings for mobile sidebar

jsha:
The CSS classes for the mobile sidebar changed recently, so use the new
classes. Also set the height of the sidebar appropriately.

This fixes a bug where the sidebar is on top of the hamburger menu,
which prevents closing it once you've opened it.

### DIFF
--- a/templates/style/rustdoc-common.scss
+++ b/templates/style/rustdoc-common.scss
@@ -98,6 +98,16 @@ div.rustdoc {
                     top: $top-navbar-height;
                 }
             }
+
+            // A later version of rustdoc uses the .sidebar.shown instead of .sidebar.mobile, and
+            // also needs a height adjustment.
+            &.shown {
+                width: 250px;
+                margin-left: 0;
+                // 45px is the size of the rustdoc mobile-topbar
+                top: 45px + $top-navbar-height !important;
+                height: calc(100vh - 45px - #{$top-navbar-height}) !important;
+            }
         }
     }
 


jsha:
For context reviewing, it's useful to look at the lines above this change - line 84 to 100.

GuillaumeGomez:
Can you show screenshots with the bug too please? It'd make my life easier to check it. ^^'

jsha:
Good point! Also, steps to reproduce:

1. Open a recently-generated page on docs.rs.
2. Click on sidebar toggle.

Expected result:

Sidebar toggle stays visible.

Actual result:

Sidebar toggle is hidden.

Buggy:

image

Fixed:

image

templates/style/rustdoc-common.scss
```diff
@@ -98,6 +98,15 @@ div.rustdoc {
                     top: $top-navbar-height;
                 }
             }
+
+            // A later version of rustdoc uses the .sidebar.shown instead of .sidebar.mobile, and
+            // also needs a height adjustment.
+            &.shown {
+                width: 250px;
+                margin-left: 0;
+                top: 45px + $top-navbar-height !important;
```
GuillaumeGomez, line 107: I think you need a `calc()` here.

templates/style/rustdoc-common.scss
```diff
@@ -98,6 +98,15 @@ div.rustdoc {
...
+                height: calc(100vh - 78px) !important;
```
GuillaumeGomez, line 108: Why the `78px`?

templates/style/rustdoc-common.scss
GuillaumeGomez: Also, why the `45px`?

templates/style/rustdoc-common.scss
jsha: 78px is 45px (rustdoc topbar) plus 33px (docsrs topbar).

templates/style/rustdoc-common.scss
jsha: 45px is the size of the rustdoc topbar.

`calc()` is not needed because this is uses [sass simplification](https://sass-lang.com/documentation/values/calculations#simplification). Basically sass does the math at compile time and emits `top: 78px` into the file.

templates/style/rustdoc-common.scss
GuillaumeGomez: Then better use `$top-navbar-height` like you did above and add a comment about the rustdoc topbar. Which now brings a question: for older crates, won't your fix break the sidebar?

templates/style/rustdoc-common.scss
GuillaumeGomez: Oh that's cool. I didn't know that. :)

GuillaumeGomez:
Please add a comment for the `45px` explaining that it's the rustdoc sidebar "top".

templates/style/rustdoc-common.scss
jsha: It won't break the sidebar because older rustdoc versions didn't use the `.shown` class.

GuillaumeGomez: Good point!

jsha:
Updated, thanks!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #1661, opened 2022-02-05
Replace custom `index::crates::Crates` with `crates_index::Crates`


Nemo157:
After the upgrade in #1660 the comment here is out of date:

https://github.com/rust-lang/docs.rs/blob/5c3b6898c34add186661b34ad0c6e947e9380a86/src/index/mod.rs#L97-L100

The new version of `crates_index` supports bare repositories so we should be able to use it directly.

jyn514:
@Nemo157 hmm, I'm not quite sure I follow - I don't see a way to walk the repo in https://docs.rs/crates-index-diff/latest/crates_index_diff/struct.Index.html, you have to know the initial commit ahead of time. And we can't hard-code the commit because it can change when crates.io squashes the index: https://github.com/rust-lang/crates.io-index/commit/dcb14b9012f15cf806bd0a61c982e9d6f76a7d63

Nemo157:
<https://docs.rs/crates-index/latest/crates_index/>, not `crates-index-diff`. That is the library we use to parse the point-in-time index to compare what is in the database against. Back when that was first implemented its iterator worked off the files in a checked out working directory, but since then has been updated to work directly off git objects.

PR #1677, opened 2022-02-13
Use `crates-index` instead of rewriting it

jyn514:
This is possible now that it supports bare checkouts. This also takes the
liberty of iterating the crates in parallel, which should give a nice speedup.

Fixes https://github.com/rust-lang/docs.rs/issues/1661

r? @Nemo157


src/utils/consistency/index.rs
```diff
@@ -1,11 +1,12 @@
+use rayon::iter::ParallelIterator;
 use super::data::{Crate, CrateName, Data, Release, Version};
 use crate::Index;
 
 pub(crate) fn load(index: &Index) -> Result<Data, anyhow::Error> {
     let mut data = Data::default();
 
-    index.crates()?.walk(|krate| {
-        data.crates.insert(
+    data.crates = index.crates()?.crates_parallel().map(|krate| {
```
Nemo157, line 8: Lol, I had written a comment that this could be `let crates = ...; Ok(Data { crates })`; but changed my mind. Looks like Clippy would prefer that though.

Nemo157 (approved):

PR #1698, opened 2022-03-16
Add GUI tests

GuillaumeGomez:
Part of #1494.

I finally took the time to wrap something together. So in this case, I took what we do for rustdoc and simply ported it to docs.rs. I added 2 GUI tests so you can take a look at what they look like. The documentation for the framework is [here](https://github.com/GuillaumeGomez/browser-UI-test/blob/master/goml-script.md).

Do you think this is the right approach or would you prefer another framework?


.github/workflows/ci.yml
```diff
@@ -81,6 +81,43 @@ jobs:
...
+          sleep 5
+          # Make sure the database is actually working
+          psql "${DOCSRS_DATABASE_URL}"
+          cargo run -- database migrate
+          cargo run -- build crate sysinfo 0.23.4
+          cargo run -- build crate sysinfo 0.23.5
+          cargo run -- build add-essential-files
+          cargo run -- start-web-server &
+          # Give the web server time to start up
+          sleep 1
```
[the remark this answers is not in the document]
jyn514: the `build crate` commands specifically - the rest are fine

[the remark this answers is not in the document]
GuillaumeGomez: Actually I need these crates as well. I'm not sure what's the best way to have something easy to reproduce. Would it be fine to put everything into a docker instead?

GuillaumeGomez:
I'm a bit stuck with the current docker situation. So to sum up what I tried to and what my issue is:

 * I want to have a docs.rs environment set up with a "clean" database with only the crates/updates we need for the GUI tests. The other point is that it doesn't "pollute" the existing DB by adding new crates.
 * It should be reproducible easily.
 * The docker image doesn't need to be rebuilt every time you make a GUI test update.

So now the problem is that I think putting everything in one docker container isn't a great idea because it makes a big image which needs to be rebuilt every time there is a change on docs.rs source code.

The other problem is that apparently, we can't start docs.rs web server in the container: it exists right away (without an error).

So first: is it the right approach? Secondly: any help would be very appreciated.

GuillaumeGomez:
So I was finally able to make it work like I wanted. I even added it to the docker-compose and I added a shell script to make it simpler to run.

GuillaumeGomez:
Maybe ping @Nemo157 ?

GuillaumeGomez:
@syphar It's "ready" now, we need to discuss about if this looks good to you or if it seems to complicated to maintain and eventually try to pick another solution.

syphar:
> @syphar It's "ready" now, we need to discuss about if this looks good to you or if it seems to complicated to maintain and eventually try to pick another solution.

Thanks for the ping! 
This topic was next on my list to dig into. 

For my understanding: rustdoc is actively using the same framework?

GuillaumeGomez:
Yes. You can see it being used in `src/test/rustdoc-gui`.

GuillaumeGomez:
I updated to the last version of the framework and added some documentation in case you want to run it locally. So with this, it should be ready.

From here, the plan is then to test as much as possible our UI, especially the top-navbar and all docs.rs pages.

.github/workflows/ci.yml
```diff
@@ -81,9 +81,15 @@ jobs:
...
           done
 
+      - name: Run GUI tests
+        run: ./dockerfiles/run-gui-tests.sh
+
       - name: Clean up the database
         run: docker-compose down --volumes
 
+    steps:
+      - uses: actions/checkout@master
```
syphar, line 91: what's this for?

syphar: ah, probably the reason for CI failing: 
```
Invalid workflow file:                          .github/workflows/ci.yml#L132                                                                                                  
You have an error in your yaml syntax on line 132
```

.github/workflows/ci.yml
```diff
@@ -81,9 +81,15 @@ jobs:
             $f || exit 1
           done
 
+      - name: Run GUI tests
```
syphar, line 84: I wonder if this should actually be a separate job, running in parallel to `tests` and `build_tests`

syphar: we probably could also use `upload-artifact` above to upload the `cratesfyi` binary above to achieve this.

gui-tests/tester.js
```diff
@@ -0,0 +1,252 @@
+// This package needs to be install:
```
syphar, line 1: how much of this is specific to docs.rs? 

skimming over it I would guess it's generic? is it the same as in rustdoc?

.github/workflows/ci.yml
syphar: or perhaps rather part of the `build_tests`, so the inital build setup is only done once, we could test this

dockerfiles/run-gui-tests.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Just in case it's running, we stop the web server.
+docker-compose stop web
+
+docker-compose up -d db s3
+
+# We add the information we need.
+cargo run -- database migrate
+cargo run -- build crate sysinfo 0.23.4
```
syphar, line 10: this won't work on mac systems, or on non-amd64 linux systems (from what I remember). 

But I would be fine leaving it as-is and fixing it later when I find the time :) 

If you want time with it, I typically use the web container to run builds via `docker compose run web build crate ...`

syphar (changes requested):

.github/workflows/ci.yml
GuillaumeGomez: So which one should I do? :)

gui-tests/tester.js
GuillaumeGomez: It's a simplified version.

dockerfiles/run-gui-tests.sh
GuillaumeGomez: Is having the time really that important here?

GuillaumeGomez: Yes please do. I did my best at trying to set up something as easy as possible to use so you don't have to install `node`, `npm` and `npm` packages but I'm not sure my dockerfile is good. ^^'

gui-tests/tester.js
GuillaumeGomez: Hopefully it might get useless in the future if I make the default runner for the framework better.

.github/workflows/ci.yml
GuillaumeGomez: Ok I'll put it into its own testsuite then.

GuillaumeGomez:
CI finally passed! \o/

PR is now ready. :)

syphar (commented):
only one last comment from my side. I wonder if we can safe time on the setup, or if caching will help a a little, 20 minutes, are quite long. 

I would like to have a short test from @Nemo157 because I can't currently run the tests on my mac

dockerfiles/run-gui-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+
+. .env
+
+cargo run -- start-web-server &
+SERVER_PID=$!
+
+docker build . -f dockerfiles/Dockerfile-gui-tests -t gui_tests
+
+# status="docker run . -v `pwd`:/build/out:ro gui_tests"
+docker-compose run gui_tests
```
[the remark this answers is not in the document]
GuillaumeGomez: No idea. Let's try without it.

GuillaumeGomez:
> only one last comment from my side. I wonder if we can safe time on the setup, or if caching will help a a little, 20 minutes, are quite long.

Yes, 20 minutes is way too long. If you know how to cache it, I'd be very happy to do it.

> I would like to have a short test from @Nemo157 because I can't currently run the tests on my mac

Is it because of this PR or something specific to your mac?

dockerfiles/run-gui-tests.sh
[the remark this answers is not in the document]
syphar: or it should be just `docker compose build ...`

syphar:
> > only one last comment from my side. I wonder if we can safe time on the setup, or if caching will help a a little, 20 minutes, are quite long.
> 
> Yes, 20 minutes is way too long. If you know how to cache it, I'd be very happy to do it.

do you know where most of the time is spent? 


> 
> > I would like to have a short test from @Nemo157 because I can't currently run the tests on my mac
> 
> Is it because of this PR or something specific to your mac?

I mean these GUI tests don't run, the other tests do. 

the `-- build crate ...` things don't run, and I have to use `docker compose run web build crate... `  

but running the test script leads to another error related to the host / container platform (I'm on an M1). 

But IMO this doesn't block this PR, I'll figure it out afterwards.

GuillaumeGomez:
Oh wow. It took only 8 minutes without `docker build`.

Nemo157:
> the `-- build crate ...` things don't run, and I have to use `docker compose run web build crate... `

I also can't use a local `build crate`, the rustup that rustwide downloads doesn't like nixos.

GuillaumeGomez:
Wow, weird. I thought using docker would prevent all these issues. ^^'

Nemo157:
rustwide does all the installation outside docker containers for some reason (it should be possible to move that all into the containers, which would probably make it work on macos just fine too).

GuillaumeGomez:
I see. Do you want it to be fixed before we merge this PR or after?

Nemo157:
It's orthogonal to this PR (and given maintenance status of rustwide unlikely to happen soon). I plan to see if I can adapt the script here to run everything inside docker-compose containers, can probably get around to it today if it's not too complicated.

GuillaumeGomez:
Thanks! :D

Nemo157:
I got it working in <https://github.com/rust-lang/docs.rs/commit/16d4957acfdad0d0b0fbac88d49b9d6521a9a96a> and managed to run them locally, but I think this can be much simpler after the refactoring in #2269, so I'd merge as is and I'll try to finish that up and update the script later.

syphar:
@Nemo157 so you think we should merge this PR now, and then improve on #2269 ?

Nemo157:
Yep

syphar (approved):

GuillaumeGomez:
Awesome! \o/

PR #1717, opened 2022-04-10
Fix dropdown menu display in top navbar

GuillaumeGomez:
Fixes https://github.com/rust-lang/docs.rs/issues/1714.

We can't use `overflow-y: hidden` because it's then impossible (unless we use `position: fixed`) to display the dropdown menus. There is an ongoing discussion in the w3c about this [here](https://github.com/w3c/csswg-drafts/issues/4092). But still far from done and even more far from implementation.


GuillaumeGomez:
Doesn't look very nice but at least not broken anymore:

Screenshot from 2022-04-10 13-26-32

syphar (approved):

PR #1719, opened 2022-04-11
Fix top navbar style

GuillaumeGomez:
I just realized that the display of the top navbar menu is broken (well, not good looking):

Screenshot from 2022-04-11 14-06-07

It comes from my changes in https://github.com/rust-lang/docs.rs/pull/1717 to allow having a different font size.

Overall, the style for the top navbar is quite the nightmare so I decided to remove the `float` properties and to use `flex` instead. The result is quite nice (imo) and fixes some small UI issues:

Screenshot from 2022-04-11 15-16-23

And it looks like this with bigger font size:

Screenshot from 2022-04-11 15-12-54

I checked other pages and they work well too.

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -28,6 +28,12 @@ div.nav-container {
     color: var(--color-navbar-standard);
     /* The font size must be specified in pixels because the height is specified in pixels. */
     font: 16px $font-family-sans;
+    position: relative;
+
+    .container, .pure-menu-horizontal {
+        position: relative;
+        height: 100%;
+    }
 
     li {
         border-left: 1px solid var(--color-border);
@@ -73,18 +79,18 @@ div.nav-container {
         }
     }
 
-    .pure-menu-right {
-        float: right;
-    }
-
     form.landing-search-form-nav {
         max-width: 1200px;
+        height: 100%;
+        display: flex;
+        flex-direction: row;
 
         #search-input-nav {
-            float: right;
             max-width: 150px;
             display: none;
             border-left: 1px solid var(--color-border);
+            height: 100%;
+            overflow-x: hidden;
 
             @media #{$media-sm} {
                 display: block;
@@ -107,7 +113,7 @@ div.nav-container {
                 font-size: 0.8em;
                 box-shadow: none;
                 background-color: var(--color-background);
-                height: 31px;
+                height: 100%;
             }
         }
 
@@ -122,8 +128,21 @@ div.nav-container {
            See https://github.com/rust-lang/docs.rs/issues/1669.
            */
        .pure-menu-item a {
-            /* 0.5 em is the padding */
-            max-height: calc(#{$top-navbar-height} - 0.5em * 2);
+            max-height: 100%;
+            overflow-x: hidden;
+            text-overflow: ellipsis;
+       }
+
+       .docsrs-logo, .pure-menu-item a {
+            padding: 8px 1em;
+       }
+
+       .pure-menu-item {
+            height: 100%;
+       }
+
+       .spacer {
+            flex-grow: 1;
        }
     }
 
--- a/templates/header/topbar_begin.html
+++ b/templates/header/topbar_begin.html
@@ -3,9 +3,6 @@
     The only piece of context it can take is `search_query`, which should
     be a string and will populate the search field if it exists
 #}
-
-{%- import "macros.html" as macros -%}
-
 <div class="nav-container">
     <div class="container">
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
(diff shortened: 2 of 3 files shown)


jsha:
In my local Chrome, with the light theme, it looks like this:

image

There are grey rectangles between the menu names. Looking the Dev Tools, it appears to be due to these rules in the light CSS:

```
::-webkit-scrollbar-thumb {
    background-color: rgba(36,37,39,0.6);
}
::-webkit-scrollbar-track {
    background-color: #ecebeb;
}
```

Those are getting applied to the `<div class="pure-menu pure-menu-horizontal">`, but I'm not sure why.

GuillaumeGomez:
Great catch! I'll check tomorrow what's going on.

templates/style/_navbar.scss
```diff
@@ -122,8 +128,21 @@ div.nav-container {
            See https://github.com/rust-lang/docs.rs/issues/1669.
            */
        .pure-menu-item a {
-            /* 0.5 em is the padding */
-            max-height: calc(#{$top-navbar-height} - 0.5em * 2);
+            max-height: 100%;
+            overflow-x: hidden;
```
jsha, line 132: I think these lines are unnecessary. When I remove these properties locally, the unwanted grey rectangles (which are tiny vertical scrollbars) go away. Also, with these properties removed, even if the menu items' font size is weird, it doesn't interfere with the main page.

jsha (changes requested):

templates/style/_navbar.scss
jsha: In fact, I think this whole CSS rule (including the comment about what happens with a locally installed old Fira Sans) should get deleted. Switching to the flex layout seems like it sufficed to fix the issue with overflowing menu titles, even without this rule.

jsha: (I just tested in a locally installed Firefox, with the Fira Sans from https://github.com/rust-lang/docs.rs/issues/1669#issuecomment-1068396583, at these commits:

1. ad8d4fe33ba6d48669175221f3b651eb5e1e78ec (i.e. right before #1711)
2. 0ae58075652454f6bd535e39fdcd7be27f846fd5 (this branch)
3. 0ae58075652454f6bd535e39fdcd7be27f846fd5 with this CSS rule removed

And I can confirm that at (1) I have the overflow issue. Both (2) and (3) fix the overflow issue, and (3) additionally fixes the scrollbar issue introduced in this branch.

BTW the scrollbar issue seems to be Chrome-only.

templates/style/_navbar.scss
```diff
@@ -28,6 +28,12 @@ div.nav-container {
     color: var(--color-navbar-standard);
     /* The font size must be specified in pixels because the height is specified in pixels. */
     font: 16px $font-family-sans;
+    position: relative;
+
+    .container, .pure-menu-horizontal {
+        position: relative;
+        height: 100%;
```
jsha, line 35: What are the `position: relative` and `height: 100%` needed for here? I think they are unnecessary.

templates/style/_navbar.scss
```diff
@@ -73,18 +79,18 @@ div.nav-container {
         }
     }
 
-    .pure-menu-right {
-        float: right;
-    }
-
     form.landing-search-form-nav {
         max-width: 1200px;
+        height: 100%;
```
jsha, line 84: I think you can get rid of this `height: 100%`. That makes the search field vertically misaligned, but you can fix that by also removing the `height: 100%` on line 92.

templates/style/_navbar.scss
```diff
@@ -73,18 +79,18 @@ div.nav-container {
...
+        display: flex;
+        flex-direction: row;
 
         #search-input-nav {
-            float: right;
             max-width: 150px;
             display: none;
             border-left: 1px solid var(--color-border);
+            height: 100%;
+            overflow-x: hidden;
```
jsha, line 93: Why do we need `overflow-x: hidden` here?

templates/style/_navbar.scss
```diff
@@ -122,8 +128,21 @@ div.nav-container {
...
+            text-overflow: ellipsis;
+       }
+
+       .docsrs-logo, .pure-menu-item a {
+            padding: 8px 1em;
+       }
+
+       .pure-menu-item {
+            height: 100%;
```
jsha, line 141: I think this is also unnecessary.

jsha (changes requested):

jsha:
Overall, great change! I am really happy to have the items show up in the same HTML order as the visual order on the page.

templates/style/_navbar.scss
GuillaumeGomez: It's because with this, we can set the element of the navbar to `height: 100%`. It prevents the bottom margin to disappear in case the font size is too big.

templates/style/_navbar.scss
GuillaumeGomez: Without this rule:

Screenshot from 2022-04-13 16-51-54

With this rule:

Screenshot from 2022-04-13 16-52-21

So we definitely need it.

templates/style/_navbar.scss
GuillaumeGomez: Without the `height: 100%`:

Screenshot from 2022-04-13 16-56-07

GuillaumeGomez:
I tested all of them and they all ended up in bad rendering in firefox. :-/

jsha:
Which type of bad rendering? The "body text is too narrow" bad rendering? Or "bottom border of menu title is not visible"?

GuillaumeGomez:
Size of top navbar items becomes chaotic:

Screenshot from 2022-04-13 16-56-07

For example, `sysinfo` item even gets its height different from the rest.

jsha:
I spent some time fiddling with this locally and came up with a solution that I think is tidier, avoids the scrollbars on Chrome, and looks right on both Chrome and Firefox. I sent it as a PR against your branch: https://github.com/GuillaumeGomez/docs.rs/pull/1. Let me know what you think!

GuillaumeGomez:
Nice, thanks a lot! Will test it locally as soon as possible.

GuillaumeGomez:
I integrated @jsha's improvements so this should now be ready!

jsha:
I've manually tested the version on this branch in Firefox and Chrome, with the old version of Fira Sans locally installed. Tests I did:

 - Scrolled vertically to make sure the navbar stayed in place
 - Opened and closed menus
 - Verified that the rustdoc content was not artificially narrow
 - Verified that the vertical bars separating menu titles went all the way to the top and bottom of the navbar
 - Verified that menu titles were approximately centered vertically

I also tested with a large font size in firefox, and a large minimum font size. The presentation wasn't perfect but it was good enough. And it didn't interere with the rustdoc content.

syphar (approved):
LGTM, definitely fixes #1725

PR #1720, opened 2022-04-11
Turn remaining top navbar "hover dropdowns" to "click dropdowns"

GuillaumeGomez:
All links in the top navbar are dropdown menus. And you need to click on them to see the dropdown menu... except for these two:

Screenshot from 2022-04-11 15-32-13

For coherency, I turned them from "hover dropdown" into "click dropdown". It'll also allow mobile device users to actually be able to see the dropdown menus for these two items.

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -1489,12 +1489,13 @@ mod tests {
                     continue;
                 }
 
-                let resp = if url.starts_with("http://") || url.starts_with("https://") {
-                    // Skip external links
-                    continue;
-                } else {
-                    web.get(&url).send()?
-                };
+                let resp =
+                    if url.starts_with("http://") || url.starts_with("https://") || url == "#" {
+                        // Skip external links
+                        continue;
+                    } else {
+                        web.get(&url).send()?
+                    };
                 let status = resp.status();
                 assert!(status.is_success(), "failed to GET {}: {}", url, status);
             }


templates/header/topbar_begin.html
```diff
@@ -48,12 +49,17 @@
                     </li>{#
 
                     The Rust dropdown menu
-                    #}<li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover pure-menu-opt">
-                        <a href="https://www.rust-lang.org/" target="_blank" class="pure-menu-link">
-                            Rust
+                    #}<li class="pure-menu-item pure-menu-has-children">
+                        <a href="#" class="pure-menu-link" aria-label="Rust">
+                            <span class="title">Rust</span>
```
jsha, line 54: Why not put the title on the `<a>` tag and skip the `<span>`? This title doesn't even show up on hover for me.

jsha (changes requested):
One comment. Otherwise looks good.

templates/header/topbar_begin.html
GuillaumeGomez: Just checked and it's indeed unnecessary. Removing it.

GuillaumeGomez:
Updated!

jsha (approved):
Looks good!

GuillaumeGomez:
I fixed the conflicts.

cc @syphar :)

syphar (approved):

Issue #1756, opened 2022-07-11
sitemap: use /latest/ URLs


jsha:
Right now https://docs.rs/sitemap.xml (and the sub-sitemaps it links to, like https://docs.rs/-/sitemap/r/sitemap.xml) uses URLs of the form https://docs.rs/rustls. However, that 302's to https://docs.rs/rustls/latest/rustls/. Per https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap "List only canonical URLs in your sitemaps. If you have two versions of a page, list in the sitemap only the one you prefer to appear in search results."

We should list the `<crate>/latest/<crate>` version directly in the sitemap to save on redirects when crawling. And we should change the 302 (temporary) to a 301 (permanent). The `/<crate>` URL used to be a temporary redirect because it would point to a different path depending on the latest version of the crate, but now it will always redirect to `/latest/` so we can make it permanent.

jsha:
Fixed in #1759

PR #1760, opened 2022-07-12
Use 301 redirects when redirect target is /latest/

jsha:
Part of #1756.

Note: I tried adding the test case below, but got 200 instead of a redirect. Any idea what I'm doing wrong?

```rust
    #[test]
    fn test_redirect_to_latest_301() {
        wrapper(|env| {
            env.fake_release().name("dummy").version("1.0.0").create()?;
            let web = env.frontend();
            assert_eq!(
                web.get("/dummy").send()?.status(),
                StatusCode::MOVED_PERMANENTLY,
            );
            Ok(())
        })
    }
```

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -71,7 +71,12 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
             url_str.push_str(query);
         }
         let url = ctry!(req, Url::parse(&url_str));
-        let mut resp = Response::with((status::Found, Redirect(url)));
+        let status_code = if vers == "latest" {
+            status::MovedPermanently
+        } else {
+            status::Found
+        };
+        let mut resp = Response::with((status_code, Redirect(url)));
         resp.headers.set(Expires(HttpDate(time::now())));
 
         Ok(resp)


jyn514:
```
assert_eq!(           web.get("/dummy").send()?.status(),           StatusCode::MOVED_PERMANENTLY,
            );
```
Reqwest follows redirects by default. There's some helper function for testing redirects you can use instead (`assert_redirect` maybe?). Took me ages to write lol

jyn514 (approved):
Lgtm with the test added, good catch!

Nemo157:
Actually, your comment on #1759 reminded me that this is not guaranteed to be permanent. Between versions a package could change its `lib.name` which would change the `target_name`.

Nemo157:
> Do we think this ever happens in practice?

Not sure, I couldn't find it in the couple of crates I knew about that actually overrode `lib.name`, they'd always used the same name.

> That would be a breaking change for anyone using the crate, right?

Yep, but that's fine over a major version bump.

jsha:
> I'm kinda inclined to merge this anyway and just have people force-refresh if the crate changes the library name ... should be super rare in practice.

I think that's correct. But the comments about caching made me realize I should put a short caching header on this to start - e.g. 24 hours. Then we can increase assuming that goes well.

src/web/rustdoc.rs
```diff
@@ -1599,6 +1599,22 @@ mod test {
...
+        wrapper(|env| {
+            env.fake_release().name("dummy").version("1.0.0").create()?;
+            let web = env.frontend();
+            let client = ClientBuilder::new()
+                .redirect(redirect::Policy::none())
+                .build()
+                .unwrap();
+            let url = format!("http://{}/dummy", web.server_addr());
+            let status = client.get(url).send()?.status();
+            assert_eq!(status, StatusCode::MOVED_PERMANENTLY);
```
jyn514, line 1613: can we also test that the target is `/dummy/latest/dummy`?

jsha: Note that I was not able to use `assert_redirect` because that does not assert a specific status code. It seemed small enough that duplicating a couple of lines was better than adding an "expected status" parameter an updating all tests.

PR #1762, opened 2022-07-12
Fix the sidebar height

### LINKED ISSUE #1761: Browser's bottom-left URL display hides last entry in sidebar
This has caught me out today a few times on some crates. I've looked but cannot see a similar issue reported at the moment.

Docs pages have lots of links on them. When a mouse is over a link, the browser (firefox in this case) shows the URL in the bottom left and that hides the last entry in the side bar even when the sidebar is scrolled to the bottom:

image

vs:

image

It is possible to keep scrolling so that the main page scrolls and the footer comes in to view and lifts everything enough to resolve the issue but I think it is possibly preferable to add ~20px of margin/padding at the bottom of the sidebar to stop this happening?

I assume this happens for everyone but I recognise there might be something off with my set up though I use firefox without any significant customisations.

Nemo157:
`calc` is parsed specially in SASS as if the interior is a plain unquoted string and requires using explicit interpolation syntax (see the "fun fact" at https://sass-lang.com/documentation/at-rules/function#plain-css-functions).

Though, that [appears to have changed in DartSASS 1.40](https://sass-lang.com/documentation/values/calculations), where the original code would have worked as-is. LibSASS hasn't had the same change applied (and probably won't since it's deprecated...).

Can test at https://www.sassmeister.com/ with:

```scss
$top-navbar-height: 32px;

#rustdoc_body_wrapper {
    .sidebar {
        top: $top-navbar-height;
        height: calc(100vh - $top-navbar-height);
    }
}

#rustdoc_body_wrapper {
    .sidebar {
        top: $top-navbar-height;
        height: calc(100vh - #{$top-navbar-height});
    }
}
```

fixes #1761

### DIFF
--- a/templates/style/rustdoc-2021-12-05.scss
+++ b/templates/style/rustdoc-2021-12-05.scss
@@ -10,7 +10,7 @@
     .sidebar {
         margin-top: 0;
         top: $top-navbar-height;
-        height: calc(100vh - $top-navbar-height);
+        height: calc(100vh - #{$top-navbar-height});
         /* Since we use `overflow-wrap: anywhere;`, we never have the need for a X scrollbar... */
         overflow-x: hidden;
 


jyn514:
> Though, that [appears to have changed in DartSASS 1.40](https://sass-lang.com/documentation/values/calculations), where the original code would have worked as-is. LibSASS hasn't had the same change applied (and probably won't since it's deprecated...).

why are we using a deprecated library? how hard would it be to switch?

jyn514 (approved):

GuillaumeGomez (approved):

Nemo157:
I took a glance at <https://lib.rs/?sass> and there doesn't appear to be anything built on DartSASS (I don't know how you would even do bindings to a Dart library, it's main consumers seem to be from transpiling it to JS). Eventually https://lib.rs/crates/grass looks like it could have potential, but it's not yet feature complete.

jyn514:
ah, fair enough. unfortunate that the only maintained libraries are experimental or only used from JS ... surprised no one else has had this problem.

Nemo157:
I actually tried throwing `grass` into the build script too, and it seems to support everything we're using, there's just what appears to me to be inconsequential differences: https://gist.github.com/Nemo157/2767f23c2541f3fb42673290a0a302c5

PR #1764, opened 2022-07-12
Support 1.62 running doc-tests on private macros

Nemo157:
Fixes test failures on nightly caused by https://github.com/rust-lang/rust/issues/97030

Pre-emptively uses https://github.com/rust-lang/rust/pull/96573 because we run a `--ignored` testrun in CI which would cause `` ```ignore`` to still fail. This works already since it currently treats `` ```no_compile`` as an unknown filetype and doesn't attempt to build it.

### DIFF
--- a/src/db/migrate.rs
+++ b/src/db/migrate.rs
@@ -11,7 +11,7 @@ use schemamama_postgres::{PostgresAdapter, PostgresMigration};
 ///
 /// Example:
 ///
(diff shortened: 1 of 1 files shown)


jyn514 (approved):

Nemo157:
Heh, actually the change was already in 1.62, but we don't run the doc-tests in CI.

PR #1768, opened 2022-07-14
Authenticate jsha to use Google Search Console

jsha:
This allows troubleshooting issues with why Google doesn't crawl docs.rs deeply enough.

Related to https://github.com/rust-lang/docs.rs/issues/1438#issuecomment-1174811787

Per https://support.google.com/webmasters/answer/9128668?visit_id=637926083810327397-791277710&rd=2:

> Search Console offers tools and reports for the following actions:
>
> - Confirm that Google can find and crawl your site.
> - Fix indexing problems and request re-indexing of new or updated content.
> - View Google Search traffic data for your site: how often your site appears in
> - Google Search, which search queries show your site, how often searchers click through for those queries, and more.
> - Receive alerts when Google encounters indexing, spam, or other issues on your site.
> - Show you which sites link to your website.
> - Troubleshoot issues for AMP, mobile usability, and other Search features.
>

Per https://support.google.com/webmasters/answer/7687615#manage-owners&zippy=%2Cadd-or-remove-verified-owners, removing this handler later will deauthenticate me.

Here's an example of the sort of report I'll be able to get from Google Search Console:

image

### DIFF
--- a/src/web/routes.rs
+++ b/src/web/routes.rs
@@ -21,6 +21,10 @@ pub(super) fn build_routes() -> Routes {
     //   https://developers.google.com/search/reference/robots_txt#handling-http-result-codes
     //   https://support.google.com/webmasters/answer/183668?hl=en
     routes.static_resource("/robots.txt", PermanentRedirect("/-/static/robots.txt"));
+    routes.static_resource(
+        "/googleb54494ba4d52c200.html",
+        super::sitemap::google_search_console_handler,
+    );
     routes.static_resource("/favicon.ico", PermanentRedirect("/-/static/favicon.ico"));
     routes.internal_page("/sitemap.xml", super::sitemap::sitemapindex_handler);
     routes.internal_page(
--- a/src/web/sitemap.rs
+++ b/src/web/sitemap.rs
@@ -148,6 +148,16 @@ pub fn about_handler(req: &mut Request) -> IronResult<Response> {
     .into_response(req)
 }
 
+// This authenticates @jsha (https://github.com/jsha) to use the Google Search Console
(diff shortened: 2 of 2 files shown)


src/web/routes.rs
```diff
@@ -21,6 +21,10 @@ pub(super) fn build_routes() -> Routes {
     //   https://developers.google.com/search/reference/robots_txt#handling-http-result-codes
     //   https://support.google.com/webmasters/answer/183668?hl=en
     routes.static_resource("/robots.txt", PermanentRedirect("/-/static/robots.txt"));
+    routes.static_resource(
+        "/googleb54494ba4d52c200.html",
```
jyn514, line 25: Where does this string come from? I clicked on the "verification flow" link in the webmaster docs, but didn't see anything that looked relevant.

jyn514 (approved):
I think this is fine.

@pietroalbini do you have strong opinions about giving @jsha access to the Google webmaster console? They're not on the docs.rs team but they are on rustdoc, and have a history of contributions. Plus I don't think there's terrible much you can do there other than tell Google not to crawl the site.

src/web/routes.rs
jsha: If you visit https://search.google.com/search-console/welcome from a logged-in Google Account, and click the "URL Prefix" box, enter "https://docs.rs", and hit "continue," you'll get a similar string.

pietroalbini:
I have no problem with giving them access to the webmaster console, *but* we should do it by having the infrastructure team setup the account and then granting jsha access to it, rather than having jsha setup their own account linked to docs.rs. I'll add that to my todo list.

PR #1773, opened 2022-07-19
Add canonical URL for rustdoc pages

jsha:
Omit the link for crates that specify their own documentation URL.

Part of #74 and #1438

### DIFF
--- a/src/web/rustdoc.rs
+++ b/src/web/rustdoc.rs
@@ -195,6 +195,7 @@ pub fn rustdoc_redirector_handler(req: &mut Request) -> IronResult<Response> {
 #[derive(Debug, Clone, PartialEq, Serialize)]
 struct RustdocPage {
     latest_path: String,
+    canonical_url: String,
     permalink_path: String,
     latest_version: String,
     target: String,
@@ -483,6 +484,27 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
 
     let latest_path = format!("/crate/{}/latest{}{}", name, target_redirect, query_string);
 
+    // Set the canonical URL for search engines to the `/latest/` page on docs.rs.
+    // For crates with a documentation URL, where that URL doesn't point at docs.rs,
+    // omit the canonical link to avoid penalizing external documentation.
+    // Note: The URL this points to may not exist. For instance, if we're rendering
+    // `struct Foo` in version 0.1.0 of a crate, and version 0.2.0 of that crate removes
+    // `struct Foo`, this will point at a 404. That's fine: search engines will crawl
+    // the target and will not canonicalize to a URL that doesn't exist.
+    let canonical_url = if krate.documentation_url.is_none()
+        || krate
+            .documentation_url
+            .as_ref()
+            .unwrap()
+            .starts_with("https://docs.rs/")
+    {
+        // Don't include index.html in the canonical URL.
+        let canonical_path = inner_path.replace("index.html", "");
+        format!("https://docs.rs/{}/latest/{}", name, canonical_path)
+    } else {
+        "".to_string()
+    };
+
     metrics
         .recently_accessed_releases
         .record(krate.crate_id, krate.release_id, target);
@@ -496,6 +518,7 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
     rendering_time.step("rewrite html");
     RustdocPage {
         latest_path,
+        canonical_url,
         permalink_path,
         latest_version,
         target,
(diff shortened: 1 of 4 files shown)


src/web/rustdoc.rs
```diff
@@ -483,6 +484,24 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
 
     let latest_path = format!("/crate/{}/latest{}{}", name, target_redirect, query_string);
 
+    let canonical_url = if krate.documentation_url.is_none()
+        || krate
+            .documentation_url
+            .as_ref()
+            .unwrap()
+            .starts_with("https://docs.rs/")
```
jyn514, line 492: should we also do this for `http://docs.rs`?

jyn514 (commented):
This looks good to me, but can we add some tests to make sure it works properly?

src/web/rustdoc.rs
```diff
@@ -483,6 +484,24 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+    {
+        format!(
+            "{}://{}/{}/latest/{}",
+            req.url.scheme(),
+            req.url.host(),
+            name,
+            inner_path
+        )
+    } else {
+        "".to_string()
```
jyn514, line 502: shouldn't this be the crate's `documentation_url`?
cc @Nemo157 I know you had concerns about this not actually being a rustdoc page

src/web/rustdoc.rs
```diff
@@ -483,6 +484,24 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+        || krate
+            .documentation_url
+            .as_ref()
+            .unwrap()
+            .starts_with("https://docs.rs/")
+    {
+        format!(
+            "{}://{}/{}/latest/{}",
+            req.url.scheme(),
+            req.url.host(),
```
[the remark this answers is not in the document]
syphar: canonical URLs can be anything that a link accepts, so we also could just omit everything except the path

src/web/rustdoc.rs
syphar: since the canonical url should point to the _same page_, I don't see how this would work for `documentation_url` where we don't even know if it's generated with rustdoc, or just some mdbook.

templates/rustdoc/head.html
```diff
@@ -3,4 +3,8 @@
 
         <link rel="search" href="/-/static/opensearch.xml" type="application/opensearchdescription+xml" title="Docs.rs" />
 
+        {%- if canonical_url -%}
+        <link rel="canonical" href="{{canonical_url | safe}}" />
```
syphar, line 7: @jyn514 wouldn't we need the same logic for other versioned pages? The docs.rs crate page comes to mind

src/web/rustdoc.rs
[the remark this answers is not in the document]
jsha: According to https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls:

> Use absolute paths rather than relative paths with the rel="canonical" link tag.
> Recommended: https://www.example.com/dresses/green/greendress.html
> Not recommended: /dresses/green/greendress.html

src/web/rustdoc.rs
```diff
@@ -483,6 +484,22 @@ pub fn rustdoc_html_server_handler(req: &mut Request) -> IronResult<Response> {
...
+        || krate
+            .documentation_url
+            .as_ref()
+            .unwrap()
+            .starts_with("https://docs.rs/")
+    {
+        format!(
+            "https://docs.rs/{}/latest/{}",
+            name,
+            inner_path
```
syphar, line 497: there is no guarantee that a page that exists in one version also exists in another version. 

That's why we use the `target-redirect` handler for the links in the dropdown. 

Should we put a link to the target-redirect here into the canonical url? 

Or replicate part of the logic in `target_redirect_handler` to show something more usable here? 

( since archive storage file-exist queries are relatively fast and local, but we probably should measure)

templates/rustdoc/head.html
jsha: IMO we shouldn't bother. Almost no one searches for the crate page, so it's not that critical to add canonicalization for it. If we find down the line that people are looking for that page more often, we can always add canonicalization later.

src/web/rustdoc.rs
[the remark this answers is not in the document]
syphar: ok, thanks for the info, didn't know that!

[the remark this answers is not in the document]
jyn514: not sure I heard an answer here - @jsha why did you choose to use `req.url.host()`?

src/web/rustdoc.rs
syphar: Ah, understood, this is probably worth a comment so nobody is confused. 

at some point we could optimize this by pointing to the specific latest version where this page exists, but this would need faster file-exist checks.

syphar:
@jsha I assume the second part of https://github.com/rust-lang/docs.rs/issues/74#issuecomment-1184900299 will be in a second PR? So `no-index` for pages which provide a documentation url?

jsha:
> I assume the second part of https://github.com/rust-lang/docs.rs/issues/74#issuecomment-1184900299 will be in a second PR? So no-index for pages which provide a documentation url?

Yes, that's correct. That part is a little trickier - I'll update on #74.

jsha:
Pushed a new revision with a test. Writing a test helped me notice we were including `index.html` unnecessarily in the canonical URLs; I removed that.

Added a comment describing why it's fine that the target might in some cases not exist.

jyn514 (approved):
Lgtm other than the nit. Thanks!

jsha:
Sure! I named it after the corresponding field in Metadata. Should I change
that too?

src/web/rustdoc.rs
```diff
@@ -1980,4 +2003,64 @@ mod test {
...
         })
     }
+
+    #[test]
+    fn canonical_url() {
+        wrapper(|env| {
+            env.fake_release()
+                .name("dummy-dash")
+                .version("0.1.0")
+                .documentation(Some("http://example.com".to_string()))
```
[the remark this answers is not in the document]
jsha: Done! I wound up not renaming the field in Metadata since it's meant to match what's in Cargo.toml.

PR #1774, opened 2022-07-21
fix(mobile): sidebar & footer layout issue

fundon:
## Before

Screen Shot 2022-07-21 at 8 05 35 AM

## After

Screen Shot 2022-07-21 at 8 13 41 AM

### DIFF
--- a/templates/style/rustdoc-2021-12-05.scss
+++ b/templates/style/rustdoc-2021-12-05.scss
@@ -33,8 +33,14 @@ div.container-rustdoc {
     }
 }
 
-div.rustdoc {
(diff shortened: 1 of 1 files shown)


jyn514 (commented):
Thanks, this has been bugging me!

@jsha @GuillaumeGomez does this look good to you?

jsha (approved):
This looks reasonable as a quick fix, particularly since it helps with built docs, but it suggests that https://github.com/rust-lang/rust/pull/98671 might be buggy; we should take a second look at that. But that doesn't get in the way of merging this.

PR #1782, opened 2022-07-26
Smarter sub-menu positioning on narrow displays

euphemism:
Hi there,

I pulled up [docs.rs](https://docs.rs/) on my phone the other day and noticed that the navbar's sub-menus behave pretty poorly.

The left edge of the sub-menus is always in line with the menu link's left edge. On narrower displays this results in the menus extending off screen, causing the page to stretch and the user to have to scroll to the right to see the entirety of the sub-menu:

https://user-images.githubusercontent.com/1179754/181129059-6c0b919e-3cf0-4b82-8860-39d0526b87b7.mp4

I have written some code to adjust the sub-menus' positions based on the current width of the page. If the sub-menu would overflow off the right side of the screen, it is instead shifted left so that it is fully visible and the page does not expand and require horizontal scrolling. If for some reason the shifting to the left would cause the sub-menu to clip with the left edge of the page it is then shifted back to the right so that the left edge is at `x = 0`.

https://user-images.githubusercontent.com/1179754/181129455-b576569f-46e1-4e2a-95b5-ebd6bb237253.mp4

Since this change necessitates the sub-menus not always being vertically aligned with the left edge of the menu links, I also tweaked the CSS for the menu links to show the drop down arrow flipped 180º when the menu's sub-menu is open. Hopefully this helps to visually link the sub-menu to its parent menu link in these cases.

I've had some difficulties getting this up and running on my M1 Mac, which I suspect is the cause of the visual differences between the two videos.

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -1,9 +1,19 @@
+const updateMenuPositionForSubMenu = (currentMenuSupplier) => {
+    const currentMenu = currentMenuSupplier();
+    const subMenu = currentMenu?.getElementsByClassName('pure-menu-children')?.[0];
+
+    subMenu?.style.setProperty('--menu-x', `${currentMenu.getBoundingClientRect().x}px`);
+}
+
 // Allow menus to be open and used by keyboard.
 (function() {
     var currentMenu;
     var backdrop = document.createElement("div");
     backdrop.style = "display:none;position:fixed;width:100%;height:100%;z-index:1";
     document.documentElement.insertBefore(backdrop, document.querySelector("body"));
+
+    addEventListener('resize', () => updateMenuPositionForSubMenu(() => currentMenu));
+
     function previous(allItems, item) {
         var i = 1;
         var l = allItems.length;
@@ -52,6 +62,9 @@
             this.blur();
         } else {
             if (currentMenu) closeMenu();
+
+            updateMenuPositionForSubMenu(() => this.parentNode);
+
             openMenu(this.parentNode);
         }
         e.preventDefault();
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -50,6 +50,12 @@ div.nav-container {
         }
     }
 
+    .pure-menu-has-children.pure-menu-active > .pure-menu-link {
+        &:after {
+            content:"\25B2";
+        }
+    }
+
     .pure-menu-link {
         font-size: 12.8px;
         font-weight: 400;
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -139,10 +139,19 @@ div.nav-container {
     }
 
     .pure-menu-children {
+        --menu-x: 0;
+
+        // #{} interpolates a SassScript expression. Using this prevents
+        // SASS from using its built-in min/max over the CSS min/max functions
+        --clamped-offset: m#{i}n(var(--menu-x), calc(100vw - 100%));
+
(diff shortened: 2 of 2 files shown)


static/menu.js
```diff
@@ -1,9 +1,66 @@
...
+    }
+
+    computeSubMenuOffset(menu) {
+        const { x: menuLeft } = menu.getBoundingClientRect();
+        const subMenu = this.getSubMenu(menu);
+
+        // Temporarily include the sub-menu in page layout in order to get the width
+        if ("block" !== window.getComputedStyle(subMenu).display) {
+            subMenu.style.display = "block";
+            subMenu.style.visibility = "hidden";
```
euphemism, line 17: Maybe the order here should be flipped to guarantee the sub-menu isn't visible before setting its display to block? Unsure if there's an actual possibility of a flicker here or not.

GuillaumeGomez:
The result looks great indeed! Big question though: wouldn't it be possible to handle this only with CSS? It's quite a lot of JS after all for such a small improvement (still great though!).

euphemism:
> The result looks great indeed! Big question though: wouldn't it be possible to handle this only with CSS? It's quite a lot of JS after all for such a small improvement (still great though!).

Hey, thanks for taking a look. It doesn’t _seem_ like all that much JS for what it does? It’s just getting some bounding information for each sub-menu and computing an offset. This only happens once for each sub-menu unless the page is resized.

I don’t know how to do this in pure CSS. There are some required pieces like the `vw` unit and `clamp` function to facilitate this, but I don’t know how to get the element bounds.

GuillaumeGomez:
We're talking about more than 50 lines of JS for a resize. I'll try to get the same result with CSS only whenever I have time. If not possible, then let's merge this.

euphemism:
Okay, I think this is the best we are going to be able to get. We still need some JS code to get the actual x-coordinate for the menu in order to compute the sub-menu's position, but now the positioning logic has been pushed into CSS.

GuillaumeGomez:
I was able to find a much simpler solution: we can switch `left: 0;` to `right: 0;` for the menus on the right. Like this, their full content will always be visible. It has the advantage of not requiring JS and involves very small CSS changes. Want to make the update?

euphemism:
Hi @GuillaumeGomez , thanks for circling back. While the suggestion for the most part works, it still doesn't cover the case of the very large dropdown on the left of the menu bar. On my phone this submenu still extends past the viewport, causing the page to expand and introduce horizontal scrolling:


https://user-images.githubusercontent.com/1179754/183978520-a93a9ff7-7dc0-4f69-9f1a-d378c481b816.mp4


The code change as it currently stands covers this case.

GuillaumeGomez:
Good catch! I'll test your solution locally for this case to see how it looks like.

GuillaumeGomez:
It doesn't fix the problem completely (because the width is set, so if the viewport width is less than 350px, then we can't do anything about it). Well in any case, your implementation is working nicely so let's go with it. Thanks!

euphemism:
Yup @ fixed `350px`. That can be somewhat mitigated by replacing `350px` with `min(350px, 100vw)`, which would force the menu to shrink to the viewport width if a very narrow display is encountered.

GuillaumeGomez:
The result is quite bad when the width gets smaller so I think the best thing to do is keeping it as is or removing some of its content. But that seems like a completely new debate. :)

PR #1800, opened 2022-08-23
Pass `rustc-args` to build dependencies too

### LINKED ISSUE #1580: Pass cfg to build script
https://github.com/rust-lang/docs.rs/pull/1543 broke proc-macro2's documentation build &mdash; see https://github.com/dtolnay/proc-macro2/issues/309.

As far as I can tell the difference is that `cargo rustdoc -- --cfg mycfg` sets mycfg during **both** the compilation of the build.rs and the invocation of rustdoc, whereas `cargo rustdoc --config 'build.rustdocflags=["--cfg", "mycfg"]'` sets it only during rustdoc and **not** the compilation of build.rs.

As far as I can tell there is now no way to set cfg for a build script build on docs.rs.

@willcrichton @jyn514

Nemo157:
This was the previous behaviour before https://github.com/rust-lang/docs.rs/pull/1559 for the default target and is most likely expected by users.

fixes #1580

### DIFF
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -281,6 +281,10 @@ impl Metadata {
             let rustflags =
                 toml::to_string(&self.rustc_args).expect("serializing a string should never fail");
             cargo_args.push(format!("build.rustflags={}", rustflags));
+            cargo_args.push("-Zhost-config".into());
+            cargo_args.push("-Ztarget-applies-to-host".into());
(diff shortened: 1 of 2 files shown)


Nemo157:
(I'm continuously disappointed that github only displays the committer-date, not the 8 month old author-date).

jyn514 (approved):
amazing, thanks!

PR #1802, opened 2022-08-24
Support querying crate contents with cratename::some::path

### LINKED ISSUE #1531: Consider splitting crate name on spaces and `::`
Right now, `!drs tokio::spawn` takes you to https://docs.rs/tokio%3A%3Aspawn, which gives a 404. Ideally, duckduckgo would fix this on their end, but I looked at https://duckduckgo.com/newbang and they don't have any way to split the search string, only paste it literally. So it might be worth doing on our end instead and rewriting `/tokio::spawn` to https://docs.rs/tokio/?search=spawn.

joshtriplett:
This also makes searches on DuckDuckGo such as `!docsrs cratename::some::path` work.

Fixes #1531

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -554,23 +554,35 @@ pub fn search_handler(req: &mut Request) -> IronResult<Response> {
     let mut conn = extension!(req, Pool).get()?;
 
     // check if I am feeling lucky button pressed and redirect user to crate page
-    // if there is a match
-    if params.contains_key("i-am-feeling-lucky") {
+    // if there is a match. Also check for paths to items within crates.
+    if params.contains_key("i-am-feeling-lucky") || query.contains("::") {
         // redirect to a random crate if query is empty
         if query.is_empty() {
             return redirect_to_random_crate(req, &mut conn);
         }
+
+        let (krate, query) = match query.split_once("::") {
+            Some((krate, query)) => (krate.to_string(), format!("?query={query}")),
+            None => (query.clone(), "".to_string()),
+        };
+
         // since we never pass a version into `match_version` here, we'll never get
         // `MatchVersion::Exact`, so the distinction between `Exact` and `Semver` doesn't
         // matter
-        if let Ok(matchver) = match_version(&mut conn, &query, None) {
+        if let Ok(matchver) = match_version(&mut conn, &krate, None) {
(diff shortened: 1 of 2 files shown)


joshtriplett:
https://github.com/rust-lang/docs.rs/issues/1531

jyn514 (changes requested):
This seems reasonable :) can you please add a test for both `tokio::spawn` and `tokio::runtime::Runtime`? Just that the redirect is correct, we don't need to assert anything about the search results. There are tests at the bottom of the file you can use as examples; you'll probably want the `assert_redirect` helper function too.

Nemo157:
It'd also be good to have a test for something like `https://docs.rs/std::Future` since the sysroot crates have a custom redirect.

jyn514:
@Nemo157 from my reading of the code, that isn't supported yet - @joshtriplett if you want to add it that's fine, but I won't block on it.

joshtriplett:
Tests added. It looks like all the existing tests use mocked sample crates, so I followed that example.

jyn514 (changes requested):
Ok, I tested and this doesn't actually do what we want. It's modifying the *search* handler on docs.rs itself. But our search handler already (mostly) supports this, it doesn't give a 404: 
image

what we actually want to modify is the top-level crate handler, which is what duckduckgo redirects to: https://docs.rs/quiche%3A%3Aconnection
image

joshtriplett:
@jyn514 So, as far as I can tell, `!docsrs PLACEHOLDER` seems to be redirecting to `https://docs.rs/releases/search?query=PLACEHOLDER`, which is what I thought I modified. That was what I was trying to fix.

Nemo157:
`!drs` and `!docsrs` are different (and personally I just type `docs.rs/foobar` directly since it's ~the same length, so I'd like it supported there).

joshtriplett:
@Nemo157 Ah, interesting. That's an odd inconsistency in DDG.

joshtriplett:
Added, plus tests; let's see if this works.

jyn514 (approved):

joshtriplett:
Currently fixing a bug in this; new PR shortly.

joshtriplett:
https://github.com/rust-lang/docs.rs/pull/1806

PR #1805, opened 2022-08-27
Fix missing vendored stylesheet insertion

GuillaumeGomez:


Nemo157:
Fallout from https://github.com/rust-lang/rust/pull/101023, the old selector no longer matched anything so the vendored css didn't get inserted anywhere.

Nemo157:
Tested locally and it works on both old and new docs 👍

PR #1809, opened 2022-08-30
add clippy book to Rust dropdown menu

llogiq:
It came to my attention that docs.rs has nothing about clippy, despite it being an official Rust component for a while. This PR puts the nightly clippy book into the Rust dropdown menu so people will have an easier time finding it.

### DIFF
--- a/templates/header/topbar_end.html
+++ b/templates/header/topbar_end.html
@@ -65,6 +65,12 @@
                                 text="The Cargo Guide",
                                 target="_blank"
                             ) }}
+
+                            {{ macros::menu_link(
+                                href="https://doc.rust-lang.org/nightly/clippy",
+                                text="Clippy Documentation",
+                                target="_blank"
+                            ) }}
                         </ul>
                     </li>
                 </ul>


GuillaumeGomez:
I'm a bit shared on this: the menu is already quite big and list learning resources. Clippy seems to be a step too far: when you use clippy, you already have knowledge about Rust. In short: it's advanced usage rather than "learning basics". Well, we'll see what the others think about it.

llogiq:
> when you use clippy, you already have knowledge about Rust

I sure hope that's not the case. Many clippy lints explicitly benefit newcomers, sometimes at the expense of `#[allow(..)]` annotations by those who know what they're doing. Ah well, your reply proves my point is that clippy is still under-advertised. 😁

llogiq:
Another point here: We have started the "clippy book", which has more detailed instructions on how to use clippy (as opposed to just describing the lints). It's a work in progress for now, but once the dust has settled, it could be a better thing to link from docs.rs.

jyn514:
I think this list should be, in order of importance:
- only official resources (rust cookbook is iffy, I would prefer to remove it)
- only documentation; docs.rs cannot be all things to all people and the point of the site is docs. That suggests we should remove the top-level link to crates.io, which seems good to me, we already link to individual crates on the /crate page. It also suggests we should remove the link to rust-lang.org, which I'm more ambivalent about - RLO is a good jumping off point for links that we don't have on docs.rs itself.
- only something you'd use while writing code. This is the one that would rule out clippy lints; clippy already links to the lint reference when it emits a lint, I'm not sure how useful it is to link it here.

That said, having links on docs.rs definitely adds to discovery. If we can instead link to a "clippy guide", or a more general page that recommends official tools, that seems good to me. But having just the lint reference seems kind of strange imo.

jyn514:
also, I implied above but didn't say it explicitly - if we reject the lint reference, then we're holding it to a higher standard than some of the existing links. We should clean up the list if we reject this PR, it shouldn't be frozen as is.

llogiq:
The nightly clippy book is at https://doc.rust-lang.org/nightly/clippy, so I'm going to update this PR to include the book instead of the lint list.

GuillaumeGomez:
I think having a nightly link is already good enough. Once the stable link is available, we'll just update it.

llogiq:
Done.

r? @jyn514 @GuillaumeGomez

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! I'll send a PR later on to remove the other links too.

jyn514:
@llogiq I think you forgot to push your changes.

llogiq:
Strange. I had pushed the update, but for some reason github failed to update the PR. Well, it seems to be up to date now.

llogiq:
Ping? How do we go from here?

jyn514:
I haven't had time to look at this, sorry. I'll try and make time this weekend.

jyn514 (approved):
This seems reasonable; I think we should still go back and remove some of the old links, but that doesn't need to block your PR.

GuillaumeGomez:
@jyn514 Done in https://github.com/rust-lang/docs.rs/pull/1820.

PR #1820, opened 2022-09-04
Remove links that shouldn't be in the top navbar

GuillaumeGomez:
As discussed in https://github.com/rust-lang/docs.rs/pull/1809#issuecomment-1232031749, some links shouldn't be in the docs.rs top navbar. The parameters are as follows:

 * Only official resources (it excludes the rust cookbook).
 * Only documentation (it excludes rust-lang.org and crates.io, and for the latter, we already link to it).

This is very much up-to-debate.

### DIFF
--- a/templates/header/topbar_end.html
+++ b/templates/header/topbar_end.html
@@ -48,19 +48,6 @@
                             ) }}
 
                             {{ macros::menu_link(
-                                href="https://rust-lang-nursery.github.io/rust-cookbook/",
-                                text="Rust Cookbook",
-                                target="_blank",
-                                extra_classes="menu-item-divided"
-                            ) }}
-
-                            {{ macros::menu_link(
-                                href="https://crates.io",
-                                text="Crates.io",
-                                target="_blank"
-                            ) }}
-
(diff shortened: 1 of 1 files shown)


templates/header/topbar_end.html
```diff
@@ -24,11 +24,6 @@
                     #}<li class="pure-menu-item pure-menu-has-children">
                         <a href="#" class="pure-menu-link" aria-label="Rust">Rust</a>
                         <ul class="pure-menu-children">
-                            {{ macros::menu_link(
-                                href="https://www.rust-lang.org/",
-                                text="Rust website",
```
jyn514, line 29: I think it makes sense to make an exception here, since it's the official site for the language.

jyn514 (approved):
Thanks! r=me once you add back rust-lang.org.

templates/header/topbar_end.html
GuillaumeGomez: :+1:

GuillaumeGomez:
Merging then. Thanks for the review!

PR #1831, opened 2022-09-13
Remove `type="text/css"` from stylesheet links

notriddle:
MDN directly recommends this in <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link>, since "CSS is the only stylesheet language used on the web."

Like <https://github.com/rust-lang/rust/pull/101023>, but for docs.rs.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1840, opened 2022-09-16
Set a minimum for last-modified in sitemap

jsha:
This should encourage Google to recrawl pages that now use the /latest/ URLs and have canonicalization links.

### DIFF
--- a/src/web/sitemap.rs
+++ b/src/web/sitemap.rs
@@ -6,7 +6,7 @@ use crate::{
     web::error::Nope,
     web::page::WebPage,
 };
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeZone, Utc};
 use iron::{
     headers::ContentType,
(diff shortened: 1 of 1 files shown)


src/web/sitemap.rs
```diff
@@ -86,6 +86,12 @@ pub fn sitemap_handler(req: &mut Request) -> IronResult<Response> {
             target_name: row.get("target_name"),
             last_modified: row
                 .get::<_, DateTime<Utc>>("release_time")
+                // On Aug 27 2022 we added `<link rel="canonical">` to all pages,
+                // so they should all get recrawled if they haven't been since then.
+                .max(DateTime::from_utc(
```
Nemo157, line 91: If you `use chrono::TimeZone` this can be
```rust
                .max(Utc.ymd(2022, 8, 28).and_hms(0, 0, 0))
```

jsha: Good tip, thanks! Done.

PR #1841, opened 2022-09-17
Update Discord invite link on about page

cailloumajor:
The invite link on https://docs.rs/about is outdated, this PR updates it to be the same as in README.md (which works, just tested).


syphar (approved):
awesome, thanks!

PR #1847, opened 2022-09-17
Update to font-awesome 0.6.2

notriddle:


syphar (changes requested):
I only found one issue, apart from that I'm fine with the change. 

Thank you! 

I also checked the new icons and all these seemed to be good new replacements. 

@Nemo157 do you have anything to add? Am I missing something?

Nemo157:
Looks good to me

syphar (approved):

PR #1857, opened 2022-09-25
add `#[track_caller]` to test wrapper

Emilgardis:
this makes it easier to identify what test failed, pointing to the caller of wrapper instead of the wrapper fn

```
panicked at 'the test failed', src/web/rustdoc.rs:line:col
```


jyn514 (approved):
Thanks!

PR #1863, opened 2022-09-29
Update comment about buggy static file serving

jsha:
In https://github.com/rust-lang/rust/pull/101702#issuecomment-1243052385 @Nemo157 pointed out that static file handling is hacky, and even pointed out to me the relevant lines of code. But after reading the comment and the issues I still didn't understand exactly what was happening, what the desired direction was, and whether there were multiple intersecting bugs.

After staring at the code for a little while I wrote up a summary to see if I properly understand, and hopefully make it clearer for the next person who comes along. Please correct what I have gotten wrong. :-)

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -163,6 +163,8 @@ impl Handler for MainHandler {
             }
         }
 
+        // This is kind of a mess.
+        //
         // Some versions of rustdoc in 2018 did not respect the shared static root
         // (--static-root-path) and so emitted HTML that linked to shared static files via a local
         // path. For instance, `<script src="../main-20181217-1.33.0-nightly-adbfec229.js">` instead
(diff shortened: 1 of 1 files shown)


Nemo157 (approved):
Looks correct to my understanding.

src/web/mod.rs
```diff
@@ -163,19 +163,42 @@ impl Handler for MainHandler {
             }
         }
 
-        // This is kind of a mess.
+        // Some versions of rustdoc in 2018 did not respect the shared static root
```
jyn514, line 166:
```suggestion
        // This is kind of a mess.
        //
        // Some versions of rustdoc in 2018 did not respect the shared static root
```

jyn514 (approved):

syphar (approved):

PR #1890, opened 2022-10-22
Add environment check to test_build_crate

jsha:
If $DOCSRS_INCLUDE_DEFAULT_TARGETS == false, this test fails when checking the list of built targets; better to provide a more useful error message.

### DIFF
--- a/src/docbuilder/rustwide_builder.rs
+++ b/src/docbuilder/rustwide_builder.rs
@@ -867,6 +867,11 @@ mod tests {
                 .map(|v| v.as_str().unwrap().to_owned())
                 .collect();
             targets.sort();
+            assert_ne!(
+                std::env::var("DOCSRS_INCLUDE_DEFAULT_TARGETS").unwrap(),
+                "false",
+                "This test requires setting $DOCSRS_INCLUDE_DEFAULT_TARGETS to true"
+            );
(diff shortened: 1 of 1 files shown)


src/docbuilder/rustwide_builder.rs
```diff
@@ -867,6 +867,11 @@ mod tests {
                 .map(|v| v.as_str().unwrap().to_owned())
                 .collect();
             targets.sort();
+            assert_ne!(
+                std::env::var("DOCSRS_INCLUDE_DEFAULT_TARGETS").unwrap(),
```
jyn514, line 871: Can we set this in the test suite rather than requiring the user to set it?

jsha: I wound up fixing this in a drive-by fix over here: https://github.com/rust-lang/docs.rs/pull/1892/files#diff-2d6b4b8b3ed45b3dfd5b9f67389371c1f06362ddc510e3add1934c0d7d614f94R282

jsha:
Closing as fixed a different way: https://github.com/rust-lang/docs.rs/pull/1892/files#diff-2d6b4b8b3ed45b3dfd5b9f67389371c1f06362ddc510e3add1934c0d7d614f94R282

PR #1926, opened 2022-11-25
Use header for canonical URLs on rustdoc pages

jsha:
Since the header on rustdoc pages is partly user-controlled, the user can inadvertantly break parsing such that Google doesn't see the `<link>` tag. Putting it in the header ensures it is reliably parsed.


syphar (approved):
LGTM

PR #1938, opened 2022-11-29
Improve instructions a bit for bash

GuillaumeGomez:
Took me a while to understand why `source` wasn't working. It's because it doesn't export variables defined in a script into the environment by default. To do so, we need to use `set -o allexport` or `set -a`. Since I think some others might have the same issue, I updated the README a bit to include this command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,9 @@ cargo build
 docker-compose up -d db s3
 # anything that doesn't run via docker-compose needs the settings defined in
 # .env. Either via `. ./.env` as below, or via any dotenv shell integration.
+# If using bash, you need to run instead:
+# `set -o allexport && . ./env && set +o allexport` otherwise they won't be
+# saved in your environment.
 . ./.env
 # Setup the database you just created
 cargo run -- database migrate


Nemo157:
This shouldn't be necessary, testing in `bash`:

```console
[nemo157@shao docs.rs]$ export | grep DOCSRS
[nemo157@shao docs.rs]$ . .env.sample
[nemo157@shao docs.rs]$ export | grep DOCSRS
declare -x DOCSRS_DATABASE_URL="postgresql://cratesfyi:password@localhost:15432"
declare -x DOCSRS_DOCKER_IMAGE="ghcr.io/rust-lang/crates-build-env/linux-micro"
declare -x DOCSRS_INCLUDE_DEFAULT_TARGETS="false"
declare -x DOCSRS_LOG="docs_rs=debug,rustwide=info"
declare -x DOCSRS_PREFIX="ignored/cratesfyi-prefix"
declare -x DOCSRS_TOOLCHAIN="nightly"
```

You may have an old `.env` file from before the sample had `export` added to make it actually export the variables into the environment.

GuillaumeGomez:
I didn't even pay attention to that. Quite convenient! Closing then as unneeded.

PR #1996, opened 2023-01-12
Fix name of docker image being uploaded

rylev:
This was a copy pasta mistake


rylev:
Going to merge this since this is just fixing a mistake in #1995

PR #1997, opened 2023-01-12
Defer script load for improved performance ✈️

sanjaiyan-dev:
Defering script loading for performance boost 🚀

Ref-: https://web.dev/efficiently-load-third-party-javascript/#defer

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -19,6 +19,9 @@
 
         <script nonce="{{ csp_nonce }}">{%- include "theme.js" -%}</script>
         {%- block css -%}{%- endblock css -%}
+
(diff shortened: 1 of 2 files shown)


syphar (changes requested):
Hi, 

sorry for the delay, thank you for the contribution! 

I'm definitely for it, 

could you only update the commit message so it's clearly visible in the header what the change is? 

moving the body up should be good enough

sanjaiyan-dev:
> Hi,
> 
> sorry for the delay, thank you for the contribution!
> 
> I'm definitely for it,
> 
> could you only update the commit message so it's clearly visible in the header what the change is?
> 
> moving the body up should be good enough

Done 🤝🏽

syphar (approved):

Issue #1084, opened 2020-10-08
docs.rs pages are left-aligned


jyn514:
This is a regression introduced by https://github.com/rust-lang/docs.rs/pull/1079 - only the /crate pages should be left-aligned, not /about and index. cc @jsha , do you think you have time to follow-up and fix this? I think a simple fix would be to only apply left-aligns to /crate pages, not sure if we have custom CSS for them yet but it shouldn't be too hard to add.

jsha:
Oops, sorry about that! I'll send a fix.

PR #1079, opened 2020-10-08
Make crates page closer to rustdoc pages.

jsha:
- Align the whole page to the left, rather than center.
- Change the "Docs.rs" link on the rustdoc page to use three cubes,
  like other pages do.

Followup from #1077

PR #2012, opened 2023-01-22
Left-align crate and release pages

jsha:
To make them more similar to doc pages. Retain centered alignment for / and /about.

Since the crate page is navigated as part of a crate's documentation, it's jarring for it to have different alignment than the documentation itself.

We first [made this change in 2020](https://github.com/rust-lang/docs.rs/pull/1079) but [reverted it due to a regression](https://github.com/rust-lang/docs.rs/issues/1084): the about and home pages weren't supposed to change.

Now, homepage stays the same:

image

About stays the same:

image

Crate page is left aligned:

image

The doc page stays left aligned:
image

### DIFF
--- a/templates/about-base.html
+++ b/templates/about-base.html
@@ -1,5 +1,9 @@
 {% extends "base.html" %}
 
+{%- block body_classes -%}
+centered
+{%- endblock body_classes -%}
+
 {% block header %}
     <div class="docsrs-package-container">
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Big :+1: for coherency!

PR #2031, opened 2023-02-04
Documentation: fix latest version redirection example

CreatorSiSo:
The current example links to https://docs.rs/clap/*/clap/struct.App.html which does not exist.
It is also not really in line with all the other examples, which do not directly link to any crate items.

image

This PR just changes the link to [https://docs.rs/clap/*](https://docs.rs/clap/*) but https://docs.rs/clap/*/clap/struct.Arg.html would be a struct that currently exists and could be used insted.

### DIFF
--- a/templates/core/about/redirections.html
+++ b/templates/core/about/redirections.html
@@ -52,7 +52,7 @@
 
                 <tr>
                     <td>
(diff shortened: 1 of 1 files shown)


jyn514 (commented):
Thanks!

CreatorSiSo:
You're welcome

PR #2043, opened 2023-02-12
Update README.md

5h4nx:
- Explicitly mentioning dev dependency
- Alternate use of docker compose command


syphar (approved):
thank you for the improvement !

PR #2060, opened 2023-02-26
Add missing centered CSS class

GuillaumeGomez:
Fixes cases like this (content is completely on the left, leaving big "empty space" on the right):

Screenshot from 2023-02-26 21-37-52

As a reminder, https://github.com/rust-lang/docs.rs/pull/1698 is still waiting for feedback to prevent regressions like this to happen. ;)

### DIFF
--- a/templates/crate/build_details.html
+++ b/templates/crate/build_details.html
@@ -5,6 +5,10 @@
     {{ macros::doc_title(name=metadata.name, version=metadata.version) }}
 {%- endblock title -%}
 
+{%- block body_classes -%}
+centered
+{%- endblock body_classes -%}
+
 {%- block topbar -%}
   {%- set latest_version = "" -%}
   {%- set latest_path = "" -%}
--- a/templates/crate/builds.html
+++ b/templates/crate/builds.html
@@ -9,6 +9,10 @@
 <link rel="canonical" href="{{ canonical_url | safe }}" />
 {%- endblock -%}
 
+{%- block body_classes -%}
+centered
+{%- endblock body_classes -%}
+
 {%- block topbar -%}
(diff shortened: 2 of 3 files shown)


Nemo157:
These pages being left-aligned was an explicit change https://github.com/rust-lang/docs.rs/pull/2012

GuillaumeGomez:
Even the search results and the non "main" crate pages?

Nemo157:
The docs.rs crate pages, yes that was definitely called out in the PR, and the search page is one of the "release pages".

Nemo157:
(Pages that looks like they could do with a bit of style work to make left-alignment look better are<https://docs.rs/crate/serde/1.0.152/builds> and <https://docs.rs/crate/serde/1.0.152/builds/709459>, those are the only ones without a left-sidebar).

GuillaumeGomez:
Okay, gonna update the PR then.

GuillaumeGomez:
Put back the left align like mentioned.

Nemo157:
cc @jsha since you have more context on what the overall plan of left aligning was.

jsha (approved):
This looks good to me. The idea with left-aligning is that the rustdoc pages themselves are left-aligned, and the crate pages are related to them, so it's jarring to go back and forth between left-aligned and right-aligned.

For the homepage and various pages that show search results, it's reasonable to be centered, IMO.

In this PR, builds and build_details should ideally be left-aligned, but given that their layout is currently messed up when left-aligned I'm fine landing this as a quick fix. We should later follow up and left-align them with appropriate padding.

GuillaumeGomez:
Ok, merging then. Thanks for the feedback!

PR #2094, opened 2023-04-06
Update the favicon with a sharper, tuned version.

lgarron:
The new version keeps the style of the previous one uses the R from the Rust logo. It has been made slightly taller to be more clear at small resolution and given a clearer border for stronger constrast on varying backgrounds.

It contains 16x16 and 32x32 versions, rendered to show well on Retina as well non-Retina screens.


lgarron:
Here are screenshots in both light and dark mode on macOS, for both Retina and non-Retina resolutions.

screenshots

They also include a version that keeps the R and its border a bit closer to the original, in case you would prefer that.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! I'll let someone with a mac confirm that it looks better before merging.

lgarron:
> Looks good to me, thanks! I'll let someone with a mac confirm that it looks better before merging.

Thanks!

Just to be clarify, the icon should look snazzier on a variety of devices, I just happen to have macOS for testing.

syphar (approved):

syphar:
LGTM, thanks!

PR #2113, opened 2023-04-16
Load release list only when needed

GuillaumeGomez:
It was something we discussed some time ago. Instead of loading the release list all the time, with this PR it's only loaded when the menu is opened. It should greatly reduce the page size on crates with a lot of releases.

Original issue: https://github.com/rust-lang/docs.rs/issues/1999

cc @Nemo157 @jsha

### DIFF
--- a/static/menu.js
+++ b/static/menu.js
@@ -5,6 +5,34 @@ const updateMenuPositionForSubMenu = (currentMenuSupplier) => {
     subMenu?.style.setProperty('--menu-x', `${currentMenu.getBoundingClientRect().x}px`);
 }
 
+function generateReleaseList(data, crateName) {
+}
+
+let loadReleases = function() {
+    const releaseListElem = document.getElementById('releases-list');
+    // To prevent reloading the list unnecessarily.
+    loadReleases = function() {};
+    if (!releaseListElem) {
+        // We're not in a documentation page, so no need to do anything.
+        return;
+    }
+    const crateName = window.location.pathname.split('/')[1];
+    const xhttp = new XMLHttpRequest();
+    xhttp.onreadystatechange = function() {
+        if (xhttp.readyState !== XMLHttpRequest.DONE) {
+          return;
+        }
+        if (xhttp.status === 200) {
+            releaseListElem.innerHTML = xhttp.responseText;
+        } else {
+            console.error(`Failed to load release list: [${xhttp.status}] ${xhttp.responseText}`);
+            document.getElementById('releases-list').innerHTML = "Failed to load release list";
+        }
+    };
+    xhttp.open("GET", `/${crateName}/releases`, true);
+    xhttp.send();
+};
+
 // Allow menus to be open and used by keyboard.
 (function() {
     var currentMenu;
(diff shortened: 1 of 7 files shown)


GuillaumeGomez:
I finished porting the code so now JS generates the same thing as jinja used to. The problem now is that some tests relied on the DOM content to check some things, which isn't possible anymore. We could however have the failing tests ported to some "front-end testing"like what https://github.com/rust-lang/docs.rs/pull/1698 proposes.

static/menu.js
```diff
@@ -53,6 +135,7 @@ const updateMenuPositionForSubMenu = (currentMenuSupplier) => {
         currentMenu = newMenu;
         newMenu.className += " pure-menu-active";
         backdrop.style.display = "block";
+        loadReleases();
```
Nemo157, line 138: This looks like it will load the releases when _any_ menu is opened, even on non-crate pages.

GuillaumeGomez:
That would indeed make it much simpler to maintain and we could directly re-use the existing template. :+1:

static/menu.js
GuillaumeGomez: I modified `loadReleases` to look if there is the element where to put the release list so that it won't make the AJAX query if it's not necessary.

GuillaumeGomez:
I applied suggestions: this path now returns HTML which is directly inserted into the DOM, allowing to greatly simplify the code and also minimize the impact on tests.

GuillaumeGomez:
@Nemo157: Anything else that needs to be done here?

templates/rustdoc/releases.html
```diff
@@ -0,0 +1,6 @@
+{% import "macros.html" as macros %}
+{% set target = "" %}
+{% set inner_path = releases[0].target_name ~ "/index.html" %}
+<ul>
```
Nemo157, line 4:
```suggestion
<ul class="pure-menu-list">
```

Without the class the list is offset inside the container

image

templates/rustdoc/topbar.html
```diff
@@ -134,11 +134,8 @@
                             <li class="pure-menu-heading">Versions</li>
 
                             <li class="pure-menu-item">
-                                <div class="pure-menu pure-menu-scrollable sub-menu" tabindex="-1">
-                                    <ul class="pure-menu-list">
-                                        {# Display all releases of this crate #}
-                                        {{ macros::releases_list(name=krate.name, releases=krate.releases, target=target, inner_path=inner_path) }}
-                                    </ul>
+                                <div class="pure-menu pure-menu-scrollable sub-menu" id="releases-list" tabindex="-1">
+                                    <span class="rotate">↻</span>
```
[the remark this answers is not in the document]
GuillaumeGomez: I'll use this one then: https://fontawesome.com/v6/icons/spinner?f=classic&s=solid

GuillaumeGomez:
Applied suggestions.

templates/rustdoc/topbar.html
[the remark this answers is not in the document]
Nemo157: It could have a little bit of left margin, but it spins centered now

image

Nemo157 (approved):

PR #2136, opened 2023-05-31
Configure and document how to have download endpoint working locally

Nemo157:
I was looking back through my old branches and noticed I had this unpushed from when I was testing the download changes.


jyn514 (approved):
:heart:

PR #2162, opened 2023-06-30
Add limits on what code will be highlighted

Nemo157:
The numbers are pretty arbitrary, I didn't actually time it but opening a 3MB `.rs` file felt like it was just about on the threshold of too long, and my machine is faster than production.

I did consider having the numbers configurable, but threading the config through was looking quite complex and I don't think it's that useful.

This also exposed another bug, if highlighting failed we weren't escaping any html in the source files, so fixed that now that failure will be much more common.

### DIFF
--- a/src/web/highlight.rs
+++ b/src/web/highlight.rs
@@ -73,14 +73,16 @@ pub fn with_lang(lang: Option<&str>, code: &str) -> String {
             } else {
                 log::error!("failed while highlighting code: {err:?}");
             }
-            code.to_owned()
+            tera::escape_html(code)
         }
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use super::{select_syntax, try_with_lang, LimitsExceeded, CODE_SIZE_LIMIT, LINE_SIZE_LIMIT};
+    use super::{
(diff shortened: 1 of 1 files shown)


src/web/highlight.rs
```diff
@@ -6,6 +6,13 @@ use syntect::{
     util::LinesWithEndings,
 };
 
+const CODE_SIZE_LIMIT: usize = 2 * 1024 * 1024;
+const LINE_SIZE_LIMIT: usize = 512;
```
GuillaumeGomez, line 10: 512 lines seems a bit short no? But more globally: maybe we should store highlighted files when they are too big?

Nemo157: That's 512 characters per line.

GuillaumeGomez: Ok I see. Let's maybe rename the constant then? Otherwise the name is pretty confusing. Something like `LINE_CHARACTERS_LENGTH_LIMIT`?

syphar (commented):
the escaping piece feels like something we should manually test, were you able to do this? 

( I'm fixing my local env and can do it too later today )

The missing config feels like something we should add at some point, similar to the max HTML size, 

but currently, I feel like using constants is fine, for system stability.

Nemo157:
Yep, it was manual testing that showed me the escaping was broken.

syphar (approved):

PR #2170, opened 2023-07-27
Sort platforms list

GuillaumeGomez:
As you can see:

image

it's currently complicated to find the correct one from the list. However, I'm not sure if it's the best approach to sort the list every time or if we should insert it correctly instead. But in the second case, that would mean updating all the entries in the DB. Not sure which solution you prefer?

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -550,7 +550,7 @@ impl MetaData {
     }
 
     fn parse_doc_targets(targets: Value) -> Vec<String> {
-        targets
+        let mut targets = targets
             .as_array()
             .map(|array| {
                 array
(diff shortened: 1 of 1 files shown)


fbstj:
probably not the place to discuss this, but is there scope to group them by platform tier?

syphar (commented):
Perhaps I'm wrong, but: 

When I look at our rustwide-builder, specifically [here](https://github.com/syphar/docs.rs/blob/89e5ed87ae389976f94fcf4358aff54ae116cfc3/src/docbuilder/rustwide_builder.rs#L439-L470), it looks like the default target is the first to be added to the list of targets, and then the other targets follow. 

The `successful_targets` vec is then stored in the database, and then used in the code you changed. 

This means that currently, the default target is the first in the list, and this would change with this PR. Or am I missing something?

GuillaumeGomez:
It would change it indeed. But I don't consider it a problem: it'll remain the default platform, no need to put it at the op of the list. I was also thinking about making the current platform text bold or something along the line (in another PR).

syphar (approved):

Issue #2134, opened 2023-05-30
Fluvio failed to compile on docs.rs


morenol:
### Crate name

fluvio

### Build failure link

https://docs.rs/crate/fluvio/0.19.1/builds/824799

### Additional details

Seems to be related to some nightly failures

Nemo157:
Your `Cargo.lock` specifies `value-bag 1.0.0-alpha.9`, that fails to build since it's about a year old and tries to use nightly features after doing compiler version detection (a terrible idea for this reason). If you update that it will work.

Nemo157:
Actually, scratch that, there is no `Cargo.lock` in the package. I have no idea why it chose `1.0.0-alpha.9`, when I try building locally it chooses the latest `1.4.0` like it should 🤔.

morenol:
unless that file is not uploaded to crates.io? maybe we can just retrigger the build?

morenol:
it seems that it is there: https://docs.rs/crate/fluvio/0.19.1/source/Cargo.lock

Nemo157:
That one is I believe generated by docs.rs during the build, we record the source afterwards (...something I've long been meaning to fix), when I download and inspect the package locally it doesn't have it:

```console
> cargo dl -e fluvio
                         crates.io index   updated
                                  fluvio   extracted fluvio 0.19.1 to fluvio-0.19.1

> ls fluvio-0.19.1
build.rs  Cargo.toml  Cargo.toml.orig  LICENSE-APACHE  README.md  src/  test-data/
```

Nemo157:
Yes, I've just queued a new build.

PR #2175, opened 2023-07-30
Allow index-update during lockfile regeneration

Nemo157:
Since the default index changed to the sparse-index we have had issues with regenerating the lockfile if the dependency list changed, we relied on the previous fetch having pulled down details for all crates to use, but now it's only guaranteed to have details on the crates in the original locked dependency list. If new versions of crates (like log 0.4.17 -> 0.4.19) changed what dependencies it used we wouldn't have the details on these dependencies to use the updated version in the lockfile.

Related to #2174 and probably #2134 and #2143.


syphar (approved):
In the end, this would slow down the build when regenerating the lockfile, but I also can't think of a different way of solving this, apart from switching cargo back to the git-index.

PR #2176, opened 2023-07-30
Indicate current documentation target

GuillaumeGomez:
As mentioned in https://github.com/rust-lang/docs.rs/pull/2170. It allows to know what is the platform being checked currently. It looks like this:

Screenshot from 2023-07-30 14-51-12

### DIFF
--- a/templates/rustdoc/topbar.html
+++ b/templates/rustdoc/topbar.html
@@ -222,9 +222,14 @@ extra whitespace unremovable, need to use html tags unaffacted by whitespace T_T
                     {%- set target_url = "/crate/" ~ metadata.name ~ "/" ~ metadata.version_or_latest ~ "/target-redirect/" ~ target ~ "/" ~ inner_path -%}
                     {%- set target_no_follow = "nofollow" -%}
                 {%- endif -%}
+                {%- if current_target is defined and current_target == target -%}
+                    {%- set current = " current" -%}
+                {%- else -%}
+                    {%- set current = "" -%}
+                {%- endif -%}
 
                 <li class="pure-menu-item">
-                    <a href="{{ target_url | safe }}" class="pure-menu-link" data-fragment="retain" rel="{{ target_no_follow }}">
+                    <a href="{{ target_url | safe }}" class="pure-menu-link{{ current | safe }}" data-fragment="retain" rel="{{ target_no_follow }}">
                         {{- target -}}
                     </a>
                 </li>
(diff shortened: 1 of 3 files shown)


templates/rustdoc/topbar.html
```diff
@@ -222,9 +222,14 @@
                     {%- set target_url = "/crate/" ~ metadata.name ~ "/" ~ metadata.version_or_latest ~ "/target-redirect/" ~ target ~ "/" ~ inner_path -%}
                     {%- set target_no_follow = "nofollow" -%}
                 {%- endif -%}
+                {%- if current_target is defined and current_target == target -%}
```
GuillaumeGomez, line 225: I did it this way because if it's not a "rustdoc page" (like crate docs.rs page, or features, etc), then the platform isn't actually used (in the URL) so no need to set `current_target`. But since we still render the topbar, we need to check if `current_target` is set before actually using it.

templates/style/_navbar.scss
```diff
@@ -349,6 +349,14 @@ div.nav-container {
         li a {
             overflow-x: hidden;
             text-overflow: ellipsis;
+
+            &.current {
+                font-weight: bold;
+                display: list-item;
+                list-style-type: disc;
+                list-style-position: inside;
```
GuillaumeGomez, line 357: Instead of creating a character or using an image, I preferred to rely on browser's provided CSS elements.

templates/style/_navbar.scss
```diff
@@ -349,6 +349,14 @@ div.nav-container {
...
+                padding-left: 4px;
```
GuillaumeGomez, line 358: I needed to reduce the padding a bit to ensure the current platform is still aligned with the others (because of the bullet).

GuillaumeGomez:
cc @Nemo157

templates/style/_navbar.scss
Nemo157: Given the bullet is decided by the browser, I assume this distance is actually browser dependent?

As an alternative maybe something like

```css
.current::before {
  content: "•";
  position: absolute;
  margin-left: -10px;
}
.current {
  font-weight: bold;
}
```

That way the element itself is always aligned, and any weirdness about the bullet size will just cause it to be mispositioned.

GuillaumeGomez: I was wondering which approach would be the best because `content` depends on the font but I suppose we have a better control over the font rather than the web browser so your solution sounds to be the best. Thanks!

GuillaumeGomez:
Used content "controlled by" font instead of the browser's bullet.

Nemo157 (approved):

PR #2181, opened 2023-07-31
Only put platforms list is less than 10, otherwise load with AJAX

GuillaumeGomez:
Based on the same principle as https://github.com/rust-lang/docs.rs/pull/2113.

This PR does actually two things: only load releases if the "big menu" is opened and not if any menu is open. It allows to only load if needed even further.

The second part is actually adding the async load for the targets if there are more than 5 (having libc in mind for that).

If this approach looks good to you, I'll finish the PR by adding tests.


GuillaumeGomez:
cc @Nemo157

static/menu.js
```diff
@@ -5,13 +5,17 @@ const updateMenuPositionForSubMenu = (currentMenuSupplier) => {
...
 
-function generateReleaseList(data, crateName) {
-}
+const loadedMenus = new Set();
 
-let loadReleases = function() {
-    const releaseListElem = document.getElementById('releases-list');
-    // To prevent reloading the list unnecessarily.
-    loadReleases = function() {};
+function loadReleases(menu, id, msg, path, extra) {
```
Nemo157, line 10:
```suggestion
function loadAjaxMenu(menu, id, msg, path, extra) {
```

src/web/crate_details.rs
```diff
@@ -420,6 +424,167 @@ pub(crate) async fn get_all_releases(
...
+    let original_path = percent_encoding::percent_decode(uri.path().as_bytes())
+        .decode_utf8()
+        .map_err(|_| AxumNope::BadRequest)?;
+    let mut req_path: Vec<&str> = original_path.split('/').collect();
+
+    let release_found = match_version_axum(&pool, &params.name, Some(&params.version)).await?;
+    trace!(?release_found, "found release");
+
+    // Remove the empty start, "releases", the name and the version from the path
+    req_path.drain(..4).for_each(drop);
```
Nemo157, line 464: Isn't this `params.path`?

templates/rustdoc/topbar.html
```diff
@@ -209,31 +209,11 @@
...
-                    {%- set current = "" -%}
-                {%- endif -%}
-
-                <li class="pure-menu-item">
-                    <a href="{{ target_url | safe }}" class="pure-menu-link{{ current | safe }}" data-fragment="retain" rel="{{ target_no_follow }}">
-                        {{- target -}}
-                    </a>
-                </li>
-            {%- endfor -%}
+            {%- if metadata.doc_targets|length < 6 -%}
```
Nemo157, line 212: Maybe bump it up to 10 for the default limit, that way only crates that have requested to have their limit increased will be impacted.

GuillaumeGomez: Do we have a constant which contains this value somewhere? Otherwise I'll add one.

src/web/crate_details.rs
GuillaumeGomez: It is. Most of this code comes from `rustdoc_html_server_handler` so I tried to keep the same format.

templates/rustdoc/topbar.html
GuillaumeGomez: We don't have a constant (found where it's set in `docbuilder/limits.rs`) so I'll create one.

src/web/routes.rs
```diff
@@ -181,7 +181,15 @@ pub(super) fn build_axum_routes() -> AxumRouter {
             get_internal(super::crate_details::crate_details_handler),
         )
         .route(
-            "/:name/releases",
+            "/platforms/:name/:version/:target/",
```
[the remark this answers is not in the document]
GuillaumeGomez: Good point!

src/web/crate_details.rs
GuillaumeGomez: I updated this to use `params.path`.

static/menu.js
GuillaumeGomez: Updated.

src/web/routes.rs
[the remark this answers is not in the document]
GuillaumeGomez: Using `/-/` as prefix was problematic for tests so instead I used `/menus/` which actually renders pretty well I think.

GuillaumeGomez:
I added test for this feature and updated the failing ones (needed to update URLs mostly). It's now ready for review!

src/web/routes.rs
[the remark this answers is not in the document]
Nemo157: That would similarly cause an issue for <https://docs.rs/menus> if someone ever publishes it. What about `/-/` caused issues for tests? It shouldn't be treated specially in any way.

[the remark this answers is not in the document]
GuillaumeGomez: It failed to query URLs starting with `/-/` saying they're blacklisted. Since normally it's used for static resources, I thought it was voluntary. Should I disable that?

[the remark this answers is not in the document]
Nemo157: It's for any internal paths so that we don't have more routes that overlap with rustdoc routes, the only place it's blacklisted is in the rustdoc handler, I did a quick test changing it and it seems to work fine.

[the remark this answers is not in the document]
GuillaumeGomez: Perfect, going to revert the URLs to use `/-/` and update the rustdoc handler.

src/web/crate_details.rs
```diff
@@ -466,6 +470,160 @@ pub(crate) async fn get_all_releases(
...
+    };
+
+    let res = PlatformList {
+        metadata: ShortMetadata {
+            name,
+            version_or_latest: version_or_latest.to_string(),
+            doc_targets,
+        },
+        inner_path,
+        use_direct_platform_links: true,
```
Nemo157, line 621: This shouldn't be unconditionally true, it should only be set when we're linking to the crate root (I think that's `inner_path == format!("{crate_name}/index.html")` by the looks of it, we normalize the path enough to not have to check multiple variations).

GuillaumeGomez: So anything to change here?

src/web/routes.rs
```diff
@@ -14,7 +14,7 @@ use axum_extra::routing::RouterExt;
 use std::convert::Infallible;
 use tracing::{debug, instrument};
 
-const INTERNAL_PREFIXES: &[&str] = &["-", "about", "crate", "releases", "sitemap.xml"];
+const INTERNAL_PREFIXES: &[&str] = &["about", "crate", "releases", "sitemap.xml"];
```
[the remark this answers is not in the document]
GuillaumeGomez: Indeed, might have confused it with something else. I'll revert this change.

GuillaumeGomez:
Fixed the `target-redirect` and updated the test checking that to ensure it's always set to the expected value. I also reverted the blacklist change I wrongly made.

src/web/crate_details.rs
```diff
@@ -1105,14 +1256,86 @@ mod tests {
...
+                .rustdoc_file("dummy/index.html")
+                .rustdoc_file("x86_64-pc-windows-msvc/dummy/index.html")
+                .rustdoc_file("x86_64-pc-windows-msvc/dummy/struct.A.html")
+                .default_target("x86_64-unknown-linux-gnu")
+                .add_target("x86_64-pc-windows-msvc")
+                .create()?;
+
+            let response = env
+                .frontend()
+                .get("/-/menus/platforms/dummy/0.4.0/x86_64-pc-windows-msvc")
```
Nemo157, line 1281:
```suggestion
                .get("/-/menus/platforms/dummy/0.4.0/x86_64-pc-windows-msvc/dummy/")
```

Nemo157: The path includes the crate as the first component after the target.

Nemo157: We should also probably keep the test for the non-ajax path too.

GuillaumeGomez: Good point! I'll do both.

GuillaumeGomez:
Updated the test.

GuillaumeGomez:
Anything else to be done here?

src/web/crate_details.rs
```diff
@@ -1105,14 +1256,115 @@ mod tests {
...
+                .name("dummy")
+                .version("0.4.0")
+                .rustdoc_file("dummy/index.html")
+                .rustdoc_file("x86_64-pc-windows-msvc/dummy/index.html")
+                .rustdoc_file("x86_64-pc-windows-msvc/dummy/struct.A.html")
+                .default_target("x86_64-unknown-linux-gnu")
+                .add_target("x86_64-pc-windows-msvc")
+                .create()?;
+
+            let response = env.frontend().get("/dummy/latest/dummy").send()?;
```
Nemo157, line 1283: We should still have a test for `/crate/dummy/latest` too, making sure that doesn't have redirects.

GuillaumeGomez: That allowed to find a bug so great idea! :)

src/web/crate_details.rs
```diff
@@ -1077,42 +1236,142 @@ mod tests {
...
                 .add_target("x86_64-pc-windows-msvc")
                 .create()?;
 
             let response = env.frontend().get("/crate/dummy/0.4.0").send()?;
             assert!(response.status().is_success());
+            check_links(response.text()?, false, false);
+            // Same test with AJAX endpoint.
+            let response = env
+                .frontend()
+                .get("/-/menus/platforms/crate/dummy/0.4.0")
```
GuillaumeGomez, line 1285: I added the test but it fails because:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource at http://localhost:3000/-/menus/platforms/crate/sysinfo/0.29.7 (“default-src”).
```

So I took a look and I think that the missing `nonce` attribute on the script is what's preventing it from making AJAX queries. However, `csp_nonce` doesn't seem to be generated for `/crate/:name/` and `/crate/:name/:version` pages. Is there a way to make it generate it if needed?

syphar: Currently rustdoc pages are excluded from CSP, from what I remember because there is a risk of docs breaking that bring their own JS.

Nemo157: You need to add `connect-src 'self'` to the policy generated in `src/web/csp.rs` to allow the XHR. The `nonce` is generated and applied to the script, but that doesn't affect what XHR's are allowed from the script, those hit the `default-src 'none'` policy.

Nemo157: (Though, that then hits a 404 because the generated URL needs some adjustments).

GuillaumeGomez: That was it, thanks a lot! I'll update the route as well to allow this one.

GuillaumeGomez:
Fixed the bug, improved the test a bit and updated the `csp` test.

src/web/crate_details.rs
```diff
@@ -496,7 +496,7 @@ pub(crate) async fn get_all_platforms(
     Extension(pool): Extension<Pool>,
     uri: Uri,
 ) -> AxumResult<AxumResponse> {
-    let is_crate_root = uri.path().starts_with("/crate/");
+    let is_crate_root = uri.path().contains("/crate/");
```
Nemo157, line 499: I think there could potentially be subpaths that contain a `/crate/` directory, like the rendered source if using something like `#[path = "/crate/foo.rs"]`.

GuillaumeGomez: Good point. I can put the whole `/-/menu/platforms/crate/` with `starts_with`. What do you think?

Nemo157: That sounds good

GuillaumeGomez: Done!

Nemo157:
Found a couple more bugs clicking around pages:

* `/crate/libc/0.2.147/source/` requests `/-/menus/platforms/crate/libc/0.2.147/source/` which 404's
* `/crate/libc/0.2.147/builds/1436` gets a list, but they have links like `/libc/0.2.147/arm-unknown-linux-musleabi/libc/1436` which search for `1436` in the docs
* `/libc/0.2.147/libc/struct.Elf32_Chdr.html` works correctly, but `/libc/0.2.147/aarch64-unknown-linux-gnu/libc/struct.Elf32_Chdr.html` gets links to paths like `/crate/libc/0.2.147/target-redirect/arm-unknown-linux-gnueabi/libc/libc/struct.Elf32_Chdr.html`, the doubled `/libc/libc/` internal path causes it to not find the target file and fallback to a search.

GuillaumeGomez:
Fixed all the URLs I could think of in `/crate`, which allowed me to discover cases where the URLs have a `target-redirect` whereas they shouldn't. You can see the cases in the tests I added.

Nemo157:
I think that LGTM now, can't find any more issues clicking around 🎉

It might be good to have @syphar or someone to confirm the CSP change is fine, I can't think of any issues it could cause but I'm not that familiar with it.

GuillaumeGomez:
:tada: Thanks for the review! I'll send a follow-up once merged about the `target-redirect` issue I mentioned.

src/web/crate_details.rs
```diff
@@ -466,12 +469,194 @@ pub(crate) async fn get_all_releases(
...
+
+            (version, "latest".to_string())
+        }
+
+        // Redirect when the requested version isn't correct
+        MatchSemver::Semver((v, _)) => {
+            // to prevent cloudfront caching the wrong artifacts on URLs with loose semver
+            // versions, redirect the browser to the returned version instead of loading it
+            // immediately
+            return redirect(&params.name, &v, &req_path, CachePolicy::ForeverInCdn);
```
syphar, line 545: this redirect is cached, but the handlers are served below the `/-/` prefix, which is not part of the invalidation we run. 

So either we should use `NoCaching` here, or serve everything below the `/krate/` prefix, or invalidate additional paths.

syphar (commented):
reading about `connect-src` I believe the CSP change is fine. 

Apart from that: 
Looking back at #2113 we used the `/:crate/` prefix so the cache invalidation would cover the release-list. While we don't return a caching header in `get_all_releases` (yet, which I definitely overlooked), we should keep CDN caching in mind also for the platform list fetching. 

And using the `/-/` prefix would bring the need to either add a new path to invalidations, or perhaps introduce a different prefix? ( I could imagine `/:crate/-/...`, if that doesn't overlap with rustdoc)

src/web/crate_details.rs
Nemo157: My alternative suggestion earlier might be good then, putting them with the rest of the non-rustdoc per-crate paths: `/crate/:name/:version/menus/...`

GuillaumeGomez:
Changes the menus URLs to start with `/crate/:name/:version/menus`.

GuillaumeGomez:
Anything else that remains to be done here?

syphar:
> Anything else that remains to be done here?

sorry for the delay, from my side I see that the routes were changed. 
We could now also start actually caching the response, but I could imagine adding this separately later, especially since most platform lists will be directly rendered into the page. 

I think a last manual test after the route change would be good, if @Nemo157 doesn't have time I could dig into this.

PR #2185, opened 2023-08-07
Change timeout to apply to the overall build process, not per-build

### LINKED ISSUE #1910: Don't spend an hour on crates that timeout during the build
Currently if we have a build that hits the 15 minute timeout (e.g. [`app-db-schema@0.1.2`](https://docs.rs/crate/app-db-schema/0.1.2/builds/672648)) we will end up spending an hour total attempting to build the release. We do 4 builds total, each with the full timeout available to them:

1. With the crates `Cargo.lock`:
    a. Generate coverage data
    b. Build docs
2. After deleting the lock:
    a. Generate coverage data
    b. Build docs

It would be better to only generate the coverage data if the build succeeded, so we would only attempt builds 1.b. and 2.b. before deciding it timed out, but that has issues

https://github.com/rust-lang/docs.rs/blob/2e5ef9b6d8f13b527436893a9a5e3e67019d5fb3/src/docbuilder/rustwide_builder.rs#L658-L660

One idea would be to skip the subsequent steps if one fails because of a timeout rather than a build error. It seems unlikely that unlocking the crate will turn a timeout into a successful build, or that `rustdoc --show-coverage` would somehow be the cause of a timeout rather than it being one of the dependencies.

Nemo157:
fixes #1910

### DIFF
--- a/.sqlx/query-a4cf6f9051e04118ca5ed67813d26c5deee3e17a11a070b6e8d0879b93e68152.json
+++ b/.sqlx/query-a4cf6f9051e04118ca5ed67813d26c5deee3e17a11a070b6e8d0879b93e68152.json
@@ -1,0 +1,20 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT json_array_length(releases.doc_targets) as length FROM releases;",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "length",
+        "type_info": "Int4"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "a4cf6f9051e04118ca5ed67813d26c5deee3e17a11a070b6e8d0879b93e68152"
+}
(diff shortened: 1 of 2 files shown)


Nemo157:
One other case this potentially impacts a lot is when a crate is published with an incorrect lockfile, to get any docs the failure with the old lockfile and then the successful build with the new lockfile will have to complete together within the timeout.

I want to add more detailed logging and more visible metrics around this to the `/crate/.../builds` pages, so authors may improve their crates configuration, but would prefer to delay that till after the work I'm trying to get completed soon on #1011.

syphar:
when we would deploy this, we wouln't have any idea how many crate builds we would break. 

I assume #2186 would help a little, but only for the builds in the time observed. Same as #1011 when we record more information about the build. 

Do we have other ways to optimize? 
For example, do we try a broken lockfile for each target?

syphar:
Without more data this change feels quite risky, what do you think about waiting for the results of #2186 and then deciding?

Nemo157:
Sounds good.

Nemo157:
Some stats from the last week:

image

The orange line is 15 minutes, the red line is 30 minutes. We did have the 99th percentile exceed 30 minutes briefly, but I think it's prettty safe as an overall timeout if it'll just cause _some_ targets to be missing.

PR #2187, opened 2023-08-09
Resurrect CBOR index handling and migration, update to migrate failed builds

Nemo157:
We had some errors logged saying that the index wasn't a sqlite database: https://rust-lang.sentry.io/issues/4382204513/events/6c800157e8484ad893508b75745255a2/

```
Error: fetching readme for deno_websocket 0.48.0
Error: fetching Cargo.toml from deno_websocket 0.48.0 (archive: true)
SqliteFailure: file is not a database
Error: Error code 26: File opened that is not a database file
```

For the impacted crates trying to browse their source files also didn't work, the list was there (since that's stored in the database) but none of the files showed source code.

Investigating on the server we can see that the index files _are_ not a database file, they're still in the CBOR encoding:

```console
> xxd /opt/docs-rs-prefix/archive_cache/sources/deno_fetch/0.66.0.zip.index | head -n1
00000000: a165 6669 6c65 73b1 7366 735f 6665 7463  .efiles.sfs_fetc
```

This only impacts crates that failed to build, because [I suggested those be filtered out from migrating](https://github.com/rust-lang/docs.rs/pull/2154#discussion_r1278544162) to avoid the error when their rustdoc-archive didn't exist, but they still have the source-archive that needs migrating.

This PR reverts the CBOR index removal, and updates the migration to also attempt migrating crates that failed to build, and doesn't report the index not existing as an error.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -3,6 +3,9 @@ ignore = [
     "RUSTSEC-2020-0036", # failure is officially deprecated/unmaintained
     # https://github.com/rust-lang/docs.rs/issues/1014
 
+    "RUSTSEC-2021-0127", # serde_cbor is unmaintained
+    # https://github.com/rust-lang/docs.rs/issues/1568
+
     "RUSTSEC-2023-0018", # rustwide -> remove_dir_all,TOCTOU / Race Condition 
     # https://github.com/rust-lang/docs.rs/issues/2074
 ]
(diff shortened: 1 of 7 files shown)


Nemo157:
I ran this on my local instance and confirmed it migrated and made it possible to browse sources on crates that had failed to build.

syphar (approved):

PR #2209, opened 2023-09-03
Add a disclaimer about how limited our feature metadata is

Nemo157:
I've seen some complaints about how limited the feature page is on discord. To make it more obvious about _why_ it's currently limited I thought it would be useful to have a small disclaimer linking to the common places that crate authors document the features.

image


GuillaumeGomez (approved):

GuillaumeGomez:
I think it's a good idea indeed.

syphar (approved):

PR #2236, opened 2023-10-02
Specify color-scheme for each theme

Dubzer:
This PR adds a CSS property color-scheme to `_schemes.scss`. 
It allows the browser to apply the appropriate default user-agent styles for the current theme. For example, to use dark scrollbar on Windows in chromium-based browsers. 

Learn more about this property: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

Here's an example of improvement in scrollbar style for dark themes on Windows in MS Edge:

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/templates/style/_themes.scss
+++ b/templates/style/_themes.scss
@@ -1,5 +1,6 @@
 // Standard white theme
 html {
+  color-scheme: light;
   --color-background-code: #f5f5f5;
   --color-background: #fff;
   --input-color: #000;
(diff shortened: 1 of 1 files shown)


Nemo157:
Interesting, this doesn't seem to do much in firefox because rustdoc sets the `scrollbar-color` overriding it (to something that's hard to see with just the thumb visible in `dark` :pensive:), and even if it didn't firefox appears to base the scrollbar color on the background-color, but chromium browsers and safari don't support setting it yet and I guess don't auto-derive a good color so this is useful in them.

Nemo157 (approved):
I'd still like @GuillaumeGomez to take a look too.

GuillaumeGomez:
I'll try to take a look by the end of the week.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #2292, opened 2023-10-23
Replace tera with rinja

GuillaumeGomez:


GuillaumeGomez:
PR is now ready! For now the rinja release hasn't been done yet. Can be done any time. :)

src/web/page/web_page.rs
```diff
@@ -115,51 +95,47 @@ macro_rules! impl_axum_webpage {
...
-                templates
-                    .templates
-                    .render(&template, &context)
-                    .map_err(Error::new)
-            };
-
-            let rendered = match rendered {
+            let DelayedTemplateRender { template, .. } = render;
+
+            // let rendered = if cpu_intensive_rendering {
```
GuillaumeGomez, line 111: I'm not sure if it's still relevant to have this code. I commented it to ensure it was ok.

templates/core/about/metadata.html
```diff
@@ -14,7 +14,7 @@ <h1>Metadata for custom builds</h1>
 
 	<p>The available configuration flags you can customize are:</p>
 
-	{% filter highlight(lang="toml") %}
+	{% filter highlight("toml") %}
```
syphar, line 17: something seems off with this page: 
grafik

templates/rustdoc/topbar.html
```diff
@@ -130,17 +125,7 @@
...
-                                                <a href="/{{ dep[0] }}/{{ dep[1] }}" class="pure-menu-link">
-                                                    {{ dep[0] }} {{ dep[1] }}
-                                                    <i class="dependencies {{ dep[2] | default(value='') }}">{{ dep[2] | default(value="") }}</i>
-                                                    {%- if dep[3] -%}
-                                                    <i> optional </i>
-                                                    {%- endif -%}
-                                                </a>
-                                            </li>
-                                        {%- endfor -%}
+                                        {% call macros::dependencies_list(link_prefix="", dependencies=krate.dependencies, if_empty="") %}
```
syphar, line 128: It seems like this shows all dependencies mistakenly as optional?

grafik

to compare:
https://docs.rs/regex/1.10.5/regex/

templates/crate/details.html
```diff
@@ -105,21 +101,7 @@
...
-                                                    {%- if dep[3] -%}
-                                                        <i> optional </i>
-                                                    {%- endif -%}
-                                                </a>
-                                            </li>
-                                        {%- endfor -%}
-                                    {%- else -%}
-                                        &mdash;
-                                    {%- endif -%}
+                                    {% call macros::dependencies_list(link_prefix="/crate", dependencies=details.dependencies, if_empty="&mdash;") %}
```
syphar, line 104: this seems to show all dependencies mistakenly as optional.

templates/crate/details.html
```diff
@@ -96,7 +92,7 @@
                         <li class="pure-menu-item">
                             <a href="https://crates.io/crates/{{ details.name }}" class="pure-menu-link"
                                 title="See {{ details.name }} on crates.io">
-                                {{ "cube" | fas }} crates.io
+                                {{ "cube"|fas(false, false, "")|safe }} Crates.io
```
syphar, line 95:
```suggestion
                                {{ "cube"|fas(false, false, "")|safe }} crates.io
```

syphar (changes requested):
This is huge! thank you! 

I really like compile time checked templates :) 

Generally there seem to be some small spacing changes, not sure if they are an issue for you? 
( they seem fine with me)

src/web/tmp.rs
```diff
@@ -0,0 +1,19 @@
+use crate::db::types::{BuildStatus, Feature};
```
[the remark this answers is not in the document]
GuillaumeGomez: Woups, forgot to remove it.

src/web/page/web_page.rs
GuillaumeGomez: Sounds goo to me! Then moving this content in there.

src/web/releases.rs
```diff
@@ -442,11 +494,12 @@ pub(crate) async fn owner_handler(Path(owner): Path<String>) -> AxumResult<impl
     .map_err(|_| AxumNope::OwnerNotFound)
 }
 
+#[derive(Template)]
+#[template(path = "releases/search_results.html")]
 #[derive(Debug, Clone, PartialEq, Serialize)]
 pub(super) struct Search {
     pub(super) title: String,
-    #[serde(rename = "releases")]
```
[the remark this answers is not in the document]
GuillaumeGomez: Yes we can!

templates/core/about/metadata.html
GuillaumeGomez: Ah yes, forgot `|safe`.

templates/crate/details.html
GuillaumeGomez: Yes indeed.

templates/crate/details.html
GuillaumeGomez: Fixed the typo.

GuillaumeGomez:
And fixed merge conflicts.

GuillaumeGomez:
Fixed the fix for JS mime types.

GuillaumeGomez:
CI passed. \o/

templates/core/about/metadata.html
syphar: @GuillaumeGomez still looks like this

syphar (changes requested):
one remaining issue

src/web/rustdoc.rs
```diff
@@ -604,12 +602,6 @@ pub(crate) async fn rustdoc_html_server_handler(
         .recently_accessed_releases
         .record(krate.crate_id, krate.release_id, target);
 
-    let target = if target.is_empty() {
-        String::new()
-    } else {
-        format!("{target}/")
```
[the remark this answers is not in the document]
GuillaumeGomez: I'm surprised as well. Do we have tests to ensure this behaviour by any chance? Because `target` is declared in the templates but I'd like to be sure it's actually doing what it's supposed to.

templates/core/about/metadata.html
GuillaumeGomez: Indeed, might be a bug in rinja. Investigating!

GuillaumeGomez:
This issue allowed me to uncover a bug in rinja. PR for fix is open [here](https://github.com/rinja-rs/rinja/pull/71).

templates/core/about/metadata.html
GuillaumeGomez: Fixed. :)

GuillaumeGomez:
Fixed the bug. Should be ready now.

GuillaumeGomez:
Forgot to add a `|safe` once again. Fixed too.

src/web/rustdoc.rs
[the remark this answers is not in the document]
syphar: this is super odd. From what I see the only place where this attribute is used is in the releases-list, where it gets contatenated with the `inner_path` without a slash in between. So we would nees this trailing slash here when the inner path doesn't start with a slash. 

But I fail to find any situation where it's either broken on master or in this PR, and I don't know why. 

So my hunch is to leave it out and see if we find breakages.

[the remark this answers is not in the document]
syphar: Ah wait, you're not even passing it into the `RustdocPage` any more. 

So, the reason we don't need it is that the releases-list is not directly rendered any more, but only via ajax, in `web::crate_details::get_all_releases`, 

where we still have this behaviour for the target string.

syphar (approved):
I can't find anything else, so I would propose we go for it. 

We probably will find some edge cases in production, but I feel safer with compile time checked templates. 

Can you release rinja and change the dependency? 

Also you can do any rebasing, if you wanted to do any.

GuillaumeGomez:
Normally it should already be rebased. Gonna make a release today hopefully.

GuillaumeGomez:
Released `rinja`, rebased this branch on top of `master`. I'll merge once CI is green then.

templates/rustdoc/topbar.html
```diff
@@ -64,56 +59,56 @@
...
+                                    <a href="{{ repository_url }}" class="pure-menu-link">
+                                        {{ "code-branch"|fas(false, false, "")|safe }} Repository
                                     </a>
                                 </li>
                             {%- endif -%}
 
                             <li class="pure-menu-item">
                                 <a href="https://crates.io/crates/{{ krate.name }}" class="pure-menu-link" title="See {{ krate.name }} in crates.io">
-                                    {{ "cube" | fas }} crates.io
+                                    {{ "cube"|fas(false, false, "")|safe }} Crates.io
```
Turbo87, line 90: @GuillaumeGomez was there a reason for reverting this back to capitalized? (see https://github.com/rust-lang/docs.rs/pull/2508)

GuillaumeGomez: No, mistake on my end. Gonna send a PR to fix it.

PR #2297, opened 2023-10-31
add sort by select-box for searching crate

linyihai:
This PR for https://github.com/rust-lang/docs.rs/issues/2152

The main features are:

- Add a drop-down selection box to the top right of the search result to support searching by category. This prototype comes from crates-io, so the main drop-down options also come from it. 
- Another improvement is that when the dropdown option is changed, the search will be triggered if there is content in the search box

In the end, I didn't find a good way to test it, so I built some packages locally, changed the drop-down options, and saw how the results returned compared to crates-io.

the finally look is:

image

### DIFF
--- a/static/keyboard.js
+++ b/static/keyboard.js
@@ -110,6 +110,18 @@
         }
     }
 
+    function handleSortByChange() {
+        const inputSearch = document.getElementById("nav-search");
+        const searchForm = document.getElementById("nav-search-form");
+        if (inputSearch.value && searchForm) {
+            searchForm.submit()
+        }
+    }
+    const searchSortBySel = document.getElementById("nav-sort");
+    if (searchSortBySel) {
+        searchSortBySel.addEventListener("change", handleSortByChange)
+    }
+
     document.onkeypress = handleKey;
     document.onkeydown = handleKey;
 })();
--- a/templates/header/topbar_begin.html
+++ b/templates/header/topbar_begin.html
@@ -8,6 +8,7 @@
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
             <form action="/releases/search"
                   method="GET"
+                  id="nav-search-form"
                   class="landing-search-form-nav {%
                   if is_latest_version is defined and not is_latest_version %}not-latest{% endif
                   %} {% if metadata.yanked %}yanked{% endif %}">
--- a/templates/header/topbar_end.html
+++ b/templates/header/topbar_end.html
@@ -67,19 +67,6 @@
                     <input id="nav-search" name="query" type="text" aria-label="Find crate by search query" tabindex="-1"
                         placeholder="Find crate" {%- if search_query %} value="{{ search_query }}" {%- endif %}>
                 </div>
-                {# The sort by select-box #}
-                <div id="search-select-nav"> 
-                    <label for="nav-sort">
-                        {{ "list" | fas }}
-                    </label>
-                    <select name="sort" id="nav-sort" aria-label="Find crate by the sort by select-box"  tabindex="-1">
-                        <option value="relevance" {%- if search_sort_by and search_sort_by == "relevance" %} selected="selected" {%- endif %}>Relevance</option>
-                        <option value="downloads" {%- if search_sort_by and search_sort_by == "downloads" %} selected="selected" {%- endif %}>All-Time Downloads</option>
-                        <option value="recent-downloads" {%- if search_sort_by and search_sort_by == "recent-downloads" %} selected="selected" {%- endif %}>Recent Downloads</option>
-                        <option value="recent-updates" {%- if search_sort_by and search_sort_by == "recent-updates" %} selected="selected" {%- endif %}>Recent Updates</option>
-                        <option value="new" {%- if search_sort_by and search_sort_by == "new" %} selected="selected" {%- endif %}>Newly Added</option>
-                    </select>
-                </div>
             </form>
         </div>
     </div>
--- a/templates/header/topbar_end.html
+++ b/templates/header/topbar_end.html
@@ -67,6 +67,19 @@
                     <input id="nav-search" name="query" type="text" aria-label="Find crate by search query" tabindex="-1"
                         placeholder="Find crate" {%- if search_query %} value="{{ search_query }}" {%- endif %}>
                 </div>
+                {# The sort by select-box #}
+                <div id="search-select-nav"> 
+                    <label for="nav-sort">
+                        {{ "list" | fas }}
+                    </label>
+                    <select name="sort" id="nav-sort" aria-label="Find crate by the sort by select-box"  tabindex="-1">
+                        <option value="relevance" {%- if search_sort_by and search_sort_by == "relevance" %} selected="selected" {%- endif %}>Relevance</option>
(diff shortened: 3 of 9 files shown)


templates/header/topbar_end.html
```diff
@@ -67,6 +67,19 @@
                     <input id="nav-search" name="query" type="text" aria-label="Find crate by search query" tabindex="-1"
                         placeholder="Find crate" {%- if search_query %} value="{{ search_query }}" {%- endif %}>
                 </div>
+                {# The Search Sort by dropdown menu#}
+                <div id="search-select-nav"> 
+                    <label for="nav-sort">
+                        {{ "list" | fas }}
+                    </label>
+                    <select name="sort" id="nav-sort">
+                        <option value="revevance" {%- if search_sort_by and search_sort_by == "revevance" %} selected="selected" {%- endif %}>Relevance</option>
```
Nemo157, line 76:
```suggestion
                        <option value="relevance" {%- if search_sort_by and search_sort_by == "relevance" %} selected="selected" {%- endif %}>Relevance</option>
```

Nemo157:
From the screenshot it looks like a different text color is being used. I'm not sure if the existing CSS has handling for a `select` in the navbar, it may just need some class applied, or it might need some extra rules added.

linyihai:
image
thanks for you suggestions, I add the same CSS into the select-box, now it looks well.

templates/header/topbar_end.html
linyihai: It looks like a spelling mistake. I'll correct it later

static/keyboard.js
```diff
@@ -110,6 +110,18 @@
         }
     }
 
+    function handleSortByChange() {
+        var inputSearch = document.getElementById("nav-search");
```
GuillaumeGomez, line 114: Please replace `var` with `const`.

static/keyboard.js
```diff
@@ -110,6 +110,18 @@
...
+        var searchForm = document.getElementById("nav-search-form");
```
GuillaumeGomez, line 115: Please replace `var` with `const`.

static/keyboard.js
```diff
@@ -110,6 +110,18 @@
...
+        if (inputSearch.value && searchForm) {
+            searchForm.submit()
+        }
+    }
+    var searchSortBySel = document.getElementById("nav-sort");
```
GuillaumeGomez, line 120: Please replace `var` with `const`.

templates/header/topbar_begin.html
```diff
@@ -8,6 +8,7 @@
         <div class="pure-menu pure-menu-horizontal" role="navigation" aria-label="Main navigation">
             <form action="/releases/search"
                   method="GET"
+                  id = "nav-search-form"
```
GuillaumeGomez, line 11:
```suggestion
id="nav-search-form"
```

templates/header/topbar_end.html
```diff
@@ -67,6 +67,19 @@
                     <input id="nav-search" name="query" type="text" aria-label="Find crate by search query" tabindex="-1"
                         placeholder="Find crate" {%- if search_query %} value="{{ search_query }}" {%- endif %}>
                 </div>
+                {# The Search Sort by dropdown menu#}
```
GuillaumeGomez, line 70: That will very likely mess up the media queries since it'll change the size of the content in the topbar.

linyihai: em, I test it on Edge web browser, the width and height of the select-box  stay the same when i change the option.

Is there anything else I need to test.

GuillaumeGomez: Sorry, that's not what I meant: reduce the width of your window. Normally, the topbar icons should never be on two lines.

linyihai: So you mean: The new drop-down box is not recommended to be placed at the topbar ?

GuillaumeGomez: I don't think it's a good idea indeed, as the topbar is already very loaded and also because this box should be closer to the element it interacts with.

linyihai: I'll try to move the drop-down box to the top-right corner of the search results, as crates.io did. But I ran into some problems (some code issues), maybe I will come out with another version next week.

GuillaumeGomez:
Thanks for the update! I'll try to review it by the end of the week.

GuillaumeGomez:
So currently it looks like this:

image

With these changes:

```diff
diff --git a/templates/style/_vars.scss b/templates/style/_vars.scss
index aa97be9d..f28a0c19 100644
--- a/templates/style/_vars.scss
+++ b/templates/style/_vars.scss
@@ -7,6 +7,7 @@ $font-family-mono: "Source Code Pro", Menlo, Monaco, Consolas, "DejaVu Sans Mono
 
 // Sizes
 $top-navbar-height: 32px; // height of the floating top navbar
+$search-result-right-left-padding: 1em; // Size of the right and left padding for the search results
 
 // Pure compatible media queries
 // usage:
diff --git a/templates/style/style.scss b/templates/style/style.scss
index 0f3d7f92..149a6789 100644
--- a/templates/style/style.scss
+++ b/templates/style/style.scss
@@ -244,6 +244,7 @@ div.landing {
 #search-select-nav {
     display: flex;
     flex-direction: column;
+    padding: 1em $search-result-right-left-padding;
 
     .item-end {
         align-self: flex-end
@@ -282,7 +283,7 @@ div.recent-releases-container {
     .release {
         display: block;
         border-bottom: 1px solid var(--color-border);
-        padding: 0.4em 1em;
+        padding: 0.4em $search-result-right-left-padding;
         color: var(--color-standard);
 
         a {
```

It now looks like this:

image

Could you apply them please?

linyihai:
> ```diff
> $search-result-right-left-padding;
> ```

No problem.

linyihai:
> So currently it looks like this:
> 
> image
> 
> With these changes:
> 
> ```diff
> diff --git a/templates/style/_vars.scss b/templates/style/_vars.scss
> index aa97be9d..f28a0c19 100644
> --- a/templates/style/_vars.scss
> +++ b/templates/style/_vars.scss
> @@ -7,6 +7,7 @@ $font-family-mono: "Source Code Pro", Menlo, Monaco, Consolas, "DejaVu Sans Mono
>  
>  // Sizes
>  $top-navbar-height: 32px; // height of the floating top navbar
> +$search-result-right-left-padding: 1em; // Size of the right and left padding for the search results
>  
>  // Pure compatible media queries
>  // usage:
> diff --git a/templates/style/style.scss b/templates/style/style.scss
> index 0f3d7f92..149a6789 100644
> --- a/templates/style/style.scss
> +++ b/templates/style/style.scss
> @@ -244,6 +244,7 @@ div.landing {
>  #search-select-nav {
>      display: flex;
>      flex-direction: column;
> +    padding: 1em $search-result-right-left-padding;
>  
>      .item-end {
>          align-self: flex-end
> @@ -282,7 +283,7 @@ div.recent-releases-container {
>      .release {
>          display: block;
>          border-bottom: 1px solid var(--color-border);
> -        padding: 0.4em 1em;
> +        padding: 0.4em $search-result-right-left-padding;
>          color: var(--color-standard);
>  
>          a {
> ```
> 
> It now looks like this:
> 
> image
> 
> Could you apply them please?

No problem.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Let's wait for @Nemo157 to check if it looks good to them as well.

Nemo157 (approved):

Issue #2389, opened 2024-01-22
Native support for docs.rs conditional compilation


epage:
Currently, docs.rs sets [`DOCS_RS` on the environment of the build](https://docs.rs/about/builds#cross-compiling) which allows build scripts to tell when running within docs.rs but this requires users to add boilerplate if they want to then use unstable features for the docs.

#954 codified in the documentation a practice of having [`--cfg docsrs` set in the metadata table](https://docs.rs/about/builds#cross-compiling).

This pattern has been long enough that it seems like we should just bake it in and always pass `--cfg docsrs` for users.  This would be both a code to pass in the flag like the user does today and a documentation update.

One reason to do something is to help break a deadlock on a problem with [`--check-cfg`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#check-cfg).  If Cargo passes `--check-cfg` to rustc by default, then instances of `#[cfg(docsrs)]` will warn.
- You can't put an `#[allow]` around a `#[cfg]`
- You could put it in the enclosing scope but, for enabling nightly features, that would be the entire lib, negating this feature
- You could add a `build.rs` to tell `cargo` that `docsrs` is a "known value" but that then slows down your build for documentation and increases boilerplate
- You could use a Cargo `feature` instead but having nightly-only features is annoying (no `cargo check --all-features`).

`rustc --check-cfg` has a list of well-known names but
- docs.rs might be removed enough from rustc that it might not make sense to bake in docs.rs-specific logic.  We could solve this by having the concept of well-known names in Cargo which has a closer relationship with docs.rs.
- Even if we do, its a convention, rather than a policy of docs.rs.  If docs.rs/ passed in `--cfg docsrs`, then it adds weight for making it a well known name.

A future extension of this could be to set a different `--cfg` for [public dependencies](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#public-dependency) of what is being built

[zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/356853-t-docs-rs/topic/.60--cfg.20docsrs.60)

PR #2390, opened 2024-01-23
Add `--cfg docsrs` to all rustdoc invocations

Urgau:
Fixes https://github.com/rust-lang/docs.rs/issues/2389

### DIFF
--- a/crates/metadata/lib.rs
+++ b/crates/metadata/lib.rs
@@ -267,7 +267,12 @@ impl Metadata {
             cargo_args.push("--no-default-features".into());
         }
 
-        let mut all_rustdoc_args = self.rustdoc_args.clone();
+        // Unconditionnaly set `--cfg docsrs` as it has become a de-facto way to
+        // distinguish docs.rs.
+        //
+        // See https://github.com/rust-lang/docs.rs/issues/2389.
+        let mut all_rustdoc_args = vec!["--cfg".into(), "docsrs".into()];
+        all_rustdoc_args.extend_from_slice(&self.rustdoc_args);
         all_rustdoc_args.extend_from_slice(rustdoc_args);
 
         // Pass `RUSTFLAGS` and `RUSTDOCFLAGS` using `cargo --config`, which handles whitespace correctly.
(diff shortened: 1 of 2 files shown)


crates/metadata/lib.rs
```diff
@@ -267,10 +267,15 @@ impl Metadata {
...
+        // Unconditionnaly set `--cfg docsrs` as it has become a de-facto way to
+        // distinguish docs.rs.
+        //
+        // See https://github.com/rust-lang/docs.rs/issues/2389.
+        let mut all_rustdoc_args = vec!["--cfg".into(), "docsrs".into()];
+        all_rustdoc_args.extend_from_slice(&self.rustdoc_args);
         all_rustdoc_args.extend_from_slice(rustdoc_args);
 
-        if !self.rustc_args.is_empty() || !all_rustdoc_args.is_empty() {
+        if !self.rustc_args.is_empty() || all_rustdoc_args.len() > 2 {
```
Nemo157, line 278: I think we might be able to get rid of this now, it was from when `--config` was unstable, we're not adding any other unstable flags below other than `-Z` ones which shouldn't need `-Zunstable-options`.

templates/core/about/builds.html
```diff
@@ -29,6 +29,13 @@ <h4 id="setting-a-readme"> <a href="#setting-a-readme">Setting a README</a> </h4
     </p>
 
     <h4 id="detecting-docsrs"> <a href="#detecting-docsrs">Detecting Docs.rs</a> </h4>
+    <p>
+        To recognize Docs.rs from your Rust code, you can test for the <code>docsrs</code> cfg, e.g.:
```
Nemo157, line 33: Should mention that this is only set in the terminal crate, not for dependencies.

crates/metadata/lib.rs
Urgau: Done.

templates/core/about/builds.html
Urgau: The documentation for `cargo rustdoc` [says](https://doc.rust-lang.org/nightly/cargo/commands/cargo-rustdoc.html#description):

> The specified target for the current package (or package specified by -p if provided) will be documented with the specified args being passed to the **final rustdoc invocation**. Dependencies will not be documented as part of this command.

----

So maybe we should also use "final rustdoc invocation"?

```diff
-        The `docsrs` cfg only applies to the "terminal" crate (aka the last crate build by Cargo).
-        It does not applies to dependencies or workspace members, only the "terminal" crate.
+        The `docsrs` cfg only applies to the final rustdoc invocation. It does not applies
+        to dependencies (even workspace one).
```

epage: `cargo rustc` also has this distinction and uses terms like in `cargo help rustc`:
- "current package" (vs "all of its dependencies")
- "final compiler invocation"

(`cargo help rustdoc` seems to gloss over this)

With a quick glance, I'm not seeing this distinction made elsewhere.

rust-lang/rfcs#3310 uses the term "root crate" for "selected packages" but our existing documentation refers to the "root crate" as the crate in the non-virtual workspace manifest.

Urgau: Thanks for the input. It seems like there isn't (yet) a name for this.

So for the purpose documenting the `docsrs` cfg I went ahead and updated the text to:

> The `docsrs` cfg only applies to the final rustdoc invocation (i.e. the crate currently being documented). It does not apply to dependencies (including workspace ones).

Nemo157 (approved):
Changes look good to me, would be good to hear if @syphar can think of any reason to not do this.

syphar (approved):
I agree, from what I see this has become the de-facto standard, so setting it by default is a good idea. 

You _could_ see this as a breaking change, but only if 
- someone used the feature name `docsrs` for something else, or 
- the feature exists for future usage, but shouldn't be used yet. 

Also having @epage on this topic makes me lean towards seeing these as edge-cases, and merging & deploying this.

epage:
Good thinking through those cases and I agree that those are edge enough cases that for a service like docs.rs it seems fine for moving forward.

PR #2412, opened 2024-02-09
Fix invalid release link generation

### LINKED ISSUE #2411: Clicking on old version in the topbar menus loads the name of the crate into the searchbar
I you visit https://docs.rs/event-listener/latest/event_listener/index.html and then click on the "4.0.3" link in the menu under the second menu in the top bar, you'll not only be taken to the "4.0.3" docs, but "event-listener" will be loaded in the search bar for you, leading to a cryptic error that will confuse you (since you didn't search for anything).

GuillaumeGomez:
Fixes #2411.

Sorry in advance for the noise but for some reason I can't run DB migrations locally so I can't run tests. I'll use the CI to check that I did correctly.


src/web/crate_details.rs
```diff
@@ -501,7 +501,7 @@ pub(crate) async fn get_all_releases(
         };
 
         inner = inner_path.join("/");
-        (target, inner.trim_end_matches('/'))
+        (target.replace('-', "_"), inner.trim_end_matches('/'))
```
Nemo157, line 504: This needs to use the crate name from the database, it's user configurable.

GuillaumeGomez:
CI passed. \o/

src/web/crate_details.rs
```diff
@@ -495,7 +496,7 @@ pub(crate) async fn get_all_releases(
                 .any(|s| Some(s) == params.target.as_ref())
         {
             inner_path.remove(0);
-            params.target.as_ref().unwrap()
+            &row.target_name
```
syphar, line 499: this seems wrong, we still need the target in the links, right? 

( I see that the column name `target_name` is confusing, since it's actually the package-name)

syphar (changes requested):
I'll do manual testing now, 

but just from reading the code I would say this will drop the target from the version links when I'm on a non-default target, since now you don't use `params.target` at all any more. 

Or am I missing something? 

Example: 
https://docs.rs/tokio_telegraf/0.5.0/i686-pc-windows-msvc/tokio_telegraf/index.html
In the version links in the dropdown we see the target name too.

syphar:
I now ran a local build, and yes, the target is missing. 

Also there seems to be another issue: 

grafik

GuillaumeGomez:
Good point. I'll add a test for that as well, thanks!

src/web/crate_details.rs
GuillaumeGomez: It was indeed wrong. The problem came from two places: the test that didn't add a target and thus this modification which came from my misunderstanding.

GuillaumeGomez:
Extended/fixed test and fixed code as well.

syphar (approved):

syphar (approved):

PR #2448, opened 2024-03-08
Add .editorconfig

Nemo157:
I keep having to fix indentation because my default html/js indent is set to 2, it might be nice to actually auto-format these files at some point, but for now at least it'd be good to tell editors what the expected setup is.


syphar (approved):
awesome, thanks, didn't realize we didn't have one yet

PR #2470, opened 2024-03-26
Faster docker development

Nemo157:
This is a set of various changes I did to make working on the UI faster while using a docker-compose instance. Overall it about halved the time between changing a `templates/style` file and being able to view the result for me after `docker compose up -d --build web`. (And allows changing the JS on the fly and just reloading the page).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*.{rs,html,js,json}]
+[*.{rs,html,js,json,scss}]
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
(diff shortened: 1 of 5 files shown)


GuillaumeGomez (approved):

Nemo157:
So this also affects CI times, it significantly slows down the clean build because of optimizing all dependencies, but speeds up the incremental build a little (though that is dominated by times that it doesn't affect), it also slightly reduces the cache size (by ~5%).

| | clean | incremental |
|-|-|-|
| before | [image](https://github.com/rust-lang/docs.rs/actions/runs/8356146418) | [image](https://github.com/rust-lang/docs.rs/actions/runs/8444633417) |
| after |[image](https://github.com/rust-lang/docs.rs/actions/runs/8451706674) | [image](https://github.com/rust-lang/docs.rs/actions/runs/8452045085) |

syphar (approved):
awesome, thanks! 

I'm fine with the changes, so you can merge when the conflicts are fixed.

PR #2478, opened 2024-04-02
Add test case with zero byte source file

Nemo157:
(draft because I cannot run build-tests locally, need to figure out how to run them from within docker at some point)


Nemo157:
Huh, so it works independently :thinking: I guess it's related to the production database state somehow. I did miss when I looked at the logs last time that we get a `docs_rs::db::add_package: Adding package into database` log line so it does complete adding the source files into an archive. So we'll have to wait and see what the production logs say once the panic is fixed.

PR #2487, opened 2024-04-06
Pin eslint back to v8

Nemo157:
v9 has a breaking change around config file formats: https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #2506, opened 2024-05-11
Handle SizeLimit error

shusann01116:
This solves #2481. Handle if `AsyncStorage::fetch_source_file` returnes `SizeLimitReached` error.

The message shown in the page will be like the screen shot.

image

### DIFF
--- a/src/web/source.rs
+++ b/src/web/source.rs
@@ -153,6 +153,7 @@ struct SourcePage {
     file: Option<File>,
     file_content: Option<String>,
     canonical_url: CanonicalUrl,
+    is_file_too_large: bool,
     is_latest_url: bool,
     use_direct_platform_links: bool,
 }
@@ -226,7 +227,7 @@ pub(crate) async fn source_browser_handler(
 
     // try to get actual file first
     // skip if request is a directory
-    let blob = if !params.path.ends_with('/') {
+    let (blob, is_file_too_large) = if !params.path.ends_with('/') {
         match storage
             .fetch_source_file(
                 &params.name,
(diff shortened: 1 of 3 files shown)


syphar (changes requested):
Awesome, thank you! 

Two changes would be awesome: 
- cold you move the warning to the right of the file list? 
- also do you mind adding a small test? something like `fetch_source_file_utf8_path` is probably enough, checking for the existance of the warning. 

I'm not 100% certain if we should change the error returned by the `SizedBuffer` to make the matching easier, but it's ok for now.

shusann01116:
Thank you for the review!
I've made changes as you mentioned.

## Moved the warning to the right of the file list

image

## Added a tiny test to ensure the message exists

I'm not sure, but I mocked a file has 50MB + 1Byte array to exceeds the default limit.

src/web/source.rs
```diff
@@ -729,4 +737,23 @@ mod tests {
             Ok(())
         });
     }
+
+    #[test]
+    fn large_file_test() {
+        wrapper(|env| {
```
syphar, line 743: you can change the config for the test, so you don't have to create a 50 MB file in memory
```suggestion
        wrapper(|env| {
            env.override_config(|config| {
                config.max_file_size = 1;
                config.max_file_size_html = 1;
            });
            // ... 

```

syphar (changes requested):
last remaining bit, then we can merge and deploy! 

I'm already looking forward to deploying the change and reducing error logs :)

src/web/source.rs
shusann01116: Thank you!
Revised the code.

syphar (approved):

PR #2560, opened 2024-07-18
Add line numbers to `crate/**/source/`

xFrednet:
Pretty much what the title says. This adds line numbers to the source view that allow the same selection and URL behavior like rustdoc.

On desktop it looks like this:

image

And the phone view also works:

image

---

I'm not an HTML/CSS expert, and this change proofed that again. It took way too long to get the proper layout. But now it seems to work now, as I wanted. :D

I wasn't able to create the layout using the `pure-*` classes, as it would separate the line numbers from the code block for small screens. That's why I added a new class. I'm also not sure if I placed the CSS in the correct files. If you want anything moved around, please let me know!

---

Closes: https://github.com/rust-lang/docs.rs/issues/2551

r? syphar

Also thank you to @syphar for giving me guidance in the issue! :heart:

### DIFF
--- a/templates/crate/source.html
+++ b/templates/crate/source.html
@@ -127,8 +127,15 @@
                 {% else %}
                     {% set file_name = "" %}
                 {% endif %}
-                <div id="source-code" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
-                    {{- file_content|highlight(file_name)|safe -}}
+                <div id="source-code-container" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
+                    <div data-nosnippet class="source-code"><pre id="line-numbers"><code>
+                        {%- for line in 1..=file_content.lines().count() -%}
+                            <a href="#{{line|safe}}" id="{{line|safe}}">{{line|safe}}</a>
+                        {%~ endfor -%}
+                    </code></pre></div>
+                    <div id="source-code" class="source-code">
+                        {{- file_content|highlight(file_name)|safe -}}
+                    </div>
                 </div>
             {%- endif -%}
         </div>
--- a/templates/crate/source.html
+++ b/templates/crate/source.html
@@ -130,7 +130,7 @@
                 <div id="source-code-container" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
                     <div data-nosnippet class="source-code"><pre id="line-numbers"><code>
                         {%- for line in 1..=file_content.lines().count() -%}
-                            <a href="#{{line|safe}}" id="{{line|safe}}">{{line|safe}}</a>
+                            <a href="#{{line}}" id="{{line}}">{{line}}</a>
                         {%~ endfor -%}
                     </code></pre></div>
                     <div id="source-code" class="source-code">
--- a/static/source.js
+++ b/static/source.js
@@ -38,4 +38,95 @@
             toggleSource(toggleSourceButton);
         });
     });
+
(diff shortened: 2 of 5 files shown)


templates/crate/source.html
```diff
@@ -127,8 +127,15 @@
                 {% else %}
                     {% set file_name = "" %}
                 {% endif %}
-                <div id="source-code" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
-                    {{- file_content|highlight(file_name)|safe -}}
+                <div id="source-code-container" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
+                    <div data-nosnippet class="source-code"><pre id="line-numbers"><code>
+                        {%- for line in 1..=file_content.lines().count() -%}
+                            <a href="#{{line|safe}}" id="{{line|safe}}">{{line|safe}}</a>
+                        {%~ endfor -%}
```
GuillaumeGomez, line 134: What is the `~` doing?

templates/crate/source.html
```diff
@@ -127,8 +127,15 @@
                 {% else %}
                     {% set file_name = "" %}
                 {% endif %}
-                <div id="source-code" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
-                    {{- file_content|highlight(file_name)|safe -}}
+                <div id="source-code-container" class="pure-u-1 pure-u-sm-17-24 pure-u-md-19-24">
+                    <div data-nosnippet class="source-code"><pre id="line-numbers"><code>
+                        {%- for line in 1..=file_content.lines().count() -%}
+                            <a href="#{{line|safe}}" id="{{line|safe}}">{{line|safe}}</a>
```
GuillaumeGomez, line 133: Since line is a number, I think `|safe` is useless here.

templates/crate/source.html
xFrednet: The `~` keeps one white space character. Otherwise, it seems to remove the line breaks as well and all line numbers are placed on the same line.

templates/crate/source.html
xFrednet: Most likely, this was inspired/taken from rustdoc. I assumed it was used to indicate the string doesn't even need to be changed? 

https://github.com/rust-lang/rust/blob/5c8488605624d67b272953bc21d41db60dbd5654/src/librustdoc/html/templates/source.html#L4

Either way, I can remove it :)

GuillaumeGomez: Oh I see. Then the `<code>` should be removed but the `~` should be kept.

GuillaumeGomez: Arf, wrong location for my comment. In this case please remove `|safe`. :)

templates/crate/source.html
GuillaumeGomez: Oh I see. Then the `<code>` should be removed but the `~` should be kept.

xFrednet: The `<code>` tag was intentional, as some css rules target `pre > code`. I can remove the `<code>` tag, but that would mean that some rules need to be duplicated. I tried to change the rules to simply target `pre` but this doesn't work directly and needs more adaptation.

So, shall I remove the tag or do we want to keep it?

GuillaumeGomez: Hum... Let's go to the simplest then and keep the `<code>` then.

GuillaumeGomez (approved):
Looks good to me, thanks!

GuillaumeGomez:
@syphar: We'll really need to add GUI tests for things like source code pages. ^^'

PR #2596, opened 2024-09-06
More template cleanup

GuillaumeGomez:


GuillaumeGomez:
Can't figure out what's wrong when removing `is_latest_version`. Gonna come back to this one later on because it seems like some dark magic is happening in there.

syphar (approved):

PR #2599, opened 2024-09-11
fix the link to the docs.rs team on rust-lang.org

mockersf:
the link to the docs.rs team on https://www.rust-lang.org should be https://www.rust-lang.org/governance/teams/dev-tools#team-docs-rs

The anchor seems to have changed from `#docs-rs` to `#team-docs-rs`


GuillaumeGomez (approved):

GuillaumeGomez:
Good catch, thanks!

PR #2623, opened 2024-09-30
Put SVG images into CSS

GuillaumeGomez:
Since CSS files are often cached, we can take advantage of this to reduce network usage.

Now how it works: we inline the SVG images into a CSS file we generate at compile-time to ensure this file only contains what we need to limit its use but also to generate equivalent Rust types so we know at compile-time that the SVG exists.

I checked on caniuse for `mask` here: https://caniuse.com/css-masks. It says that more than 89% of current web browsers supports it unprefixed (without `-webkit-`). With the prefix, it goes up to more than 96%. Just in case, I used both (thanks to HTML variables, we don't need to duplicate the SVG for both `mask`) so we should be on the safe side with more than 96% of our potential users being able to see the icons.

I compared the HTML size on https://docs.rs/sysinfo/0.30.12/sysinfo/index.html. On current docs.rs, the transferred HTML is 52KB, with this PR is goes down to 40KB.


src/web/csp.rs
```diff
@@ -54,7 +54,7 @@ impl Csp {
         // the MIME type to allow loading favicons.
         //
         // Images from other HTTPS origins are also temporary allowed until issue #66 is fixed.
-        result.push_str("; img-src 'self' https:");
+        result.push_str("; img-src 'self' https: data:");
```
GuillaumeGomez, line 57: This change allows to have inlined SVG in the CSS.

GuillaumeGomez: If it's an issue, we can instead use URLs and have the SVG images stored in files.

GuillaumeGomez:
@syphar: Any idea why there are sqlx errors?

build.rs
```diff
@@ -80,11 +84,194 @@ fn main() -> Result<()> {
...
+    icon_str: &str,
+    type_name: String,
+    code_output: &mut String,
+    css_output: &mut String,
+    icon_kind: &str,
+) {
+    let css_class = format!("f-a_{icon_name}_{icon_kind}");
+    css_output.push_str(&format!(
+        "\
+.{css_class} {{
```
syphar, line 124: what I wonder: 

isn't a CSS distribution with the icons something that font-awesome already provides? That we can just use?

syphar:
not finished yet with all review, but one bigger question

build.rs
```diff
@@ -80,11 +84,194 @@ fn main() -> Result<()> {
     write_known_targets(out_dir)?;
     compile_syntax(out_dir).context("could not compile syntax files")?;
 
+    println!("cargo::rustc-check-cfg=cfg(icons_out_dir)");
```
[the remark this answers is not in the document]
GuillaumeGomez: It's used in `src/lib.rs` to import the generated Rust file.

syphar (changes requested):
Thank you for the work! 

I tried to test it locally, but I don't see any icons, probably because the fontawesome.css request leads to a 404

syphar:
Next thing from manual testing: 

new: 
grafik

old: 
grafik


it seems like the clipboard icon is missing there? 

it's also missing on the crate details page: 
grafik


The rest of the tests look good

GuillaumeGomez:
Found the issues:
1. The CSS file where I added `svg-clipboard` was not included in all docs.rs pages. So I moved it into `font-awesome.css` instead because I already modified it (for the fonts paths) to make it work with docs.rs. And since it's an icon, made more sense to put it there.
2. The SVG image contained a jinja comment that I forgot to update to HTML comment.

We should really add more GUI tests. ^^'

syphar (approved):

Issue #2733, opened 2025-02-02
Name of crates overlapping with their description


pvk29:
Longer names of crates overlapping with their description on desktop.
![Image](https://github.com/user-attachments/assets/4c873d6a-bf90-4aba-96b7-6b129262c434) #

GuillaumeGomez:
Thanks! Taking a look.

PR #2734, opened 2025-02-03
Fix display of long crates name in release list

GuillaumeGomez:
Fixes https://github.com/rust-lang/docs.rs/issues/2733.

It now looks like this:

image

Gonna send a GUI test once I added something missing into the framework to make the test simpler.

### DIFF
--- a/templates/core/home.html
+++ b/templates/core/home.html
@@ -36,7 +36,7 @@ centered
                     <strong>Recent Releases</strong>
                 </a>
                 <a href="/releases/feed" title="Atom feed">
-                    {{ crate::icons::IconSquareRss.render_solid(false, false, "") }}
+                    {{- crate::icons::IconSquareRss.render_solid(false, false, "") -}}
                 </a>
             </div>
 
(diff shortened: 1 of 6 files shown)


templates/core/home.html
```diff
@@ -54,16 +54,16 @@ <h1 class="brand">{{ crate::icons::IconCubes.render_solid(false, false, "") }} D
                             class="release">
                             <div class="pure-g">
                                 <div class="pure-u-1 pure-u-sm-6-24 pure-u-md-5-24 name">
-                                    {{ release.name }}-{{ release.version }}
+                                    {{- release.name }}-{{ release.version -}}
```
GuillaumeGomez, line 57: I wonder if we should just enable the "auto-trim" feature...

GuillaumeGomez:
I added a GUI test as well.

GuillaumeGomez:
Forgot to push so now it's done. :sweat_smile:

syphar:
thanks! 

When I test, the change looks different than the screenshot?

syphar:
grafik

syphar:
also the gui tests fail

GuillaumeGomez:
Ah right, might need to update the node version.

As for the fix, you need to hover the item to actually see it in action. ;)

syphar (approved):

PR #2752, opened 2025-02-28
Increase the width of search input in navbar

hamidrezakp:
This PR just removes the limitation of navbar search input width, allowing it to extend all the way to the end of flex container.

Before:
before

After:
after

### DIFF
--- a/templates/style/_navbar.scss
+++ b/templates/style/_navbar.scss
@@ -132,7 +132,6 @@ div.nav-container {
                 box-shadow: none;
                 background-color: var(--color-background);
                 height: 100%;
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
On smaller width (like 600px), it makes the layout too wide, adding a scrollbar on the x axis. So needs to have some extra changes. But it's a good idea nonetheless!

GuillaumeGomez:
I included your commit in https://github.com/rust-lang/docs.rs/pull/2761 and added the extra missing parts. Closing this PR. Thanks for your contribution!

PR #2756, opened 2025-03-04
Fix pagination for recent failures

Cerber-Ursi:
Currently, on the "recent failures" page, pagination links are broken - they lead to something like https://docs.rs/releases/recent_failures/2 instead of https://docs.rs/releases/recent-failures/2. I've attempted to change the corresponding string representation - it seems to be not used otherwise.

### DIFF
--- a/src/web/releases.rs
+++ b/src/web/releases.rs
@@ -2182,4 +2182,69 @@ mod tests {
             Ok(())
         })
     }
+
+    #[test]
+    fn recent_failures_correct_pagination_links() {
+        async_wrapper(|env| async move {
+            for i in 0..RELEASES_IN_RELEASES + 1 {
+                env.fake_release()
+                    .await
+                    .name("failed")
+                    .version(&format!("0.0.{}", i))
+                    .build_result_failed()
+                    .create()
+                    .await?;
+            }
+
+            let web = env.web_app().await;
+
+            let response = web.get("/releases/recent-failures").await?;
+            assert!(response.status().is_success());
+
+            let page = kuchikiki::parse_html().one(response.text().await?);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

GuillaumeGomez:
Just realized: can you add a test to ensure that:

1. We generate the right link
2. The linked page is actually the expected one

please?

Cerber-Ursi:
I was hoping to avoid installing the whole thing for UI testing, to be honest :) But this request is reasonable (after all, this change now is not checked automatically at all), so I'll look into it.

Cerber-Ursi:
Actually, thinking about it, I'm not totally sure how this should be approached. To test pagination, we need, well, pages, that is, we need a large enough amount of failed builds. Will it be normal to add some known-bad builds to `run-gui-tests.sh`?

And by the way, how it is ever expected to be launched locally? As-is, that is, as described in README, it fails saying that `DOCSRS_PREFIX is missing` when doing `cargo run -- database migrate`, and I guess this is expected, since this part of the script doesn't use `.env`-file, it explicitly moves it away.

Cerber-Ursi:
Aha, thanks - this appears to be easier then I thought. Updated the source branch, hopefully this is OK.

GuillaumeGomez (approved):
Awesome work, thanks a lot!

Cerber-Ursi:
Forgot about formatting, now updated.

GuillaumeGomez (approved):

GuillaumeGomez:
All good, thanks!

Issue #2787, opened 2025-04-05
Source view has many scrollbars


Alexendoo:
e.g. on https://docs.rs/crate/cargo/latest/source/Cargo.toml.orig it's gathering quite the collection

![Image](https://github.com/user-attachments/assets/d9a200f6-6073-47ab-a49c-682d6271eac1)

Alexendoo:
Occurs on Chrome & Firefox. If you're on Firefox Linux you may need to enable the "always show scrollbars" setting as it can default to overlay scrollbars. I would guess it's similar on macOS

Alexendoo:
It's from setting `overflow: scroll` on `#source-code-container` and `#source-code`, the actual scrolling is happening in the `code` element so removing these should be fine. In browsers where the scrollbar takes up room in the document (is not an overlay) `scroll` forces them to be visible

https://github.com/rust-lang/docs.rs/blob/47a493fbca82f65cbdc5809a72b179f6872cdfa8/templates/style/style.scss#L915-L918

https://github.com/rust-lang/docs.rs/blob/47a493fbca82f65cbdc5809a72b179f6872cdfa8/templates/style/style.scss#L943-L946

GuillaumeGomez:
Wanna send a fix then maybe? You already seem to have the solution. 😉

Alexendoo:
I looked at https://github.com/rust-lang/docs.rs?tab=readme-ov-file#getting-started and went with a comment instead, but actually yeah since it works in the inspector I don't need to build a local version

PR #2792, opened 2025-04-07
Fix source code UI bugs

GuillaumeGomez:
Two of the bugs can be seen here:

Screenshot From 2025-04-07 16-44-09

The last one is that the cursor was not the right one when hovering the toggle button, it's now fixed too.

I think it might help with https://github.com/rust-lang/docs.rs/issues/2787 but since I can't reproduce the bug, not sure...

Looks this now:

image


GuillaumeGomez:
And CI passed. :D

syphar (approved):
thank you for the fix!

PR #2795, opened 2025-04-09
Remove `csp_nonce` field from all template types

GuillaumeGomez:
With the new version of `askama` came the runtime values, which allows to remove a loooot of fields. :D


syphar (approved):
super nice!

PR #2857, opened 2025-07-02
Move about page tabs system in header dropdown

GuillaumeGomez:
It now looks like this:

Screenshot From 2025-07-02 15-27-56

It greatly simplifies the about pages and it also directly gives access to all pages without needing to go to the about pages index page.

### DIFF
--- a/src/web/build_details.rs
+++ b/src/web/build_details.rs
@@ -8,7 +8,7 @@ use crate::{
         extractors::{DbConnection, Path},
         file::File,
         filters,
-        page::templates::{RenderRegular, RenderSolid},
+        page::templates::{RenderBrands, RenderRegular, RenderSolid},
     },
 };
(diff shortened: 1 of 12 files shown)


syphar (commented):
can we horizontally align the text in the dropdown menu?

GuillaumeGomez:
Even simpler, added an icon:

image

syphar:
> Even simpler, added an icon:
> 
> image

I meant that I expected the _text_ to be aligned too. no matter the icon size

GuillaumeGomez:
Ah I see. Done:

image

GuillaumeGomez:
> testing this I also saw there are some discrepancies where the queue is full width, while other release-lists are not.

Ah? Do you have a screenshot? Might be simpler to open an issue and ping me on it. :)

PR #2883, opened 2025-08-07
Open a PR if SQLX files need an update

GuillaumeGomez:
Apparently we can open a PR on some PR events without a token so let's try it. :)


syphar (approved):
let's try 😅

GuillaumeGomez:
Let's see how it goes.

PR #2892, opened 2025-08-12
Add github actions yml files checker

GuillaumeGomez:
Considering for now all updates I made to the sqlx worflow file failed, it would be nice to be able to actually check it ahead of time.


GuillaumeGomez:
Nice, gonna merge this one then let's send the fix for the sqlx CI. :)

PR #2893, opened 2025-08-12
Fix sqlx workflow syntax

GuillaumeGomez:


GuillaumeGomez:
At least now the workflow syntax is valid. :)

PR #3071, opened 2025-12-07
(DRAFT) ci: try freeing some disk space for gui tests

syphar:
alternative approaches that would take more time: 
- use the docker images from the docker CI jobs instead of building our own 
- figure out where exactly the disk space is used


syphar:
@GuillaumeGomez I think this will help :)

GuillaumeGomez:
CI passed so should be good. Thanks!

PR #3076, opened 2025-12-11
ugprade to rust 1.92

syphar:


syphar:
it seems like the docker images aren't ready yet :)

PR #3077, opened 2025-12-11
add crate-name validation to web endpoints

syphar:
for now, copy/paste the [subcrate we want to have published](https://github.com/rust-lang/crates.io/pull/12435) for clarity. 

Then this PR migrates most web handlers working with crate names to use the newtype first. This makes it possible to decline a bunch of requests without any database query. 

Only handler where we needed a little more work was the `rustdoc_redirector_handler` where we have to handle some non-crate cases before we can validate. 

I intentionally limited the migration to just a first block, otherwise the new types leak into the whole codebase. 
And I want to do the migration step by step.


GuillaumeGomez:
Normally you don't need to push sqlx files, our CI will handle it. ;)

crates/crates_io_validation/README.md
```diff
@@ -0,0 +1,6 @@
+# crates_io_validation
+
+This package contains some low-level validation functions that are used across
+the code base.
+
+These are mostly in their own crate to be shared with docs.rs.
```
GuillaumeGomez, line 6: Is the plan to make it used by crates.io?

GuillaumeGomez: (and cargo eventually?)

syphar: > Is the plan to make it used by crates.io?

yes, plan is let them have come iterations on the design, and then we publish the crate. 

> (and cargo eventually?)

Cargo accepts more things because of third party registries. So sharing logic between crates.io & docs.rs will happen more, but not sure how much forther that can gop

PR #3079, opened 2025-12-12
DRAFT/POC: proposal for splitting docs.rs into sub-crates / a workspace

syphar:
This is a proposal how I could imagine dosc.rs split into workspaces. ( please ignore the commit messages :) )

I would love to get your feedback @GuillaumeGomez beforeI spend more time on this. Perhaps better to dig into this locally to get a feeling about the structure. 

Compile time wise it's worlds, mostly when you're only working on one sub-crates. In any case you won't ever have to compile gix/crates-index/diff etc when you're just working on the builder or webserver. Also working on the watcher you'll not have to pay compile time for static assets ETag generation. 

There is still optimization potential where some libraries could work with less dependencies

### what's still to be done here: 
- [ ] test release / sqlx migrations
- [ ] follow up & integrate the last changes from the main branch 
- [ ] fix consistency check
- [ ] re-add CLI commands (not sure if all), some things perhaps into a separate CLI bin-crate, some things perhaps into the builder / watcher binaries. 
- [ ] uncomment & fix tests, probably with a test helper library. 
- [ ] i have an idea about a better split between the build-queue, and the builders 
- [ ] ( if we didn't migrate to ECS until then), perhaps create an additional legacy `cratesfyi` binary that acts as the old one.

### DIFF
--- a/crates/bin/docs_rs_builder/src/utils/queue_builder.rs
+++ b/crates/bin/docs_rs_builder/src/utils/queue_builder.rs
@@ -1,10 +1,8 @@
 use crate::config::Config;
 use crate::docbuilder::rustwide_builder::{PackageKind, RustwideBuilder};
 use anyhow::{Context as _, Result};
-use docs_rs_build_queue::BuildQueue;
 use docs_rs_context::Context;
 use docs_rs_database::types::krate_name::KrateName;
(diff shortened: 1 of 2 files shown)


syphar:
@GuillaumeGomez first design feedback? :) 

( when you have time, not time critical )

PR #3086, opened 2025-12-20
extract s3 storage into subcrate

syphar:
most notable in here: **i decided to drop the database-storage**. 

IMO we can just use an in-memory storage for unit-tests, and people locally testing should use minio. 

Generally I feel the cut between subcrates will evolve when I extract more, especially around testing. 

There are some other improvements to the storage I want to do (like streaming uploads), but I'll do them separately.


GuillaumeGomez:
Thanks!

PR #3095, opened 2025-12-23
fix clippy errors

syphar:


crates/lib/docs_rs_cargo_metadata/src/metadata.rs
```diff
@@ -144,16 +144,19 @@ struct DeserializedMetadata {
 #[derive(Deserialize)]
 struct DeserializedResolve {
     root: String,
+    #[allow(dead_code)]
```
GuillaumeGomez: If unused, why keeping them?

syphar: valid, removed.

GuillaumeGomez:
Thanks!

PR #3096, opened 2025-12-23
move shared crate-details functions into database crate

syphar:
will be needed by the watcher & the builder binaries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2003,8 +2003,10 @@ name = "docs_rs_database"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "docs_rs_env_vars",
  "docs_rs_opentelemetry",
+ "docs_rs_types",
  "futures-util",
  "hex",
  "opentelemetry",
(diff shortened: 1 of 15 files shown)


crates/lib/docs_rs_cargo_metadata/src/metadata.rs
```diff
@@ -144,16 +144,19 @@ struct DeserializedMetadata {
 #[derive(Deserialize)]
 struct DeserializedResolve {
     root: String,
+    #[allow(dead_code)]
```
GuillaumeGomez: If it's unused, can't they be removed?

syphar: valid, removed in https://github.com/rust-lang/docs.rs/pull/3095, 

rebases this PR on https://github.com/rust-lang/docs.rs/pull/3095

GuillaumeGomez:
I think the conflict comes from the same code removed in the same location, maybe wrong rebase? ;)

PR #3097, opened 2025-12-23
Update to new `askama` 0.15 version

GuillaumeGomez:
New 0.15 version: https://github.com/askama-rs/askama/releases/tag/v0.15.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,11 +317,11 @@ dependencies = [
 
 [[package]]
 name = "askama"
-version = "0.14.0"
+version = "0.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f75363874b771be265f4ffe307ca705ef6f3baa19011c149da8674a87f1b75c4"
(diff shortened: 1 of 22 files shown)


syphar (approved):

syphar:
@GuillaumeGomez there is a new error when running tests, after this PR: 

```
⚠️  templates/releases/releases.html:18:4: block `topbar` was already called at `templates/releases/search_results.html:8:4` so the previous one will be ignor
e
⚠️  templates/releases/releases.html:6:4: block `header` was already called at `templates/releases/search_results.html:4:4` so the previous one will be ignore
d   templates/releases/releases.html:18:4: block `topbar` was already called at `templates/releases/search_results.html:8:4` so the previous one will be ignor
e
⚠️  templates/releases/releases.html:6:4: block `header` was already called at `templates/releases/search_results.html:4:4` so the previous one will be ignore
```

GuillaumeGomez:
Not errors, warnings. It's "ok", but we might want to rewrite the templates a bit to not have them anymore. We just abuse some things. :3

PR #3102, opened 2025-12-25
add security middleware to block malicious paths

syphar:
There were a bunch of sentry errors from someone trying to break in into our system via relativ paths, or invalid UTF-8. 

Instead of trying to handle these errors in the whole codebase, we just fully block these URLs. 

I also thought about using the CDN to block these, but then decided to have it here for consistency. 

If the middleware gets too big, we could think about this: 
https://www.fastly.com/documentation/guides/next-gen-waf/using-ngwaf/signals/using-system-signals/

related but not needed: 
https://github.com/rust-lang/docs.rs/pull/3101

### DIFF
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -17,6 +17,7 @@ pub mod page;
 mod releases;
 mod routes;
 pub(crate) mod rustdoc;
+mod security;
 mod sitemap;
 mod source;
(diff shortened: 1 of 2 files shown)


syphar:
Will check the test failure later

GuillaumeGomez (approved):

PR #3103, opened 2025-12-25
remove heapless dependency from postcard

syphar:
removes ~7 dependencies, also solves #3099


GuillaumeGomez (approved):

GuillaumeGomez:
I like this end of the year gift. =D

PR #3108, opened 2025-12-29
Update `askama` to `0.15.1`

GuillaumeGomez:


GuillaumeGomez:
The `let mut value` thing is just thanks to a fix I wrote for `0.15.1` which keeps arguments mutability info.

PR #3113, opened 2026-01-03
extract webserver into subcrate

syphar:
This will be the biggest PR of the whole bunch. A good part of the filecount are the moved, unchanged assets.

There is a new approach around the global app context I'm trying here, will adapt the other binaries too when finished. 

Next steps are: 
- docs.rs-cli stuff into subcrate (migrations, some of the small commands to update priorities, ....)
- perhaps the legacy binary into a subcrate too, at least clean up


syphar:
ping @GuillaumeGomez only in case you didn't see it. 

( take your time if you need it :) )

GuillaumeGomez (approved):

GuillaumeGomez:
Nah, just waiting for you to be done with changes. :)

PR #3120, opened 2026-01-07
extract admin-CLI & legacy cratesfyi binary into subcrate

syphar:
last 2 extractions. Keeps the legacy binary just for the daemon, until we have the new infra, or I migrated the old server to the separate binaries or even the docker images. 

Next steps after this: 
- update the readme when / where needed 
- look into reducing dependencies in some cases, when crates depend on binary crates.


syphar:
@GuillaumeGomez ready for review now.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3125, opened 2026-01-09
Introduce new TestEnvironment usable by all binary crates

syphar:
- migrate the test environment structs for all binary crates into a single one 
- library crates keep (for now) their smaller custom test envs 
- introduce `AppConfig` trait (needed for the shared test env), migrate all config structs to use it

smaller design changes: 
- `TestEnvironment` can `Deref` into `Context`, so some duplicated methods go away 
- some attributes on the `TestEnvironment` are now private, people need to use the deref or the accessor methods.

not in here, but I want to do it separately: 
- more config builders, currently test config building is all over the place 
- more docstrings  / design explanation


syphar:
@GuillaumeGomez next refactor / cleanup :)

PR #3134, opened 2026-01-12
show average build durations in build lists, crate details

syphar:
This shows the build duration in the build-list for a release, and an average duration (this release or all release) in the crate-details page. 

I've put logic into database views for some of the future things I want to do with this. 

grafik

grafik

note: there were some html errors in `crate/builds.html` I just fixed in this PR. 

### next things after this PR

My idea is, if it works performance wise, to add a column to the build queue where you see an estimated build-time for each entry. Logic would be to just start at the first entry, current time, and add up the historical averages for the crate to estimate the build time for each new release.

We could also show some ecosystem wide statistics to users, or crate-level graphs to the crate authors (like, historical build durations for your crate).

### DIFF
--- a/crates/bin/docs_rs_web/Cargo.toml
+++ b/crates/bin/docs_rs_web/Cargo.toml
@@ -48,7 +48,6 @@ font-awesome-as-a-crate = { path = "../../lib/font-awesome-as-a-crate" }
 futures-util = { workspace = true }
 getrandom = "0.3.1"
 http = { workspace = true }
-humantime = "2.3.0"
 itertools = { workspace = true }
 lol_html = "2.0.0"
 mime = { workspace = true }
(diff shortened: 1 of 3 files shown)


crates/bin/docs_rs_web/Cargo.toml
```diff
@@ -48,6 +48,7 @@ font-awesome-as-a-crate = { path = "../../lib/font-awesome-as-a-crate" }
 futures-util = { workspace = true }
 getrandom = "0.3.1"
 http = { workspace = true }
+humantime = "2.3.0"
```
GuillaumeGomez: Do we really need a new crate for that? ^^'

syphar: you're right, I'm always torn between not duplicating functions that are already built by other people, and reducing dependency count 😅 

I dropped it and replaced it with a small custom impl

GuillaumeGomez:
Thanks! :)

PR #3137, opened 2026-01-13
fix static file serving in local dev, fix edge case in prod

syphar:
fixes a production edge case that would have lead to wrong cache headers when someone locally deletes files from the `static` or `vendor` directories that were there at built-time. 

Also we're fixing a bug when you were locally running the docs.rs webserver from the repository-root. In this case the static assets weren't found, because they don't live in the repo root any more. We fallback to `CARGO_MANIFEST_DIR` now. 

I added a unit-test for the wrong cache headers in this edge-case, and a first "real" integration test at the repo root that just so happens to test the static asset fetching in development. More integration tests will be added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,19 +66,19 @@ docker compose up --wait db s3
 # allow downloads from the s3 container to support the /crate/.../download endpoint
 mcli policy set download docsrs/rust-docs-rs
 # Setup the database you just created
-cargo run -- database migrate
+cargo run --bin docs_rs_admin -- database migrate
 # Update the currently used toolchain to the latest nightly
 # This also sets up the docs.rs build environment.
 # This will take a while the first time but will be cached afterwards.
-cargo run -- build update-toolchain
+cargo run --bin docs_rs_builder -- build update-toolchain
 # Build a sample crate to make sure it works
-cargo run -- build crate regex 1.3.1
+cargo run --bin docs_rs_builder -- build crate regex 1.3.1
 # This starts the web server but does not build any crates.
 # It does not automatically run the migrations, so you need to do that manually (see above).
-cargo run -- start-web-server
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Write a unit test which starts the server from top level and query a static resource?

syphar:
> Write a unit test which starts the server from top level and query a static resource?

now after having slept, I can see it. 

starting a test from the top level would either mean re-introducing the top-level crate, 

~~but perhaps I can change the CWD for a single test-case~~

GuillaumeGomez:
Yep, changing the cwd was what I had in mind.

syphar:
> Yep, changing the cwd was what I had in mind.

but cwd is always per-process, so with parallel test execution you might break stuff, or we have to set threads=1. 

But I have another, cleaner idea. 

I'll inject the root dir into the router builder, and write a test on a "broken" router.

GuillaumeGomez:
No I mean, when you use `Command`, you can give it a cwd. That's what I had in mind.

syphar:
@GuillaumeGomez added some tests, I think this is the better approach.

GuillaumeGomez:
Does that mean it works wherever we start the server?

syphar:
> Does that mean it works wherever we start the server?

in a dev project, yes. 

in the built binary, or docker container, no.

the fallback is `CARGO_MANIFEST_DIR`, which is the path on the developer machine making the build. 

So it might work whereever you try in this repo :

GuillaumeGomez:
Awesome! Then please update docs as well then I'll approve. :)

syphar:
@GuillaumeGomez like this?

GuillaumeGomez:
Yes exactly, thanks!

PR #3148, opened 2026-01-16
remove unnecessary wrapper around storage method

syphar:
not sure how I missed that so long.


GuillaumeGomez (approved):

GuillaumeGomez:
Less code for same results? That makes me very happy, well done!

PR #3152, opened 2026-01-20
add CLI command to repackage file-storage to archive-storage

syphar:
there are **362_872** releases out of **1_913_132** left that still store single files on the S3 bucket, and not a zip file.

Since rebuilding would take quite some time, and also wouldn't be guarenteed to work for old releases, I think the best idea is to repackage the files. 

So 
- download all files 
- create & upload the ZIP 
- switch the archive-storage marker 
- delete the single files

I'll try the migration one-by-one first and manually check if there are any issues. 

After all old releases are migrated, we can drop a whole lot of code related to the old storage, G
also we'll be able to provide ZIP downloads for these old releases too.


GuillaumeGomez:
Please add instructions about it (in a README I suppose).

syphar:
> Please add instructions about it (in a README I suppose).

For whom? 

This a temporary command, and I'll drop it when the run is finished, probably in a couple of weeks. 

Or am I missing something?

GuillaumeGomez:
No, *I* was missing something. Nevermind then! Thanks. :)

syphar:
> No, _I_ was missing something. Nevermind then! Thanks. :)

re-reading the PR i see I forgot to mention that in the PR descrition. 

" temporary" is only mentioned in the CLI structs / docstring

PR #3153, opened 2026-01-21
Remove dependencies

GuillaumeGomez:
The first commit removes an unused dependency. The other commits though "inline" the dependency when it was only for very minor things.  It's all up-to-debate (well except the first commit :smile:). :)


GuillaumeGomez:
I think there isn't much more to be done after this PR on this track. Might have missed some things but at least all the obvious ones were removed.

PR #3155, opened 2026-01-21
fix build performance tracing

syphar:
I wanted to look into performance traces for builds, and discovered they are all over the place. 

Mainly: the main `docbuilder.build_package` trace is not connected to the child traces. 

I'm not 100% sure if these changes solve it completely, but I definitely discovered these issues, and they are fixed by the PR: 

- we had logic in place that would force _all_ builds to be traced, and only applies the sampling ratio to web requests. **This didn't work**, at least since the multi-crate migration. I validated what's in `ctx.operation()` and `ctx.name()` and fixed the filter. 
- we have our custom `spawn_blocking` function that takes care of passing on the parent span into the task because tokio doesn't. I realized that `block_on` has the same issue. For now I just used the new fixed handle for builds. If it solves the issues there, I'll move the `spawn_blocking` fn into the handler too, and update the whole codebase.

### DIFF
--- a/crates/lib/docs_rs_utils/src/runtime_ext.rs
+++ b/crates/lib/docs_rs_utils/src/runtime_ext.rs
@@ -1,0 +1,27 @@
+use std::ops::Deref;
+use tokio::runtime;
+use tracing::Instrument as _;
+
+/// Newtype around `tokio::runtime::Handle` that adds
+/// missing integration with tracing spans.
+pub struct Handle(runtime::Handle);
+
+impl Handle {
+    pub fn block_on<F: Future>(&self, future: F) -> F::Output {
+        runtime::Handle::block_on(self, future.in_current_span())
+    }
+}
+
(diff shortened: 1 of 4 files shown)


crates/lib/docs_rs_utils/src/runtime_ext.rs
```diff
@@ -0,0 +1,27 @@
...
+/// missing integration with tracing spans.
+pub struct Handle(runtime::Handle);
+
+impl Handle {
+    pub fn block_on<F: Future>(&self, future: F) -> F::Output {
+        runtime::Handle::block_on(self, future.in_current_span())
+    }
+}
+
+impl From<runtime::Handle> for Handle {
```
GuillaumeGomez, line 15: Thanks for the by-hand writing for both trait impls. :)

I wonder if we could turn `Deref` into an std proc-macro.

PR #3158, opened 2026-01-22
storage repackage: parallel downloads, more logging

syphar:
trying the first conversions worked (yay!) but felt a little slow. 

So we can: 
- parallelize downloads 
- emit some logs to see where the remaining time is spent.


GuillaumeGomez:
Nice, well done. :)

PR #3160, opened 2026-01-23
block requests with `#` in the path (likely XSS probes)

syphar:
fixes: 
- https://rust-lang.sentry.io/issues/7195650500/?environment=production&project=5499376&query=is%3Aunresolved&referrer=issue-stream
- https://rust-lang.sentry.io/issues/7199830736/?environment=production&project=5499376&query=is%3Aunresolved&referrer=issue-stream

decoded path:  `/minidumper/latest/#</script><teste>`

### DIFF
--- a/crates/bin/docs_rs_web/src/middleware/security.rs
+++ b/crates/bin/docs_rs_web/src/middleware/security.rs
@@ -25,6 +25,12 @@ pub(crate) async fn security_middleware(
         return StatusCode::NOT_ACCEPTABLE.into_response();
     }
 
+    // `#` is never allowed in any rustdoc URLs, even encoded
+    if path.contains('#') {
+        warn!(%uri, "detected `#` in server-side request path");
+        return StatusCode::NOT_ACCEPTABLE.into_response();
+    }
+
     next.run(req).await
 }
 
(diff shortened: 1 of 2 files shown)


crates/bin/docs_rs_web/src/middleware/security.rs
```diff
@@ -25,6 +25,12 @@ pub(crate) async fn security_middleware(
         return StatusCode::NOT_ACCEPTABLE.into_response();
     }
 
+    // `#` is never allowed in any rustdoc URLs, even encoded
+    if path.contains('#') {
```
GuillaumeGomez, line 29: Just to be sure, that won't block anchors right? Like:

```
/whatever/#main-content
```

In here `main-content` is an actual element on the page and is used as an anchor to scroll to it.

syphar: 1. the browser never sends anchors to the server 
2. even if someone sends it (plain-text), axum will filter it out before handing the rest of the path to us 
3. so we only handle the cases where someone uses url-encoding to _hide_ the anchor, hoping we re-add it to a redirect URL for the user.

GuillaumeGomez: Perfect, thanks for the explanations!

PR #3162, opened 2026-01-23
move builder build_package performance transaction parent, add spans 

syphar:
So, I really hope that this fixes the issue. 

Generally: 
- I moved the transaction parent up one level, so everything we do for one build is in that transaction. Missing previously: toolchain updates, builder docker image download / update
- I used the new block_on with tracing in the build queue 
- After having slept I now feel like `Deref` is wrong for this use-case, there are cases where people would accidentally call the wrong method. 
- I fixed the default log-directive (though this directive is overwritten anyways on the server). 

This time I locally ran the old build-server (via daemon), and new (via builder-binary), and in both cases I only saw the parent-transaction I want, and never the parent-less child spans we see in prod.


syphar:
@GuillaumeGomez this can be reviewed now

PR #3178, opened 2026-02-05
Don't keep old failed builds in the build queue

syphar:
We have two main kinds of build-errors: 
1. "standard" rustc build errors, which always ended up in the proper tables 
2. "errors before the build even starts" (like, a broken cargo.toml file, fails when fetching dependencies). 

Since we also add in-progress builds to the database, we can see both error types in the proper interfaces. 

But (My guess) from the times before that, we wanted _some_ record of the failed build. And these were these queue entries where we tried more than X times. 

This leads to around ~5k entries in the queue table of old failed builds. 

I don't think we need these any more and can keep a clean queue table without any issues. 

I also fixed my `TestMetrics` / `CollectedMetrics` testing stub, which returned the wrong values when we queried the values multiple times.


syphar:
not ready for review yet, will fix the test first

syphar:
@GuillaumeGomez ready for review now

GuillaumeGomez:
Looks good to me, thanks!

PR #3179, opened 2026-02-05
web/statics, replace phf with sorted array & binary_search

syphar:
After looking into this deeper for another thing, I realized that using phf here is a little overengineered, and we can totally just use binary-search here. Disk i/o will always eat the slight advantage, as minimal as it might be, and we can save some compile time.


syphar:
checking the test failures, locally they passed

GuillaumeGomez:
Thanks!

PR #3181, opened 2026-02-06
use workspace.package attributes where possible

syphar:
not sure how I missed this initially :)


GuillaumeGomez:
Never realized we could do that. :o

PR #3182, opened 2026-02-06
fix auth header format in github repository update

syphar:
Looking at the intro here: https://docs.github.com/en/graphql/guides/introduction-to-graphql

I think this was the issue with the github repo updater


GuillaumeGomez:
Thanks!

PR #3185, opened 2026-02-08
add missing index in repositories table, order updates by oldest first

syphar:
- I saw the index was missing but the column was queried. With the exception of small tables, all columns used in `WHERE` statements should have an index. 
- I want to update the repositories first that were updated the earliest. 

current state from prod: 
```
cratesfyi=> select min(updated_at), max(updated_at) from repositories;
              min              |              max
-------------------------------+-------------------------------
 2026-02-04 16:28:14.709552+00 | 2026-02-08 04:24:43.370584+00
(1 row)
```

( This is better than yesterday when I checked, so I think we're still making enough progress with updates, despite the rate limiting)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3187, opened 2026-02-08
add dev-util to import release-docs from production for local testing

syphar:
I'm currently not sure yet if this is worth the LOC, 

but since building the builder & running builds is quite slow on my machine, this helps me to quickly have valid docs in a previously empty database. ( in my personal dev workflow containers / images / volumes are regularly deleted)

You can run it by just `cd` into the directory of the binary crate, and then just 
```
cargo run -- regex 
# or 
cargo run -- regex '~1' 
```

( the version identifier understands the same things as docs.rs semver itself). 

For now it only works for successful builds, and it likely will fail for older versions. 

@GuillaumeGomez what do you think? Does it work for you? I remember you also thought about something similar, but simpler. 

I didn't want to add it to the `admin` cli, since it's not for the server, but only for local usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ cargo run --bin docs_rs_builder -- build update-toolchain
 # Build a sample crate to make sure it works
 cargo run --bin docs_rs_builder -- build crate regex 1.3.1
 # if you don't want to run the builder, you can import a release from docs.rs itself
-cargo run --bin docs_rs_import_release -- regex latest
+cargo run -p docs_rs_import_release -- regex latest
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Sounds like a good idea. Please add docs.

syphar:
> Sounds like a good idea. Please add docs.

@GuillaumeGomez I now mention it in the readme. 

Did the command work for you too?

GuillaumeGomez:
Yup, all good for me now, thanks!

PR #3188, opened 2026-02-08
store crate-count on repositories, deprioritize bigger workspaces

syphar:
- store the crate-count on the `repositories` table. Updated via CLI, or after each build when we finish the release. 
- after we processed all changes in the index watcher, we to a bulk activity where we deprioritize the workspaces we find. 

caveats: 
- this will only work when we actually find the repo in github/gitlab etc. Otherwise we don't have a column in the `repositories` table. I didn't want to base everything onto the `releases.repository` column. ( happy to discuss)
- some edge cases depending on where and how you use the build-queue classes. I didn't want the deprioritization logic in `process_version_added`, because the query is nontrivial. 

### deploy checklist 
- after deploy, one of the repo-stats backfill commands have to be run to the point that the crate-counts are filled initially. 
- I can probably also remove some of the hardcoded crate priorities via prefixes.


GuillaumeGomez:
To be noted: would be nice to have the config documented somewhere.

PR #3189, opened 2026-02-08
migrate remaining places to KrateName NewType

syphar:
- removed `Deref` from the type
- I wasn't 100% certain about the package metadata, so I left it as-is.


GuillaumeGomez:
Thanks!

PR #3190, opened 2026-02-09
gracefully handle 404 errors from crates.io

syphar:
otherwise we are getting sentry reports.


syphar:
@GuillaumeGomez now ready

PR #3196, opened 2026-02-09
 more detailed error handling in registry-api, show crates.io errors to user

syphar:
Today I gave the registry api code a little love. 

generally I wanted: 
- retry only when it's worth retrying (= on 5xx & connection errors) 
- parse the crates.io api error responses, and show the message to the user. For now I assume that 4xx responses are json / can be parsed, and 5xx are plain text. 
- use the status code we get from crates.io as our own status code. 
- use a enum as error type, so we can use the specific error information in the search-handler. 
- to move some tests from integration-like (where we tested crates.io api things via the search-handler in `web::handlers::releases`), to more unit-like tests directly on the `RegistryApi` struct. 

In the code I discovered some code that tried to parse error messages out out successful responses. I assume [that this is legacy, and dropped it](https://blog.rust-lang.org/2024/02/06/crates-io-status-codes/). [To be safe, I also asked]([#t-crates-io > api , error responses](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/api.20.2C.20error.20responses/with/572909326)). 

( I didn't put any time in design for the error message, feel free to improve)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3197, opened 2026-02-10
registry-api: gracefully handle releases not found

syphar:
I remember some edge cases when this can happen, one of them is definitely when the release was deleted after it was queued, but before it was built.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3198, opened 2026-02-10
also remove deleted crates/versions from the queue

syphar:
thinking about edge cases in https://github.com/rust-lang/docs.rs/pull/3197 I actually thought I already did this change. 

It seemed like not.


GuillaumeGomez:
Thanks!

PR #3200, opened 2026-02-11
move build profiles into workspace cargo.toml

syphar:
I wasn't aware where the right place is :) 

Checking the output of `cargo build -vv`, the profiles aren't used when defined in `.cargo/config.toml`, but _are_ used when defined in the workspace `Cargo.toml`


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3201, opened 2026-02-11
store rustwide command-error in build-table

syphar:
References #878 

mainly: store the actual rustwide `CommandError` kind in our build-result table. Through that we can directly see timeouts, OOM, ... 
At some point also have statistics. 

I want to store it for a bit before I show it anywhere.


GuillaumeGomez:
Thanks!

PR #3203, opened 2026-02-12
move log config to struct, support disabling stdout for builds

syphar:
this is triggered by @marcoieni asking us for logging less to syslog for cost reasons. 

Especially the build logs inside `rustwide::logging::capture` are also stored on S3, so we don't need to additionally log them to stdout. 

While I'm on it, move the logging config to a config struct.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3204, opened 2026-02-13
add workspace lints, forbid dbg! usage, remove current usages

syphar:
after not having the build-logs in the cratesfyi log stream I saw that I actually left a `dbg!` macro call in one of the main web handlers... 

Of course this contributes quite much to our currently high log usage. 

To be very save, this adds config for workspace lints, and sets `warn` for `dbg!` and `todo!` (we have `-D warnings` in `just lint`)

I also removed them in tests for now, [while there is `allow-dbg-in-tests`](https://doc.rust-lang.org/stable/clippy/lint_configuration.html#allow-dbg-in-tests), I couldn't quickly figure out how to best have a workspace-level `clippy.toml`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,10 @@ url = { version = "2.1.1", features = ["serde"] }
 walkdir = "2"
 zip = { version = "7.0.0", default-features = false, features = ["bzip2"] }
 
+[workspace.lints.clippy]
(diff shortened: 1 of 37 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3205, opened 2026-02-14
update build-time metric to be more detailed

syphar:
- smaller buckets in the lower end of build times (and less for longer times) 
- attribute so we can differentiate between success/failure/error


GuillaumeGomez:
Thanks!

PR #3218, opened 2026-03-01
Handle JSON generation like other errors instead of bailing out

GuillaumeGomez:
I tested one of the failed crate on docs.rs and couldn't reproduce the bug locally, so I assume the build could be retried and therefore this error should be handled like the others.


syphar:
Also the tests fail for proc-macro crates, remembering now that [there was an edge case about json&proc-macros](https://rust-lang.sentry.io/issues/7178942819/?environment=production&project=5499376&query=is%3Aunresolved&referrer=issue-stream) I don't remember?

GuillaumeGomez:
Let's keep things as is then. I saw an issue in sentry listing this but maybe I misunderstood the problem. Anyway, closing. :)

PR #3221, opened 2026-03-04
chore: update deployment workflows and add ECS action

marcoieni:
Update the deployment workflows by using latest best practices:
- Use GitHub OIDC vs `aws_secret_access_key` (after this PR is merged, we can delete the AWS secrets from the GitHub Actions secrets)
- Use official docker github actions, which we noticed are more robust in github actions wrt just running `docker build`
- Introduce caching to take advantage of cargo chef in CI

Based on https://github.com/rust-lang/bors/blob/main/.github/actions/deploy-ecs/action.yml

GitHub OIDC support was added in https://github.com/rust-lang/simpleinfra/pull/1006

Since the docker containers are not used yet, it is safe to merge this PR and experiment with it 👍

## Note on the github action
In the previous approach you were using a github action managed by the infra team, while now you don't anymore.

If there's a preference from various teams that the infra team creates a new shared github action (or update the existing one) we'll do it. But for now I would prefer to start with a local github action, so that we can iterate faster.

## Other Docker images

Do we need to upload other images to ECR? If yes, we can do it in another PR.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,19 +8,19 @@ jobs:
   prod:
     name: Production
     runs-on: ubuntu-latest
+    environment: production
+    concurrency: production
+    permissions:
+      id-token: write
+      contents: read
     steps:
-      - uses: actions/checkout@v6
-
-      - name: Build the Docker image
-        run: docker build -t docs-rs-web -f dockerfiles/Dockerfile --target web-server .
-
-      - name: Upload the Docker image to ECR (production)
-        uses: rust-lang/simpleinfra/github-actions/upload-docker-image@master
+      - name: Checkout repository
+        uses: actions/checkout@v6
         with:
-          image: docs-rs-web
-          repository: docs-rs-web
-          region: us-west-1
-          aws_access_key_id: "${{ secrets.aws_access_key_id }}"
-          aws_secret_access_key: "${{ secrets.aws_secret_access_key }}"
-          redeploy_ecs_cluster: rust-ecs-prod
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -8,19 +8,16 @@ jobs:
...
-        run: docker build -t docs-rs-web -f dockerfiles/Dockerfile --target web-server .
-
-      - name: Upload the Docker image to ECR (production)
-        uses: rust-lang/simpleinfra/github-actions/upload-docker-image@master
+      - name: Deploy to ECS
+        uses: ./.github/actions/deploy-ecs
         with:
-          image: docs-rs-web
-          repository: docs-rs-web
-          region: us-west-1
```
marcoieni, line 22: the region here was wrong. I checked the [actions tab](https://github.com/rust-lang/docs.rs/actions/workflows/deploy.yml) and this workflow was never triggered.

GuillaumeGomez:
Looks good to me, thanks! Let's wait for @syphar to check as well before merging.

syphar (approved):
as a first version to play around totally fine. 

things for later: 
- deploy a specific branch to the staging cluster for testing 
- build/deploy the watcher, builder
- run a "release" script / task after deploy (database migrations etc)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3221/45578b4c48b10df249fecdf4db38525917dfeeea..0da30f0c936ea0adcd71d7c7032a2535cfab8c0f)*

PR #3226, opened 2026-03-07
storage: convert some blocking i/o to non-blocking

syphar:
When working on something else I discovered that there were places where I used blocking i/o functionality inside async functions. 

It wasn't too big of an issue, since this is only used on the build-servers, so blocking the tokio runtime threads is not a huge problem, but for other things I'm planning it would be good to fix ( streaming / parallel uploads). 

I now validated that all remaining sync i/o is inside `spawn_blocking`. 

For one function (upload a whole local folder) I also now introduced paralellism. 

The change to our own `spawn_blocking` was necessary so we can use it in the same way as the original tokio `spawn_blocking` ( = use the function to spawn a task, that then start running, and join it later).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2356,6 +2356,7 @@ dependencies = [
  "test-case",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-util",
  "tracing",
  "walkdir",
  "zip",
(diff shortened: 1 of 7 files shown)


syphar:
@GuillaumeGomez tests pass now

PR #3239, opened 2026-03-09
Fix syntax highlighting for rust `union` keyword

GuillaumeGomez:
Fixes https://github.com/rust-lang/docs.rs/issues/3175.


GuillaumeGomez:
CI passed. :)

GuillaumeGomez:
Ah right, the regex was expecting a `{`, so if you have generics, it didn't work. Fixed it.

PR #3244, opened 2026-03-13
start caching build-status json

syphar:
Seeing some current request stats (~60 rps in peaks), I thought about this again.

Main users are: 
- crates.io frontend to show links or not. 
- shields.io 

I believe we can cache this the same way we cache docs, release lists, details etc. 

( also i created a PR to crates.io for an optimization https://github.com/rust-lang/crates.io/pull/13174 )

### DIFF
--- a/crates/bin/docs_rs_web/src/handlers/status.rs
+++ b/crates/bin/docs_rs_web/src/handlers/status.rs
@@ -4,27 +4,20 @@ use crate::{
     extractors::{DbConnection, rustdoc::RustdocParams},
     match_release::match_version,
 };
-use axum::{
-    Json, extract::Extension, http::header::ACCESS_CONTROL_ALLOW_ORIGIN, response::IntoResponse,
-};
+use axum::{Json, http::header::ACCESS_CONTROL_ALLOW_ORIGIN, response::IntoResponse};
 
 pub(crate) async fn status_handler(
     params: RustdocParams,
     mut conn: DbConnection,
 ) -> impl IntoResponse {
     (
-        Extension(CachePolicy::NoStoreMustRevalidate),
         [(ACCESS_CONTROL_ALLOW_ORIGIN, "*")],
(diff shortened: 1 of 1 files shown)


syphar:
Though, thinking about it, 

if we can cache or not probably depends on the use-case? 
if people use it to get an actual live-status?

GuillaumeGomez:
Not sure it's a good idea to cache a build-status indeed. ^^'

crates/bin/docs_rs_web/src/handlers/status.rs
```diff
@@ -4,45 +4,42 @@ use crate::{
...
-                })?
-                .into_version();
+                })?;
 
-            let json = Json(serde_json::json!({
-                "version": version.to_string(),
-                "doc_status": rustdoc_status,
-            }));
+            let mut response = Json(serde_json::json!({
+            "version": matched_release.release.version,
```
GuillaumeGomez, line 33: Broken indent btw. :p

syphar: that's rustfmt 😅

syphar:
> Not sure it's a good idea to cache a build-status indeed. ^^'

My thought here is: 
everything that changes the build status in any important way also purges the CDN for this key. 
So, it's the same cache we're using for crate-details, or the `/latest/` path. 

And that takes ~250ms globally in fastly's network. 

So for the currently known use-cases, that's totally fine, since we don't need in-progres status. 

I probably should think through some of the in-progress responses for this, but my hunch is that it's fine, assuming the invalidation. 

high level, after the build is finished, the status only rarely changes, which makes it a good candidate for caching. And prefer the CDN responding in these cases over having to run more webservers ( and databases ) over time.

syphar:
closing this for now. 

We might have to rethink this when the load on the endpoint gets higher. 

( IMO it's totally fine for the current users if the build-status is <30s outdated)

PR #3248, opened 2026-03-15
fix deadlock between in-progress builds & delete-crate/release

syphar:
I'm quite sure this is the reason why we sometimes stop consuming from the index. And it's when we get deletes while we have an in-progress build. 

What happens today (simplified):
- Builder txn: locks queue row (FOR UPDATE SKIP LOCKED) and keeps it locked while build runs.
- Delete txn: locks releases/metadata rows, then tries DELETE FROM queue ... and waits on builder’s queue lock.
- Builder’s build path (different DB txn/conn) may try touching releases and waits on delete txn.
- Now each side waits on the other -> deadlock.

For now the simple solution is to make the delete wait for the build to be finished. I added a configurable timeout. 

I want to postpone the better solution to when we have something event-based between crates.io & docs.rs, where I then could easily NACK some events, and handle them later.

### DIFF
--- a/.sqlx/query-1105d2dff6e153dbc8fce074f6cc37f3cd187b93228f403dc2d1d2bebc37c334.json
+++ b/.sqlx/query-1105d2dff6e153dbc8fce074f6cc37f3cd187b93228f403dc2d1d2bebc37c334.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT id FROM queue WHERE name = $1 AND version = $2 FOR UPDATE",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Int4"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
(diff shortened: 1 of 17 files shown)


syphar:
fixing CI, sec

.sqlx/query-1105d2dff6e153dbc8fce074f6cc37f3cd187b93228f403dc2d1d2bebc37c334.json
GuillaumeGomez, line 1: As a reminder, it's not necessary to commit your sqlx files as we have github actions doing it for us (it's fine if they're changed locally while working on it though :p).

GuillaumeGomez:
Nice, thanks!

.sqlx/query-1105d2dff6e153dbc8fce074f6cc37f3cd187b93228f403dc2d1d2bebc37c334.json
syphar: you're right, I remember something. But it's separate PRs? or new commits to the PR branch?

GuillaumeGomez: Once merged, github actions open a new PR if needed.

PR #3250, opened 2026-03-15
mark in-progress builds as canceled when we restart

syphar:
This solves an annoyance that existed since I added https://github.com/rust-lang/docs.rs/pull/2533. 

When I want to update the docs.rs binaries, and would just run `update-docs.rs`, then we have leftover `in-progress` records in the `builds` table. Our process would just exit. The build itself was fine, since postgres would just unlock the record, and another builder (or this one, restarted) would pick it up again. 

But: the in-progress record would stay in the database. 

This is a small workaround for this. The queue guarantees that these only can be one build for each release, so we can simply mark existing in-progress build records as error. 

[This doesn't solve the issue that we'll still have the old `docker run` process running in the background, but that's for a separate PR.](https://github.com/rust-lang/docs.rs/issues/3251)

Note: **this issue would have become a bigger one with more automatic deployments for the new infra**.


GuillaumeGomez:
Thanks!

PR #3257, opened 2026-03-16
fix: archive index backfill / walk-dir should skip over NotFound errors

syphar:
This can happen when moko evicts entries, while we are backfilling them. 

More likely before https://github.com/rust-lang/docs.rs/pull/3253, but still can happen.


GuillaumeGomez:
Thanks!

PR #3258, opened 2026-03-16
show build duration & estimated length for in-progress builds

syphar:


syphar:
@GuillaumeGomez do you have time to help with the styling ? Otherwise I dig into it a little more with more time

GuillaumeGomez:
Leave it to me. :)

If I don't do it this evening, gonna do it tomorrow.

GuillaumeGomez:
Here's the new look:

Screenshot From 2026-03-16 23-21-07

Since they are built concurrently, I removed the numbers on the left, since the information isn't relevant.

syphar:
> Since they are built concurrently, I removed the numbers on the left, since the information isn't relevant.

There is (kind of) an order, because some builds were started first, so in-progress is kind of FIFO. 

But I'm fine with removing them, the time passed will also give the overview.

syphar:
@GuillaumeGomez thanks! 

I tweaked some small things, now it's ready for review

GuillaumeGomez (approved):

PR #3259, opened 2026-03-17
sitemaps: add "recent" sitemap, use last-build-date instead of release date

syphar:
- in case of rebuilds, we previously didn't update the dates in the sitemap, which means there weren't (quickly) re-crawled. 
- I saw some recommendation to provide a separate "recent" sitemap, which would make google crawl fresh or updated entries earlier.

### DIFF
--- a/crates/bin/docs_rs_web/src/handlers/sitemap.rs
+++ b/crates/bin/docs_rs_web/src/handlers/sitemap.rs
@@ -38,13 +38,7 @@ pub(crate) async fn sitemapindex_handler() -> impl IntoResponse {
     let sitemaps: Vec<char> = ('a'..='z').collect();
     let today = Utc::now().date_naive();
     let recent_sitemaps = (0..RECENT_SITEMAP_DAYS)
-        .map(|days| {
-            today
-                .checked_sub_days(Days::new(days))
-                .expect("underflow when building recent sitemap dates")
-                .format("%F")
-                .to_string()
-        })
+        .map(|days| (today + Days::new(days)).format("%F").to_string())
         .collect();
 
     SitemapIndex {


crates/bin/docs_rs_web/src/handlers/sitemap.rs
```diff
@@ -33,8 +36,21 @@ impl_axum_webpage! {
 
 pub(crate) async fn sitemapindex_handler() -> impl IntoResponse {
     let sitemaps: Vec<char> = ('a'..='z').collect();
+    let today = Utc::now().date_naive();
+    let recent_sitemaps = (0..RECENT_SITEMAP_DAYS)
+        .map(|days| {
+            today
+                .checked_sub_days(Days::new(days))
+                .expect("underflow when building recent sitemap dates")
```
GuillaumeGomez: Instead of panicking, couldn't we return an error?

crates/bin/docs_rs_web/src/handlers/sitemap.rs
```diff
@@ -51,41 +67,49 @@ const SITEMAP_HEADER: &[u8] = br#"<?xml version="1.0" encoding="UTF-8"?>
...
-    mut conn: DbConnection,
-) -> AxumResult<impl IntoResponse> {
-    if letter.len() != 1 {
-        return Err(AxumNope::ResourceNotFound);
-    } else if let Some(ch) = letter.chars().next()
-        && !(ch.is_ascii_lowercase())
-    {
-        return Err(AxumNope::ResourceNotFound);
-    }
+fn render_sitemap_item(item: SitemapItem) -> AxumResult<Bytes> {
```
GuillaumeGomez: Wouldn't it be better to call this function with `SitemapItemXml` directly?

crates/bin/docs_rs_web/src/handlers/sitemap.rs
```diff
@@ -132,115 +141,273 @@ pub(crate) async fn sitemap_handler(
...
+}
+
+pub(crate) async fn sitemap_handler(
+    Path(letter): Path<String>,
+    conn: DbConnection,
+) -> AxumResult<impl IntoResponse> {
+    if letter.len() != 1 {
+        return Err(AxumNope::ResourceNotFound);
+    } else if let Some(ch) = letter.chars().next()
+        && !(ch.is_ascii_lowercase())
```
GuillaumeGomez: I don't think the parens are needed after `!` here, right?

crates/bin/docs_rs_web/src/handlers/sitemap.rs
syphar: not sure if worth it? 

[this can only happen when `checked_sub_days` leads to an invalid date](https://docs.rs/chrono/latest/chrono/struct.NaiveDate.html#method.checked_sub_days). since we're subtracting a fixed number (1..=6) days from today, we're safe.

GuillaumeGomez: Then no need for the `checked_` variant I suppose? :)

crates/bin/docs_rs_web/src/handlers/sitemap.rs
syphar: I'll have another look how this could be improved

syphar:
this is ready for the next review, thanks ! @GuillaumeGomez

crates/bin/docs_rs_web/src/handlers/sitemap.rs
```diff
@@ -12,118 +12,105 @@ use axum::{
...
+    SitemapIndex,
     content_type = "application/xml",
 }
 
 pub(crate) async fn sitemapindex_handler() -> impl IntoResponse {
     let sitemaps: Vec<char> = ('a'..='z').collect();
-
-    SitemapIndexXml { sitemaps }
+    let today = Utc::now().date_naive();
+    let recent_sitemaps = (0..RECENT_SITEMAP_DAYS)
```
GuillaumeGomez, line 40: We could store the iterator instead of creating a `Vec` but can be done in a follow-up. Or we could directly do that in the template.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3284, opened 2026-04-07
override build priorities for some workspaces

### LINKED ISSUE #3276: Tokio 1.51.0 not built after 24 hours
Hey, I noticed that Tokio's latest release is still pending after 24 hours. I was wondering if it's possible to bump its priority?

syphar:
Fixes #3276 

the current automatic workspace deprio groups crates by repository, counts the crates, and will trigger deprioritization when there are more than 20 crates in that repo. 

This is what we want for (I assume) most workspaces, which are released together most of the time. These clog the queue and we want them on prio -1. 

But there are exceptions we learned about in the recent "full queue" event: [`tokio`](https://github.com/tokio-rs/tokio), or [`rustcrypto/utils`](https://github.com/RustCrypto/utils), where we have many crates in one repo, but they aren't released together, but only separately. So these we don't need to deprioritize, and in the situations where the queue is full, this is rather annoying for the maintainers. 

So this PR will enable us to override the priority for a repository / workspace, for the current cases just to "revert it back" to high priority. I can also use that feature to deprioritize workspaces _further down_. We're doing this for `swc_%` at the moment, I think it was because it's a huge workspace and wasn't even in production, or self-hosts their docs. I could imagine we can even drop the old build-priorities thing and instead just use this.

### DIFF
--- a/.sqlx/query-1fcb4371909a27b059e3dba7c0791b40f873618036c55c679e5854dbd4cbf91e.json
+++ b/.sqlx/query-1fcb4371909a27b059e3dba7c0791b40f873618036c55c679e5854dbd4cbf91e.json
@@ -1,0 +1,15 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n                    UPDATE queue\n                    SET priority = $2\n                    WHERE name = $1\n                    ",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Int4"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "1fcb4371909a27b059e3dba7c0791b40f873618036c55c679e5854dbd4cbf91e"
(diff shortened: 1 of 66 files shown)


syphar:
fyi @Darksonn @tarcieri 

With this I can bump your workspaces back up so they are not getting sniped down.

GuillaumeGomez (changes requested):
Please add docs. :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3284/1c19f2212991786e781a3c56390dd7243c1de2b5..d538635faa46cf77887b2c6477091ff9eccddd29)*

syphar:
done

GuillaumeGomez (approved):

GuillaumeGomez:
Awesome, thanks!

PR #3285, opened 2026-04-08
builder: support cpu-core assignment for builders

syphar:
Using https://github.com/rust-lang/rustwide/pull/120. 

My current assumption is that the quota-based cpu-limit makes our builds slower than they should be. This new feature allows us to assign the builder so specific cores. When we assign cores, docker won't stop and build because of the quota on one of the "virtual" cores. The rest should be handled by the OS scheduler.

Caveat: 
webserver & watcher are not limited to the other (= non-builder) cores. I assume that the OS scheduler will put them onto more free cores when the build-cores are full. Watcher has higher CPU usage when it calculates the index-diff. Webserver has relatively high cpu usage (for a webserver) due to rustdoc html rewriting and quite a bit of compression / decompression. 

### deployment 
I want to keep both options in the beginning, so I can just switch one of the builders and see if anything is odd, before changing the rest. 

### NOTE 
This PR includes a new rustwide release, [which includes an internal change into how we run containers](https://github.com/rust-lang/rustwide/pull/115). I don't think anything will break, but if things feel odd, we should remember that this was done too.


GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #3286, opened 2026-04-08
reduce compression level inside zip files

syphar:
I did some tests. 

going from level 6 (the default) to level 3 gives us **25% performance improvement for 10% bigger zip files**.

### DIFF
--- a/crates/lib/docs_rs_storage/src/archive_index.rs
+++ b/crates/lib/docs_rs_storage/src/archive_index.rs
@@ -850,7 +850,9 @@ mod tests {
             for i in 0..file_count {
                 archive.start_file(
                     format!("testfile{i}"),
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Do you have the numbers for each level?

Also: is it more important to generate zip files faster or to reduce storage?

syphar:
> Do you have the numbers for each level?
> 

I did measurements for 6 (default), 3 (this value), and 1. 
The improvement for 1 was much less than hoped, so I left it at 3. 

> Also: is it more important to generate zip files faster or to reduce storage?

Right now: faster zip files :)

GuillaumeGomez:
For posterity sake, might be worth having a nice table with each level impact on speed and size. Well, I love numbers but feel free to ignore me. :rofl:

PR #3287, opened 2026-04-08
perf/storage: start streaming zip/index uploads, parallel directory upload

syphar:
I wanted to do streaming uploads for quite some time, and hoped it could be as cool as the streaming downloads. 

But sadly no :) all S3 APIs need a known content-length before you start the upload. This makes it hard for example to have a local buf and just compress it while uploading. 

Only way around that is multipart uploads, but these are comprex, and a part has ( I think) a minimum size of 5 MiB anyways, which we would have to buffer. Since most files are smaller than that, we can also just buffer the whole compressed file and then upload it. 

Where the stream works well is when you have a local file. I updated our zip&index method to use tempfiles and stream these to S3. 

With the new API it was also easy to optimize `store_all` (upload all files from a directory) to compress & upload directly, and in parallel, where before we first loaded all files into memory, compress then, and then upload them. 

Let's see if new places come up

### DIFF
--- a/crates/bin/docs_rs_admin/src/main.rs
+++ b/crates/bin/docs_rs_admin/src/main.rs
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
+
 mod rebuilds;
 mod repackage;
 #[cfg(test)]
--- a/crates/bin/docs_rs_admin/src/main.rs
+++ b/crates/bin/docs_rs_admin/src/main.rs
@@ -1,5 +1,3 @@
-#![recursion_limit = "256"]
-
 mod rebuilds;
 mod repackage;
 #[cfg(test)]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2287,6 +2287,8 @@ dependencies = [
  "flate2",
  "futures-util",
  "http 1.4.0",
+ "http-body 1.0.1",
+ "http-body-util",
  "itertools 0.14.0",
  "mime",
  "moka",
--- a/crates/bin/docs_rs_admin/src/repackage.rs
+++ b/crates/bin/docs_rs_admin/src/repackage.rs
@@ -108,7 +108,6 @@ pub async fn repackage(
 /// repackage contents of a S3 path prefix into a single archive file.
 ///
 /// Not performance optimized, for now it just tries to be simple.
-#[instrument(skip(storage))]
 async fn repackage_path(
     storage: &AsyncStorage,
     prefix: &str,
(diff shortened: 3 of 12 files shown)


crates/bin/docs_rs_admin/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
```
GuillaumeGomez: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3287/81cf3c23ec31c866c60877a32a705c4dbb76696e..90447e41ad89bb9921face4c6209a8581d90c2e6)*

Why this change?

syphar: compiler error that asked for it.

syphar: let me check if it still happens :)

syphar:
```
 1  error: queries overflow the depth limit!
   |
   = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`docs_rs_admin`)
   = note: query depth increased by 130 when computing layout of `{async fn body of repackage::repackage()}`
```

GuillaumeGomez: That is worrying. Might be better to fix the recursion rather than increase the limit considering the impact on compile-time.

syphar: how would you fix such a thing? I'm clueless where to start :) 
it feels like something deep in the async machinery.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Mark-Simulacrum:
You may want to talk to @Kobzol, he's played around with various S3 upload strategies in rustc-perf recently. I think we settled on https://docs.rs/object_store/ for it. That seems like it has some support for multi-part uploads, though I don't know if we're using them in rustc-perf.

Their docs do seem to confirm some of the multi-part limitations you're suggesting:

> Most stores require that all parts excluding the last are at least 5 MiB, and some further require that all parts excluding the last be the same size, e.g. [R2](https://developers.cloudflare.com/r2/objects/multipart-objects/#limitations). Clients wanting to maximise compatibility should therefore perform writes in fixed size blocks larger than 5 MiB.

https://docs.rs/object_store/latest/object_store/trait.MultipartUpload.html#tymethod.put_part

Kobzol:
In rustc-perf we pre-compress the files before uploading them to S3, so no direct streaming is involved there (though I try to hide latency by compressing multiple files concurrently on a background tokio blocking worker thread pool). The object_store crate seems to work fine for the uploads, and it doesn't have so many dependencies as the official AWS crates. We are not using multi-part uploads in rustc-perf at the moment, or at least not explicitly.

PR #3289, opened 2026-04-09
storage: start streaming zip/index uploads, parallel directory upload

syphar:
fixed version of https://github.com/rust-lang/docs.rs/pull/3287 

the first two commits are the ones from the original PR (cherry-picked), the last commit is the updated functionality. 

Generally I tried to go more in the direction of the s3 sdk itself. When using it with a local tempfile, it can just stream-upload it, including retries. 

The upload was still hanging when testing this directly with S3 (not with minio). My current assumption is that this is a bug between the s3 sdk & s3 itself. The sdk switches to a streaming checksum calculation when uploading a local file, which doesn't complete in my tests. Easiest workaround: calculate the checksum ourselves (or disable checksums, which I don't like). 

When I see the data from the updated `uploaded_files` metric  we can see if we could actually remove our  often we retry uploads, currently I believe we can drop our own retries and rely on the sdk-internal retries.


syphar:
checking the failing tests

syphar:
@GuillaumeGomez I would still love to do this, 

what do you think? is anything missing? 

I'm also happy to split stuff up.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me from what I can understand. :)

PR #3293, opened 2026-04-13
block more encoded path traversals

syphar:
- https://rust-lang.sentry.io/issues/7195650500/?query=is%3Aunresolved&referrer=issue-stream
- https://rust-lang.sentry.io/issues/7318119675/?query=is%3Aunresolved&referrer=issue-stream

This fixes one of the cases, I'm not sure how to fix the others yet. 

image

Notable: 
in the URL we don't have any encoded chars, but in the path we try to get from storage, we have them. 

Generally: 
when I finished the repackage, and all is in archive storage, these errors won't happen any more, sqlite will just return 404.

### DIFF
--- a/crates/bin/docs_rs_web/src/middleware/security.rs
+++ b/crates/bin/docs_rs_web/src/middleware/security.rs
@@ -41,7 +41,11 @@ fn validate_path(initial_path: &str) -> Result<()> {
 }
 
 fn validate_decoded_path(path: &str) -> Result<()> {
-    if path.contains("/../") || path.ends_with("/..") {
+    if path.contains("/../")
+        || path.ends_with("/..")
+        || path.contains("//\\../")
+        || path.contains("\\..\\")
+    {
         bail!("path traversal attempt");
     }
 
@@ -61,6 +65,7 @@ mod tests {
         testing::{AxumResponseTestExt as _, AxumRouterTestExt as _},
     };
     use axum::{Router, middleware, routing::get};
+
     use test_case::test_case;
     use tower::ServiceBuilder;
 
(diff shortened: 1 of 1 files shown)


crates/bin/docs_rs_web/src/middleware/security.rs
```diff
@@ -41,7 +41,11 @@ fn validate_path(initial_path: &str) -> Result<()> {
 }
 
 fn validate_decoded_path(path: &str) -> Result<()> {
-    if path.contains("/../") || path.ends_with("/..") {
+    if path.contains("/../")
```
GuillaumeGomez, line 44: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3293/b7fa86c53f780e3a301e8f1a8a16c5ba38012e62..917fd76d153bb866b85fd939642e45d5569ca39d)*

Is there any path that could contain `..`? If, might be easier to just look for that.

syphar: Inside the rustdoc html? Not sure 

But inside source archives, probably somewhere.

syphar: in the end, you can build filenames like `sources/some.file..ext` if you want and put that into the source archive.

syphar: btw, with the new infra the target picture is probably just to use the fastly WAF product (
signal sciences), they have automated filters for stuff like this.

crates/bin/docs_rs_web/src/middleware/security.rs
```diff
@@ -41,7 +41,11 @@ fn validate_path(initial_path: &str) -> Result<()> {
...
+        || path.ends_with("/..")
```
GuillaumeGomez, line 45: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3293/b7fa86c53f780e3a301e8f1a8a16c5ba38012e62..844d8148257f0aac165177a8f55ec0d15feef89e)*

So I guess the equivalent with backlashes is also needed?

syphar: I'm not really sure, didn't see it, 

but just added it.

GuillaumeGomez (approved):

PR #3294, opened 2026-04-14
Remove ambiguity from feature flag with no dependencies.

kevincox:
For the longest time I thought that it meant that the feature flag was defined but unused, not enabling any features. I just realized that it was actually talking about dependencies. So I figured it was good to make this more clear to avoid other people being confused like I was.

### DIFF
--- a/crates/bin/docs_rs_web/templates/crate/features.html
+++ b/crates/bin/docs_rs_web/templates/crate/features.html
@@ -108,7 +108,7 @@
                                     {%- endfor -%}
                                 </ul>
                             {%- else -%}
-                                <p>This feature flag does not enable additional features.</p>
+                                <p>This feature flag does not enable additional feature flags.</p>
                             {%- endif -%}
                         {%- endfor -%}
                     {% endif -%}


crates/bin/docs_rs_web/templates/crate/features.html
```diff
@@ -108,7 +108,7 @@ <h3 id="{{ feature.name }}">{{ feature.name }}{%- if is_default  %} (default){%-
                                     {%- endfor -%}
                                 </ul>
                             {%- else -%}
-                                <p>This feature flag does not enable additional features.</p>
+                                <p>This feature flag does not enable additional feature flags.</p>
```
GuillaumeGomez, line 111: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3294/b7fa86c53f780e3a301e8f1a8a16c5ba38012e62..c87ce45f8d55b10abadbb3679b29905f03aebe4f)*

Features can be both a feature flag or an optional crate, so not sure it removes ambiguity.

PR #3296, opened 2026-04-15
also block XSS attempts in security middleware

syphar:
errors like 
https://rust-lang.sentry.io/issues/7195650500/events/69171fe8e54b4cf5b7ee4fa7d29718e0/?query=is%3Aunresolved&referrer=previous-event

### DIFF
--- a/crates/bin/docs_rs_web/src/handlers/rustdoc.rs
+++ b/crates/bin/docs_rs_web/src/handlers/rustdoc.rs
@@ -2631,12 +2631,12 @@ mod test {
             assert_eq!(
                 latest_version_redirect(
                     "tungstenite",
-                    "/tungstenite/0.10.0/tungstenite/?search=String+-%3E+Message",
+                    "/tungstenite/0.10.0/tungstenite/?search=String+-%7B+Message",
                     &env.web_app().await,
                     env.config()
                 )
                 .await?,
-                "/crate/tungstenite/latest/target-redirect/tungstenite/?search=String+-%3E+Message",
+                "/crate/tungstenite/latest/target-redirect/tungstenite/?search=String+-%7B+Message",
             );
             Ok(())
         });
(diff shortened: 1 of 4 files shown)


crates/bin/docs_rs_web/src/handlers/rustdoc.rs
```diff
@@ -3388,8 +3388,8 @@ mod test {
         let web = env.web_app().await;
 
         web.assert_redirect_cached_unchecked(
-            "/minidumper/latest/%3c%2f%73%63%72%69%70%74%3e%3c%74%65%73%74%65%3e",
-            "/minidumper/latest/%3C/script%3E%3Cteste%3E",
+            "/minidumper/latest/%7d%2f%73%63%72%69%70%74%7b%7d%74%65%73%74%65%7b",
+            "/minidumper/latest/%7D/script%7B%7Dteste%7B",
```
syphar, line 3392: @GuillaumeGomez can you imagine a rustdoc usecase where it's ok to have `<` or `>` in the URL? 


Then I would have to rethink this rule :) 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3296/f868061cda2ecb7e9b92e79ed3419cb3dd5f62d7..d3227bea97c0eae8e0a733d5fbedcec79eb906fb)*

GuillaumeGomez: For `impl<>` maybe? But I think we encode them, no?

syphar: > For `impl<>` maybe? 

do you have an example? 

> But I think we encode them, no?

sure, but this PR would block these. 

( the examples from the test here would contain encoded `<script>` tags, which I assume are always malicious)


So I'm wondering if we could just block it and see, 
or if we should do nothing and accept the s3 / server errors, until I finished the repackage. 
( the archives / index don't have the issue, it's only S3 that raises service errors for the legacy storage)

GuillaumeGomez: Here you go: https://doc.rust-lang.org/nightly/std/convert/trait.Into.html#impl-Into%3CU%3E-for-T

syphar: but that's in the fragement, so it would never be on the server.

GuillaumeGomez: Then it's all good. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3299, opened 2026-04-16
start storing peak-memory-usage in builds table

syphar:
for now it's just the max of all sub-builds (all targets, json builds), I think that is what we and the authors would need most of the time. 

Theoretically we could emit log entries with more details, but I want to think about that longer. 

Use case for us and authors is mostly to be able to judge what memory limit is needed in the sandboxes. 

First just store, then we look at it, UI later.


crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -36,20 +36,61 @@ use docsrs_metadata::{BuildTargets, DEFAULT_TARGETS, HOST_TARGET, Metadata};
...
     io::BufReader,
+    ops::AddAssign,
     path::Path,
     sync::{Arc, LazyLock},
     time::Instant,
 };
 use tracing::{debug, error, info, info_span, instrument, warn};
 
+#[derive(Default)]
+struct ProcessStatistics(RustwideProcessStats);
```
syphar: we could get rid of this NewType with https://github.com/rust-lang/rustwide/pull/122

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3299/f868061cda2ecb7e9b92e79ed3419cb3dd5f62d7..f433a90d2f78dd1920815736b3684fcaee01af55)*

GuillaumeGomez: Maybe we should do that first? Lot of trait impls for conversions.

syphar: this is not critical, if you agree that this has a place in rustwide, we should do that :)

syphar: waiting on https://github.com/rust-lang/rustwide/pull/124

syphar: updated

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -1132,19 +1138,24 @@ impl RustwideBuilder {
...
-                info!("continuing anyways.");
-                None
-            }
-        };
+        let (coverage_stats, doc_coverage) =
+            match self.get_coverage(target, build, metadata, limits) {
+                Ok((stats, cov)) => (stats, cov),
+                Err(err) => {
+                    info!("error when trying to get coverage: {}", err);
+                    info!("continuing anyways.");
```
GuillaumeGomez: Maybe put this message directly in the previous one so it's only one log and prevent the (very unlikely) case where another log is added in between?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3299/f868061cda2ecb7e9b92e79ed3419cb3dd5f62d7..857ff217eab26eac67521a16572a93e9657a9859)*

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -1210,6 +1229,10 @@ impl RustwideBuilder {
             std::fs::rename(old_dir, new_dir)?;
         }
 
+        let main_build_stats = result.as_ref().ok().cloned().unwrap_or_default();
```
GuillaumeGomez, line 1234: Why does it need to be `cloned()`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3299/f868061cda2ecb7e9b92e79ed3419cb3dd5f62d7..857ff217eab26eac67521a16572a93e9657a9859)*

syphar: so `main_build_stats` is not a ref, so `merge` can merge it.

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
syphar: this is coming from the main branch. 

Merged it now.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3300, opened 2026-04-16
Add base for alert support

GuillaumeGomez:
This is a POC to add support for alerts. For now, it always generates an alert so you can give it a try. Even without JS, you can hide the alert. If JS is enabled when closing the alert, it will store the information and not show this alert again from this point.

### DIFF
--- a/crates/bin/docs_rs_web/src/lib.rs
+++ b/crates/bin/docs_rs_web/src/lib.rs
@@ -41,3 +41,11 @@ pub(crate) static GLOBAL_ALERT: Option<GlobalAlert> = Some(GlobalAlert {
     fa_icon: "exclamation-triangle",
 });
 */
+
+pub(crate) fn get_alert() -> Option<(&'static str, usize)> {
+    Some((
+        "We are changing which targets are built by default on May 1st.<br>\
+        <a href=\"https://blog.rust-lang.org/2026/04/04/docsrs-only-default-targets/\">Learn more</a>",
+        1,
+    ))
+}
--- a/crates/bin/docs_rs_web/templates/header/topbar_end.html
+++ b/crates/bin/docs_rs_web/templates/header/topbar_end.html
@@ -123,3 +123,15 @@
         </div>
     </div>
 </div>
+
+{%- if let Some((alert, alert_id)) = crate::get_alert() -%}
+    <input type="checkbox" id="docsrs-alert-input" data-id={{alert_id|safe}}>
+    <label id="docsrs-alert-hide" for="docsrs-alert-input" title="Close alert message">
+        {{- crate::icons::IconXmark.render_solid(false, false, "") -}}
+    </label>
+    <div id="docsrs-alert">
+        {{- crate::icons::IconCircleInfo.render_solid(false, false, "") -}}
+        <span class="alert-text">{{- alert|safe -}}</span>
+    </div>
+    <script type="text/javascript">{%- include "alert.js" -%}</script>
+{%- endif -%}
--- a/crates/bin/docs_rs_web/templates/alert.js
+++ b/crates/bin/docs_rs_web/templates/alert.js
@@ -1,0 +1,16 @@
+(function() {
+    const alertCheckbox = document.getElementById("docsrs-alert-input");
+    alertCheckbox.onchange = () => {
+        // If the user clicks on the "close button", we save this info in their local storage.
+        window.localStorage.setItem("hide-alert-id", alertCheckbox.getAttribute("data-id"));
+    };
+
+    const info = window.localStorage.getItem("hide-alert-id");
+    if (info !== null) {
+        const alertId = alertCheckbox.getAttribute("data-id");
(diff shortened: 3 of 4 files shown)


crates/bin/docs_rs_web/src/lib.rs
```diff
@@ -41,3 +41,11 @@ pub(crate) static GLOBAL_ALERT: Option<GlobalAlert> = Some(GlobalAlert {
     fa_icon: "exclamation-triangle",
 });
 */
+
+pub(crate) fn get_alert() -> Option<(&'static str, usize)> {
```
syphar, line 45: fyi: 
I prepared parts of https://github.com/rust-lang/docs.rs/pull/3274 so we can include the alerts in that caching pipeline. When https://github.com/rust-lang/docs.rs/pull/3274 is merged etc I'm happy to update this PR to do it. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3300/12ec53c5f3d0c89b6a3996e0d089eb4bdd494c11..5bd5b3a6b6605cba96b8ccbbe2823eb1ed062b45)*

crates/bin/docs_rs_web/templates/header/topbar_end.html
```diff
@@ -123,3 +123,15 @@
...
+{%- if let Some((alert, alert_id)) = crate::get_alert() -%}
+    <input type="checkbox" id="docsrs-alert-input" data-id={{alert_id|safe}}>
+    <label id="docsrs-alert-hide" for="docsrs-alert-input" title="Close alert message">
+        {{- crate::icons::IconXmark.render_solid(false, false, "") -}}
+    </label>
+    <div id="docsrs-alert">
+        {{- crate::icons::IconCircleInfo.render_solid(false, false, "") -}}
+        <span class="alert-text">{{- alert|safe -}}</span>
+    </div>
+    <script type="text/javascript">{%- include "alert.js" -%}</script>
```
syphar, line 136: we probably should expose `alert.js` similar to `menu.js` as separate ( cacheable ) asset: 

```html
<script async src="/-/static/menu.js?{{ build_slug }}"></script>
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3300/12ec53c5f3d0c89b6a3996e0d089eb4bdd494c11..5bd5b3a6b6605cba96b8ccbbe2823eb1ed062b45)*

crates/bin/docs_rs_web/templates/header/topbar_end.html
```diff
@@ -123,3 +123,15 @@
         </div>
     </div>
 </div>
+
+{%- if let Some((alert, alert_id)) = crate::get_alert() -%}
```
syphar, line 127: IMO: many people reach docs.rs via search, or crates.io 

right now I can only see the alert on the homepage. 

I would expect it to show on _all_ pages? 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3300/12ec53c5f3d0c89b6a3996e0d089eb4bdd494c11..5bd5b3a6b6605cba96b8ccbbe2823eb1ed062b45)*

syphar (changes requested):

crates/bin/docs_rs_web/templates/header/topbar_end.html
GuillaumeGomez: It's on all pages for me. Did you close it by any chance?

syphar: I didn't close it, opened another page, and didn't see it. 

But let me recheck later.

syphar: yes, works, not sure why not initially.

syphar (changes requested):
this might also interfere with caching, what do you think? 

If these alerts are rare, we could just invoke a full cdn purge each time we set one. 

otherwise we'll have to find some way to load the alert as partial (differently cached)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3300/12ec53c5f3d0c89b6a3996e0d089eb4bdd494c11..5bd5b3a6b6605cba96b8ccbbe2823eb1ed062b45)*

crates/bin/docs_rs_web/templates/header/topbar_end.html
GuillaumeGomez: Good point.

GuillaumeGomez:
For this part, you're the expert so whatever you think is better.

syphar:
> For this part, you're the expert so whatever you think is better.

a partial loaded via separate URL. 

So we have a "static" part (script / html), which is always the same, can be cached as long as the page itself (forever). 

And a "dynamic" part loaded from a separate URL (like the version lists), 
which is either "no alerts" or the alert.

PR #3303, opened 2026-04-19
update rustdoc json about page

syphar:
since https://github.com/rust-lang/docs.rs/pull/3052 we directly stream the content from the bucket to the client, and don't redirect anymore.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3312, opened 2026-04-26
Fix indent of docs.rs build page code example

GuillaumeGomez:
Fixes this bug (on https://docs.rs/about/builds):

image

### DIFF
--- a/crates/bin/docs_rs_web/templates/core/about/builds.html
+++ b/crates/bin/docs_rs_web/templates/core/about/builds.html
@@ -32,19 +32,19 @@
             <h4 id="detecting-docsrs"> <a href="#detecting-docsrs">Detecting Docs.rs</a> </h4>
             <p>
                 To recognize Docs.rs from your Rust code, you can test for the <code>docsrs</code> cfg, e.g.:
-                {% filter dedent(None)|highlight("rust") -%}
-                    #[cfg(docsrs)]
-                    mod documentation;
+                {% filter highlight("rust") -%}
+#[cfg(docsrs)]
+mod documentation;
                 {%- endfilter %}
                 The `docsrs` cfg only applies to the final rustdoc invocation (i.e. the crate currently
                 being documented). It does not apply to dependencies (including workspace ones).
             </p>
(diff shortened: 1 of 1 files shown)


crates/bin/docs_rs_web/templates/core/about/builds.html
```diff
@@ -42,9 +42,9 @@ <h4 id="detecting-docsrs"> <a href="#detecting-docsrs">Detecting Docs.rs</a> </h
             <p>
                 To recognize Docs.rs from <code>build.rs</code> files, you can test for the environment variable <code>DOCS_RS</code>, e.g.:
                 {% filter dedent(3)|highlight("rust") -%}
```
syphar:
```suggestion
                {% filter dedent(5)|highlight("rust") -%}
```

I think the `dedent` filter is for that. 

So we can have nicely indented templates while the page also looks good. 

My guess is that I probably forgot to update the dedent when refactoring templates. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3312/af671b6a3a38ead3f52ca2b7a675331b51e89289..e5c36c023fc49e3cbdceace6c5cc21dd6657be7f)*

syphar (changes requested):

crates/bin/docs_rs_web/templates/core/about/builds.html
GuillaumeGomez: I think it's way too fragile as is. Well, I suppose I can write a GUI test to catch a regression but still...

GuillaumeGomez: I should remove the whole `dedent` thing actually I think. I don't see any added value here for having "nice indent" when this code is not part of the indent in the first place.

syphar (approved):

crates/bin/docs_rs_web/templates/core/about/builds.html
syphar: valid

PR #3314, opened 2026-04-26
extend archive index with lists, use in source view for tree

syphar:
- right now we're storing a big JSON structure in `releases.files`, containing a list of all files in the source archive. 
- It's used to generate the file tree in our source handlers. 
- But: we already have the file list available in the archive index that belongs to the source archive, so we can use that and be much more efficient. 

So this PR: 
- updates the storage lib to provide access to file lists in the index, and prefix filters
- the source handlers now use the index, if archive storage is true. 

Next steps after this: 
- perhaps stop storing the file list for new builds / releases 
- stop passing around the file-lists in results from storage, builders, ... 
- or, if repackaging is done, directly remove the column when we remove the legacy storage.


crates/bin/docs_rs_web/src/utils/mod.rs
```diff
@@ -3,10 +3,48 @@ pub(crate) mod html_rewrite;
...
 use chrono::{DateTime, NaiveDate, Utc};
+use docs_rs_mimes as mimes;
+use docs_rs_storage::FolderEntry;
 use docs_rs_utils::rustc_version::parse_rustc_date;
 
+pub trait FolderEntryExt {
+    fn icon(&self) -> Safe<String>;
+}
+
+impl FolderEntryExt for FolderEntry {
```
GuillaumeGomez: Instead of creating a new trait for this, why not a function taking a `FolderEntry` as argument instead?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3314/e9cfb228d6083cd9c89e8c56abb89a67a26c3e6c..36f667e0bfcbbbcf67ff2697d4d7c6d913144aa3)*

syphar: valid, updated it.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3328, opened 2026-05-08
Add information about crate builds

### LINKED ISSUE #3318: Builds status page should be more clear
If you go to: https://docs.rs/crate/zbus/latest/builds. You will see:

Image

Seems like everything is ok. But then when you click on it and go look https://docs.rs/crate/zbus/latest/builds/3186413/x86_64-pc-windows-gnu.txt, you will see that it failed to compile, which is also not easily visible:

Image

So there are two things to be done here:

1. Have 3 status for the doc builds: all green, some failed (maybe a ⚠️ sign for this case?) and all failed (a red cross mark?).
2. Also have a status on each build log on the file icon (a checkmark or a cross mark).

GuillaumeGomez:
Fixes https://github.com/rust-lang/docs.rs/issues/3318.

The build logs now look like this:

Screenshot From 2026-05-09 02-07-46

The page listing builds now take into account if all builds succeeded or not (and adds a warning sign in such a case).

I also tweaked the `gui-test/tester.js` to make it possible to run only some tests, quite convenient. :3


GuillaumeGomez:
@syphar It's ready for review. Not sure why `sqlx` testsuite is not happy. When running `cargo sqlx prepare --workspace` locally, nothing changes. Dark magic.

syphar:
> @syphar It's ready for review. Not sure why `sqlx` testsuite is not happy. When running `cargo sqlx prepare --workspace` locally, nothing changes. Dark magic.

with sqlx some thing are odd, `just sqlx-prepare` should work.

crates/lib/docs_rs_types/src/build_status.rs
```diff
@@ -33,6 +35,7 @@ mod tests {
     #[test_case(BuildStatus::Success, "success")]
     #[test_case(BuildStatus::Failure, "failure")]
     #[test_case(BuildStatus::InProgress, "in_progress")]
+    #[test_case(BuildStatus::InProgress, "partial_failure")]
```
syphar:
```suggestion
    #[test_case(BuildStatus::PartialFailure, "partial_failure")]
```

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE builds_logs(
```
syphar, line 1: I think we should try to use _less_ JSON fields for structured data. They are good for document-like (unstructured) things. 

Here, I imagine a table like `(id INT PRIMARY KEY, build_id INT, log_filename TEXT, build_status build_status)`. 

( `build_id` indexed as foreign key)

crates/lib/docs_rs_database/src/releases.rs
```diff
@@ -627,6 +627,22 @@ where
     Ok(())
 }
 
+pub async fn add_build_logs(
+    conn: &mut sqlx::PgConnection,
+    build_id: BuildId,
+    build_logs: Vec<(String, bool)>,
+) -> Result<()> {
+    sqlx::query!(
+        "INSERT INTO builds_logs(build_id, logs)
```
syphar: this can still be one query, even when we use one row for each log-entry: 

```rust
pub async fn add_build_logs(
    conn: &mut sqlx::PgConnection,
    build_id: BuildId,
    build_logs: Vec<(String, bool)>,
) -> Result<()> {
    let (messages, successes): (Vec<String>, Vec<bool>) = build_logs.into_iter().unzip();

    sqlx::query!(
        "INSERT INTO builds_logs(build_id, message, success)
         SELECT $1, * FROM UNNEST($2::text[], $3::bool[])",
        build_id as _,
        &messages as &[String],
        &successes as &[bool],
    )
    .execute(&mut *conn)
    .await?;

    Ok(())
}
```

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -782,13 +784,19 @@ impl RustwideBuilder {
...
+                    let successful = res.successful();
                     self.blocking_storage.store_one(build_log_path, res.build_log)?;
-                    for (target, log) in target_build_logs {
+                    build_logs.push((format!("{default_target}.txt"), successful));
+                    for (target, (log, successful)) in target_build_logs {
                         let build_log_path = format!("build-logs/{build_id}/{target}.txt");
                         self.blocking_storage.store_one(build_log_path, log)?;
+                        build_logs.push((format!("{target}.txt"), successful));
                     }
+                    self.runtime.block_on(add_build_logs(&mut async_conn, build_id, build_logs))?;
```
syphar, line 795: can we extend one of the builder-tests to check if the log filenames are added to the db with their respective states?

syphar (changes requested):
thank you! 

I have one design change in mind, happy to discuss. 


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3328/37872fe2560111535811b11b3dc8df9e553cfcb4..8f2ddcc2ee53dfbe33c5ff04bfc99ec9f3700fb7)*

crates/lib/docs_rs_types/src/build_status.rs
GuillaumeGomez: Woups. ^^'

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.up.sql
GuillaumeGomez: That makes the table a lot bigger considering we'll need multiple entries for the same build. Well, doesn't matter I guess. Rewriting it then...

GuillaumeGomez:
Apart from CI (weird) errors, ready for review.

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -1510,6 +1527,16 @@ mod tests {
         assert!(row.source_size > 0);
         assert!(row.documentation_size.unwrap() > 0);
         assert!(row.memory_peak.unwrap() > 10 * 1024 * 1024); // 10 MiB, in my test it was > 100 MiB
+        assert_eq!(
+            row.logs,
+            Some(vec![
```
syphar: it seems like this test fails. 

Perhaps just ordering?

crates/bin/docs_rs_web/src/handlers/build_details.rs
```diff
@@ -127,16 +136,29 @@ pub(crate) async fn build_details_handler(
...
-            .map_ok(|path| {
-                path.strip_prefix(&prefix)
-                    .expect("since we query for the prefix, it has to be always there")
-                    .to_owned()
-            })
-            .try_collect()
-            .await?;
+
+        // A list of `(path, build_successful)`.
+        let all_log_filenames: Vec<(String, Option<bool>)> = if let Some(logs) = row.logs
```
syphar, line 141: I would expect a test covering the old behaviour (where logs are just the files in that folder), and the new one (where the table is the source). 

This wouldn't have to be on the handler, you could also extract that functionality into a separate method and test both ways on it. 

We should perhaps also update the fake-release to do it the new way? Or support both ways?

crates/bin/docs_rs_web/src/handlers/builds.rs
```diff
@@ -190,7 +190,19 @@ async fn get_builds(
             builds.id as "id: BuildId",
             builds.rustc_version,
             builds.docsrs_version,
-            builds.build_status as "build_status: BuildStatus",
+            CASE
```
syphar, line 193: I would expect a test covering this new functionality. 

Perhaps this needs also fake-release/fake-build updates? 

It's already not a handler, but this `get_builds`, which could make it easier?

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    id INTEGER NOT NULL,
```
syphar, line 2:
```suggestion
    id INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
```

This saves you the whole `_seq` / `SEQUENCE` / `nextval` stuff below.

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS build_logs_build_id_idx;
+DROP TABLE builds_logs;
```
syphar, line 2: reverting 

`ALTER TYPE build_status ADD VALUE IF NOT EXISTS 'partial_failure' AFTER 'failure';`

is missing here.

crates/bin/docs_rs_web/src/handlers/build_details.rs
syphar: I see the new behaviour is covered in the gui-tests, and the old behaviour then in the old non-gui tests here.

crates/bin/docs_rs_web/src/handlers/builds.rs
syphar: I see the new behaviour is covered in the gui-tests, and the old behaviour then in the old non-gui tests here.

crates/bin/docs_rs_web/src/handlers/build_details.rs
syphar: generally I feel like gui-tests should test gui, and behaviour should be tested separately, but I think this should block this merge.

crates/bin/docs_rs_web/src/handlers/builds.rs
syphar: generally I feel like gui-tests should test gui, and behaviour should be tested separately, but I think this should block this merge.

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.up.sql
syphar: ( also doesn't really block the approval / merge, it works both ways)

syphar (changes requested):
I updated sqlx & pushed to your branch, the one build test is failing, perhaps just ordering? 

I had other comments, but they aren't blocking the merge, and feel more philosophical than real issues. 

One addition (I can also make later) is then updating the fake-release etc to match the "new" way. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3328/e50152ed411bb913753b1dfd203f22cb8711f097..66923824f2795acd2a07a65bcb2abc0432ca2084)*

crates/lib/docs_rs_database/migrations/20260508232425_build-logs.down.sql
GuillaumeGomez: We cannot revert an enum. Fun postgresql times. =D

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
GuillaumeGomez: It worked locally for me. So I suppose the insert is not always maintained. Interesting... Well, gonna make it work in all cases.

crates/bin/docs_rs_web/src/handlers/build_details.rs
syphar: i decided not to refactor this handler, we can do that later.

syphar:
@GuillaumeGomez want to have a look at the last commit? 

adding the build-logs to the test-fakes surfaced some bugs, which I solved then.

GuillaumeGomez:
Looks good to me, thanks for the help! :)

syphar (approved):
thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3328/4789ca453c92a37dee27aff720db0a706c6f69fc..04803b1e3e5c11ef869c91070d008eb9c5b94e6f)*

PR #3338, opened 2026-05-17
storage: use Bytes instead of Vec<u8> to prevent some cloning

syphar:
Preparation for some parallelization, where I'll have to upload the same blob to more than one location. 

our `upload_stream` API already uses `Bytes`, so this just makes the rest of the storage API do the same.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3349, opened 2026-05-22
WIP: watcher: add optional SQS subscription for crate index changes

syphar:
idea is: 
- infra has to provide the SQS FIFO queue, permissions for docs.rs, and credentials for crates.io
- merge & release https://github.com/rust-lang/docs.rs/pull/3357 , so crates.io can use the structs to start publishing the events 
- finish up https://github.com/rust-lang/docs.rs/pull/3378 as prep-work for this. 
- when we can finish up this PR, first let the SQS consumption just run in parallel (dry-run / log) to git index 
- if no errors / all fine, switch over to sqs

not in here yet, but better for prod: 
a graceful shutdown for the watcher / subscriber, so when shutting down the daemon we'll wait until in-flight events are handled. 

I'll add this while this PR is in dry-run mode.


crates/lib/docs_rs_crates_io/src/events.rs
```diff
@@ -0,0 +1,207 @@
...
+
+use chrono::{DateTime, Utc};
+use std::fmt;
+
+/// A change that can happen to a crate on our index.
+#[derive(Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug)]
+#[serde(tag = "type", content = "payload", rename_all = "snake_case")]
+pub enum IndexChangeV1 {
+    /// A crate version was added.
+    Added(CrateVersion),
```
syphar, line 9: note to myself: 
this needs either `pub` or accessors

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3349/8ee6f397030443c235425b833b539a961e01c1c5..c0436b166914cf1326d6e7104961ea9a477e0378)*

crates/bin/docs_rs_watcher/src/subscriber.rs
```diff
@@ -0,0 +1,499 @@
...
+                time::sleep(WAIT_TIME).await;
+                continue;
+            }
+        };
+        metrics
+            .sqs_messages_received_total
+            .add(messages.len() as u64, &[]);
+
+        for message in messages {
+            match handle_message_body(context, config, metrics, message.body.as_deref()).await {
```
syphar, line 114: can we have some graceful shutdown here? 

otherwise, if the watcher is stopped while handling a message, the msg will be in-flight in SQS, so sqs will wait delivering _any_ message for the visibility timeout. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3349/964c90a45859fd8f9076e3cbe58470a5a4f4d4f7..dd9c9d59c8a1c2e1e3f846a7e608e5e1423eb0e5)*

PR #3356, opened 2026-06-03
storage: use s3 multipart upload for files > 100 MiB

syphar:
the current method we're using (`put_object`) works fine most of the time, but has a limitation: 

it only allows files < 5 GiB. 

Especially with the reduced default targets this is more of an edge case, but it is blocking me from my repackaging efforts, where I'm trying to convert old releases into archives, so I can drop the non-archive logic from the codebase. We have a handful of crates where the zipfile with the html files will become > 5 GiB. 

AWS recommends using multipart uploads for anything > 100 MiB, so let's do that generally. It also enables us to do parallel uploads. 

I'm also dropping our custom retry logic, in favor of the SDK internal retries that we configure too.


syphar:
cc @GuillaumeGomez do you have time for a review here?

crates/lib/docs_rs_storage/src/lib.rs
```diff
@@ -25,3 +25,7 @@ pub use utils::{
     file_list::get_file_list,
     storage_path::{rustdoc_archive_path, rustdoc_json_path, source_archive_path},
 };
+
+pub(crate) const KIB: u64 = 1024;
```
GuillaumeGomez: They only seem to be used in one file, so why not putting these consts in this file directly?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3356/7a6c393e89ad794e8555625722326ca4aa60051e..d4afcd94a9d21501467cc4952a1f05a6dbd7aa1e)*

syphar: i had another new thing in this PR, that I then removed to keep it small. In there, I was using these constants. 

But let me revert that part , so we can keep it small

crates/bin/docs_rs_admin/src/repackage.rs
```diff
@@ -51,23 +51,23 @@ pub async fn repackage(
 
     let mut algs: HashSet<CompressionAlgorithm> = HashSet::new();
 
-    if let Some((_rustdoc_file_list, alg)) = repackage_path(
+    if let Some((_rustdoc_file_list, alg)) = Box::pin(repackage_path(
```
syphar, line 54: this was necessary because we again had the recursion limit error

( where the error message says to `#![recursion_limit = "256"]`)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3356/7a6c393e89ad794e8555625722326ca4aa60051e..8b23b18f445a4a9e0d6a0ba46ee8afea7a0b25f6)*

crates/lib/docs_rs_storage/src/config.rs
```diff
@@ -118,7 +118,7 @@ impl AppConfig for Config {
             max_file_size_html: env("DOCSRS_MAX_FILE_SIZE_HTML", 50 * 1024 * 1024)?,
             #[cfg(any(test, feature = "testing"))]
             s3_bucket_is_temporary: false,
-            network_parallelism: env("DOCSRS_NETWORK_PARALLELISM", 8usize.min(cores))?,
+            network_parallelism: env("DOCSRS_NETWORK_PARALLELISM", 8usize.min(cores))?.max(1),
```
GuillaumeGomez, line 121: Just to be sure I understand: the minimum is the number of CPU cores, and the max is 1. So `network_parallelism` will always be `1`, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3356/7a6c393e89ad794e8555625722326ca4aa60051e..8b23b18f445a4a9e0d6a0ba46ee8afea7a0b25f6)*

syphar: - without env set, it's the smaller value of 8 or the number of cores. 
- with the env set, `.max(1)` protects against someone setting the env to zero

GuillaumeGomez: Shouldn't it be `.min(1)` then?

syphar: now you make me question reality :) 

`0.min(1)` will be zero again, right?

GuillaumeGomez: Oooooh, it works the other way around. I see, sorry for the confusion. ^^'

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3364, opened 2026-06-09
storage: re-add retries for s3 delete_prefix

syphar:
Digging into an error with the `repackage` command, I realized the `DeleteObjects` API only retries for errors in the main api call, and doesn't retry single failed objects on mixed responses (where some deletes fail, and some succeed). 

See https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_Examples

Here we add it back. 

On top, I'm adding parallelization if we have to delete more than 1000 objects.

### DIFF
--- a/crates/lib/docs_rs_storage/src/backends/s3.rs
+++ b/crates/lib/docs_rs_storage/src/backends/s3.rs
@@ -589,10 +589,10 @@ impl S3Backend {
             }
 
             let mut retry_keys = Vec::with_capacity(errs.len());
-            for err in errs {
-                let key = err.key;
-                let code = err.code;
-                let message = err.message;
+            for aws_sdk_s3::types::Error {
+                key, code, message, ..
+            } in errs
+            {
                 let retryable = code
                     .as_deref()
                     .is_some_and(|c| RETRYABLE_DELETE_OBJECT_ERROR_CODES.contains(&c));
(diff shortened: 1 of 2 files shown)


crates/lib/docs_rs_storage/src/backends/s3.rs
```diff
@@ -539,37 +554,91 @@ impl StorageBackendMethods for S3Backend {
...
+                    errs
+                );
+            }
 
-                anyhow::bail!("deleting from s3 failed");
+            let mut retry_keys = Vec::with_capacity(errs.len());
+            for err in errs {
+                let key = err.key;
+                let code = err.code;
+                let message = err.message;
```
GuillaumeGomez:
```suggestion
            for Error { key, code, message, .. } in errs {
```

The error type is likely wrong and I don't know if there are other fields, but that gives you an idea. :p

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3364/11da9628dad8856fd3144c2af0debff423648fe0..bdbf9ea3ab832953bdae65e82b8d5e56349e0626)*

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3367, opened 2026-06-12
update links to use the updated branch names main instead of master

darricheng:
This PR just does some cleanup of links in the readme to update links to repositories that have their `master` branch updated to `main`. 

Note: none of the links were broken, I just thought that it would be better to link directly than rely completely on GitHub's auto-redirect.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice catch, thanks!

PR #3369, opened 2026-06-13
storage: delete_prefix, more retries, reduce concurrency

syphar:
it seems like I'm hitting S3 with too many requests, even when I reduce the network concurrency. 

So here I'm 
- doing more retries, in the main `delete_objects` request 
- remove concurrent delete calls. 

( error message I'm seeing is "SlowDown: Please reduce your request rate")


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3378, opened 2026-06-20
build queue: refactor & cache priority-finding for new releases

syphar:
prep-work for the SQS ingestion from crates.io so we support a higher load in event ingestion without all these queries. 

- unifies the priority logic in one struct that is easily testable. 
- We still need a call (`reevaluate_priorities`) every now and then, which would update priorities of already queued releases, if the underlying data changes (patterns, workspace sizes, ...), but that's more optional. 
- While I was on it, I also added  a builder to the `FakeGithubStats` struct and improved some code quality in tests. 
- There's a pending refactor / some duplications with test-environments, but I want to postpone that change. 
- I want to directly store regex in the pattern-based priorities later, but for now I want this PR to be easily revertible.


GuillaumeGomez:
Can you not include `sqlx` file changes in your PR please? The server regenerate them, and it would make the review much simpler for me without having all these file files to ignore. :')

syphar:
> Can you not include `sqlx` file changes in your PR please? The server regenerate them, and it would make the review much simpler for me without having all these file files to ignore. :')

this will only work after merge, so CI on the PR will always be red? or am I missing something?

syphar:
but a valid point, let me think about a good solution. I could imagine one when we have the new infra, not sure with the current infra.

GuillaumeGomez:
Isn't the CI regenerate them? Because I don't commit my sqlx changes anymore as they never work for some reason. ^^'

syphar:
> Isn't the CI regenerate them? Because I don't commit my sqlx changes anymore as they never work for some reason. ^^'

not on PRs, as it looks like: https://github.com/rust-lang/docs.rs/pull/3443 

https://github.com/rust-lang/docs.rs/pull/3442 might help a little

GuillaumeGomez:
Hum, maybe we should do that. Would simplify things a lot.

syphar:
@GuillaumeGomez I rebased on `main`, and regenerated the sqlx files.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me!

PR #3392, opened 2026-07-02
Fix release description hover jitter

### LINKED ISSUE #173: Recent Releases shaking when descriptions are too long
When the description is too long, it will show full description on hovering.
However, when there're several too long descriptions, they will make the Recent Releases block shaking like following gif

01ac9e85-7cc4-457f-y844-ea62cff29482

Herrtian:
## Summary
- keep release descriptions truncated on hover so release rows do not change height
- add `title` text so full descriptions remain available without layout shifts
- extend the long description GUI test to cover hover row height

Fixes #173

## Validation
- cargo check -p docs_rs_web
- cargo fmt --all --check
- git diff --check

### DIFF
--- a/crates/bin/docs_rs_web/templates/core/home.html
+++ b/crates/bin/docs_rs_web/templates/core/home.html
@@ -58,7 +58,7 @@ centered
                                     {{- release.name }}-{{ release.version -}}
                                 </div>
                                 <div class="pure-u-1 pure-u-sm-14-24 pure-u-md-16-24 description"
-                                    title="{{ release.description.as_deref().unwrap_or_default() }}">
+                                    title="{{ release.description.as_deref().unwrap_or_default()|escape }}">
                                     {{- release.description.as_deref().unwrap_or_default() -}}
                                 </div>
 
--- a/crates/bin/docs_rs_web/templates/core/home.html
+++ b/crates/bin/docs_rs_web/templates/core/home.html
@@ -57,7 +57,8 @@ centered
                                 <div class="pure-u-1 pure-u-sm-6-24 pure-u-md-5-24 name">
                                     {{- release.name }}-{{ release.version -}}
                                 </div>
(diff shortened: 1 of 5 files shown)


crates/bin/docs_rs_web/templates/core/home.html
```diff
@@ -57,7 +57,8 @@ <h1 class="brand">{{ crate::icons::IconCubes.render_solid(false, false, "") }} D
                                 <div class="pure-u-1 pure-u-sm-6-24 pure-u-md-5-24 name">
                                     {{- release.name }}-{{ release.version -}}
                                 </div>
-                                <div class="pure-u-1 pure-u-sm-14-24 pure-u-md-16-24 description">
+                                <div class="pure-u-1 pure-u-sm-14-24 pure-u-md-16-24 description"
+                                    title="{{ release.description.as_deref().unwrap_or_default() }}">
```
GuillaumeGomez: Nice idea but doesn't work if the description contains a double quote. Double quotes need to be replaced.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3392/760d9e004b3b995747c14196f539df4bbf3e8a98..bb623ffe8c771263595fb8fc5c4171bcd7cd4e4a)*

GuillaumeGomez (changes requested):
Test looks good, however like mentioned, we cannot simply add the description like that in case it contains double quotes.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3392/760d9e004b3b995747c14196f539df4bbf3e8a98..bb623ffe8c771263595fb8fc5c4171bcd7cd4e4a)*

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3401, opened 2026-07-07
gui-tests: freeze puppeteer version, switch to puppeteer provided chromium

syphar:
this should fix the gui-tests, while keeping the stable environment. 

Everything JS / node related is just improvised, feel free to provide a better way.


package.json
```diff
@@ -2,5 +2,8 @@
   "name": "deps",
   "dependencies": {
     "browser-ui-test": "^0.24.0"
+  },
+  "overrides": {
```
GuillaumeGomez, line 6: The `25.1.0` doesn't work?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3401/0c4a2aded4069c2595bb0421e49cc08e21043e69..b318e9fd96e9e94ac1e139930fc49c826eddd31d)*

syphar: @GuillaumeGomez currently this is trial-and-error, errors change, it never works fully. 

( which is why this is draft still)

syphar: super time consuming too, testing the docker container locally is even slower than waiting for CI :) 

( also I don't know much about puppeteer / headless chrome etc)

GuillaumeGomez: Welcome to the club. :')

GuillaumeGomez: Still the same question. ;)

package-lock.json
GuillaumeGomez, line 1: Oh also, `npm` doesn't "respect" the lock file unlike `Cargo.lock`. It's mostly a minimum version thing rather than a strict requirement.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3401/e8a46b7b5a511a097ae8494b8de99c649fdda2f3..69116dee0f3db0d258fa3ee634be623ff8c23408)*

package.json
syphar: I don't follow? 

25.3 is the most recent version ?

package-lock.json
syphar: I don't follow? what does this mean? 

What should be changed here?

package.json
syphar: why should we choose 25.1?

GuillaumeGomez: I'm asking if there is any reason to pick another version in short. Normally `npm install` will instead `25.3` automatically too.

package-lock.json
GuillaumeGomez: Mostly that this file is completely useless and doesn't need to be committed.

syphar: kk, removing it

package.json
syphar: not knowing much about puppeteer at all, 

the reason was mostly wanting to freeze the version, so it doesn't randomly start breaking. 

Or can we assume that it will always continue working if we don't rely on the system chromium anymore?

GuillaumeGomez: In rust we pinned a specific chromium version to go around this problem. For now it seems to work.

package-lock.json
syphar: @GuillaumeGomez works now without package-lock

GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #3414, opened 2026-07-21
remove legacy non-archive storage

syphar:
the repackage is done, so all historical releases are in zip files now (without having to rebuild everything).  

This also means the zip download works for the whole history. 

I'm not yet dropping the `releases.files` and `releases.archive_storage` columns to be able to revert this change. That's also why we still write `TRUE` into `releases.archive_storage`. 

I'll do that later when all is fine.


syphar:
@GuillaumeGomez also rebased on main, good enough for review?

GuillaumeGomez (approved):

GuillaumeGomez:
This one was easy to review. ^^'

PR #3415, opened 2026-07-23
add content-length to streamed file downloads for browsers / progress

syphar:
This is mainly for our zip / json downloads. Right now, the browser can't show any download progress because we don't send a `Content-Length` header. 

Thing is: we only know the original content length from S3, which becomes obsolete when we decompress the stream while serving it to the user. 

But that doesn't matter for the zip / json downloads, since we use `get_raw_stream`, which doesn't do any decompression. 

This was also a bug with our "don't cache big objects" logic, where we assumed that the S3 content length is also what we return to the user / the CDN.

### DIFF
--- a/crates/bin/docs_rs_web/src/file.rs
+++ b/crates/bin/docs_rs_web/src/file.rs
@@ -10,7 +10,7 @@ use axum::{
 };
 use axum_extra::{
     TypedHeader,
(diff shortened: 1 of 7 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #3417, opened 2026-07-24
add retries to fastly api requests, use same retry logic for registry api

syphar:
to fix errors like [this sentry error](https://rust-lang.sentry.io/issues/7478288604/?project=5499376&query=is%3Aunresolved&referrer=issue-stream) 

- fastly api requests weren't retried before, we have some sentry errors where content was outdated because of that. 
- registry api retries where just retried for all possible errors. the reqwest-retry logic only retries transient errors, which would reduce load on crates.io. 

With this middleware we get retries just for "transient" errors, and niceties like jittering. 

When the repackage & https://github.com/rust-lang/docs.rs/pull/3414 is done we then can remove our custom `retry_async` method. 

the default retry strategy is explained here: 
- https://docs.rs/reqwest-retry/latest/reqwest_retry/fn.default_on_request_success.html 
- https://docs.rs/reqwest-retry/latest/reqwest_retry/fn.default_on_request_failure.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2169,6 +2169,8 @@ dependencies = [
  "mockito",
  "opentelemetry",
  "reqwest",
+ "reqwest-middleware",
+ "reqwest-retry",
  "tokio",
  "tracing",
  "url",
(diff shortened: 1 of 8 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3430, opened 2026-08-05
builder: unify target/doc finding, support json file in get_coverage

syphar:
This is coming from [a bugfix in rustdoc which changes behaviour for the coverage output](https://github.com/rust-lang/rust/pull/159411). 

Before we always got the coverage output in stdout. Now, the `-o` arg is used, which is set to `./target/doc` by `cargo` when it calls rustdoc. 

Without dropping to raw rustdoc, the only option is to switch our coverage-reading to reading the local file. 

While I'm on it, I unified the logic to find the doc output directory,  and the logic to find a single json file in the output dir. 

The legacy stdout logic I left in for now, so we can deploy this change with the old frozen nightly version, and then unfreeze nightly as we wish. Later I can remove that part too. 

*this also fixes the currently failing CI on `main`*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-*/ignored
+**/ignored
 /.env
 /.docker.env
 /src/web/badge/Cargo.lock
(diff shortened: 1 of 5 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3438, opened 2026-08-11
builder: only pass scrape-examples for html builds

syphar:
This is partially a regression from https://github.com/rust-lang/docs.rs/pull/3430. 

When a crate has examples, and we tried to generate the coverage, we'll get multiple json files. We don't want coverage for examples. 

The actual issue seems to be that we added `-Zrustdoc-scrape-examples` to all builds, but we only want this for the html build. 

[This is the related sentry error](https://rust-lang.sentry.io/issues/7178942819/events/6eabe7ce74ba4f3c93ac419380fb86c9/?project=5499376&query=is%3Aunresolved&referrer=previous-event), I'll re-run builds for all the crates where rustdoc json failed in the last days. This will also fix coverage for these.


GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3440, opened 2026-08-11
fix rustdoc-json & coverage for proc-macro crates with custom rustc_args

syphar:
Weird corner-case. I was digging into [cases where we still don't get rustdoc-json for some crates](https://rust-lang.sentry.io/issues/7178942819/?query=is%3Aunresolved&referrer=issue-stream). 

In https://github.com/rust-lang/docs.rs/commit/3f92a79e12a98c77a15c003a810bdba3e9cdf3ca we added `-Zhost-config`, `-Ztarget-applies-to-host`, and `--config host.rustflags={}` to also pass custom `rustc_args` to build scripts. This works because we always pass `--target` to doc builds. 

In case of proc-macro builds it's different: we don't pass `--target`, so the args above lead to our custom `build.rustdocflags` config being ignored, including custom `rustdocflags` from the metadata, or `--output-format json` from us. 

This removes `target-applies-to-host` for proc-macro crates. 

*for now I wouldn't try to rebuild rustdoc-json, since it's super tricky to find the affected crates. This could be a part of a general backfill*.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3441, opened 2026-08-11
fix json / coverage builds with custom -Z rustdoc-scrape-examples given

syphar:
Fix errors like [this](https://rust-lang.sentry.io/issues/7178942819/events/3ea4bc256f564e80a92520e37fea1933/?query=is%3Aunresolved&referrer=previous-event). 

Since https://github.com/rust-lang/docs.rs/pull/3438 we only pass `-Z rustdoc-scrape-examples` for HTML builds. But crates can use docs.rs metadata to add `-Z rustdoc-scrape-examples` themselves. This breaks the coverage & json builds for them.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3442, opened 2026-08-11
replace per-crate sqlx folders with just the workspace level

syphar:
I remember trying this in the very beginning of sqlx, and it didn't work. let's try again with a newer sqlx version. 

This reduces the amount of sqlx files in PRs by quite a bit (942 files => 164), so should help with reviews a little. I have another idea on top, but that will only work with the docker-based deploys / new infra. 

**Review: only the middle commit (the justfile update) is necessary, the rest was autogenerated, and eventual issues are covered in CI.**. 

See also https://github.com/transact-rs/sqlx/issues/3362 , https://github.com/transact-rs/sqlx/pull/4376


GuillaumeGomez (approved):

GuillaumeGomez:
Middle commit changes look good to me. :)

PR #3450, opened 2026-08-18
storage/s3: reduce unnecessary retries, add cleanup for legacy storage

syphar:
When running the `repackage` command (that converted legacy single-file to archive-storage) I had regular errors in the cleanup phase. Since [that was was only tried _after_ everything else is finished, that wasn't an issue for the conversion itself](https://github.com/syphar/docs.rs/blob/eaa7f6af7403a361377c58f82237c2961aec3be4/crates/bin/docs_rs_admin/src/repackage.rs#L113-L115). 

The error I was was from the AWS SDK, and about a mismatchin the XML schema when running `delete_prefix`. Currently I don't know why that error happened. 

But: I checked an bucket-inventory (= file-list for the whole bucket), and there are plenty of remaining rustdoc or source files. 

This command tries to clean these up. Likely I'll run into the same XML encoding errors, but no I'll print the file-list / error list to be able to debug these better. 

Before actually running the delete, I'll `--dry-run` the command, and inspect the resulting file-list.


syphar:
r? @GuillaumeGomez 

audit failure fixed in https://github.com/rust-lang/docs.rs/pull/3451

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #3459, opened 2026-08-18
refactor justfiles / test commands 

syphar:
I tried to revisit all the just commands, coming from my current usage under linux, also thinking about my past usage as mac user.  https://github.com/rust-lang/docs.rs/pull/3474, https://github.com/rust-lang/docs.rs/pull/3472, https://github.com/rust-lang/docs.rs/pull/3475, and https://github.com/rust-lang/docs.rs/pull/3473 were actually intermediate steps I pulled out into separate PRs. 

My goal is actually that I'm using all these commands myself, instead of my shell history. 

Main points: 
- by default on a dev machine, the binaries should be directly used, locally via `cargo run -p`, on all operating systems. You can switch to using the containers if you want. 
- the builder should do `cargo run` only on linux, otherwise use the builder docker container, since our builder doesn't work otherwise. 
- gui-tests should run without docker on dev machines, only in CI we want all in docker to know it works. In any case, I dropped the `Dockerfile.gui-tests` and we'll always run the tests on the runner. With this, locally iterating on the gui-tests works super smooth. 
- I want to easily run the CI checks locally via a recipe, when fixing stuff. 

Generally I see this just as "the first idea", and will iterate over the next weeks when I use them. 

Other small changes: 
- all these super small `cli-xx-` recipes are dropped, people can just use `just cli SOME ARGS`. Downside: no recipe autocomplete. Not sure yet what I like more.


syphar:
( waiting for actions to work again, they are broken at the moment)

package-lock.json
```diff
@@ -0,0 +1,400 @@
+{
```
GuillaumeGomez: Do we want to add this file?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3459/79f294f14a07662f909cb6c5d1db6d34a2754d8c..b4a78bc61fdbe014cd5ef625d3f026342e32e45a)*

GuillaumeGomez (approved):

PR #3473, opened 2026-08-26
drop unused fuzzystrmatch postgres extension

syphar:
we don't use that since we migrated to using the crates.io search api a couple of years ago. 

This also means we don't need the custom `Dockerfile-postgres`. 

I adapted the initial migration, and made the reverse migration a noop, 
because otherwise our testing would fail creating the extension in the new docker postgres image, since we don't create that with `install_extensions.sql` any more.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice!

PR #3479, opened 2026-08-27
add single-crate consistency check, using the sparse index

syphar:
thinking about the future where the git index might be only updated once an hours, we need another source for the consistency check. 

So with this, when the db is outdated for a single crate, we can just sync that. 

I imagine that we could also run a scheduler similar to the repo stats. 

the removed `queue.attempt` was obsolete code that's not needed any more

( test failure is unrelated)


GuillaumeGomez:
Failed CI.

syphar:
> Failed CI.

that test failure is unrelated

crates/bin/docs_rs_watcher/src/consistency/db.rs
```diff
@@ -64,6 +60,49 @@ pub(super) async fn load(
     Ok(crates)
 }
 
+pub(super) async fn load_single(conn: &mut sqlx::PgConnection, name: &KrateName) -> Result<Crates> {
```
GuillaumeGomez: Wouldn't it be better to return `Result<Option<Crate>>` instead since we always return 0 or 1?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3479/79f294f14a07662f909cb6c5d1db6d34a2754d8c..72f1c1449373e17da3382a6132a03eec522208e6)*

syphar: initially I would have answered we need the list, so we can reuse the same `calculate_diff` algo that we have for the whole index. 

But then I rememnered: 
- the algorithm just needs an `Iterator`  
- and `Option` implements that.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3481, opened 2026-08-30
fix: Fix pluralization in rebuild queue message

Bishops-exe:
Not really tested, but should be good to go


GuillaumeGomez:
Thanks for the fox! A test would be appreciated indeed.

syphar (changes requested):
see message above

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3481/fc88a933f9d2529d4808afcce3ba76deeee42512..d25d2536e2b2a4ea6fe921955d780e88accd54c1)*

PR #3498, opened 2026-09-03
update tracing log format, remove unnessary rustwide log fields

syphar:
- our internal `Pretty` format used the tracing-subscriber `Full` format, which is quite noisy. 
- I now added support for all log-formats that `tracing-subscriber` supports by default. 
- for local cli usage, `compact` will be the default, which puts the msg/span fields at the end 
- the default for the log-format is now determined by `is_terminal` or not. So we can drop the explicit setting on our server, and will have nicers logs in `docs_rs_admin`, or other CLI calls. 
- `rustwide` build logs came via `tracing_log::LogTracer`, which appended 4 additional fields to each line. We don't need most of these, so I added  a custom adapter that will just emit the log-message and target to tracing, and drop the rest. This only affects tests, local CLI / dev usage, etc. On our production build-servers, the log-tracer was disabled anyways, since we don't emit the build logs, just capture and upload them. But: this will be much more interesting when the build-cli for CI makes progress, so the output is actually usable. 

### next steps after this 
- reduce the level from `INFO` to `DEBUG` for some spans (for example internal function calls in rustwide) 
- remove _all_ fields from spans (via `instrument(skip_all)`), only add back what we really want to see in the logs, and _never use the debug representation of structs in logs_. 
- move some fields from spans to events, so they get logged once, and not again in all child-method-calls. 

### tracing log format 
 With an active `build{crate=example}` span, the previous full formatter
 repeated the span fields before each bridged message:

 ```text
 2026-09-03T12:00:00Z  INFO build{crate=example}: log: Compiling example
 ```

 The compact formatter keeps the span name before the message and appends
 the span fields after it:

 ```text
 2026-09-03T12:00:00Z  INFO build: log: Compiling example crate=example
 ```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2115,7 +2115,6 @@ dependencies = [
  "tokio",
  "toml",
  "tracing",
- "tracing-log",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


syphar:
Damn, some compile errors / test errors. 

Will fix them tomorrow or later or next week, then ping you again

GuillaumeGomez (approved):

PR #3480, opened 2026-08-27
WIP: refactor builder / rustwide integration 

syphar:

PR #3511, opened 2026-09-09
build_limits: Default::default for defaults, rename `new` to `from_config`

syphar:
starting to pull out things from https://github.com/rust-lang/docs.rs/pull/3480 

- use `Default::default` for the default values 
- rename `::new` to `::from_config` to make it more explicit

### DIFF
--- a/crates/lib/docs_rs_build_limits/src/limits.rs
+++ b/crates/lib/docs_rs_build_limits/src/limits.rs
@@ -15,16 +15,28 @@ pub struct Limits {
     pub max_log_size: usize,
 }
 
-impl Limits {
-    pub fn new(config: &Config) -> Self {
+impl Default for Limits {
+    fn default() -> Self {
         Self {
             // 3 GB default default
-            memory: config.build_default_memory_limit.unwrap_or(3 * GB),
+            memory: 3 * GB,
             timeout: Duration::from_secs(15 * 60), // 15 minutes
             targets: crate::DEFAULT_MAX_TARGETS,
             networking: false,
             max_log_size: 100 * 1024, // 100 KB
         }
+    }
+}
+
+impl Limits {
+    pub fn from_config(config: &Config) -> Self {
+        let mut limits = Limits::default();
+
+        if let Some(memory_limit) = config.build_default_memory_limit {
+            limits.memory = memory_limit;
+        }
+
+        limits
     }
 
     pub async fn for_crate(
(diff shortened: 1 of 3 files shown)


crates/lib/docs_rs_build_limits/src/limits.rs
```diff
@@ -15,24 +15,36 @@ pub struct Limits {
     pub max_log_size: usize,
 }
 
-impl Limits {
-    pub fn new(config: &Config) -> Self {
+impl Default for Limits {
+    fn default() -> Self {
```
GuillaumeGomez, line 19: It's not used so can be removed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3511/3461bba35589eefed0895950604b575bc9502394..e20ba89d2ac6b382ce3fada2a91a2941d822c00f)*

syphar: I want to / will use it in the new PR / build CLI

syphar: there I need the default limits, without having the builder config, 

then perhaps possible to overwrite with args

GuillaumeGomez: Then add it when needed. ;)

syphar: as often with big PRs, my idea here is just to reduce PR size by pulling out stuff that is ready / reviewable. 

On top: IMO `default` is more explicit / clearer than `new`?

GuillaumeGomez: > as often with big PRs, my idea here is just to reduce PR size by pulling out stuff that is ready / reviewable.

It's not great when trying to understand why something was added but won't block the merge for this.

> On top: IMO `default` is more explicit / clearer than `new`?

I tend to never use `Default`, no particular reason and no opinion beyond that.

syphar: > > as often with big PRs, my idea here is just to reduce PR size by pulling out stuff that is ready / reviewable.
> 
> 
> 
> It's not great when trying to understand why something was added but won't block the merge for this.
> 
> 

If you prefer it like that, but as part of a bigger PR, we can do that too.

GuillaumeGomez: 7 lines for implementing the `Default` trait won't change much (although I appreciate your consideration). ;)

syphar: it's also for me when re-reviewing my own changes and trying to grasp everything in my head

GuillaumeGomez (approved):

PR #3512, opened 2026-09-10
add retries to github/gitlab repo graphql api

syphar:
[we are seeing errors like this](https://rust-lang.sentry.io/issues/7630632742/?query=is%3Aunresolved&referrer=issue-stream&sort=recommended): 
```
GitHub GraphQL response status: 502 Bad Gateway
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

Where retries would likely help. 

We don't retry on rate-limiting errors, what the default would be in the library we're using. 
The repo-updater tasks currently run once an hour, and when we reach rate limits, we just stop updating repos and continue on the next run.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3513, opened 2026-09-11
builder: report error when we can't fetch owner data from crates.io

syphar:
let's upgrade this to an error, I think we should get a sentry report when this happens, it's not normal, and should be fixed.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3514, opened 2026-09-12
repo-stats: retry on non-standard HTTP 499 too

syphar:
in [sentry we see http 499 coming from github](https://rust-lang.sentry.io/issues/7630632742/?query=is%3Aunresolved&referrer=issue-stream): 

```
GitHub GraphQL response status: 499 <unknown status code>
```

Some research implies it's transient (to me). 

We could also have some shared retry policy where other parts of the codebase (registry-api) would also know how to handle the status 499, but for now I'll leave it just here, where I actually saw it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2398,6 +2398,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
+ "test-case",
  "tokio",
  "tracing",
 ]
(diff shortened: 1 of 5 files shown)


syphar:
test failure is unrelated, due to minio being archived / taken down. Only available via license key now. 

CI / local testing fix in https://github.com/rust-lang/docs.rs/pull/3515/

syphar:
test failure is a flaky test, unrelated to this change

GuillaumeGomez (approved):

GuillaumeGomez:
That one was quick to review. Weird that they have their own HTTP status code. "internet"

Anyway, thanks! =D

PR #3515, opened 2026-09-12
switch from minio to RustFS for local & CI testing

syphar:
This fixes CI on `main`, and in other PRs. 

Minio seem to have become commercial: 
- https://www.min.io/ 
- https://github.com/minio/minio
( also, their docker image is gone). 

The new thing (_AIStor_, who comes up with these names?) needs a license key to run. 

[`rustfs`](https://github.com/rustfs/rustfs) seems to be [S3 compatible](https://docs.rustfs.com/en/reference/s3-compatibility), big enough in usage, apache2 licensed, and works.

PR #3516, opened 2026-09-12
fix source-browser with filenames that have literal `%` in their name

syphar:
minio fix from https://github.com/rust-lang/docs.rs/pull/3515 only that the tests pass. 

- [sentry error](https://rust-lang.sentry.io/issues/7726770013/events/14a3d97c2887440aa0a17257c95f28a5/?query=is%3Aunresolved&referrer=previous-event) 
- [source folder with the file](https://docs.rs/crate/create-tauri-app/4.6.0/source/templates/_base_/src-tauri/capabilities/)


GuillaumeGomez (approved):

GuillaumeGomez:
Fun times with URLs. ^^' Thanks!

PR #3517, opened 2026-09-12
cache 404 errors inside release docs

syphar:
more a micro optimization, but it's annoying me that these reach our origin server, while we know the result only will change after a rebuild. 

Same for unknown about pages. 

### data 
This is how many 404 requests we get over 6h. Mostly from bots. 

image


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3518, opened 2026-09-12
Faster test database with DDL replay

syphar:
Looking at how much time tests take, a good part of that is creating the test database / environment, and tearing it down again. 

### Our old design: 
- we run everything inside one database 
- each test case gets its own database schema 
- we run all migrations on the schema 
- then run the test
- then run all the reverse-migrations (needed) 

### A classic other approach
- create a postgres _template database_ once, using the migrations 
- for each test, create a new database using that template (fast-ish) 
- run the test 
- drop the test specific db again 

`sqlx::test` uses that approach, crates.io did so previously, but dropped it recently [for another approach](https://github.com/rust-lang/crates.io/pull/13668): 

### the new approach: DDL replay

- inside a single database, create a _template schema_ using the migrations 
- generate & store a DDL (schema-creation SQL) from that schema 
- each test gets a new schema (as before) 
- we use the DDL to set that schema up (faster than running the migrations) 
- run the test 
- drop the test schema again 

Here **we change docs.rs to use the same approach**, with one difference: we still run the reverse-migrations after each test. We could make that optional and only run it in CI, but my tests showed that it doesn't make a big difference (<1s). 

Using `cargo nextest` as test runner enables us to set up the test schema only once, and not once per test-process. A fallback makes it possible to still run the tests with `cargo test`, though a little bit slower, since every test process will regenerate the template schema.

### design notes: 
This adds a dependency, where developers have to have a _matching_ postgres clients on their system. For CI, the `docker compose exec` workaround is fine, or for complicated dev setups. But for the normal case I actually assume that developers have a postgres client to work with. 

Alternatives would be libraries like [`elefant_tools`](https://docs.rs/elefant-tools/latest/elefant_tools/fn.copy_data.html), but there aren't many (only this), and they are not really maintained. Seeing the test speed improvement, that's an OK price to pay IMO to have the postgres client on your computer, which you probably had anyways to `psql` into the db. 

### performance, on my machine 

| Variant                              | duration |
| ------------------------------------ | -------- |
| migrations + `cargo test` (baseline) | 94s      |
| migrations + `cargo nextest`         | 70s      |
| DDL replay + `cargo nextest`         | **21s**  |

Most of the speed difference probably comes from creating the test-schema with a complete DDL instead of running the historical migrations.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #3519, opened 2026-09-12
Reduce missing/corrupt SQLite archive-index errors caused by overlapping cache cleanup and repair

syphar:
This is motivated by [sentry errors we see regularly](https://rust-lang.sentry.io/issues/7622197286/?query=&referrer=issue-stream&sort=recommended) (_error returned from database: (code: 14) unable to open database file_), and **similar errors leading to flaky tests**. Also keep in mind: the storage engine only reports to sentry when the problem still happens after 5 retries. 

The solutions here are more or less guesses about where these race conditions happen. Also, this PR is more about quickfixes, hoping these reduce the issue. 

I'm thinking about a redesign separately, the whole thing feels too complex for me, of course it's also a central part of our webserver. 

### details
- Await eviction cleanup instead of spawning detached deletion tasks. Explicit invalidations leave file cleanup to the caller, avoiding a second deletion that could remove a replacement.
- Invalidate stale manager entries before repair, then reopen the file inside the coalesced initializer. This allows immediate repair while reusing files another request has already repaired,
instead of acting on the original `force_redownload` decision.
- Retry the first failed operation without purging. This gives a concurrently published replacement a chance to succeed; subsequent failures still trigger purge within the existing five-attempt
limit.
- Delete the index, WAL, and SHM files concurrently, preserving the existing error handling. This might slow down get/set acccess to the moka cache, when it does a maintenance run. It shouldn't happen too often, we do a background maintenance run every 30s. 

Successful fast-path lookups remain without a path lock. These changes reduce avoidable deletion and repair churn but do not eliminate all filesystem races.

### more tests 
- more unit-tests for old/new edge cases 
- some optional stress-tests, seeing that they failed on main and are fixed here.


syphar:
@GuillaumeGomez 
I'm happy to provide more context, details, explanations, if that's what's needed here. 

Thanks for your help!

crates/lib/docs_rs_storage/src/archive_index.rs
```diff
@@ -408,14 +412,18 @@ impl Cache {
...
-            let to_delete = path.with_extension(format!("{ARCHIVE_INDEX_FILE_EXTENSION}-{ext}"));
-            let _ = fs::remove_file(&to_delete).await;
-        }
+    async fn remove_local_index(index_path: impl AsRef<Path>) -> Result<()> {
+        let index_path = index_path.as_ref();
+        let wal = index_path.with_extension(format!("{ARCHIVE_INDEX_FILE_EXTENSION}-wal"));
+        let shm = index_path.with_extension(format!("{ARCHIVE_INDEX_FILE_EXTENSION}-shm"));
 
-        if let Err(err) = fs::remove_file(&path).await
+        let (_, _, index_result) = tokio::join!(
```
GuillaumeGomez, line 420: Can there ever be a race condition here? Like a file is being read while it's being deleted.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3519/e8fb233564a3ee1e9c38c788329242e4a290f636..3263e701b30e1c0d3ac1040d9a6e2923fe76fe1a)*

GuillaumeGomez (approved):
Apart from my worry, looks good to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3519/e8fb233564a3ee1e9c38c788329242e4a290f636..3263e701b30e1c0d3ac1040d9a6e2923fe76fe1a)*

crates/lib/docs_rs_storage/src/archive_index.rs
syphar: Deleting just deletes from the folder / directory. 

If the file as already open, the read will work normally for long enough.

syphar: so, I checked again :) 

this might happen, but our repair-cycles would handle that.

syphar: ( edit: back to the original comment. is true). 

example:
 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0463dbe4a3d538118f57de7ba350e426

PR #3531, opened 2026-09-16
use ByteSize for byte size rendering / parsing, add same to duration

syphar:
This started of as a part of the new `docs_rs_build` CLI, where I wanted to parse & render byte-sizes & durations consistently. 

The lead to doing that for `docs_rs_admin` too, which will make my life easier when using the admin CLI to work with crate limits. 

Then I thought about configuration, where the `.env` files will also be much easier to read when we use human-readable values for durations & sizes. 

Then I fixed all the errors. Also, the `Duration` type offers some nice constructors since rust 1.91, so I changed that where I saw it. The `docs_rs_env_vars` change was necessary to handle the `Err` type from `ByteSize` / `FromStr` with our helpers. 

These are not yet all places, and not yet the perfect design (I tried to prevent a `ByteSize` newtype for now).


GuillaumeGomez:
Is it really worth it to add a new dependency for a type we could implement in less than 100 lines?

syphar:
> Is it really worth it to add a new dependency for a type we could implement in less than 100 lines?

I admit I can't judge the compile time difference, it's probably more than these 100 lines with the parsing & rendering, but even when it's 300/400 lines, plus tests?

crates/bin/docs_rs_builder/src/docbuilder/rustwide_builder.rs
```diff
@@ -1165,7 +1162,7 @@ impl RustwideBuilder {
         ]);
 
         let mut storage = LogStorage::new(log::LevelFilter::Info);
-        storage.set_max_size(limits.max_log_size());
+        storage.set_max_size(limits.max_log_size().as_u64() as usize);
```
GuillaumeGomez, line 1164: Btw, I'm very uncomfortable with such casts. I don't expect `docs.rs` to even run on 32 bits systems, but still. You can ignore this comment, it's just me expressing my worries. ^^'

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/docs.rs/3531/250275a3606094729aef7129211030d238eb1ab1..cb7cbdf8840f5d9fcaeb01afcbc12c9e378e8ccf)*

GuillaumeGomez:
I don't even think we need 100 lines. We need `gb`, `mb` and `b` constructors. And we can implement `Deref` (or `repr(transparent)`, this one might be better in this case) on it and basically we have all we need.

syphar:
> I don't even think we need 100 lines. We need `gb`, `mb` and `b` constructors. And we can implement `Deref` (or `repr(transparent)`, this one might be better in this case) on it and basically we have all we need.

I replaced the external crate with a simplified custom implementation, with less unnecessary features or speed, and with two added things that the lib didn't have.

GuillaumeGomez (approved):

PR #3532, opened 2026-09-17
WIP: load, cache & show std replacements

syphar:


syphar:
This is how [crates.io renders the same warning](https://crates.io/crates/lazy_static): 
image

Generally I feel like I want to find a way that we don't have to make the topbar bigger. 

But not sure if this is worth a dropdown? or detail page like for the abnormalities?

PR #3533, opened 2026-09-17
fix builder documentation_size metric, improve build_time

syphar:
Looking at datadog, I realized the documentation-size is broken: 
image

Some digging shows that is since https://github.com/rust-lang/docs.rs/pull/3009 / 2682ed47 (november 2025!), where we started using buckets defined as MiB, where we still reported bytes... :( 

To fix that, I first switched 
- `ArchiveStatistics` to use `ByteSize`, same as 
- `finish_release` and `finish_build`

So it's clearer what is reported, and the decision how to store is done in `docs_rs_database`. 

`DOCUMENTATION_SIZE_BUCKETS` are also now defined with `ByteSize`, and reporting the `documentation_size` is only done via a small helper method (taking `ByteSize` of course). That should make mistakes much harder in this area. I fixed the metric unit to [what the opentelemetry semantic conventions define](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestbodysize). 

The `crate_details` we fetch for our web handlers now also use `ByteSize` for explicitness. 

While I was on it, I did the same for the `build_time` metric.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks! I'm sure there is a way to init the const slices without having to write all values by hand but don't have time to think about it right now. Maybe later. =D

PR #3534, opened 2026-09-17
use const initializer blocks for build-time & doc size metric buckets

syphar:
the comment by @GuillaumeGomez in https://github.com/rust-lang/docs.rs/pull/3533 made me think. Never used const initializer blocks, but this would work. 

- the doc-size buckets are the same before & after 
- the build-time buckets are slightly differnt, but I think in a good way. 

Question for us is if that's better readable or not :) Let's discuss :) 

If that's accepted I'll revisit the other histogram bucket definitions & metrics.


GuillaumeGomez (approved):

GuillaumeGomez:
Readibility-wise, it's a bit less good, however the code comments explain what's going on and there is no typo issue possibility anymore (eventually logic ones). Nice work! I'll let it up to you whether you prefer this version or the previous one. In both cases you have my approval. :)

PR #3537, opened 2026-09-19
Make the build of font-awesome-as-a-crate deterministic

vadorovsky:
The order of traversing the `fontawesome-free-*-desktop/svgs` directory depends on the filesystem. Furthermore, the results were store in a `HashMap`. That caused the builds to be non-determinitic (producing different binaries) across different machines.

Fix that by sorting the entries and storing them in a `BTreeMap`.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

